PR #42, opened 2019-03-28
Add permission-setting for MacOS pipeline failure

erikma:


Examples/DotNetCoreBuild/validate-build.sh
```diff
@@ -5,6 +5,7 @@ readonly ARGS="$@"
 
 function run_build {
     local additionalArgs="$@"
+    chmod +x $MY_DIR/validate-build-kext.sh
     $MY_DIR/validate-build-kext.sh /logsDirectory:"$bxlLogDir" /o:"$bxlObjDir" $ARGS $additionalArgs
```
almiliMSFT:
```suggestion
/bin/bash "$MY_DIR/validate-build-kext.sh" /logsDirectory:"$bxlLogDir" /o:"$bxlObjDir" $ARGS "$@"
```
and then delete lines 7 and 8

erikma: Got it. Also fixing a similar pattern in build.sh I added previously

PR #55, opened 2019-03-29
Store per-build cache lookup fingerprints to improve strong fingerprint cache miss analysis

anniefu:
Previously, FingerprintStoreAnalyzer/CacheMissV2 only stored and compared fingerprints that were computed at execution time. Storing fingerprints is relatively expensive as paths are left expanded and pip fingerprints can be large.

This can cause confusing analysis for non-deterministic pips that might change their observed inputs from build-to-build and have different path sets build-to-build. That results in pips computing a different strong fingerprint at cache lookup time when a path set from the cache is used vs. at execution time when the pip is re-run and its path set is re-observed.

BuildXL Changes

- New "FingerprintStore.CacheLookup" store that writes directly into "Logs/Fingerprints/FingerprintStore.CacheLookup"
- -- This is a per-build store that starts empty/non-existent at the beginning of the build and only gets used on strong fingerprint misses
- -- On strong fingerprint misses only, one fingerprint computed at cache lookup time is stored in FingerprintStore.CacheLookup for that pip, the strong fingerprint computed using the path set that matches the path set for that pip from the previous build is preferred
- Moves the copied store "Logs/BuildXL/FingerprintStore" to "Logs/Fingerprints/FingerprintStore", the build-over-build persisted version remains in "EngineCache/FingerprintStore"
- Add "/storeFingerprints:ExecutionFingerprintsOnly" flag to revert back to previous state where only fingerprints computed at execution time were stored

FingerprintStoreAnalyzer Changes

- If a "FingerprintStore.CacheLookup" store exists, whenever a strong fingerprint miss is analyzed, the analyzer will compare the entry from the new build's "CacheLookup" store to the old build's "Execution" store. Assuming pip's are being cached correctly, this is very similar to what actually happens at cache lookup time.
- If "FingerprintStore.CacheLookup" does not exist || weak fingerprint miss || entry missing from "CacheLookup", the analyzer falls back on the execution-time "FingerprintStore"

### DIFF
--- a/Public/Src/Engine/Scheduler/Tracing/RuntimeCacheMissAnalyzer.cs
+++ b/Public/Src/Engine/Scheduler/Tracing/RuntimeCacheMissAnalyzer.cs
@@ -53,7 +53,7 @@ namespace BuildXL.Scheduler.Tracing
 
                 if (option.Mode == CacheMissMode.Local)
                 {
-                    possibleStore = FingerprintStore.CreateSnapshot(logTarget.FingerprintStore, loggingContext);
+                    possibleStore = FingerprintStore.CreateSnapshot(logTarget.ExecutionFingerprintStore, loggingContext);
                 }
                 else
                 {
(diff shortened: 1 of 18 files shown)


Public/Src/Engine/Scheduler/Tracing/RuntimeCacheMissAnalyzer.cs
```diff
@@ -122,6 +122,12 @@ public sealed class RuntimeCacheMissAnalyzer : IDisposable
         /// </summary>
         private ConcurrentDictionary<PipId, PipCacheMissInfo> m_pipCacheMissesDict;
 
+        /// <summary>
+        /// A previous build's <see cref="FingerprintStore"/> that can be used for cache miss comparison.
+        /// This may also be a snapshot of the current build's main <see cref="FingerprintStore"/> at the beginning of the build.
+        /// </summary>
+        public FingerprintStore PreviousFingerprintStore => m_fingerprintStoreToCompare;
```
semihokur: >m_fingerprintStoreToCompare [](start = 60, length = 27)

could you rename that as well to m_previousFingerprintStore

PR #56, opened 2019-03-29
Update CommandLineEscaping to escape semicolons on non-Windows platforms

almiliMSFT:
Some tools receive arguments in the form of semicolon-separated values.  For example, `csc`'s `/define:` switch:
```
csc.exe /define:DEBUG;TRACE
```
When running on a non-Windows platform, such values must be escaped because the semicolon ends the current command, i.e.,
```bash
dotnet csc.dll /define:"DEBUG;TRACE"
```
This PR:
  - [x] updates the `CommandLineEscaping` class to handle this case, and
  - [x] updates the csc runner (`csc.dsc`) to temporarily manually add quotes around the value for the `/define:` switch

### DIFF
--- a/Public/Sdk/Public/Managed/Tools/Csc/csc.dsc
+++ b/Public/Sdk/Public/Managed/Tools/Csc/csc.dsc
@@ -76,7 +76,10 @@ export function compile(inputArgs: Arguments) : Result {
 
         Cmd.option("/langversion:", args.languageVersion),
 
-        Cmd.option("/define:",       args.defines ? args.defines.join(";")    : undefined),
+        // TODO: uncoment the following line and delete the line after it once a new LKG is published
+        // Cmd.option("/define:",       args.defines ? args.defines.join(";") : undefined), 
+        ...addIf((args.defines || []).length > 0, Cmd.rawArgument(`/define:"${args.defines.join(';')}"`)),
+
         Cmd.option("/nowarn:",       args.noWarnings ? args.noWarnings.map(n => n.toString()).join(",") : undefined),
         Cmd.flag("/warnaserror",     args.treatWarningsAsErrors),
         Cmd.option("/warnaserror-:", (args.treatWarningsAsErrors && args.warningsNotAsErrors) ? args.warningsNotAsErrors.join(",") : undefined),
--- a/Public/Src/Utilities/Utilities/CommandLineEscaping.cs
+++ b/Public/Src/Utilities/Utilities/CommandLineEscaping.cs
@@ -235,13 +235,26 @@ namespace BuildXL.Utilities
 
         private static bool ContainsDelimiter(string s)
         {
-            foreach (char c in s)
+            if (OperatingSystemHelper.IsUnixOS)
             {
-                if (c <= ' ')
+                foreach (char c in s)
                 {
-                    if (c == ' ' || c == '\t')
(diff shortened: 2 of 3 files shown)


Public/Src/Utilities/Utilities/CommandLineEscaping.cs
```diff
@@ -235,16 +235,29 @@ private static int Repeat(StringBuilder builder, string s, int count)
 
         private static bool ContainsDelimiter(string s)
         {
-            foreach (char c in s)
+            if (OperatingSystemHelper.IsUnixOS)
             {
-                if (c <= ' ')
```
almiliMSFT: It's hard for me to believe that this optimization has any noticeable effect, but since it was already there, I didn't want to get rid of it when running on Windows (hence the copy-pasted `foreach` loops)

Public/Src/Utilities/Utilities/CommandLineEscaping.cs
```diff
@@ -235,16 +235,29 @@ private static int Repeat(StringBuilder builder, string s, int count)
...
                     {
                         return true;
                     }
                 }
             }
+            else
+            {
+                foreach (char c in s)
+                {
+                    if (c <= ' ')
```
almiliMSFT: It's hard for me to believe that this optimization has any noticeable effect, but since it was already there, I didn't want to get rid of it when running on Windows (hence the copy-pasted `foreach` loops)

Public/Sdk/Public/Managed/Tools/Csc/csc.dsc
```diff
@@ -76,7 +76,10 @@ export function compile(inputArgs: Arguments) : Result {
 
         Cmd.option("/langversion:", args.languageVersion),
 
-        Cmd.option("/define:",       args.defines ? args.defines.join(";")    : undefined),
+        // TODO: uncoment the following line and delete the line after it once a new LKG is published
+        // Cmd.option("/define:",       args.defines ? args.defines.join(";") : undefined), 
+        ...addIf((args.defines || []).length > 0, Cmd.rawArgument(`/define:"${args.defines.join(';')}"`)),
```
dannyvv: >d.rawArgume [](start = 52, length = 11)

I think the fix would be better to have a new enum for argument escaping that matches Unix'es Process launch encodings....  and have that algorithm be the default on the mac.

Okay for now but just sharing if we hit more of these in the future.

Public/Src/Utilities/Utilities/CommandLineEscaping.cs
```diff
@@ -235,16 +235,29 @@ private static int Repeat(StringBuilder builder, string s, int count)
 
         private static bool ContainsDelimiter(string s)
         {
-            foreach (char c in s)
+            if (OperatingSystemHelper.IsUnixOS)
```
dannyvv: >atingSystemHelper.IsUnixOS) [](start = 20, length = 27)

Okay for one, but if we hit more of these we should add a UnixOsCommandLine encoding scheme. This is in `\Public\Src\Pips\Dll\Operations\PipDataFragmentEscaping.cs`

Public/Sdk/Public/Managed/Tools/Csc/csc.dsc
almiliMSFT: Indeed.

PR #57, opened 2019-03-29
Add helper script to copy CMake dependencies for building in the cloud

marcelolynch:
Add a batch script that fetches the needed dependencies from the local machine and puts them in a special folder. This is used for internal cloud builds while developing the CMake front end.

### DIFF
--- a/Shared/Scripts/CopyCMakeDeps.bat
+++ b/Shared/Scripts/CopyCMakeDeps.bat
@@ -1,0 +1,353 @@
+@echo off
+set __NAME=%0
+if "%1"=="/?" goto :help
+
+:processArgs
+set "__ARG=%1"
+if "%__ARG%"=="/v" (
+    set __VERBOSE=1
+    shift
+    goto :processArgs
+)
+if "%__ARG%"=="/c" (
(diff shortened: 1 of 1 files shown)


Shared/Scripts/CopyCMakeDeps.bat
```diff
@@ -0,0 +1,353 @@
...
+    set SOURCE="%%%LOCATION_VARIABLE%%%"
+) else (
+    call where /Q %EXECUTABLE%
+    if !ERRORLEVEL! NEQ 0 (
+        call :ReportNotFoundExecutable %EXECUTABLE% %LOCATION_VARIABLE% %DESCRIPTION%
+        exit /b 1
+    )
+    for /f "usebackq delims=" %%i in (`where /F "%EXECUTABLE%"`) do set SOURCE="%%~i"
+)
+XCOPY /E /D /y /q %SOURCE% %DESTINATION% > NUL
```
dannyvv: >XCOPY  [](start = 0, length = 6)

There is also an option to use robocopy... That one has a /MIR option and if the folders are large you can pass /MT to get multi-threaded copy going.

PR #58, opened 2019-03-29
Update CreatePr helper for github

dannyvv:


dannyvv:
@anniefu Can't add you as a reviewer yet.

CreatePr.cmd
```diff
@@ -117,13 +117,31 @@ REM ************************************************************
...
+			echo --------------------------------------------------------------------------------------
+			echo - FAILURE  :-(  Could not determine a git 'remote' that is not Microsoft/BuildXL.git -
+			echo --------------------------------------------------------------------------------------
+			exit /b 1
+			goto :Error
+		)
+
 	:ForcePushToServer
-		call git push --force origin HEAD:%PR_BRANCH_NAME%
+		call git push --force %PR_REMOTE_NAME% HEAD:%PR_BRANCH_NAME%
```
almiliMSFT: Why push to "upstream" and not "origin"?  I don't think we should be doing this.  This `CreatePr.cmd` script is available for external user and they will not have permissions to do this.  But even we should follow best practices and creating branches in our own forks instead of in Microsoft/BuildXL.git

dannyvv: Because I don't want to be a branch manager and FI/RI I clone master from the original repo, work on that and then push to my remote. So this code is now generic regarding if you cloned the main repo or if you cloned your fork, this will always push to your fork and there doesn't have to be a convention on how you name your remotes.

almiliMSFT: I just realized that, unsurprisingly, I didn't interpret this batch script correctly, so in fact `PR_REMOTE_NAME` indeed refers to the name of the remote that is not Microsoft/BuildXL.git, which is great.

PR #59, opened 2019-03-29
Public LKG builds on macOS

Kr1sso:
- [x] Adjusted tooling to work out of the box with the latest public LKG on macOS
- [x] Enabled running Nuget via Mono to create deployable Nuget packages for our CI loops build on macOS

Among other things this allows us to run `/bxl.sh "/f:output='Out/Bin/release/pkgs/*'" /sandboxKind:macOSKext  /q:ReleaseDotNetCoreMac --internal` now and create Nuget packages for deployment on our CI pipelines.

Users can now also out of the box run `./bxl.sh --minimal /sandboxKind:macOSKext --load-kext` and get a fully sandboxed and cacheable build if they have SIP disabled, as the public LKG ships with an unsigned release KEXT.


bxl.sh
```diff
@@ -25,9 +25,15 @@ function findMono() {
...
-    local BUILDXL_LKG_VERSION=$(grep "BUILDXL_LKG_VERSION" "$MY_DIR/Shared/Scripts/BuildXLLkgVersion.cmd" | cut -d= -f2 | tr -d '\r')
-    local BUILDXL_LKG_FEED_1="https://1essharedassets.pkgs.visualstudio.com/_packaging/BuildXL/nuget/v3/index.json"
+    local LKG_FILE="BuildXLLkgVersionPublic.cmd"
+
+    if [[ -n "$arg_Internal" ]]; then
+        local LKG_FILE="BuildXLLkgVersion.cmd"
+    fi
+
+    local BUILDXL_LKG_VERSION=$(grep "BUILDXL_LKG_VERSION" "$MY_DIR/Shared/Scripts/$LKG_FILE" | cut -d= -f2 | tr -d '\r')
+    local BUILDXL_LKG_NAME=$(grep "BUILDXL_LKG_NAME" "$MY_DIR/Shared/Scripts/$LKG_FILE" | cut -d= -f2 | perl -pe 's/net472/osx-x64/g' | tr -d '\r')
```
dannyvv: >perl - [](start = 104, length = 6)

not `egrep`?

Public/Sdk/Public/Managed/Tools/NuGet/pack.dsc
```diff
@@ -125,9 +125,10 @@ export function pack(args: Arguments): PackResult {
 
     let execArgs = <Transformer.ExecuteArguments>{
         tool: args.tool || tool,
-        tags: (args.tags || ["package"] ),
+        tags: (args.tags || ["package"]),
         arguments: arguments,
         workingDirectory: outDir,
+        allowUndeclaredSourceReads: Context.getCurrentHost().os === "macOS",
```
dannyvv: >allowUndeclaredSourceReads [](start = 8, length = 26)

too much to track mono safely?

bxl.sh
Kr1sso: Tools sometimes slightly differ in behavior on Darwin/BSD in comparison to other systems. I used perl as the expected output should be stable across systems 🐭

Public/Sdk/Public/Managed/Tools/NuGet/pack.dsc
Kr1sso: Indeed, the actual worker (Nuget) is tracked, Mono just rages all over the place as the child it is executing (Nuget) does all those file accesses. We still depend on system folders and files, so when updates and env changes, we invalidate and repackage, so we do on source file changes so this should be enough.

PR #60, opened 2019-03-30
Underbuild: Fix handling of directory symlink

narasamdya:
Given a file symlink `D:\dir1\junction\directory_symlink\file.lnk` such that 
- junction points to `D:\dir1\dir2`, 
- directory_symlink points to `dir3\dir4`, 
- file.lnk points to `..\file.txt`
the next immediate target is `D:\dir1\junction\dir3\file.txt`

The prefix junctions should not be resolved, while the prefix directory symlinks should be resolved.


Public/Src/Utilities/UnitTests/Storage/FileUtilitiesTests.cs
```diff
@@ -678,6 +678,56 @@ public void TestResolveSymlinkWithDirectorySymlinkOrJunction(bool useJunction, b
             XAssert.ArePathEqual(expectedFinalPath, maybeResult.Result);
         }
 
+        [Trait("Category", "WindowsOSOnly")] // TODO for non Windows
```
almiliMSFT: Why TODO?  Junctions don't exist on non-Windows platforms

Public/Src/Utilities/Native/IO/Windows/FileSystem.Win.cs
```diff
@@ -3903,92 +3926,88 @@ private Possible<string> TryGetFinalPathByHandle(string path)
         /// with ..\target\file2.txt results in a correct path, i.e., repo\target\file2.txt. The same reasoning can be given for repo\source\symlink1.link, and its resolution results in
         /// a non-existent path target\file1.txt.
         /// </remarks>
-        public Possible<string> ResolveReparsePointPrefixes(string path)
+        public Possible<string> TryResolveReparsePointRelativeTarget(string path, string relativeTarget)
```
almiliMSFT: It would be good to have some unit tests for this method alone

PR #63, opened 2019-04-01
Increase DownloadResolver timeout and log error when canceled

edgarrs:
Fixes the issue where the public builds fails when the download resolver times out or is canceled during evaluation phase.

### DIFF
--- a/Public/Src/FrontEnd/Download/DownloadResolver.cs
+++ b/Public/Src/FrontEnd/Download/DownloadResolver.cs
@@ -327,7 +327,7 @@ namespace BuildXL.FrontEnd.Download
             }
             catch (TaskCanceledException e)
             {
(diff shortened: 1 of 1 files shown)


Public/Src/FrontEnd/Download/DownloadResolver.cs
```diff
@@ -324,8 +325,10 @@ private async Task<EvaluationResult> TryDownloadFileToDiskAsync(DownloadData dow
                         new FileInfo(downloadFilePathAsString).Length);
                 }
             }
-            catch (TaskCanceledException)
+            catch (TaskCanceledException e)
             {
+                string message = m_context.CancellationToken.IsCancellationRequested ? "Download manually canceled" : e.Message;
```
smera: >d" [](start = 113, length = 2)

nit: final '.' missing

PR #64, opened 2019-04-01
[MacSandbox] Extract bundle identifier into an xcconfig

almiliMSFT:
This will be useful because being able to control bundle identifier and buildxl class prefix from a single place will allow us to easily build and simultaneously load multiple versions of our kext (e.g., one for LKG and one for tests)


Public/Src/Sandbox/MacOs/BundleInfo.xcconfig
```diff
@@ -0,0 +1,2 @@
+BXL_BUNDLE_IDENTIFIER = com.microsoft.buildxl.sandbox
```
dannyvv: BUILDXL_ to be consistent

Public/Src/Sandbox/MacOs/projects.dsc
```diff
@@ -29,7 +30,7 @@ namespace Sandbox {
         Transformer.SealSourceDirectoryOption.allDirectories);
 
     export function build(args: Args): Result {
```
dannyvv: Not sure how many of these we have, but perhaps it makes sense to put a `buildXCode` function in `Sdk.BuildXL`?

Public/Src/Sandbox/MacOs/Sandbox/Src/BuildXLSandboxShared.hpp
```diff
@@ -259,10 +259,24 @@ inline bool HasAllFlags(const T source, const T bitMask)
 
 #pragma mark Macros and defines
 
-#define BXL_CLASS(name) com_microsoft_buildxl_ ## name
-#define BuildXLSandbox BXL_CLASS(Sandbox)
-#define kBuildXLBundleIdentifier "com.microsoft.buildxl.sandbox"
-#define kBuildXLSandboxClassName "com_microsoft_buildxl_Sandbox" // this string must match the value of BuildXLSandbox
+#ifndef BXL_BUNDLE_IDENTIFIER
```
dannyvv: `BUILDXL_` for consistency with windows?

Public/Src/Sandbox/MacOs/BundleInfo.xcconfig
almiliMSFT: will rename

Public/Src/Sandbox/MacOs/Sandbox/Src/BuildXLSandboxShared.hpp
almiliMSFT: will rename

Public/Src/Sandbox/MacOs/projects.dsc
almiliMSFT: currently it's just this one, so I'd say we do this once it becomes more pressing

PR #66, opened 2019-04-01
Implement GRPC copy beyond localhost

kewilli:
Still adding tests, but this contains the core changes.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Text;
+using BuildXL.Cache.ContentStore.Hashing;
+using BuildXL.Cache.ContentStore.Interfaces.Distributed;
+using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
+using BuildXL.Cache.ContentStore.Utils;
+
+namespace BuildXL.Cache.ContentStore.Distributed.Utilities
+{
+    /// <summary>
+    /// Path transformer for distributed file copies.
+    /// </summary>
+    public class GrpcDistributedPathTransformer : IAbsolutePathTransformer
+    {
+        private static readonly string _localMachineName = Environment.MachineName;
+        internal const string BlobFileExtension = ".blob";
+
+        /// <inheritdoc />
+        public AbsolutePath GeneratePath(ContentHash contentHash, byte[] contentLocationIdContent)
+        {
+            string contentHashString = contentHash.ToHex();
+            var pathRoot = new AbsolutePath(Encoding.UTF8.GetString(contentLocationIdContent));
+            return pathRoot / contentHash.HashType.ToString() / contentHashString.Substring(0, 3) / (contentHashString + BlobFileExtension);
+        }
+
+        /// <inheritdoc />
+        public byte[] GetLocalMachineLocation(AbsolutePath cacheRoot)
+        {
+            if (!cacheRoot.IsLocal)
+            {
+                throw new ArgumentException($"Local cache root must be a local path. Found {cacheRoot}.");
+            }
+
+            if (!cacheRoot.GetFileName().Equals(Constants.SharedDirectoryName))
+            {
+                cacheRoot = cacheRoot / Constants.SharedDirectoryName;
+            }
+
+            string networkPathRoot = cacheRoot.Path.Replace(":", "$");
+
+            return
+                Encoding.UTF8.GetBytes(
+                    Path.Combine("\\\\" + _localMachineName, networkPathRoot).ToUpperInvariant());
+        }
+
+        /// <inheritdoc />
+        public byte[] GetPathLocation(AbsolutePath path)
+        {
+            var tempPath = path;
+            while (tempPath != null && !tempPath.Path.EndsWith(Constants.SharedDirectoryName, StringComparison.OrdinalIgnoreCase) || tempPath.IsRoot)
+            {
+                tempPath = tempPath.GetParentPath<AbsolutePath>();
+            }
+
+            if (tempPath.IsRoot)
+            {
+                throw new ArgumentException("Rooted path does not match a Cache root");
+            }
+
+            return Encoding.UTF8.GetBytes(tempPath.Path.ToUpperInvariant());
+        }
+    }
+}
--- a/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
@@ -44,7 +44,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.Utilities
 
             return
                 Encoding.UTF8.GetBytes(
-                    Path.Combine("\\\\" + _localMachineName, networkPathRoot).ToUpperInvariant());
+                    Path.Combine(@"\\" + _localMachineName, networkPathRoot).ToUpperInvariant());
         }
 
         /// <inheritdoc />
--- a/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcFileCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcFileCopier.cs
@@ -21,6 +21,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.Utilities
     /// </summary>
     public class GrpcFileCopier : IAbsolutePathFileCopier
     {
+        private const int DefaultGrpcPort = 7089;
         private Context _context;
         private int _grpcPort;
 
@@ -48,7 +49,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.Utilities
 
             CopyFileResult copyFileResult = null;
             // Contact hard-coded port on source
-            using (var client = GrpcCopyClient.Create(System.Net.IPAddress.Any.ToString(), 7089))
+            using (var client = GrpcCopyClient.Create(System.Net.IPAddress.Any.ToString(), DefaultGrpcPort))
             {
                 copyFileResult = await client.CopyFileAsync(_context, contentHash, destinationPath, cancellationToken, contentSize);
             }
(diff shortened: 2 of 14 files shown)


Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
```diff
@@ -0,0 +1,67 @@
...
+            if (!cacheRoot.GetFileName().Equals(Constants.SharedDirectoryName))
+            {
+                cacheRoot = cacheRoot / Constants.SharedDirectoryName;
+            }
+
+            string networkPathRoot = cacheRoot.Path.Replace(":", "$");
+
+            return
+                Encoding.UTF8.GetBytes(
+                    Path.Combine("\\\\" + _localMachineName, networkPathRoot).ToUpperInvariant());
```
dcwuser: Minor: Prefer @"\\" to "\\\\".

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcFileCopier.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            var contentHash = new ContentHash(hashType, HexUtilities.HexToBytes(hashLiteral));
+
+            return (host, contentHash);
+        }
+
+        /// <inheritdoc />
+        public Task<CopyFileResult> CopyToAsync(AbsolutePath sourcePath, Stream destinationStream, long expectedContentSize, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
```
dcwuser: >throw [](start = 12, length = 5)

Will this screw with hash-while-streaming?

kewilli: Absolutely. This still needs to be done. If trusted hash is off, then this will not be called.

---
In reply to: [271087833](https://github.com/Microsoft/BuildXL/pull/66#discussion_r271087833) [](ancestors = 271087833)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -19,10 +21,40 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
     /// An implementation of a CAS copy helper client based on GRPC.
     /// TODO: Consolidate with GrpcClient to deduplicate code. (bug 1365340)
     /// </summary>
-    public class GrpcCopyClient : IShutdown<BoolResult>
+    public sealed class GrpcCopyClient : IShutdown<BoolResult>
     {
+        private static ConcurrentDictionary<(string, int), GrpcCopyClient> _clientPool = new ConcurrentDictionary<(string, int), GrpcCopyClient>();
```
dcwuser: >string [](start = 45, length = 6)

Note that creating a custom type that defines its own GetHashCode and Equals is usually faster than defaults. (In fact, it may default to reference equality, which will just mean you will never find a dictionary entry. Check this.)

SergeyTeplyakov: System.ValueTuple is a struct with correct value semantic, so this should be totally fine.

---
In reply to: [271089260](https://github.com/Microsoft/BuildXL/pull/66#discussion_r271089260) [](ancestors = 271089260)

SergeyTeplyakov: I would suggest to use named tuple. It is hard to tell from the definition what's the key.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -19,10 +21,40 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+    public sealed class GrpcCopyClient : IShutdown<BoolResult>
     {
+        private static ConcurrentDictionary<(string, int), GrpcCopyClient> _clientPool = new ConcurrentDictionary<(string, int), GrpcCopyClient>();
+
+        /// <summary>
+        /// Use an existing GRPC client if possible, else create a new one.
+        /// </summary>
+        /// <param name="host">Name of the host for the server (e.g. 'localhost').</param>
+        /// <param name="grpcPort">GRPC port on the server.</param>
+        public static GrpcCopyClient Create(string host, int grpcPort)
```
SergeyTeplyakov: nit: could you please move the factory after all the fields (and a private constructor (I suppose we have one).

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -19,10 +21,40 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+        {
+            Console.WriteLine($"Creating GRPC client for {host},{grpcPort}");
+            if (_clientPool.TryGetValue((host, grpcPort), out var existingClient))
```
SergeyTeplyakov: To be honest, I would prefer no caching at all instead of having a cache that is effectively a memory leak (a cache without invalidation policy).

Lack of caching means that the app is just slow. But the current version can cause some super weird errors because of port exhaustion or because of some other issues..

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -60,6 +94,7 @@ public async Task<BoolResult> ShutdownAsync(Context context)
             }
 
             ShutdownCompleted = true;
+            _clientPool.TryRemove((_host, _grpcPort), out var copyClient);
```
SergeyTeplyakov: Should we assert that TryRemove returns true?

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcFileCopier.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <inheritdoc />
+        public Task<FileExistenceResult> CheckFileExistsAsync(AbsolutePath path, TimeSpan timeout, CancellationToken cancellationToken)
+        {
+            // TODO: Implement!
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        public async Task<CopyFileResult> CopyFileAsync(AbsolutePath sourcePath, AbsolutePath destinationPath, long contentSize, bool overwrite, CancellationToken cancellationToken)
```
dannyvv: >CopyFileAsync [](start = 42, length = 13)

It seems that there is an assumptoin that this path is from the cache folder (i.e. extension `.blob`) should we document or validate this?

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcFileCopier.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            var host = segments.First();
+            var hashLiteral = segments.Last();
+            if (hashLiteral.EndsWith(GrpcDistributedPathTransformer.BlobFileExtension))
+            {
+                hashLiteral = hashLiteral.Substring(0, hashLiteral.Length - GrpcDistributedPathTransformer.BlobFileExtension.Length);
+            }
+            var hashTypeLiteral = segments.ElementAt(segments.Count - 1 - 2);
+
+            if (!Enum.TryParse<HashType>(hashTypeLiteral, out HashType hashType))
```
dannyvv: >hashType [](start = 71, length = 8)

Should we check that the length of the hash matches the length expected based on the type?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dannyvv: Agree in principle, but if you look closely this is keyed off of host and port. Perhaps we should consider making the GrpcCopyClient a weak reference so it has a chance to be collected, and we can reinstate when needed.

---
In reply to: [271135869](https://github.com/Microsoft/BuildXL/pull/66#discussion_r271135869) [](ancestors = 271135869)

dannyvv: Just noticed that `ShutDownAsync` removes it. So if we have a guarantee that even in error cases and timeouts and cancellation that shutdown method is called, this should be no worse than all the other caches we have created. I'd still love to see at some point all these caches reported in a single object that we can measure holistically for memory and usage with well known lifetimes.

---
In reply to: [271395644](https://github.com/Microsoft/BuildXL/pull/66#discussion_r271395644) [](ancestors = 271395644,271135869)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -135,15 +172,8 @@ private async Task<T> RunClientActionAndThrowIfFailedAsync<T>(Context context, F
         /// <inheritdoc />
         public void Dispose()
         {
-            Dispose(false);
-            GC.SuppressFinalize(this);
+            // noop for now
```
dannyvv: Shouldn't at least the `_clientPool` be emptied?

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcFileCopier.cs
kewilli: This is validated by the `ContentHash` constructor

---
In reply to: [271394465](https://github.com/Microsoft/BuildXL/pull/66#discussion_r271394465) [](ancestors = 271394465)

dannyvv: Thanks!

---
In reply to: [271469202](https://github.com/Microsoft/BuildXL/pull/66#discussion_r271469202) [](ancestors = 271469202,271394465)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: Removing _clientPool

---
In reply to: [271135924](https://github.com/Microsoft/BuildXL/pull/66#discussion_r271135924) [](ancestors = 271135924)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: Removing _clientPool

---
In reply to: [271397083](https://github.com/Microsoft/BuildXL/pull/66#discussion_r271397083) [](ancestors = 271397083)

PR #67, opened 2019-04-01
Add process "weight" to control scheduling parallel processes

anniefu:
BuildXL has a limited number of process slots that determine how many processes can run concurrently during the build. Previously, all process pips took up one slot, no matter how resource-heavy the process was. 

This PR adds the concept of process "weight" that processes can use to indicate they require more resources/slots than average and should not be run in parallel with as many other processes.

Changes
- Add "weight" field to Process, which represents how many process slots that process needs to run
- The total "weight" of all processes running concurrently must be less than the number of available process slots (dictated by /maxproc and possibly RAM available)
- "weight" defaults to 1, converts values < 1 to 1, and considers values >= available process slots to mean the process should run alone

[AB#1469609](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1469609)

### DIFF
--- a/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
@@ -126,7 +126,7 @@ namespace Transformer {
          * Default is globally controlled by the sandbox configuration
          * TODO: input isolation is not implemented
          */
-         containerIsolationLevel?: ContainerIsolationLevel;
+        containerIsolationLevel?: ContainerIsolationLevel;
 
         /**
          * The policy to apply when a double write occurs.
--- a/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
@@ -163,10 +163,11 @@ namespace Transformer {
         disableCacheLookup?: boolean;
 
         /**
-         * The weighted value of this pip when limiting concurrency of process pips.
-         * The default weight for a process pip is 1, and the max valid weight is the max concurrent processes allowed in the build.
-         * The higher the weight, the fewer process pips that can run in parallel.
-         * No additional processes can be run concurrently while the sum of weights of all running processes is >= the build's weight limit (which is the max valid weight).
+         * The # of process slots this process requires when limiting concurrency of process pips.
+         * The total weight of all proceses running concurrently must be less than or equal to the # of available process slots.
+         * The # of available process slots is typically a function of the number of cores on the machine, but can also be limited by runtime resource exhaustion or be set per-build by configuration.
+         * Valid input range for the weight is [min Int32, max Int32] though all values will be effectively coerced to fit within [1, # of process slots]
+         * If a given weight is greater than or equal to # of available process slots, the process will run alone.
          */
         weight?: number;
     }
--- a/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
@@ -10,21 +10,21 @@ namespace Transformer {
 
     @@public
     export interface ExecuteArguments extends ExecuteArgumentsCommon {
-        /** Regular process pips that make calls to one or more service
+        /** Regular process pips that make calls to one or more service 
           * pips should use this field to declare those dependencies
           * (so that they don't get scheduled for execution before all
           * the services have started). */
         servicePipDependencies?: ServiceId[];
 
-        /** Whether to grant the read/write permissions of this pip to
-          * the declared service pips (permissions are granted only
+        /** Whether to grant the read/write permissions of this pip to 
+          * the declared service pips (permissions are granted only 
           * throughout the lifetime of this pip). */
         delegatePermissionsToServicePips?: PermissionDelegationMode;
     }
-
+    
     /** Different options for delegating permissions of a process to a service pip. */
     @@public
-    export const enum PermissionDelegationMode {
+    export const enum PermissionDelegationMode { 
         /** Don't grant any permissions at all. */
         none,
 
@@ -33,7 +33,7 @@ namespace Transformer {
 
         /** Grant permissions permanently, i.e., until the service pip terminates. */
         permanent
-    }
+    } 
 
     @@public
     export interface ExecuteArgumentsCommon extends ExecuteArgumentsComposible {
@@ -83,7 +83,7 @@ namespace Transformer {
 
         /** Regex that would be used to extract warnings from the output. */
         warningRegex?: string;
-
+        
         /** Regex that would be used to extract errors from the output. */
         errorRegex?: string;
 
@@ -116,7 +116,7 @@ namespace Transformer {
 
         /** Whether this process should run in an isolated container (i.e. filesystem isolation)
          * When running in a container, the isolation level can be controlled by 'containerIsolationLevel' field.
-         * Note: this is an experimental feature for now, use at your own risk
+         * Note: this is an experimental feature for now, use at your own risk 
          * Default is globally controlled by the sandbox configuration
          */
         runInContainer?: boolean;
@@ -126,7 +126,7 @@ namespace Transformer {
          * Default is globally controlled by the sandbox configuration
          * TODO: input isolation is not implemented
          */
-        containerIsolationLevel?: ContainerIsolationLevel;
+         containerIsolationLevel?: ContainerIsolationLevel;
 
         /**
          * The policy to apply when a double write occurs.
@@ -134,12 +134,12 @@ namespace Transformer {
          */
         doubleWritePolicy?: DoubleWritePolicy;
 
-        /** Whether this process should allow undeclared reads from source files. A source
+        /** Whether this process should allow undeclared reads from source files. A source 
          * file is considered to be a file that is not written during the build.
          * Note: this option turns static enforcements based on source file declarations into dynamic
          * ones. The downside is that potential build errors will be reported later in time, during the
          * execution phase, and only based on runtime observations. This means that even if there is something
-         * wrong, BuildXL may not see it if the running build doesn't hit it. The advise is: statically declare all
+         * wrong, BuildXL may not see it if the running build doesn't hit it. The advise is: statically declare all 
          * sources if possible, and only use this option for the sources where static predictions are not available. */
         allowUndeclaredSourceReads?: boolean;
 
@@ -161,6 +161,14 @@ namespace Transformer {
          * for a pip. The pip will still be stored to the cache for sake of supporting distribution.
          */
         disableCacheLookup?: boolean;
+
+        /**
+         * The weighted value of this pip when limiting concurrency of process pips.
+         * The default weight for a process pip is 1, and the max valid weight is the max concurrent processes allowed in the build.
+         * The higher the weight, the fewer process pips that can run in parallel.
+         * No additional processes can be run concurrently while the sum of weights of all running processes is >= the build's weight limit (which is the max valid weight).
+         */
+        weight?: number;
     }
 
     @@public
@@ -186,7 +194,7 @@ namespace Transformer {
         zeroIsSuccess = 1,
         zeroOr255IsSuccess,
     }
-
+    
     @@public
     export interface UnsafeExecuteArguments {
         untrackedPaths?: (File | Directory)[];
--- a/Public/Src/Engine/Scheduler/Distribution/Worker.cs
+++ b/Public/Src/Engine/Scheduler/Distribution/Worker.cs
@@ -394,10 +394,11 @@ namespace BuildXL.Scheduler.Distribution
         }
 
         /// <summary>
-        /// Try acquire given resources on the worker
+        /// Try acquire given resources on the worker. This must be called from a thread-safe context to prevent race conditions.
         /// </summary>
         internal bool TryAcquire(RunnablePip runnablePip, out WorkerResource? limitingResource, double loadFactor = 1)
         {
+            Contract.Requires(runnablePip.PipType == PipType.Ipc || runnablePip.PipType == PipType.Process);
             Contract.Ensures(Contract.Result<bool>() == (limitingResource == null), "Must set a limiting resource when resources cannot be acquired");
 
             if (!IsAvailable)
@@ -421,54 +422,28 @@ namespace BuildXL.Scheduler.Distribution
                 loadFactor = 1;
             }
 
-            var processRunnablePip = (ProcessRunnablePip) runnablePip;
-            var acquiredSlots = m_acquiredProcessSlots;
-            var totalSlotsNeeded = acquiredSlots + processRunnablePip.Process.Weight;
-
+            var processRunnablePip = runnablePip as ProcessRunnablePip;
             // If a process has a weight higher than the total number of process slots, still allow it to run as long as there are no other
             // processes running (the number of acquired slots is 0)
-            bool processSlotsAcquired = false;
-            if (acquiredSlots == 0
-                && Interlocked.CompareExchange(ref m_acquiredProcessSlots, processRunnablePip.Process.Weight, 0) == 0)
+            if (AcquiredProcessSlots != 0 && AcquiredProcessSlots + processRunnablePip.Process.Weight > (EffectiveTotalProcessSlots * loadFactor))
             {
-                processSlotsAcquired = true;
-            }
-            else if (totalSlotsNeeded <= (EffectiveTotalProcessSlots * loadFactor)
-                && Interlocked.CompareExchange(ref m_acquiredProcessSlots, totalSlotsNeeded, acquiredSlots) == acquiredSlots)
-            {
-                processSlotsAcquired = true;
+                limitingResource = WorkerResource.AvailableProcessSlots;
+                return false;
             }
 
-            Contract.Assert(processRunnablePip != null);
             StringId limitingResourceName = StringId.Invalid;
-            if (processSlotsAcquired && processRunnablePip.TryAcquireResources(m_workerSemaphores, GetAdditionalResourceInfo(processRunnablePip), out limitingResourceName))
+            if (processRunnablePip.TryAcquireResources(m_workerSemaphores, GetAdditionalResourceInfo(processRunnablePip), out limitingResourceName))
             {
+                Interlocked.Add(ref m_acquiredProcessSlots, processRunnablePip.Process.Weight);
                 OnWorkerResourcesChanged(WorkerResource.AvailableProcessSlots, increased: false);
                 runnablePip.AcquiredResourceWorker = this;
                 limitingResource = null;
                 return true;
             }
 
-            // Failure cases
-            if (processSlotsAcquired)
-            {
-                // Release the acquired slots
-                Interlocked.Add(ref m_acquiredProcessSlots, -processRunnablePip.Process.Weight);
-
-                if (limitingResourceName == m_ramSemaphoreNameId)
-                {
-                    limitingResource = WorkerResource.AvailableMemoryMb;
-                }
-                else
-                {
-                    limitingResource = WorkerResource.CreateSemaphoreResource(limitingResourceName.ToString(runnablePip.Environment.Context.StringTable));
-                }
-            }
-            else
-            {
-                limitingResource = WorkerResource.AvailableProcessSlots;
-            }
-
+            limitingResource = limitingResourceName == m_ramSemaphoreNameId
+                ? WorkerResource.AvailableMemoryMb
+                : WorkerResource.CreateSemaphoreResource(limitingResourceName.ToString(runnablePip.Environment.Context.StringTable));
             return false;
         }
 
--- a/Public/Src/Engine/Scheduler/Distribution/Worker.cs
+++ b/Public/Src/Engine/Scheduler/Distribution/Worker.cs
@@ -421,27 +421,54 @@ namespace BuildXL.Scheduler.Distribution
                 loadFactor = 1;
             }
 
-            if (AcquiredProcessSlots >= (EffectiveTotalProcessSlots * loadFactor))
+            var processRunnablePip = runnablePip as ProcessRunnablePip;
+            var acquiredSlots = m_acquiredProcessSlots;
+            var totalSlotsNeeded = acquiredSlots + processRunnablePip.Process.Weight;
+
+            // If a process has a weight higher than the total number of process slots, still allow it to run as long as there are no other
+            // processes running (the number of acquired slots is 0)
+            bool processSlotsAcquired = false;
+            if (acquiredSlots == 0
+                && Interlocked.CompareExchange(ref m_acquiredProcessSlots, processRunnablePip.Process.Weight, 0) == 0)
             {
-                limitingResource = WorkerResource.AvailableProcessSlots;
-                return false;
+                processSlotsAcquired = true;
+            }
+            else if (totalSlotsNeeded <= (EffectiveTotalProcessSlots * loadFactor)
+                && Interlocked.CompareExchange(ref m_acquiredProcessSlots, totalSlotsNeeded, acquiredSlots) == acquiredSlots)
+            {
+                processSlotsAcquired = true;
             }
 
-            var processRunnablePip = runnablePip as ProcessRunnablePip;
+            Contract.Assert(processRunnablePip != null);
             StringId limitingResourceName = StringId.Invalid;
-            if (processRunnablePip != null &&
-                processRunnablePip.TryAcquireResources(m_workerSemaphores, GetAdditionalResourceInfo(processRunnablePip), out limitingResourceName))
+            if (processSlotsAcquired && processRunnablePip.TryAcquireResources(m_workerSemaphores, GetAdditionalResourceInfo(processRunnablePip), out limitingResourceName))
             {
-                Interlocked.Increment(ref m_acquiredProcessSlots);
                 OnWorkerResourcesChanged(WorkerResource.AvailableProcessSlots, increased: false);
                 runnablePip.AcquiredResourceWorker = this;
                 limitingResource = null;
                 return true;
             }
 
-            limitingResource = limitingResourceName == m_ramSemaphoreNameId
-                ? WorkerResource.AvailableMemoryMb
-                : WorkerResource.CreateSemaphoreResource(limitingResourceName.ToString(runnablePip.Environment.Context.StringTable));
+            // Failure cases
+            if (processSlotsAcquired)
+            {
+                // Release the acquired slots
+                Interlocked.Add(ref m_acquiredProcessSlots, -processRunnablePip.Process.Weight);
+
+                if (limitingResourceName == m_ramSemaphoreNameId)
+                {
+                    limitingResource = WorkerResource.AvailableMemoryMb;
+                }
+                else
+                {
+                    limitingResource = WorkerResource.CreateSemaphoreResource(limitingResourceName.ToString(runnablePip.Environment.Context.StringTable));
+                }
+            }
+            else
+            {
+                limitingResource = WorkerResource.AvailableProcessSlots;
+            }
+
             return false;
         }
 
@@ -510,7 +537,7 @@ namespace BuildXL.Scheduler.Distribution
             {
                 Contract.Assert(processRunnablePip.Resources.HasValue);
 
-                Interlocked.Decrement(ref m_acquiredProcessSlots);
+                Interlocked.Add(ref m_acquiredProcessSlots, -processRunnablePip.Process.Weight);
 
                 var resources = processRunnablePip.Resources.Value;
                 m_workerSemaphores.ReleaseResources(resources);
--- a/Public/Src/Engine/Scheduler/Distribution/Worker.cs
+++ b/Public/Src/Engine/Scheduler/Distribution/Worker.cs
@@ -421,7 +421,7 @@ namespace BuildXL.Scheduler.Distribution
                 loadFactor = 1;
             }
 
-            var processRunnablePip = runnablePip as ProcessRunnablePip;
+            var processRunnablePip = (ProcessRunnablePip) runnablePip;
             var acquiredSlots = m_acquiredProcessSlots;
             var totalSlotsNeeded = acquiredSlots + processRunnablePip.Process.Weight;
 
--- a/Public/Src/Engine/Scheduler/PipExecutor.cs
+++ b/Public/Src/Engine/Scheduler/PipExecutor.cs
@@ -2061,10 +2061,22 @@ namespace BuildXL.Scheduler
                     Contract.Assert(pipCacheMiss.Value.CacheMissType != PipCacheMissType.Invalid, "Must have valid cache miss reason");
                     environment.Counters.IncrementCounter((PipExecutorCounter)pipCacheMiss.Value.CacheMissType);
 
-                    Logger.Log.ScheduleProcessPipCacheMiss(
-                        operationContext,
-                        cacheableProcess.Description,
-                        runnableFromCacheResult.Fingerprint.ToString());
+                    if (processRunnable.Process.Weight > 1)
+                    {
+                        Logger.Log.ProcessPipCacheMissWithProcessWeight(
+                            operationContext,
+                            cacheableProcess.Description,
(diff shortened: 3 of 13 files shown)


mpysson:
Make sure to tag your commit message with the associated Azure Boards work item

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
```diff
@@ -126,20 +126,20 @@ namespace Transformer {
          * Default is globally controlled by the sandbox configuration
          * TODO: input isolation is not implemented
          */
-        containerIsolationLevel?: ContainerIsolationLevel;
+         containerIsolationLevel?: ContainerIsolationLevel;
```
mpysson: >  [](start = 8, length = 1)

nit - looks like convention in the file is 4 spaces per tab. You've got 9 here

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -329,7 +338,8 @@ public sealed partial class Process : Pip
             TimeSpan? nestedProcessTerminationTimeout = null,
             AbsentPathProbeInUndeclaredOpaquesMode absentPathProbeMode = AbsentPathProbeInUndeclaredOpaquesMode.Unsafe,
             DoubleWritePolicy doubleWritePolicy = DoubleWritePolicy.DoubleWritesAreErrors,
-            ContainerIsolationLevel containerIsolationLevel = ContainerIsolationLevel.None)
+            ContainerIsolationLevel containerIsolationLevel = ContainerIsolationLevel.None,
+            int weight = 1)
```
mpysson: > 1 [](start = 24, length = 2)

Does a negative weight have any meaning? If no I suggest adding a contract

Public/Src/FrontEnd/Script/Ambients/Transformers/AmbientTransformer.Process.cs
```diff
@@ -437,6 +439,13 @@ private void ProcessExecuteArguments(Context context, ObjectLiteral obj, Process
 
             // TODO: Regex. Should we follow ECMA, C#, JavaScript?
 
+            // Weight.
+            var weight = Converter.ExtractInt(obj, m_weight);
```
mpysson: >ExtractInt [](start = 35, length = 10)

Why isn't Converter.ExtractOptionalInt() necessary here?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ProcessWeightTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            ShouldLogSchedulerStats = true;
+        }
+
+        [Theory]
+        [InlineData(-1)] // Make sure a weights of < 1 are treated the same as a weight of 1
+        [InlineData(0)]
+        [InlineData(1)]
+        public void CanRunUpToMaxProcessesInParallel(int defaultWeight)
+        {
+            CreateAndScheduleProcessWithWeight(defaultWeight, numProcesses: Configuration.Schedule.MaxProcesses);
```
mpysson: >MaxProcesses [](start = 99, length = 12)

Does the test configuration set MaxProcesses or does it flow in based on the machine's core count? I'd suggest pinning in here in the class since the tests depend on its value

anniefu:
I went on the work item and added the GitHub PR as related work. I assume that is equivalent, but there's nothing in the GitHub UI to indicate it unfortunately.

---
In reply to: [478784432](https://github.com/Microsoft/BuildXL/pull/67#issuecomment-478784432) [](ancestors = 478784432)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
```diff
@@ -161,6 +161,14 @@ namespace Transformer {
          * for a pip. The pip will still be stored to the cache for sake of supporting distribution.
          */
         disableCacheLookup?: boolean;
+
+        /**
+         * The weighted value of this pip when limiting concurrency of process pips.
+         * The default weight for a process pip is 1, and the max valid weight is the max concurrent processes allowed in the build.
```
dannyvv: >s 1, [](start = 49, length = 4)

should we make the default weight 100?

That would allow for us to fold the 'lightweight' pips in this same concept?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ProcessWeightTests.cs
anniefu: All these tests do things relative to MaxProcesses value set, so it should be fine if the value fluctuates based off machine.

Manually setting the configuration does trump the machine's configuration and I have a test for that below with "TotalAllowedWeightIsLimitedByConfigurationMaxProcesses"



---
In reply to: [271088728](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271088728) [](ancestors = 271088728)

Public/Src/FrontEnd/Script/Ambients/Transformers/AmbientTransformer.Process.cs
anniefu: Oh, maybe that's why the builds are failing :)

---
In reply to: [271087913](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271087913) [](ancestors = 271087913)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
dannyvv: We can of course then for the 'slots' to reserve since that probably will have to remain whole numbers.... `Math.max (weight /100)`

---
In reply to: [271089886](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271089886) [](ancestors = 271089886)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
```diff
@@ -421,27 +421,54 @@ internal bool TryAcquire(RunnablePip runnablePip, out WorkerResource? limitingRe
...
+                    limitingResource = WorkerResource.AvailableMemoryMb;
+                }
+                else
+                {
+                    limitingResource = WorkerResource.CreateSemaphoreResource(limitingResourceName.ToString(runnablePip.Environment.Context.StringTable));
+                }
+            }
+            else
+            {
+                limitingResource = WorkerResource.AvailableProcessSlots;
```
mpysson: >limitingResource = WorkerResource.AvailableProcessSlots; [](start = 16, length = 56)

I'm thinking it would be good to get the number of process slots pips use into one of their log messages to help debug the "why aren't more pips running?" case. Sticking it in the pip description may be overkill. What do you think about putting it in the ProcessPipCacheMiss message?

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
mpysson: @semihokur to comment on that

Public/Src/Engine/Scheduler/Distribution/Worker.cs
```diff
@@ -421,27 +421,54 @@ internal bool TryAcquire(RunnablePip runnablePip, out WorkerResource? limitingRe
                 loadFactor = 1;
             }
 
-            if (AcquiredProcessSlots >= (EffectiveTotalProcessSlots * loadFactor))
+            var processRunnablePip = runnablePip as ProcessRunnablePip;
+            var acquiredSlots = m_acquiredProcessSlots;
+            var totalSlotsNeeded = acquiredSlots + processRunnablePip.Process.Weight;
```
dannyvv: This might be nicer abstracted in a method `getNrOfSlotsForPip(runnablePip)`

Public/Src/Pips/Dll/Operations/Process.cs
dannyvv: > = 1) [](start = 22, length = 5)

We have the default coded in too many places...

Can we make this a required field -or- nullable?

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -776,6 +790,7 @@ internal override void InternalSerialize(PipWriter writer)
             writer.Write((byte)ProcessAbsentPathProbeInUndeclaredOpaquesMode);
             writer.Write((byte)DoubleWritePolicy);
             writer.Write((byte)ContainerIsolationLevel);
+            writer.Write(Weight);
```
dannyvv: Since it seems the weight is slots, perhaps use `WriteCompact`?

dannyvv: and readCompact

---
In reply to: [271091377](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271091377) [](ancestors = 271091377)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ProcessWeightTests.cs
mpysson: It'd be a waste for these tests to spawn 80 child processes when run on powerful machines. I also worry about variability in the tests making them more error prone or harder to debug

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
anniefu: I don't know much about our lightweight pips feature, and it is relatively difficult to find in the codebase and is omitted from our wiki.

Do you have any pointers to a description/the implementation of lightweight?

---
In reply to: [271090152](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271090152) [](ancestors = 271090152,271089886)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
```diff
@@ -421,27 +421,54 @@ internal bool TryAcquire(RunnablePip runnablePip, out WorkerResource? limitingRe
                 loadFactor = 1;
             }
 
-            if (AcquiredProcessSlots >= (EffectiveTotalProcessSlots * loadFactor))
+            var processRunnablePip = runnablePip as ProcessRunnablePip;
```
dannyvv: > as  [](start = 48, length = 4)

It seems the code assumes it can only happen with a `ProcessRunnalbePip`, so please use a cast, so it fails with a CastException rather than a NullRefException later.

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
anniefu: Do you mean comment on what is the "max concurrent processes" allowed in the build?

The max valid value here is technically max int.

I can add a more detailed description here, but probably not so detailed as to actually explain what max concurrent processes is throughout the build, as that would be like full paragraph at least. 

---
In reply to: [271090887](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271090887) [](ancestors = 271090887)

narasamdya: >a process pip is 1 [](start = 34, length = 18)

What's the meaning of pip with weight 0?

Public/Src/Engine/Scheduler/Distribution/Worker.cs
anniefu: I can add a cast instead. 

There was actually already an implicit assumption that only processRunnablePips would get here due to the processRunnable != null check on what used to be line 432 and the Contract.Ensures at the top of the function.

---
In reply to: [271091922](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271091922) [](ancestors = 271091922)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
anniefu: - "weight" defaults to 1, converts values < 1 to 1, and considers values >= available process slots to mean the process should run alone

There are some tests to double check this behavior also.

---
In reply to: [271093105](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271093105) [](ancestors = 271093105)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
anniefu: Do you mean this message? 




---
In reply to: [271090384](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271090384) [](ancestors = 271090384)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
dannyvv: They are mostly used for Drop & IPC pips. They have their own lightweight queue in the schedule at the moment. I know we ahve an `IsLight` in the graph (on edges) as well as `IsLight` on the optoins., I'd need to read up on the specific implementation as well. @almilli: Could you chime in if it is worth reducing the concept count, or if they are really different concneps...

---
In reply to: [271091814](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271091814) [](ancestors = 271091814,271090152,271089886)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ProcessWeightTests.cs
anniefu: Fair, I can limit them to some constant.

---
In reply to: [271091494](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271091494) [](ancestors = 271091494)

Public/Src/Pips/Dll/Operations/Process.cs
anniefu: slots might be fn(#cores) which users wont know beforehand, so users might just want to set weight to some value like 9999999 to make sure a process runs solo.

---
In reply to: [271091402](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271091402) [](ancestors = 271091402,271091377)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
anniefu: I am going to keep it here purely because this area is race-condition prone and I'd rather have all the possible interacting variables as transparent as possible. 

(There is actually still a race condition between reading EffectiveTotalProcessSlots here and the CompareExchange with m_acquiredProcessSlots, but it seemed relatively niche and I didn't want to add a lock to prevent it).

---
In reply to: [271090931](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271090931) [](ancestors = 271090931)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
narasamdya: I was thinking that the default weight should be 0, and the allowed range would be [0, 1). Then the weight can be computed something like 1/(1-v). I also don't think that we need to mention that the max valid weight is the max concurrent process because one can just change that max. The weight should be relative so to say.

---
In reply to: [271093105](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271093105) [](ancestors = 271093105)

narasamdya: When thinking about weight I'm a bit accustomed with [0, 1). Or something more like what Danny's mentioned above. Perhaps @almilli can give some insight.

---
In reply to: [271096384](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271096384) [](ancestors = 271096384,271093105)

mpysson: I meant to tag semih regarding potentially folding this into the lightweight pips concept. My comment didn't get associated with the right chain

---
In reply to: [271092886](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271092886) [](ancestors = 271092886,271090887)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ProcessWeightTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            // Schedule one over max concurrent processes
+            CreateAndScheduleProcessWithWeight(defaultWeight, numProcesses: Configuration.Schedule.MaxProcesses + 1);
+
+            RunScheduler().AssertSuccess();
+            AssertProcessConcurrencyWeightLimited(true);
+        }
+
+        [Fact]
+        public void UseProcessWeightToRunAlone()
```
narasamdya: >UseProcessWeightToRunAlone [](start = 20, length = 26)

From jus the AssertProcessConcurrency...(...), I find it hard to distinguish this test from `CannotRunGreaterThanMaxProcessesInParallel`. Could we have a better assertion?

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
anniefu: How does one reason about the relative cost of a pip with a weight of 0.2 compared to a pip with a weight of 0?

The integer multiplier here is closely aligned with CloudBuild/QBuild/QTest weight concepts according to:
[https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2F1eswiki.com%2Fwiki%2FCBFAQ%23I_am_getting_ExceededMaxBuildTime_exception_while_running_buddy_build.2C_What_should_I_do_to_resolve_this_issue.3F&data=02%7C01%7CAnnie.Fu%40microsoft.com%7Cccbc5e5d4cb348c384ed08d6b22f7891%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636892314363790059&sdata=lTJrSIMbQ%2BSkva%2B7U%2Ba8RnDnxhRpi5zZB96LF8Dv0AU%3D&reserved=0](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2F1eswiki.com%2Fwiki%2FCBFAQ%23I_am_getting_ExceededMaxBuildTime_exception_while_running_buddy_build.2C_What_should_I_do_to_resolve_this_issue.3F&data=02%7C01%7CAnnie.Fu%40microsoft.com%7Cccbc5e5d4cb348c384ed08d6b22f7891%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636892314363790059&sdata=lTJrSIMbQ%2BSkva%2B7U%2Ba8RnDnxhRpi5zZB96LF8Dv0AU%3D&reserved=0) 


---
In reply to: [271097510](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271097510) [](ancestors = 271097510,271096384,271093105)

anniefu: Perhaps the value of 0 could be used to indicate lightweight processes. 

---
In reply to: [271094619](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271094619) [](ancestors = 271094619,271091814,271090152,271089886)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ProcessWeightTests.cs
anniefu: Can you suggest an assertion you prefer?

This test checks that I can specify a "solo" pip by using the weight feature. 

CannotRunGreaterThanMaxProcessesInParallel is a sanity check that we have a limit to the amount of processes you can run in paralle.

---
In reply to: [271098853](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271098853) [](ancestors = 271098853)

dannyvv:
@almiliMSFT is added to the review. #Closed

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
almiliMSFT: I don't think we still use lightweight pips anywhere.

They were originally introduced before we had IPC pips, and when all our drop pips (those are the pips that talk to DropDaemon) were running as regular processes.  Since those pips just sit around and do nothing while waiting to hear back from DropDaemon, we wanted to able to schedule a whole bunch of them at a time.

Now all drop pips run as IPC pips, which get their own queue and their own default limits/priorities, which makes them completely independent of the settings used for process pips.

Danny's idea makes perfect sense to me.

Public/Src/Engine/Scheduler/Distribution/Worker.cs
dannyvv: np.

---
In reply to: [271096310](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271096310) [](ancestors = 271096310,271090931)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
almiliMSFT: DScript doesn't support floating-point numbers, so, practically speaking, we should stick with integers here.

Public/Src/Engine/Scheduler/Distribution/Worker.cs
almiliMSFT: In that case just add a `Contract.Requires` to this method

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
mpysson: I think it's actually more complicated than that. Semih is tracking an issue where a concurrency change to the light pips introduced a perf regression. Please sync up with him before modifying the meaning of lightweight pips.

---
In reply to: [271102594](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271102594) [](ancestors = 271102594)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
```diff
@@ -421,27 +421,54 @@ internal bool TryAcquire(RunnablePip runnablePip, out WorkerResource? limitingRe
                 loadFactor = 1;
             }
 
-            if (AcquiredProcessSlots >= (EffectiveTotalProcessSlots * loadFactor))
+            var processRunnablePip = runnablePip as ProcessRunnablePip;
+            var acquiredSlots = m_acquiredProcessSlots;
```
almiliMSFT: Why don't use use the `AcquiredProcessSlots` property instead of reading the fields directly?  The property will read it using `Volatile.Read`, which might in some cases give you a more up-to-date value

Public/Src/Engine/Scheduler/Distribution/Worker.cs
anniefu: Yeah I actually added a Contract.Assert down below, I just forgot about it.

---
In reply to: [271104950](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271104950) [](ancestors = 271104950)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
dannyvv: That might work. Note I don't want to hold the PR on this. Just trying to go through the thought process to avoid multiple concepts trying to do the same thing :)

---
In reply to: [271100050](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271100050) [](ancestors = 271100050,271094619,271091814,271090152,271089886)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
anniefu: Added a contract requirement instead.

---
In reply to: [271093475](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271093475) [](ancestors = 271093475,271091922)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
anniefu: Ended up going back to the old format because it turns out all of this function is being protected by a mutex on the calling side.

---
In reply to: [271105791](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271105791) [](ancestors = 271105791)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
anniefu: After offline discussion, we are going to keep the semantics in-line with quickbuild, qtest, and cloudbuild and have it be an integer that defaults to 1.

---
In reply to: [271099495](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271099495) [](ancestors = 271099495,271097510,271096384,271093105)

anniefu: I am not going to account for lightweight pips in this PR, but I think the scheme of having weight represent # of process slots a pip desires, and having it default to 1 should leave us enough flexibility to do lightweight pips in the future.

We could do some negative scheme like Aleks suggested in his other comment, or just use the value of 0 as the actual value of 0, which means there will always be enough process slots to run the pip, since available process slots is never < 1. 

---
In reply to: [271470356](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271470356) [](ancestors = 271470356,271100050,271094619,271091814,271090152,271089886)

anniefu: I don't think we differentiate the "lightness" of lightweight pips right now, so I would say actually just using 0 as the value for lightweight pips might be the way to go.

We could eliminate the lightweight queue altogether and just schedule pips with a weight of 0, meaning those pips need 0 process slots and will always run unless restricted by some additional semaphore.

That being said, I am going to defer handling lightweight pips to outside of this PR.

---
In reply to: [271103676](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271103676) [](ancestors = 271103676)

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -2061,10 +2061,22 @@ private static void ReportFileAccesses(ExecutionResult processExecutionResult, F
...
-                        operationContext,
-                        cacheableProcess.Description,
-                        runnableFromCacheResult.Fingerprint.ToString());
+                    if (processRunnable.Process.Weight > 1)
+                    {
+                        Logger.Log.ProcessPipCacheMissWithProcessWeight(
+                            operationContext,
+                            cacheableProcess.Description,
+                            runnableFromCacheResult.Fingerprint.ToString(),
+                            processRunnable.Process.Weight.ToString());
```
semihokur: I do not like logging this when doing the cache lookup. in distributedbuilds, we use different workers for cache lookup and execute. I'd like to see the weight of the process in the worker where we execute the pip

anniefu: Do you know of a good execution time message for me to add to? Otherwise I can just add a brand new one, but it seemed wasteful.

I was also thinking of adding a field to critical path logging, only if process weight is > 1

---
In reply to: [271480547](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271480547) [](ancestors = 271480547)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
semihokur: If the question is whether light pips should be integrated into the weight concept, I'd like to keep them separate for now until Office gets rid of light esrp pips. 

---
In reply to: [271098357](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271098357) [](ancestors = 271098357,271092886,271090887)

Public/Src/Engine/Scheduler/PipExecutor.cs
anniefu: Ok I made a new log message that logs at executeprocess time. Decided not to append to critical path info for now.

---
In reply to: [271481873](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271481873) [](ancestors = 271481873,271480547)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
semihokur: We should first wait for Office to stop using light process pips for wrong purposes. It is good to separate them for now. 

---
In reply to: [271105044](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271105044) [](ancestors = 271105044,271102594)

Public/Src/Engine/Scheduler/PipExecutor.cs
semihokur: sounds good!

---
In reply to: [271481873](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271481873) [](ancestors = 271481873,271480547)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ProcessWeightTests.cs
anniefu: Resolving after offline discussion.

---
In reply to: [271100350](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271100350) [](ancestors = 271100350,271098853)

Public/Src/Pips/Dll/Builders/ProcessBuilder.cs
```diff
@@ -103,6 +103,12 @@ public class ProcessBuilder : IDisposable
 
         private Dictionary<StringId, ProcessSemaphoreInfo> m_semaphores;
 
+        /// <summary>
+        /// The weighted value of this pip when limiting concurrency of process pips.
+        /// The higher the weight, the fewer process pips that can run in parallel.
+        /// </summary>
+        public int? Weight { get; set; } = null;
```
ljcollins25: >Weight [](start = 20, length = 6)

Should we error if this is negative or zero somewhere? I wonder if we should allow fractional values to represent lighter pips.

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
dannyvv: Thanks @AnnieFu & @Seokur Sounds good to me.

---
In reply to: [271491641](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271491641) [](ancestors = 271491641,271098357,271092886,271090887)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
mpysson: yes. i see you added a separate message. That's probably fine too as long as this doesn't get widely used. We may want to revisit consolidating it into an existing message to save space if it is used at lot.

---
In reply to: [271094601](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271094601) [](ancestors = 271094601,271090384)

Public/Src/Pips/Dll/Builders/ProcessBuilder.cs
anniefu: We decided to just coerce < 1 to be 1. Because these represent process slots, I don't think anything < 1 is conceptually valid. 

Aleks said Dominoscript doesn't support fractional values, so it'd have to just be larger ints. 

We can use 0 in the future to indicate lightweight pips/pips that should just run regardless of whatever else is happening.


---
In reply to: [271499419](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271499419) [](ancestors = 271499419)

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -263,6 +263,15 @@ internal static string PipWriteFileFailedMessage(string pipDescription, string p
...
         internal abstract void ScheduleProcessPipCacheMiss(LoggingContext loggingContext, string pipDescription, string fingerprint);
 
+        [GeneratedEvent(
+            (ushort)EventId.ProcessPipProcessWeight,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Verbose,
+            Keywords = (int)Events.Keywords.UserMessage,
+            EventTask = (ushort)Events.Tasks.Storage,
+            Message = "[{pipDescription}] Executing process with process weight: {weight}.")]
+        internal abstract void ProcessPipProcessWeight(LoggingContext loggingContext, string pipDescription, int weight);
```
ljcollins25: >ProcessPipProcessWeight [](start = 31, length = 23)

Do we not have an executing process message already that this can be added to?

anniefu: Actually, no not really. Some messages related to preparing to execute a pip and some messages related to post-execution, but nothing for just executing.

---
In reply to: [271528643](https://github.com/Microsoft/BuildXL/pull/67#discussion_r271528643) [](ancestors = 271528643)

PR #68, opened 2019-04-01
Hide exception on console when concurrent invocations overlap output directories

mpysson:
Exception stack is moved to the log file for sake of debugging. Generally it's sufficient to just tell users they are invoking two concurrent builds with overlapping output directories. 

Fixes [AB#1469779](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1469779)

### DIFF
--- a/Public/Src/Engine/Dll/Tracing/Log.cs
+++ b/Public/Src/Engine/Dll/Tracing/Log.cs
@@ -1399,8 +1399,8 @@ If you can't update and need this feature after July 2018 please reach out to th
             EventLevel = Level.Error,
             Keywords = (int)(Events.Keywords.UserMessage | Events.Keywords.InfrastructureError),
             EventTask = (int)Events.Tasks.Engine,
-            Message = "Failed to acquire a lock to prevent directory deletion: {0}")]
-        public abstract void FailedToAcquireDirectoryDeletionLock(LoggingContext context, string innerException);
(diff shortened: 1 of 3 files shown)


Public/Src/Engine/Dll/Tracing/Log.cs
```diff
@@ -1703,14 +1703,29 @@ public abstract partial class Logger
             Message = "{ShortProductName} failed to run because only 'case-insensitive' file-systems are currently supported on non-windows hosts.")]
         public abstract void ErrorCaseSensitiveFileSystemDetected(LoggingContext context);
 
+        public void BusyOrUnavailableOutputDirectories(LoggingContext context, string objectDirectoryPath, string exception)
```
dannyvv: >BusyOrUnavailableOutputDirectories [](start = 20, length = 34)

Didn't see any callers added to this method...

mpysson: It's a bit of a trick. This new method has the same method signature of the old one. So callers that were previously logging the one message are now logging the shorter one on the console and then longer one for the log file.

---
In reply to: [271098653](https://github.com/Microsoft/BuildXL/pull/68#discussion_r271098653) [](ancestors = 271098653)

PR #70, opened 2019-04-02
[drop] Add handling of AbsentFileHash files

olkononenko:
Currently drop does not play nice with missing files. However, it possible for customers to pass legitimately missing files into drop api (e.g., temp files under shared opaques). The PR makes the following changes:
1) Change drop daemon logic to handle files with AbsentFile hash:
  - if it's a source file -> fail with a user error (so we would not trigger build re-run in CB)
  - if it's an output file -> just skip it
2) Missing files that do not have a special hash will result in an internal error. 
3) Some refactoring around addArtifactToDrop method (now we consistently log start/end messages)

[AB#1474273](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1474273)
[AB#1485849](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1485849)


narasamdya:
I think it's better to also fail on dropping non-existent output file. Otherwise, users may think that the drop is OK, but in fact some of the expected output files were not dropped.

Public/Src/Tools/UnitTests/DropDaemon/DropOperationTests.cs
```diff
@@ -393,6 +393,39 @@ public void TestAddDirectoryToDropWithFilters(string filter)
...
+           
+            var ipcProvider = IpcFactory.GetProvider();
+            var bxlApiClient = new Client(ipcProvider.GetClient(ipcProvider.CreateNewConnectionString(), new ClientConfig()));
+
+            WithSetup(
+                dropClient, 
+                (daemon, etwListener) =>
+                {
+                    // only hash and file rewrite count are important here; the rest are just fake values
+                    var addArtifactsCommand = Program.ParseArgs($"addartifacts --ipcServerMoniker {daemon.Config.Moniker} --file non-existent-file.txt --dropPath remote-file-name.txt --hash VSO0:010000000000000000000000000000000000000000000000000000000000000000_-1 --fileId 12345:{(isSourceFile ? 0 : 1)}", new UnixParser());
```
narasamdya: >VSO0:010000000000000000000000000000000000000000000000000000000000000000_-1 [](start = 190, length = 74)

Can we use WellKnownHash toString?

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -145,6 +145,20 @@ internal static string PipWriteFileFailedMessage(string pipDescription, string p
             string reason,
             string message);
 
+        [GeneratedEvent(
+            (ushort)EventId.PipIpcFailedDueToInvalidInput,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Error,
+            Keywords = (int)(Events.Keywords.UserMessage | Events.Keywords.UserError),
+            EventTask = (ushort)Events.Tasks.PipExecutor,
+            Message = "IPC operation '{operation}' could not be executed via IPC moniker '{moniker}'.  Reason: {reason}. Error: {message}")]
```
almiliMSFT: This user message is exactly the same as for `PipIpcFailed`.  To reduce potential confusion, I would change this message to include something about "InvalidInput"

almiliMSFT:
> I think it's better to also fail on dropping non-existent output file. Otherwise, users may think that the drop is OK, but in fact some of the expected output files were not dropped. Happy to discuss further (waiting on this)

I don't have all the context, but at first look it does seem sketchy to ignore absent files.  How can this happen anyway?  I though that the only way for this to happen is when a user adds some (possibly shared) opaque directory and BuildXL in the list of files for that directory has some files that are actually missing; if that's the case, maybe this logic of filtering those files out should be moved to BuildXL, so that when DropDaemon asks for files belonging to a directory, BuildXL skips those that are missing.

Public/Src/Tools/DropDaemon/Program.cs
```diff
@@ -518,77 +519,103 @@ private static void SetupThreadPoolAndServicePoint()
...
+            (Regex[] initializedFilters, string filterInitError) = InitializeDirectoryFilters(directoryFilters);
+            if (filterInitError != null)
+            {
+                return new IpcResult(IpcResultStatus.ExecutionError, filterInitError);
+            }
 
-                daemon.Logger.Verbose("[ADDARTIFACTS] " + result);
-                return result;
-            });
+            var dropFileItems = Enumerable
```
ljcollins25: >dropFileItems [](start = 16, length = 13)

Rename: dropFileItemsKeyedByIsAbsent

Public/Src/Tools/UnitTests/DropDaemon/DropOperationTests.cs
```diff
@@ -393,6 +393,101 @@ public void TestAddDirectoryToDropWithFilters(string filter)
...
+           
+            var ipcProvider = IpcFactory.GetProvider();
+            var bxlApiClient = new Client(ipcProvider.GetClient(ipcProvider.CreateNewConnectionString(), new ClientConfig()));
+
+            WithSetup(
+                dropClient, 
+                (daemon, etwListener) =>
+                {
+                    // only hash and file rewrite count are important here; the rest are just fake values
+                    var hash = FileContentInfo.CreateWithUnknownLength(ContentHashingUtilities.CreateSpecialValue(1)).Render();
```
narasamdya: >ContentHashingUtilities.CreateSpecialValue(1) [](start = 71, length = 45)

Is WellKnownHash not accessible? Can it be made accessible?

Public/Src/Engine/Scheduler/WellKnownContentHashes.cs
```diff
@@ -41,4 +41,22 @@ public static class WellKnownContentHashes
...
+    /// <summary>
+    /// Methods for handling constants from WellKnownContentHashes. 
+    /// </summary>
+    public static class WellKnownContentHashUtilities
+    {
+        /// <summary>
+        /// Checks whether <paramref name="hash"/> is an AbsentFile hash. Note: this method does a hash algorithm agnostic check 
+        /// </summary>
+        /// <remarks>
+        /// ContentHash contains not only the hash itself, but also the also the algorithm used to create the hash.
```
narasamdya: >but also [](start = 59, length = 8)

typo

Public/Src/Tools/DropDaemon/Program.cs
```diff
@@ -518,77 +519,103 @@ private static void SetupThreadPoolAndServicePoint()
...
+                initializedFilters);
+
+            if (error != null)
+            {
+                return new IpcResult(IpcResultStatus.ExecutionError, error);
+            }
+
+            var groupedDirectoriesContent = dropDirectoryMemberItems.ToLookup(f => WellKnownContentHashUtilities.IsAbsentFileHash(f.Hash));
+
+            // we allow missing files inside of directories only if those files are output files (e.g., optional or temporary files)
```
narasamdya: >we allow missing files inside of directories only if those files are output files [](start = 15, length = 81)

As far as I understand from yesterday's discussion,  we only want to do this only for opaque (or shared opaque) directories. For partially or fully sealed directories where users explicitly specify the members of those directories, inclusion of absent files should trigger errors.

narasamdya: I'm going to approve this PR. But I still think that the work is incomplete. For partially sealed and fully sealed directories, where users can specify the files explicitly, drop should fail if those files are absent.

It's up to you to complete it or not. Also, I think adding a server API to know the seal directory kind is not difficult.

Public/Src/Tools/UnitTests/DropDaemon/DropOperationTests.cs
olkononenko: addressed

PR #71, opened 2019-04-02
Update CacheFactory to ignore invalid JSON fields in config

JuanCarlosGI:
A problem arose when an old version of cache received a config file from a new version of GBR which emitted a config flag that didn't exist in the old cache bits.

To fix this, we will now ignore invalid fields in the cache config.

### DIFF
--- a/Public/Src/Cache/VerticalStore/Interfaces/CacheFactory.cs
+++ b/Public/Src/Cache/VerticalStore/Interfaces/CacheFactory.cs
@@ -290,6 +290,8 @@ namespace BuildXL.Cache.Interfaces
                 }
             }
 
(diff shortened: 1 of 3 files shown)


Public/Src/Cache/VerticalStore/Interfaces/CacheFactory.cs
```diff
@@ -290,24 +290,6 @@ public static string Serialize(this ICacheConfigData cacheData)
                 }
             }
 
-            foreach (string key in cacheData.Keys)
-            {
```
dcwuser: Add comment saying that we used to validate that every supplied key had a corresponding defined property, but that broke scenarios where new config was rolled out incrementally, so we removed it. Comments like this help future maintainers.

PR #73, opened 2019-04-02
Fix mismatch in the reporting context name between Detours and sandboxed process reports

narasamdya:
Fix mismatch in the reporting context name between Detours and sandboxed process reports.

This bug causes sandboxed process reports to report unknown operations. The mismatch happens when the operation is ZwSetFileInformation with Dispose class, which is used for file deletion.

Some clean-up is also performed in this area.


almiliMSFT:
Maybe shorten the title before merging so that it doesn't get truncated

PR #76, opened 2019-04-02
Retry TryCreateHandleToFile only on timeout or sharing violation

almiliMSFT:
[AB#1495708](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1495708)

### DIFF
--- a/Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
+++ b/Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
@@ -382,7 +382,6 @@ namespace BuildXL.Engine.Cache.Artifacts
             using (Counters.StartStopwatch(LocalDiskContentStoreCounter.TryDiscoverTime_OpenProbeHandle))
             {
                 OpenFileResult openResult = default;
-
                 Helpers.RetryOnFailure(
                     lastAttempt =>
                     {
@@ -400,7 +399,8 @@ namespace BuildXL.Engine.Cache.Artifacts
                         }
 
                         return openResult.Succeeded;
-                    });
+                    },
+                    numberOfAttempts: OperatingSystemHelper.IsUnixOS ? 1 : 3);
 
                 if (!openResult.Succeeded)
                 {
@@ -477,8 +477,7 @@ namespace BuildXL.Engine.Cache.Artifacts
                     Contract.Assert(handle != null && !handle.IsInvalid);
 
                     // We have seen few cases where closing of the stream throws an ERROR_INCORRECT_FUNCTION native error.
-                    // This happened 4 times in office meta build and could be a result of opening a non-file stream while coding a BuildXL Spec build
-                    // file. Or a failing stable drive.
+                    // This could be a result of opening a non-file stream while coding a BuildXL Spec build file.
                     //
                     // Either way, catch the exception and make it an recoverable one.
                     try
--- a/Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
+++ b/Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
@@ -393,14 +393,17 @@ namespace BuildXL.Engine.Cache.Artifacts
                             openFlags,
                             out handle);
 
-                        if (!openResult.Succeeded && openResult.Status == OpenFileStatus.Timeout)
+                        if (openResult.Status == OpenFileStatus.Timeout ||
+                            openResult.Status == OpenFileStatus.SharingViolation)
                         {
                             Tracing.Logger.Log.TimeoutOpeningFileForHashing(m_loggingContext, expandedPath);
+                            return false; // returning "not done", i.e., do retry
                         }
-
-                        return openResult.Succeeded;
-                    },
-                    numberOfAttempts: OperatingSystemHelper.IsUnixOS ? 1 : 3);
+                        else
+                        {
+                            return true; // returning "done", i.e., don't retry
+                        }
+                    });
 
                 if (!openResult.Succeeded)
                 {
--- a/Public/Src/Utilities/Native/IO/Unix/FileSystem.Unix.cs
+++ b/Public/Src/Utilities/Native/IO/Unix/FileSystem.Unix.cs
@@ -816,7 +816,7 @@ namespace BuildXL.Native.IO.Unix
         {
             try
             {
-                FileAttributes attributes = TryGetFileAttributes(path);
+                FileAttributes attributes = File.GetAttributes(path);
 
                 if ((attributes & FileAttributes.ReparsePoint) == 0)
                 {


Public/Src/Utilities/Native/IO/Unix/FileSystem.Unix.cs
```diff
@@ -816,7 +816,7 @@ private static Possible<ReparsePointType> GetReparsePointType(string path)
         {
             try
             {
-                FileAttributes attributes = TryGetFileAttributes(path);
+                FileAttributes attributes = File.GetAttributes(path);
```
almiliMSFT: unrelated change: no need to call `TryGetFileAttributes` which catches/rethrows exceptions since the same is already done here.

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
```diff
@@ -400,7 +399,8 @@ private Possible<Unit> CreateSymlinkIfNotExistsOrTargetMismatch(AbsolutePath sym
                         }
 
                         return openResult.Succeeded;
-                    });
+                    },
+                    numberOfAttempts: OperatingSystemHelper.IsUnixOS ? 1 : 3);
```
almiliMSFT: I don't know what's the rational for retrying this operation.  

For example, here's a possible (and realistic) call stack up to this point:
```
at BuildXL.Engine.Cache.Artifacts.LocalDiskContentStore.TryCreateHandleToFile(String expandedPath, Boolean forUsnQuery, Boolean createHandleWithSequentialScan) in \.\Public\Src\Engine\Cache\Artifacts\LocalDiskContentStore.cs:line 393
   at BuildXL.Engine.Cache.Artifacts.LocalDiskContentStore.TryDiscoverAsync(FileArtifact fileArtifact, ExpandedAbsolutePath path, Boolean ignoreKnownContentHash, Boolean createHandleWithSequentialScan) in \.\Public\Src\Engine\Cache\Artifacts\LocalDiskContentStore.cs:line 467
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at BuildXL.Engine.Cache.Artifacts.LocalDiskContentStore.TryDiscoverAsync(FileArtifact fileArtifact, ExpandedAbsolutePath path, Boolean ignoreKnownContentHash, Boolean createHandleWithSequentialScan)
   at BuildXL.Scheduler.Artifacts.FileContentManager.TryLoadAvailableContentAsync(OperationContext operationContext, PipInfo pipInfo, Boolean materializingOutputs, IReadOnlyList`1 filesAndContentHashes, Action`1 onFailure, Action`2 onContentUnavailable, Boolean onlyLogUnavailableContent, PipArtifactsState state) in \.\Public\Src\Engine\Scheduler\Artifacts\FileContentManager.cs:line 2291
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at BuildXL.Scheduler.Artifacts.FileContentManager.TryLoadAvailableContentAsync(OperationContext operationContext, PipInfo pipInfo, Boolean materializingOutputs, IReadOnlyList`1 filesAndContentHashes, Action`1 onFailure, Action`2 onContentUnavailable, Boolean onlyLogUnavailableContent, PipArtifactsState state)
   at BuildXL.Scheduler.Artifacts.FileContentManager.TryLoadAvailableContentAsync(OperationContext operationContext, PipInfo pipInfo, PipArtifactsState state) in \.\Public\Src\Engine\Scheduler\Artifacts\FileContentManager.cs:line 2173
   at BuildXL.Scheduler.Artifacts.FileContentManager.PlaceFilesAsync(OperationContext operationContext, PipInfo pipInfo, PipArtifactsState state, Boolean materialize) in \.\Public\Src\Engine\Scheduler\Artifacts\FileContentManager.cs:line 1958
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at BuildXL.Scheduler.Artifacts.FileContentManager.PlaceFilesAsync(OperationContext operationContext, PipInfo pipInfo, PipArtifactsState state, Boolean materialize)
   at BuildXL.Scheduler.Artifacts.FileContentManager.TryMaterializeArtifactsCore(PipInfo pipInfo, OperationContext operationContext, PipArtifactsState state, Boolean materializatingOutputs, Boolean isDeclaredProducer) in \.\Public\Src\Engine\Scheduler\Artifacts\FileContentManager.cs:line 1548
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.MoveNext()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
``` 
If you look at `FileContentManager.TryLoadAvailableContentAsync`, you can see that it's perfectly fine if `LocalDiskContentStore.TryDiscoverAsync` doesn't succeed (e.g., because the file in question doesn't exist), in which case retrying to open file handle here become a significant waste of time.

mpysson:
>                         openResult = FileUtilities.TryCreateOrOpenFile(

The condition for backoff & retry used to be this, based on the open result:
if (!openResult.Succeeded && openResult.Status == OpenFileStatus.Timeout)

A recent change (312d51d6) made it retry for all failures. Based on the commit description I think the intent was to only retry for sharing violations.

---
Refers to: Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs:388 in 1805bef. [](commit_id = 1805bef47374ed07ad98acbeb4fda8a76130fc4f, deletion_comment = False)

mpysson:
>                         openResult = FileUtilities.TryCreateOrOpenFile(

I think this is going to be a problem on Windows as well. We should make a fix that addresses both operating systems.
@narasamdya

---
In reply to: [479251730](https://github.com/Microsoft/BuildXL/pull/76#issuecomment-479251730) [](ancestors = 479251730)

---
Refers to: Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs:388 in 1805bef. [](commit_id = 1805bef47374ed07ad98acbeb4fda8a76130fc4f, deletion_comment = False)

almiliMSFT:
> I think this is going to be a problem on Windows as well. We should make a fix that addresses both operating systems.

Indeed.  I'll add a new parameter to `LocalDiskContentStore.TryCreateHandleToFile` so that the callers can decide whether this should be retried or not; the default should be "no retrying" and then Iman can tell us which caller should invoke this with retrying set to true

almiliMSFT:
zsss

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
almiliMSFT: > I think this is going to be a problem on Windows as well. We should make a fix that addresses both operating systems.

Indeed.  I'll add a new parameter to `LocalDiskContentStore.TryCreateHandleToFile` so that the callers can decide whether this should be retried or not; the default should be "no retrying" and then Iman can tell us which caller should invoke this with retrying set to true

narasamdya:
Does retry cause catastrophic failure on Mac? Or does it just cause perf. regression?

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
narasamdya: The retry was necessary in the scenario where we have to hash hardlinks to the same object file. One thread can try to flush the page cache, while the other can try to open a handle to that object file for hashing. Failure in hashing a file is considered internal error, and thus we need to retry.

I think the retry should only be done if there's sharingviolation or timeout. Perhaps I was generalizing the code too much.


---
In reply to: [271538549](https://github.com/Microsoft/BuildXL/pull/76#discussion_r271538549) [](ancestors = 271538549)

almiliMSFT: changed it to retry only on sharing violation or timeout

PR #80, opened 2019-04-04
Add support for building MSBuild projects in isolation

smera:
Add support so the MSBuild resolver can build projects in isolation, where project references and projects become first class citizens and project boundaries are enforced.
This becomes the default mode, with an option to fallback to the legacy mode, based on /buildProjectReferences=false.

Note: this will temporarily break Orchard Core validation due to missing implicit project references. I'll fix that once this PR gets in.

### DIFF
--- a/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
+++ b/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
@@ -33,7 +33,7 @@ namespace BuildXL.FrontEnd.MsBuild
     internal sealed class PipConstructor
     {
         private readonly FrontEndContext m_context;
-        private readonly ConcurrentDictionary<ProjectWithPredictions, ProcessOutputs> m_processOutputsPerProject = new ConcurrentDictionary<ProjectWithPredictions, ProcessOutputs>();
+        private readonly ConcurrentDictionary<ProjectWithPredictions, MSBuildProjectOutputs> m_processOutputsPerProject = new ConcurrentDictionary<ProjectWithPredictions, MSBuildProjectOutputs>();
         
         // Only used if resolverSettings.EnableTransitiveProjectReferences = true
         private readonly ConcurrentBigMap<ProjectWithPredictions, IReadOnlySet<ProjectWithPredictions>> m_transitiveDependenciesPerProject = new ConcurrentBigMap<ProjectWithPredictions, IReadOnlySet<ProjectWithPredictions>>();
@@ -46,7 +46,6 @@ namespace BuildXL.FrontEnd.MsBuild
         private static readonly string[] s_commonArgumentsToMsBuildExe =
         {
             "/NoLogo",
-            "/p:buildprojectreferences=false", // TODO: We wouldn't need this when MSBuild starts supporting a 'build just this project' mode
             "/p:TrackFileAccess=false", // Turns off MSBuild's Detours based tracking. Required always to prevent Detouring a Detour which is not supported
             "/p:UseSharedCompilation=false", //  Turn off new MSBuild flag to reuse VBCSCompiler.exe (a feature of Roslyn csc compiler) to compile C# files from different projects
             "/m:1", // Tells MsBuild not to create child processes for building, but instead to simply execute the specified project file within a single process
@@ -61,6 +60,14 @@ namespace BuildXL.FrontEnd.MsBuild
 
         private PathTable PathTable => m_context.PathTable;
         private FrontEndEngineAbstraction Engine => m_frontEndHost.Engine;
+
+        /// <summary>
+        /// The name of the output cache file of the given pip, if the corresponding process is built in isolation
+        /// </summary>
+        /// <remarks>
+        /// This file is created in an object directory unique to this pip, so there is no need to make this name unique
+        /// </remarks>
+        internal const string OutputCacheFileName = "output.cache";
 
         /// <nodoc/>
         public PipConstructor(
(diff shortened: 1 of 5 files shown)


Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -271,16 +297,32 @@ public bool TrySchedulePipForFile(ProjectWithPredictions project, QualifierId qu
...
                 }
+
+                // If the dependency was built in isolation, this project needs to access the generated cache files
+                if (projectOutputs.BuildsInIsolation)
+                {
+                    var outputCache = projectOutputs.OutputCacheFile;
+                    processBuilder.AddInputFile(outputCache);
+                    using (argumentsBuilder.StartFragment(PipDataFragmentEscaping.NoEscaping, string.Empty))
+                    {
+                        argumentsBuilder.Add(PipDataAtom.FromString("/irc:"));
```
erikma: /irc is an MSBuild param? Should add a comment what it is for, maybe link to a doc page or the doc comments in code to make it easier to cross-reference

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -527,6 +580,21 @@ private bool TryAddMsBuildArguments(ProjectWithPredictions<AbsolutePath> project
                     project.FullPath.GetName(m_context.PathTable).ToString(m_context.StringTable));
             }
 
+            // Pass the output result cache file if present
+            if (outputResultCacheFile != AbsolutePath.Invalid)
+            {
+                using (pipDataBuilder.StartFragment(PipDataFragmentEscaping.NoEscaping, string.Empty))
+                {
+                    pipDataBuilder.Add(PipDataAtom.FromString("/orc:"));
```
erikma: same - document /orc

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
dannyvv: Perhaps it would be also clearer since it is programatically generated to use the fully expanded commandline argument

---
In reply to: [271992530](https://github.com/Microsoft/BuildXL/pull/80#discussion_r271992530) [](ancestors = 271992530)

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
cdmihai: Or use long name.

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
smera: Yes, brand new! :) This is part of the new 'build in isolation' mode, available in the latest MSBuild. Not sure if there are docs yet, @cdmihai ? Happy to link to that, otherwise I can just add a regular comment, summarizing what it does

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
smera: I know it probably doesn't move the perf needle, but I was intendedly using the short version to not unnecessarily bloat the response file (for a big repo, we may have many of these entries).
If this is about documenting this better, I'd prefer to just add a comment explaining what this is 

---
In reply to: [272271223](https://github.com/Microsoft/BuildXL/pull/80#discussion_r272271223) [](ancestors = 272271223)

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
cdmihai: Yup `MSBuild /?` has documentation on them. There's also the build in isolation section in the spec: https://github.com/Microsoft/msbuild/blob/master/documentation/specs/static-graph.md#isolated-builds

cdmihai: But I suggest you just use the long name of the parameters

Public/Src/FrontEnd/UnitTests/MsBuild/SchedulingTests/MsBuildProcessBuilderTests.cs
```diff
@@ -225,5 +225,68 @@ public void RunInContainerFlagSetsTheRightProcessFlag()
...
+
+            var arguments = RetrieveProcessArguments(testProj);
+
+            // A project that is not built in isolation shouldn't specify cache files, nor /isolate
+            Assert.DoesNotContain("/orc", arguments);
+            Assert.DoesNotContain("/isolate", arguments);
+            // A project that is not built in isolation has to rely on /p:buildprojectreferences=false
+            Assert.Contains("/p:buildprojectreferences=false", arguments);
+        }
     }
```
cdmihai: Maybe also an integration test asserting that msbuild nodes actually fail with isolation constraints on undeclared dependencies?

cdmihai:
Not supper clear, do you allow a graph to have both isolated and non isolated nodes? Is it worth documenting it anywhere?

Public/Src/FrontEnd/UnitTests/MsBuild/SchedulingTests/MsBuildProcessBuilderTests.cs
smera: I can try to add something like that in a subsequent PR (I've been trying to push this one for the last 3 hours and now I'm only 1 validation out :)).
My initial thoughts on negative tests like those is that I was not after testing that MSBuild does what it is expected to do, at least not as my primary goal. But a test like the one you suggest sounds good as a safety net. 

---
In reply to: [272368151](https://github.com/Microsoft/BuildXL/pull/80#discussion_r272368151) [](ancestors = 272368151)

PR #81, opened 2019-04-04
Change MsBuildResolver to support multiple entry points

jbayardo:
Enabling multiple entry points in the MSBuildResolver is required to support the VcPkg integration.

This functionality breaks anyone using the MSBuildResolver and the `fileNameEntryPoint` setting, as these are now relative paths to the root of the project instead of atoms. The name has been changed as well.

### DIFF
--- a/Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
+++ b/Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
@@ -142,7 +142,7 @@ interface MsBuildResolver extends ResolverBase, UntrackingSettings {
      * Optional file name for the project or solution that should be used to start parsing (under the root traversal)
      * If not provided, BuildXL will try to find a candidate under the root traversal
      */
-    fileNameEntryPoint?: PathAtom;
+    fileNameEntryPoints?: RelativePath[];
     
     /**
      * Targets to execute on the entry point project. If not provided, the default targets are used.
--- a/Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
+++ b/Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
@@ -139,8 +139,11 @@ interface MsBuildResolver extends ResolverBase, UntrackingSettings {
     msBuildSearchLocations?: Directory[];
 
     /**
-     * Optional file name for the project or solution that should be used to start parsing (under the root traversal)
-     * If not provided, BuildXL will try to find a candidate under the root traversal
+     * Optional file paths for the projects or solutions that should be used to start parsing. These are relative 
+     * paths with respect to the root traversal.
+     *
+     * If not provided, BuildXL will attempt to find a candidate under the root traversal. If more than one candidate
+     * is available, the process will fail.
      */
     fileNameEntryPoints?: RelativePath[];
     
--- a/Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
+++ b/Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
@@ -22,7 +22,7 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         /// <summary>
         /// Path to the entry point project to parse
         /// </summary>
-        public string ProjectToParse { get; }
+        public IReadOnlyCollection<string> ProjectToParse { get; }
 
         /// <summary>
         /// Path to the file were the graph will be serialized to
@@ -54,15 +54,15 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
 
         /// <nodoc/>
         public MSBuildGraphBuilderArguments(
-            string enlistmentRoot, 
-            string projectToParse, 
+            string enlistmentRoot,
+            IReadOnlyCollection<string> projectToParse, 
             string outputPath, 
             IReadOnlyDictionary<string, string> globalProperties, 
             IReadOnlyCollection<string> mSBuildSearchLocations, 
             IReadOnlyCollection<string> entryPointTargets)
         {
             Contract.Requires(!string.IsNullOrEmpty(enlistmentRoot));
-            Contract.Requires(!string.IsNullOrEmpty(projectToParse));
+            Contract.Requires(projectToParse != null);
             Contract.Requires(!string.IsNullOrEmpty(outputPath));
             Contract.Requires(mSBuildSearchLocations != null);
             Contract.Requires(entryPointTargets != null);
@@ -80,7 +80,7 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         {
             return 
 $@"Enlistment root: {EnlistmentRoot}
-Project entry point: {ProjectToParse}
+Project entry point: {string.Join(" ", ProjectToParse)}
 Serialized graph path: {OutputPath}
 Global properties: {(GlobalProperties == null ? "null" : string.Join(" ", GlobalProperties.Select(kvp => $"[{kvp.Key}]={kvp.Value}")))}
 Search locations: {string.Join(" ", MSBuildSearchLocations)}";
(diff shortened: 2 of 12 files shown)


Public/Src/FrontEnd/MsBuild/MsBuildWorkspaceResolver.cs
```diff
@@ -277,11 +277,11 @@ private BuildParameters.IBuildParameters RetrieveBuildParameters()
             return buildParameters;
         }
 
-        private bool TryRetrieveParsingEntryPoint(out AbsolutePath parsingEntryPoint)
+        private bool TryRetrieveParsingEntryPoint(out IEnumerable<AbsolutePath> parsingEntryPoints)
         {
-            if (m_resolverSettings.FileNameEntryPoint.HasValue)
+            if (m_resolverSettings.FileNameEntryPoints != null && m_resolverSettings.FileNameEntryPoints.Count > 0)
```
dannyvv: Could just use elvis operator: `m_resolverSettings.FileNameEntryPoints?.Count`

Public/Src/FrontEnd/MsBuild/MsBuildWorkspaceResolver.cs
```diff
@@ -277,11 +277,11 @@ private BuildParameters.IBuildParameters RetrieveBuildParameters()
             return buildParameters;
         }
 
-        private bool TryRetrieveParsingEntryPoint(out AbsolutePath parsingEntryPoint)
+        private bool TryRetrieveParsingEntryPoint(out IEnumerable<AbsolutePath> parsingEntryPoints)
```
edgarrs: >IEnumerable<AbsolutePath> [](start = 54, length = 25)

Why are you changing this to IEnumerable if you are still validating if Count ==1 (line 299) and returning an IEnumerable with a single item?

Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
```diff
@@ -54,15 +54,15 @@ public sealed class MSBuildGraphBuilderArguments
...
+            string enlistmentRoot,
+            IReadOnlyCollection<string> projectToParse, 
             string outputPath, 
             IReadOnlyDictionary<string, string> globalProperties, 
             IReadOnlyCollection<string> mSBuildSearchLocations, 
             IReadOnlyCollection<string> entryPointTargets)
         {
             Contract.Requires(!string.IsNullOrEmpty(enlistmentRoot));
-            Contract.Requires(!string.IsNullOrEmpty(projectToParse));
+            Contract.Requires(projectToParse != null);
```
edgarrs: >projectToParse [](start = 30, length = 14)

Check if the collection is empty too

Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
```diff
@@ -22,7 +22,7 @@ public sealed class MSBuildGraphBuilderArguments
         /// <summary>
         /// Path to the entry point project to parse
         /// </summary>
-        public string ProjectToParse { get; }
+        public IReadOnlyCollection<string> ProjectToParse { get; }
```
edgarrs: >ProjectToParse [](start = 43, length = 14)

ProjectsToParse?

Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
```diff
@@ -54,15 +54,15 @@ public sealed class MSBuildGraphBuilderArguments
 
         /// <nodoc/>
         public MSBuildGraphBuilderArguments(
-            string enlistmentRoot, 
-            string projectToParse, 
+            string enlistmentRoot,
+            IReadOnlyCollection<string> projectToParse,
```
edgarrs: >projectToParse [](start = 40, length = 14)

plural: projectsToParse

Public/Src/FrontEnd/MsBuild/MsBuildWorkspaceResolver.cs
jbayardo: Because of line 284. If the user specified entry points within the resolver settings, then we certainly want to use those. Otherwise, we are finding projects ourselves; whether we want to build everything or fail by default is a reasonable discussion.

I kept the previous behavior because it seems better to build less unless asked to.

Public/Src/FrontEnd/MsBuild/MsBuildWorkspaceResolver.cs
jbayardo: It seems much more declarative to do the check

Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
```diff
@@ -142,7 +142,7 @@ interface MsBuildResolver extends ResolverBase, UntrackingSettings {
      * Optional file name for the project or solution that should be used to start parsing (under the root traversal)
```
smera: >Optional file name for the project or solution that should [](start = 7, length = 58)

Change comment appropriately

Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
```diff
@@ -142,7 +142,7 @@ interface MsBuildResolver extends ResolverBase, UntrackingSettings {
      * Optional file name for the project or solution that should be used to start parsing (under the root traversal)
      * If not provided, BuildXL will try to find a candidate under the root traversal
      */
-    fileNameEntryPoint?: PathAtom;
+    fileNameEntryPoints?: RelativePath[];
```
smera: >RelativePath [](start = 26, length = 12)

And please add in the comment relative to what these paths are going to be interpreted

Public/Src/FrontEnd/MsBuild/MsBuildWorkspaceResolver.cs
smera: +1 to Danny's comment

Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
```diff
@@ -22,7 +22,7 @@ public sealed class MSBuildGraphBuilderArguments
         /// <summary>
         /// Path to the entry point project to parse
```
smera: >Path to the entry point project to parse [](start = 12, length = 40)

Stale doc here as well

Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
```diff
@@ -48,27 +48,27 @@ public sealed class MSBuildGraphBuilderArguments
...
+            string enlistmentRoot,
+            IReadOnlyCollection<string> projectsToParse, 
             string outputPath, 
             IReadOnlyDictionary<string, string> globalProperties, 
             IReadOnlyCollection<string> mSBuildSearchLocations, 
             IReadOnlyCollection<string> entryPointTargets)
         {
             Contract.Requires(!string.IsNullOrEmpty(enlistmentRoot));
-            Contract.Requires(!string.IsNullOrEmpty(projectToParse));
+            Contract.Requires(projectsToParse != null && projectsToParse.Count > 0);
```
smera: >rse != null && projectsToParse.Count > 0 [](start = 42, length = 40)

suggest to use the elvis operator here as well

Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
```diff
@@ -80,7 +80,7 @@ public override string ToString()
         {
             return 
 $@"Enlistment root: {EnlistmentRoot}
-Project entry point: {ProjectToParse}
+Project entry point: {string.Join(" ", ProjectsToParse)}
```
smera: >point [](start = 14, length = 5)

points

smera: >{strin [](start = 21, length = 6)

I'd add square brackets so it denotes a list

smera:
>         /// Optional file name for the project or solution that should be used to start parsing (under the root traversal)

stale doc

---
Refers to: Public/Src/Utilities/Configuration/Resolvers/IMsBuildResolverSettings.cs:58 in 40a148f. [](commit_id = 40a148fd6c29f3f02c619cb72ee1390e45a72a2b, deletion_comment = False)

PR #83, opened 2019-04-04
[WIP] Apply special timestamps to all build outputs

almiliMSFT:
The idea is that all build outputs, no matter which pip produced them, have the special timestamps as long as they are under a shared opaque directory.  

Before this change, pips that explicitly declare one or more shared output directories, their outputs that are under any of those directories are marked with the special timestamps.

What's possible is that there are some other pips (e.g., CopyFile or WriteFile) that produce additional files into those same shared opaque directories.  Those files were not marked with the special timestamps; consequently, the scrubber fails to delete them before the next run, which can cause problems.  With this PR, all build outputs that end up being under any declared output directory will have the special timestamps.

### DIFF
--- a/Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
+++ b/Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
@@ -1974,139 +1974,116 @@ namespace BuildXL.Scheduler.Artifacts
                 // we still are mandated to finish materializing if possible and eventually complete the materialization task.
                 using (operationContext.StartOperation(PipExecutorCounter.FileContentManagerPlaceFilesDuration))
                 {
-                    using (var sharedOpaqueOutputsWrapper = Pools.FileArtifactSetPool.GetInstance())
+                    for (int i = 0; i < state.MaterializationFiles.Count; i++)
                     {
-                        // Collect all shared opaque outputs, so we can make sure they are flagged appropriately as such
-                        // after materialization succeeds for each of them
-                        var allSharedOpaqueOutputs = sharedOpaqueOutputsWrapper.Instance;
-                        if (pipInfo.UnderlyingPip is Process process)
-                        {
-                            foreach (var sharedOpaqueDirectory in process.DirectoryOutputs.Where(directory => directory.IsSharedOpaque))
+                        MaterializationFile materializationFile = state.MaterializationFiles[i];
+                        FileArtifact file = materializationFile.Artifact;
+                        FileMaterializationInfo materializationInfo = materializationFile.MaterializationInfo;
+                        ContentHash hash = materializationInfo.Hash;
+                        PathAtom fileName = materializationInfo.FileName;
+                        AbsolutePath symlinkTarget = materializationFile.SymlinkTarget;
+                        bool allowReadOnly = materializationFile.AllowReadOnly;
+                        int materializationFileIndex = i;
+
+                        state.PlacementTasks.Add(Task.Run(
+                            async () =>
                             {
-                                foreach (var fileArtifact in ListSealedDirectoryContents(sharedOpaqueDirectory))
+                                // Wait for the prior version of the file artifact to finish materialization
+                                await materializationFile.PriorArtifactVersionCompletion;
+
+                                if (Context.CancellationToken.IsCancellationRequested)
                                 {
-                                    allSharedOpaqueOutputs.Add(fileArtifact);
+                                    state.SetMaterializationFailure(fileIndex: materializationFileIndex);
+                                    success = false;
+                                    return;
                                 }
-                            }
-                        }
 
-                        for (int i = 0; i < state.MaterializationFiles.Count; i++)
-                        {
-                            MaterializationFile materializationFile = state.MaterializationFiles[i];
-                            FileArtifact file = materializationFile.Artifact;
-                            FileMaterializationInfo materializationInfo = materializationFile.MaterializationInfo;
-                            ContentHash hash = materializationInfo.Hash;
-                            PathAtom fileName = materializationInfo.FileName;
-                            AbsolutePath symlinkTarget = materializationFile.SymlinkTarget;
-                            bool allowReadOnly = materializationFile.AllowReadOnly;
-                            int materializationFileIndex = i;
+                                Possible<ContentMaterializationResult> possiblyPlaced;
 
-                            state.PlacementTasks.Add(Task.Run(
-                                async () =>
+                                using (var outerContext = operationContext.StartAsyncOperation(PipExecutorCounter.FileContentManagerTryMaterializeOuterDuration, file))
+                                using (await m_materializationSemaphore.AcquireAsync())
                                 {
-                                    // Wait for the prior version of the file artifact to finish materialization
-                                    await materializationFile.PriorArtifactVersionCompletion;
-
-                                    if (Context.CancellationToken.IsCancellationRequested)
+                                    if (m_host.CanMaterializeFile(file))
                                     {
-                                        state.SetMaterializationFailure(fileIndex: materializationFileIndex);
-                                        success = false;
-                                        return;
-                                    }
-
-                                    Possible<ContentMaterializationResult> possiblyPlaced;
-
-                                    using (var outerContext = operationContext.StartAsyncOperation(PipExecutorCounter.FileContentManagerTryMaterializeOuterDuration, file))
-                                    using (await m_materializationSemaphore.AcquireAsync())
-                                    {
-                                        if (m_host.CanMaterializeFile(file))
+                                        using (outerContext.StartOperation(PipExecutorCounter.FileContentManagerHostTryMaterializeDuration, file))
                                         {
-                                            using (outerContext.StartOperation(PipExecutorCounter.FileContentManagerHostTryMaterializeDuration, file))
-                                            {
-                                                var possiblyMaterialized = await m_host.TryMaterializeFileAsync(file, outerContext);
-                                                possiblyPlaced = possiblyMaterialized.Then(origin =>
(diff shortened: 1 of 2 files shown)


almiliMSFT:
Just for early feedback.  Tests are obviously missing, and I'll add them after the feedback

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -1974,139 +1974,116 @@ private async Task<bool> PrepareDirectoriesAsync(PipArtifactsState state, Operat
                 // we still are mandated to finish materializing if possible and eventually complete the materialization task.
                 using (operationContext.StartOperation(PipExecutorCounter.FileContentManagerPlaceFilesDuration))
                 {
-                    using (var sharedOpaqueOutputsWrapper = Pools.FileArtifactSetPool.GetInstance())
+                    for (int i = 0; i < state.MaterializationFiles.Count; i++)
```
almiliMSFT: The only change in this file is basically to not compute `allSharedOpaqueOutputs` and not apply timestamps (what used to be line 2089).  The diff is big only because indentation changed (because there is one fewer `using` block now)

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3524,21 +3524,29 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
             }
         }
 
-        private static void FlagSharedOpaqueOutputs(IPipExecutionEnvironment environment, ExecutionResult executionResult)
+        private void FlagSharedOpaqueOutputs(IPipExecutionEnvironment environment, ProcessRunnablePip process)
```
almiliMSFT: Here, in addition to `SharedDynamicDirectoryWriteAccesses` we apply the special timestamps to all outputs that are within some output directory

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5354,6 +5364,14 @@ void IFileContentManagerHost.ReportMaterializedArtifact(in FileOrDirectoryArtifa
             m_pipOutputMaterializationTracker.ReportMaterializedArtifact(artifact);
         }
 
+        private void MakeSharedOpaqueOutputIfNeeded(AbsolutePath path, bool force = false)
+        {
+            if (force || PipGraph.IsPathUnderOutputDirectory(path))
```
almiliMSFT: Here I only check if the path is under an output directory; I could amend this check to additionally check if that directory is a **shared opaque** directory

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5316,6 +5324,8 @@ void IFileContentManagerHost.ReportContent(FileArtifact artifact, in FileMateria
                     existence = PathExistence.ExistsAsFile;
                 }
 
+                MakeSharedOpaqueOutputIfNeeded(artifact.Path);
```
almiliMSFT: Here, every time content is reported, if it corresponds to an output file and the file exists we apply the timestamps if that path is under an output directory

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3524,21 +3524,29 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
             }
         }
 
-        private static void FlagSharedOpaqueOutputs(IPipExecutionEnvironment environment, ExecutionResult executionResult)
+        private void FlagSharedOpaqueOutputs(IPipExecutionEnvironment environment, ProcessRunnablePip process)
         {
+            // Select all declared output files
+            IEnumerable<(AbsolutePath, bool)> paths = process.Process
```
smera: >AbsolutePath, bool [](start = 25, length = 18)

named tuple?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5354,6 +5364,14 @@ void IFileContentManagerHost.ReportMaterializedArtifact(in FileOrDirectoryArtifa
             m_pipOutputMaterializationTracker.ReportMaterializedArtifact(artifact);
         }
 
+        private void MakeSharedOpaqueOutputIfNeeded(AbsolutePath path, bool force = false)
+        {
+            if (force || PipGraph.IsPathUnderOutputDirectory(path))
+            {
+                SharedOpaqueOutputHelper.EnforceFileIsSharedOpaqueOutput(path.ToString(Context.PathTable));
```
smera: >EnforceFileIsSharedOpaqueOutput [](start = 41, length = 31)

Maybe now a better name would be 'EnforceFileIsUnderSharedOpaqueOutput'

Public/Src/Engine/Scheduler/Scheduler.cs
smera: Scoping this down so we do this for shared opaques only makes sense to me. Sounds like the less intrusive change, considering that all this timestamp flagging is relatively hacky to being with.

---
In reply to: [272298426](https://github.com/Microsoft/BuildXL/pull/83#discussion_r272298426) [](ancestors = 272298426)

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3524,21 +3524,29 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
             }
         }
 
-        private static void FlagSharedOpaqueOutputs(IPipExecutionEnvironment environment, ExecutionResult executionResult)
+        private void FlagSharedOpaqueOutputs(IPipExecutionEnvironment environment, ProcessRunnablePip process)
         {
+            // Select all declared output files
+            IEnumerable<(AbsolutePath, bool)> paths = process.Process
+                .FileOutputs
+                .Select(f => (f.Path, false));
```
smera: Maybe I'm missing something, but why do you need to do this for files, if you are already calling MakeSharedOpaqueOutputIfNeeded under ReportContent for every file being reported?

Public/Src/Engine/Scheduler/Scheduler.cs
smera: Just double checking. Are you sure ReportContent is always called? At least for the shared opaque case, there were cases where pips were failing in an early stage and stuff was not being reported yet, but files were already produced. There is a comment under FlagSharedOpaque... about using the raw data that comes from detours that contains dynamic writes, that was because of this.

---
In reply to: [272298907](https://github.com/Microsoft/BuildXL/pull/83#discussion_r272298907) [](ancestors = 272298907)

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
narasamdya: Is the flagging of shared opaque outputs now taken care of by Scheduler's post process?

---
In reply to: [272297613](https://github.com/Microsoft/BuildXL/pull/83#discussion_r272297613) [](ancestors = 272297613)

almiliMSFT: yes

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: agreed

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: That's why there is another place where this is called. That other place is `FlagSharedOpaqueOutputs`, which is called from `PostProcess`

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: will change

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: indeed

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: I'm not 100% sure this is needed either.

My understanding is that this code path exists so that it flags files even if the pip failed, which is why I thought I'd include all outputs.

When the pip doesn't fail, this code path is indeed not necessary

smera: From what I can see, if you leave this code as is (which seems fine, in the sense that even when pips are terminated early, this will run, since that's working for explicit shared opaque content), then you don't need the change you added in ReportContent, since that step will always re-do work.

almiliMSFT: That's true for the case when the pip runs and produces the files; when the pip is replayed from cache, then only `ReportContent` gets called.

smera: As far as I remember I also modified how the file materialization works to ensure timestamps there as well

almiliMSFT: You did, but I removed it in this PR.

The reason I removed it is because there I don't have access to PipGraph or something that would let me ask if a give path is under any output directory.  I could add a method to `IFileContentManagerHost` that would allow me to do this though

smera: I think one other reason to keep the timestamp enforcement during materialization is: a file that is not under a shared opaque is cached, then there is an unrelated new pip defining a shared opaque that includes the file. The first pip is replayed from cache. We need to make sure the timestamp is now changed.
Not sure if all this happens when reporting content. Not 100$ sure, but if I remember correctly, it doesn't.

PR #84, opened 2019-04-04
Add the Azure Boards badge to the README

AndyGerlicher:
Add the Azure Boards badge for the board used to track the work for this repository. Fixes [AB#1502732](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1502732)


AndyGerlicher:
Would this be useful for anyone? I was thinking I would just close it. This was created on my behalf via some automation. Good to know it exists and if people like it we should include it.

erikma:
Maybe... it's slightly confusing for external folks as we are also using public GitHub Issues, and they cannot reach the Azure board, obviously.

erikma:
On second thought, probably best to just cut because of that.

dannyvv:
I don't object, but I don't think it would be usefull for me.

---
In reply to: [480036116](https://github.com/Microsoft/BuildXL/pull/84#issuecomment-480036116) [](ancestors = 480036116)

PR #85, opened 2019-04-04
Failed USN read gracefully when read request not supported

narasamdya:
When read request is not supported because 
**journal query error 'FailedToOpenVolumeHandle': Failed to open a volume handle for the volume '\\?\Volume{15467674-7855-44df-a9a2-e27782706177}\'**

USN read request on that volume results in a catastrophic error:
**Exception:BuildXL.Native.IO.NativeWin32Exception (0x80070032): Native: DeviceIoControl(FSCTL_READ_FILE_USN_DATA) for ReadFileUsnByHandle failed (0x32: The request is not supported)**

With this change, we simply failed gracefully, by returning null. Consequently, the tracker will stop tracking because missing to track a single important file can result in unfaithful data in the tracker state.

[AB#1496397](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1496397)


dannyvv:
I filed a bug in VSO, can you link it?

narasamdya:
It's already linked in the description.

---
In reply to: [480093692](https://github.com/Microsoft/BuildXL/pull/85#issuecomment-480093692) [](ancestors = 480093692)

PR #87, opened 2019-04-04
Make Process Detours tests fast

narasamdya:
The slow tests seem to be caused by creating large formatted strings in `FileAccessManifestTreeTests`. 

With such formatted strings, the tests take 3:30 - 4 minutes. Without such strings, the tests take 30s.

To repro, run `bxl Test.BuildXL.Processes.Detours.dsc`

[AB#1502822](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1502822)

### DIFF
--- a/Public/Src/Engine/UnitTests/Processes.Detours/FileAccessManifestTreeTest.cs
+++ b/Public/Src/Engine/UnitTests/Processes.Detours/FileAccessManifestTreeTest.cs
@@ -606,7 +606,10 @@ namespace Test.BuildXL.Processes.Detours
 
         private static void TestManifestRetrieval(IEnumerable<ValidationData> validationData, FileAccessManifest fam)
         {
-            string famDescription = Environment.NewLine + string.Join(Environment.NewLine, fam.Describe());
+            foreach (var line in fam.Describe())
+            {
+                Console.WriteLine(line);
+            }
 
             byte[] manifestTreeBytes = fam.GetManifestTreeBytes();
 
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/UnitTests/Processes.Detours/FileAccessManifestTreeTest.cs
```diff
@@ -606,7 +606,10 @@ public void TestSolution10000()
 
         private static void TestManifestRetrieval(IEnumerable<ValidationData> validationData, FileAccessManifest fam)
         {
-            string famDescription = Environment.NewLine + string.Join(Environment.NewLine, fam.Describe());
+            foreach (var line in fam.Describe())
+            {
+                Console.WriteLine(line);
+            }
```
narasamdya: It's fine to print this unconditionally up-front.

almiliMSFT: Or even not print it at all

Public/Src/Engine/UnitTests/Processes.Detours/FileAccessManifestTreeTest.cs
```diff
@@ -627,44 +630,32 @@ private static void TestManifestRetrieval(IEnumerable<ValidationData> validation
...
-                XAssert.IsTrue(success, string.Format("Unable to find path in manifest.{0}", famDescription));
+                XAssert.IsTrue(success, "Unable to find path in manifest");
                 XAssert.AreEqual(
                     unchecked((uint)dataItem.PathId),
                     pathId,
-                    string.Format(
-                        "PathId for '{0}' did not match.{1}",
-                        dataItem.Path,
-                        famDescription));
+                    string.Format("PathId for '{0}' did not match", dataItem.Path));
```
almiliMSFT: there is no reason calling `string.Format` here because `XAssert.AreEqual` (as well as all other `XAssert` methods) already accept `format` + `args`, so you could just write
```csharp
XAssert.AreEqual(
    unchecked((uint)dataItem.PathId),
	pathId,
	"PathId for '{0}' did not match",
    dataItem.Path);
```

Public/Src/Engine/UnitTests/Processes.Detours/FileAccessManifestTreeTest.cs
narasamdya: We need it for debugging in case of error.

Public/Src/Engine/UnitTests/Processes.Detours/FileAccessManifestTreeTest.cs
narasamdya: Good catch.

PR #88, opened 2019-04-04
Avoid IOException when hardlink creation failed with access denied

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Interfaces/FileSystem/AbsFileSystemExtension.cs
```diff
@@ -17,6 +17,34 @@ public static class AbsFileSystemExtension
...
+            {
+                attributes = default;
+                return false;
+            }
+
+            try
+            {
+                attributes = fileSystem.GetFileAttributes(path);
+            }
+            catch(FileNotFoundException)
```
dannyvv: >h( [](start = 16, length = 2)

Space

dannyvv: >FileNotFoundException [](start = 18, length = 21)

Why not Anauthorizedaccessexcption?

SergeyTeplyakov: Thanks.

SergeyTeplyakov: In this case, we want to avoid the race if the file is deleted after the file existence is check to avoid `FileNotFoundExcpetion` to be propagated to the client. `UnauthorizedException` is a different type of issue and it should be handled by the client separately.

PR #90, opened 2019-04-05
Add try finally block to reset m_isSerializationInProgress

norahuang:
When exception happens during the serialization, m_isSerializationInProgress can't be reset. So the table will stay in the state of SerializationInProgress and no new item can be added to it. The crash in the bug happened in this case. Adding the finally block can reset the m_isSerializationInProgress and solve the problem.

Fixes [AB#1419013](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1419013)

### DIFF
--- a/Public/Src/Utilities/Utilities/StringTable.cs
+++ b/Public/Src/Utilities/Utilities/StringTable.cs
@@ -1100,14 +1100,14 @@ namespace BuildXL.Utilities
             try
             {
                 Serialize(
-                writer,
-                new SerializedState()
-                {
-                    Count = Count,
-                    NextId = m_nextId,
-                    StringSet = m_stringSet,
-                    ByteBuffers = m_byteBuffers,
-                });
+                    writer,
+                    new SerializedState()
+                    {
+                        Count = Count,
+                        NextId = m_nextId,
+                        StringSet = m_stringSet,
+                        ByteBuffers = m_byteBuffers,
+                    });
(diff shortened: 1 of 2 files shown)


Public/Src/Utilities/Utilities/StringTable.cs
```diff
@@ -1106,7 +1108,11 @@ public void Serialize(BuildXLWriter writer)
                     StringSet = m_stringSet,
                     ByteBuffers = m_byteBuffers,
                 });
```
semihokur: nit. indent

narasamdya:
"When exception happens during the serialization, m_isSerializationInProgress can't be reset" -- should we not simply fail the build when this happens?

norahuang:
> "When exception happens during the serialization, m_isSerializationInProgress can't be reset" -- should we not simply fail the build when this happens?



> "When exception happens during the serialization, m_isSerializationInProgress can't be reset" -- should we not simply fail the build when this happens?

Do you mean that we should catch the exception and let the build pass? In that case the build will success without caching, will that be ok?

narasamdya:
I mean, the resulting build is "BUILD ERROR".

---
In reply to: [480970203](https://github.com/Microsoft/BuildXL/pull/90#issuecomment-480970203) [](ancestors = 480970203)

PR #91, opened 2019-04-05
Retry and logging support for grpc mode

semihokur:


Public/Src/Engine/Dll/Distribution/DistributionHelpers.cs
```diff
@@ -148,6 +150,35 @@ internal static string GetServiceName(string ipAddress, int port)
             return string.Format(CultureInfo.InvariantCulture, "{0}::{1}", ipAddress, port);
         }
 
+        internal static string GetExecuteDescription(IList<long> semiStableHashes)
```
dannyvv: There is a lot of string concatenation and conversion going on. 
If this gets called a lot It would be more effecient to create a string builder and use StringBuilder.Format to the the x16 conversion so we don't allocate a string for each hash, and then another for them joined and then another that puts execute in front. 

Same applies for other method where even more string concatenation happens.

Again if this is for failed pips or once per build its okay, but if this is for every pip it is probably worth optimizing

Public/Src/Engine/Dll/Distribution/Grpc/GrpcConstants.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace BuildXL.Engine.Distribution.Grpc
+{
+    internal static class GrpcConstants
```
dannyvv: Did you consider adding the const timeouts in the ConnectionManger here as well?

Public/Src/Engine/Dll/Distribution/Grpc/GrpcConstants.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace BuildXL.Engine.Distribution.Grpc
+{
+    internal static class GrpcConstants
+    {
+        public const string TraceIdKey = "traceid-bin";
```
ljcollins25: >-bin [](start = 49, length = 4)

Why "-bin"?

Public/Src/Engine/Dll/Distribution/Grpc/ServerInterceptor.cs
```diff
@@ -0,0 +1,81 @@
...
+                    traceId = new Guid(kvp.ValueBytes).ToString();
+                }
+                else if (kvp.Key == GrpcConstants.BuildIdKey)
+                {
+                    senderBuildId = kvp.Value;
+                }
+            }
+
+            if (!string.IsNullOrEmpty(senderBuildId) && senderBuildId != m_buildId)
+            {
```
ljcollins25: >{ [](start = 12, length = 1)

Probably a good idea to log before throwing exception so we make sure to get something logged on the server.

Public/Src/Engine/Dll/Distribution/Grpc/ServerInterceptor.cs
```diff
@@ -0,0 +1,81 @@
...
+                throw new RpcException(
```
ljcollins25: >RpcException [](start = 26, length = 12)

Is throwing an exception, how this should be done? Did you verify that the exception makes it back to the client? Often in cache code, we just get a somewhat unhelpful message about exception being logged on server without text indicating the details of what happened.

Public/Src/Engine/Dll/Distribution/DistributionHelpers.cs
ljcollins25: This is probably negligible when compared to the amount of string concatenation we do in the cache code and logging in general.

---
In reply to: [272719241](https://github.com/Microsoft/BuildXL/pull/91#discussion_r272719241) [](ancestors = 272719241)

Public/Src/Engine/Dll/Distribution/Grpc/ServerInterceptor.cs
semihokur: Discussed in-person. Basically, when we throw an exception on the server side, a failed message is sent to the sender with the exception details

---
In reply to: [272744985](https://github.com/Microsoft/BuildXL/pull/91#discussion_r272744985) [](ancestors = 272744985)

Public/Src/Engine/Dll/Distribution/Grpc/GrpcConstants.cs
semihokur: when you send binary data in the headers, you need to put '-bin' suffix to indicate that the value is binary. 

---
In reply to: [272744034](https://github.com/Microsoft/BuildXL/pull/91#discussion_r272744034) [](ancestors = 272744034)

PR #95, opened 2019-04-05
Update Transformer.seal* calls to use new object-style api

dannyvv:


dannyvv:
This PR will have to wait untill we have updated the Public LKG and deployed it to cloudbuild.

PR #96, opened 2019-04-05
Apply special timestamps to all build outputs under shared opaque dirs

almiliMSFT:
The idea is that all build outputs, no matter which pip produced them, have the special timestamps as long as they are under a shared opaque directory.

Before this change, pips that explicitly declare one or more shared output directories, their outputs that are under any of those directories are marked with the special timestamps.

What's possible is that there are some other pips (e.g., CopyFile or WriteFile) that produce additional files into those same shared opaque directories. Those files were not marked with the special timestamps; consequently, if those files become stale, the scrubber will fail to delete them before the build, which can cause problems downstream. 

With this PR, all build outputs that end up being under any declared shared opaque output directory will have the special timestamps.

[AB#1472511](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1472511)


Public/Src/Engine/Scheduler/Graph/PipGraph.Serialization.cs
```diff
@@ -26,7 +26,7 @@ private sealed class SerializedState
             public readonly ConcurrentBigMap<ModuleId, NodeId> Modules;
             public readonly ConcurrentBigMap<FileArtifact, NodeId> PipProducers;
             public readonly ConcurrentBigMap<DirectoryArtifact, NodeId> OpaqueDirectoryProducers;
-            public readonly ConcurrentBigSet<AbsolutePath> OutputDirectoryRoots;
+            public readonly ConcurrentBigMap<AbsolutePath, bool> OutputDirectoryRoots;
```
smera: >bool [](start = 59, length = 4)

nit: maybe a comment here saying what this boolean is about?

almiliMSFT: agreed

Public/Src/Engine/Scheduler/Artifacts/IFileContentManagerHost.cs
```diff
@@ -114,6 +114,11 @@ public interface IFileContentManagerHost
         /// </summary>
         void ReportMaterializedArtifact(in FileOrDirectoryArtifact artifact);
 
+        /// <summary>
+        /// Callback to host to report that a file has been placed to disk.
+        /// </summary>
+        void ReportFileArtifactPlaced(in FileArtifact artifact);
```
smera: >ReportFileArtifactPlaced [](start = 13, length = 24)

Nice

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3529,21 +3529,29 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
             }
         }
 
-        private static void FlagSharedOpaqueOutputs(IPipExecutionEnvironment environment, ExecutionResult executionResult)
+        private void FlagSharedOpaqueOutputs(IPipExecutionEnvironment environment, ProcessRunnablePip process)
         {
+            // Select all declared output files
+            IEnumerable<(AbsolutePath path, bool isKnownToBeUnderSharedOpaque)> paths = process.Process
+                .FileOutputs
+                .Select(f => (path: f.Path, isKnownToBeUnderSharedOpaque: false));
```
smera: >false [](start = 74, length = 5)

Is this actually always false? Thinking in the case where a pip may declare a file under one of its own shared opaques

narasamdya:
Suppose that I have a pip P that specifies explicitly that D\f is its output file. It doesn't say anything about D. Another pip Q says that D is its shared output directory. Does the timestamp of the output D\f of P get modified?

It sounds like all outputs under any shared opaque directories will be scrubbed. Is it the case?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3529,21 +3529,29 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
...
+                var sharedOpaqueDirOutputs = process.ExecutionResult
+                    .SharedDynamicDirectoryWriteAccesses
+                    .SelectMany(kvp => kvp.Value);
+
+                paths = paths.Concat(sharedOpaqueDirOutputs.Select(path => (path, isKnownToBeUnderSharedOpaque: true)));
+            }
+
+            foreach (var tuple in paths)
+            {
+                MakeSharedOpaqueOutputIfNeeded(tuple.path, force: tuple.isKnownToBeUnderSharedOpaque);
```
narasamdya: >force [](start = 59, length = 5)

Why do we need this `force` field? For those in `SharedDynamicDirectoryWriteAccesses`, why don't we simply mark them as shared opaque outputs? And only check the graph for those in `FileOutputs`.

Examples/DotNetCoreBuild/test/TestDirectories/testDirectories.dsc
```diff
@@ -32,11 +32,13 @@ export const explicitOutputsByDifferentPipsAreAllowedInSharedOpaqueDirectory = !
```
almiliMSFT: > Suppose that I have a pip P that specifies explicitly that D\f is its output file. It doesn't say anything about D. Another pip Q says that D is its shared output directory. Does the timestamp of the output D\f of P get modified?

Yes, the timestamps of `D/f` do get modified in this case.

> It sounds like all outputs under any shared opaque directories will be scrubbed. Is it the case?

The scrubbing only happens if there is a file on disk that is not part of the pip graph; "real" shared opaque outputs are never part of the pip graph so they get scrubbed (this is exactly how it's always been).  Explicitly declared outputs that are already on disk will not get scrubbed because they are explicitly declared in the pip graph.  So the main purpose of this change is so BuildXL deletes stale outputs that are under shared opaque directories, i.e.,:
  - there once was an explicitly declared output in the pip graph
  - a build was executed and produced a corresponding file on disk (that file gets its timestamps modified with this PR)
  - now that output is removed from the pip graph and a new build is started
    - before this new build is started, the old file on disk is scrubbed because it is not part of the pip graph any more and it does have the special timestamps (which make it look like it was a shared opaque output)

Public/Src/Engine/Scheduler/Scheduler.cs
olkononenko: >isKnownToBeUnderSharedOpaque: false [](start = 44, length = 35)

I think it's possible to declare a an output file inside an opaque output:
file: foo\1.txx
directory: foo
If I'm correct, I do not think we can blindly set 'under shared opaque' for all the file outputs

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: Sure, I could break it up and do it twice: once for `SharedDynamicDirectoryWriteAccesses` (which get marked immediately, without checking) and once for `FileOutputs` (which are first checked and them marked if needed)

I did it this way so that I can concat them all together and uniformly call `MakeSharedOpaqueOutputIfNeeded`.  I can change it to what you suggested though

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: `isKnownToBeUnderSharedOpaque` is only relevant when the value is `true`; when the value is `false`, the `MakeSharedOpaqueOutputIfNeeded` method will explicitly check every path to see if it is under a shared opaque directory. 

I'll remove this flag since it appears to be very confusing.

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: Removed `force`

Examples/DotNetCoreBuild/test/TestDirectories/testDirectories.dsc
narasamdya: Thanks for the explanation.

---
In reply to: [273167983](https://github.com/Microsoft/BuildXL/pull/96#discussion_r273167983) [](ancestors = 273167983)

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: removed `force` and the linq stuff

PR #98, opened 2019-04-05
Clean up grpc copy logic, support compression

dcwuser:


Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace BuildXL.Cache.ContentStore.Service.Grpc
+{
+
+    // This class is required for decompressing data streamed via grpc
```
SergeyTeplyakov: Use <summary>?

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
...
+
+    // This class is required for decompressing data streamed via grpc
+    // because the compression api insists on reading from a stream and
+    // the grpc api exposes a read-one-chunk method instead of a stream.
+
+    internal class BufferedReadStream : Stream
+    {
+        public BufferedReadStream(Func<Task<byte[]>> reader)
+        {
+            if (reader is null)
```
SergeyTeplyakov: nit: Contract.Requires(reader != null) for consistencies' sake?

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                throw new ArgumentNullException(nameof(reader));
+            }
+            _reader = reader;
+        }
+
+        private readonly Func<Task<byte[]>> _reader;
```
SergeyTeplyakov: nit: we don't have specific guidelines, but in most cases we put the fields on top.

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
...
+        private int _readPointer; // the next index to be read from the storage buffer
+
+        private int _position; // total bytes that have been read
+        private int _length; // total bytes that have been ingested
+
+        public override bool CanRead => true;
+
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            Debug.Assert(!(buffer is null));
```
SergeyTeplyakov: Contract.Asssert instead

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
```diff
@@ -247,18 +247,17 @@ message SelectorData {
...
+    int64 offset = 5;
     CopyCompression compression = 6;
 }
 
 message CopyFileResponse {
-    ResponseHeader Header = 1;
-    bytes Content = 2;
-    int64 Index = 3;
+    bytes content = 1;
+    int64 index = 2;
```
dannyvv: >index [](start = 10, length = 5)

Curious why camlCasing these but keeping the other types PascalCasing...

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        public override long Seek(long offset, SeekOrigin origin) => throw new InvalidOperationException();
+
+        public override void SetLength(long value) => throw new NotImplementedException();
+
+        public override Task FlushAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+        public override void Flush() { }
+
+        public override long Length => _length;
```
dannyvv: >_length [](start = 39, length = 7)

To me this seems unexpected behavior if I were using this class.
My understanding is that the Length property of a stream gives how much you can read and use that as a guide to read from the stream.
In this implementation it seems that we are updating the length as we are reading, so the other way around...

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            }
+
+            Debug.Assert(totalCount == count);
+            return totalCount;
+        }
+
+        public override int Read(byte[] buffer, int offset, int count)
+        {
+            return ReadAsync(buffer, offset, count).Result;
```
dannyvv: >Result [](start = 52, length = 6)

`.GetAwaiter().GetResult()` would be better.

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
dcwuser: proto convention is Java: members smallCamel and types BigCamel. The code generator uses .NET convention when it generates C#.

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Threading.Tasks;
+
+namespace BuildXL.Cache.ContentStore.Service.Grpc
+{
+
+    // This class is required for decompressing data streamed via grpc
+    // because the compression api insists on reading from a stream and
+    // the grpc api exposes a read-one-chunk method instead of a stream.
+
+    internal class BufferedReadStream : Stream
```
SergeyTeplyakov: It seems that the logic is relatively involved, so maybe it make sense to create at least a few tests to cover the basic functionality.

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
...
+            return ReadAsync(buffer, offset, count).Result;
+        }
+
+        public override bool CanWrite => false;
+
+        public override void Write(byte[] buffer, int offset, int count) => throw new InvalidOperationException();
+
+        public override bool CanSeek => false;
+
+        public override long Seek(long offset, SeekOrigin origin) => throw new InvalidOperationException();
```
SergeyTeplyakov: Should it be not supported exception instead?
(I'm not sure, to be honest...)

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedWriteStream.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            Debug.Assert(!(buffer is null));
+            Debug.Assert(offset >= 0);
+            Debug.Assert(count >= 0);
+            Debug.Assert(count <= buffer.Length - offset);
+
+            int readPointer = offset;
+            int totalCount = 0;
+            while (totalCount < count)
+            {
```
dannyvv: > { [](start = 11, length = 2)

We should probably check the cancellation token inside the loop

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedWriteStream.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        private readonly byte[] _storage; // buffer to store content before written
+        private readonly Func<byte[], int, int, Task> _writer; // method to write
+        private int _writePointer; // pointer to next index in buffer to be written to
+        private int _position; // total bytes written
+
+        public override bool CanWrite => true;
+
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cts)
```
SergeyTeplyakov: nit: Should it be 'cancellationToken' instead of cts?

(I think 'cts' usually means 'CancellationToken*Source*'.

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedWriteStream.cs
```diff
@@ -0,0 +1,112 @@
...
+    // the grpc api exposes an write-one-chunk method instead of a stream.
+
+    // This class is not safe for simultaneous use by multiple threads.
+
+    // Be sure to call Flush when done writing, otherwise the last bit
+    // of content will (probably) not be processed, because it will (probably)
+    // not fill the buffer.
+    internal class BufferedWriteStream : Stream
+    {
+        public BufferedWriteStream(byte[] storage, Func<byte[], int, int, Task> writer)
```
dannyvv: >writer [](start = 80, length = 6)

Would it be useful for the write to take the cancellation token?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -29,6 +28,11 @@ public sealed class GrpcCopyClient : IShutdown<BoolResult>
         private readonly string _host;
         private readonly int _grpcPort;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether compression may be used during transmission.
```
SergeyTeplyakov: It seems that the comment is out of sync. It just returns the indication but can't be changed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
             };
         }
 
+        private async Task<OpenStreamResult> GetFileStreamAsync (Context context, ContentHash hash)
+        {
+            // For now, take the first store. Change this to interate through all stores
+            // until content is found.
+            IStreamStore store = _contentStoreByCacheName.First().Value as IStreamStore;
```
dannyvv: >First [](start = 58, length = 5)

Since you handle differnt items below. is it guarnatted to have at least one contentstore? Else `.FirstOrDefault()?.Value` would be more robust.

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
dcwuser: Agree this is unconventional. We can't return conventional Length b/c we don't have it until we read to end. Idea here was to keep track of how much we had ingested. (This is almost but not quite Position; since we have an internal buffer we may have ingested a bit more than has actually been read.) I used this value in debugging but it is never called by the production code. I could (a) leave as is, (b) give this property a different name (and throw NotSupported for Length), or (c) not expose this quantity at all (and throw NotSupported for Length). Preference?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
...
-        private async Task StreamContentAsync(OpenStreamResult openStreamResult, IServerStreamWriter<CopyFileResponse> responseStream, DateTime startTime)
+        private async Task<(long, long)> StreamContentAsync(Stream reader, byte[] buffer, IServerStreamWriter<CopyFileResponse> responseStream, CancellationToken cts = default(CancellationToken))
+        {
+            Debug.Assert(!(reader is null));
+            Debug.Assert(!(responseStream is null));
+
+            long chunks = 0L;
+            long bytes = 0L;
+            while (true)
+            {
```
dannyvv: > { [](start = 11, length = 2)

We should probably check the cancellation token inside the loop.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
...
+                    switch (result.Code)
+                    {
+                        case OpenStreamResult.ResultCode.ContentNotFound:
+                            headers.Add("Exception", "ContentNotFound");
+                            headers.Add("Message", $"Requested content {hash} not found.");
+                            break;
+                        case OpenStreamResult.ResultCode.Error:
+                            Debug.Assert(result.Exception != null);
+                            headers.Add("Exception", result.Exception.GetType().Name);
+                            headers.Add("Message", result.Exception.Message);
```
SergeyTeplyakov: Do we want to return the stack trace? Maybe for critical errors only?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
...
+                            break;
+                        case OpenStreamResult.ResultCode.Success:
+                            Debug.Assert(result.Stream != null);
+                            long size = result.Stream.Length;
+                            headers.Add("FileSize", size.ToString());
+                            if ((request.Compression == CopyCompression.Gzip) && (size > ContentStore.Grpc.CopyConstants.DefaultBufferSize)) {
```
SergeyTeplyakov: nit: inconsistent formatting (brace on the same line)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
...
+                bytes += chunkSize;
+                chunks++;
+
+                Console.WriteLine(chunks);
+            }
+
+            return (chunks, bytes);
+        }
+
+        /*
```
dannyvv: >/* [](start = 8, length = 2)

Would you mind removing the commented out code?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
...
-                    await StreamContentAsync(openStreamResult, responseStream, startTime);
-                }
-                else
-                {
-                    await responseStream.WriteAsync(
-                        new CopyFileResponse
+                    // Send the content.
+                    if (result.Succeeded)
+                    {
+                        byte[] buffer = new byte[ContentStore.Grpc.CopyConstants.DefaultBufferSize];
```
SergeyTeplyakov: question: do we want to pull this? Or the lifetime is more complicated and not a scoped based?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
             };
         }
 
+        private async Task<OpenStreamResult> GetFileStreamAsync (Context context, ContentHash hash)
```
SergeyTeplyakov: nit: formatting - extra space after the method name and before the opening brace.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
...
+                    ByteString content = ByteString.CopyFrom(bf, offset, count);
+                    CopyFileResponse response = new CopyFileResponse() { Content = content, Index = chunks };
+                    await responseStream.WriteAsync(response);
+                    bytes += count;
+                    chunks++;
+                }
+            ))
+
+            {
+                using (Stream compressionStream = new GZipStream(grpcStream, System.IO.Compression.CompressionLevel.Fastest, true))
```
dannyvv: >GZipStream [](start = 54, length = 10)

It might be worth comparing the perf with the built-in gzip accorss full framework, .net Core as well as the GZip stream from ICSharpCode.SharpZipLib.GZip.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dannyvv: :+1

---
In reply to: [272773954](https://github.com/Microsoft/BuildXL/pull/98#discussion_r272773954) [](ancestors = 272773954)

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
dcwuser: Fixed

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -153,97 +156,174 @@ public async Task<GetStatsResponse> GetStatsAsync(GetStatsRequest request, Cance
...
-                        new CopyFileResponse
-                        {
-                            Header = new ResponseHeader(startTime, false, (int)CopyFileResult.ResultCode.SourcePathError, $"'{name}' is an invalid cache name."),
-                        });
-                }
+                    // Figure out response headers.
+                    CopyCompression compression = CopyCompression.None;
+                    Metadata headers = new Metadata();
+                    headers.Add("ContentHash", hash.ToString());
+                    switch (result.Code)
```
SergeyTeplyakov: maybe it make sense to add a default section that will throw.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -83,74 +87,166 @@ public async Task<BoolResult> ShutdownAsync(Context context)
...
-                {
-                    return await CopyToAsync(context, contentHash, stream, cancellationToken, fileSize, enableCompression);
-                }
-            }
-            catch (Exception ex) when (ex is UnauthorizedAccessException || ex is InvalidOperationException)
-            {
-                return new CopyFileResult(CopyFileResult.ResultCode.DestinationPathError, ex, $"Error copying to file {destinationPath.Path}");
-            }
+            Func<Stream> streamFactory = () => new FileStream(destinationPath.Path, FileMode.Create, FileAccess.Write, FileShare.None, ContentStore.Grpc.CopyConstants.DefaultBufferSize, FileOptions.Asynchronous | FileOptions.SequentialScan);
+            return CopyToAsync(context, hash, streamFactory, ct);
```
dannyvv: >         [](start = 65, length = 8)

stray spaces

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
dcwuser: Fixed

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
SergeyTeplyakov: I'm curious, should we switch to the new convention or should we keep using the old "unconventional" style?

---
In reply to: [272773030](https://github.com/Microsoft/BuildXL/pull/98#discussion_r272773030) [](ancestors = 272773030)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -83,74 +87,166 @@ public async Task<BoolResult> ShutdownAsync(Context context)
...
+                CopyFileRequest request = new CopyFileRequest()
                 {
                     TraceId = context.Id.ToString(),
-                    HashType = (int)contentHash.HashType,
-                    ContentHash = contentHash.ToByteString(),
-                    // TODO: If `Drive` is expected to be the drive of the file on the source machine, then this should have nothing to do with the destination's drive
-                    Drive = "B",
+                    HashType = (int) hash.HashType,
+                    ContentHash = hash.ToByteString(),
+                    Drive = "X",
```
dannyvv: > "X" [](start = 27, length = 4)

Not sure what this one means... Just checking this isn't `X:\` on windows... :)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -83,74 +87,166 @@ public async Task<BoolResult> ShutdownAsync(Context context)
...
+                }
 
-                bytesReceived = await StreamContentAsync(stream, response.ResponseStream);
+                // Check header collection for server-side errors.
+                string exception = null;
+                string message = null;
+                CopyCompression compression = CopyCompression.None;
+                foreach(Metadata.Entry header in headers)
+                {
+                    if (header.Key == "exception")
```
SergeyTeplyakov: maybe use switch? In this case we check the key over and over again. Not very important, but maybe will be easier to read.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -83,74 +87,166 @@ public async Task<BoolResult> ShutdownAsync(Context context)
...
+                    if (header.Key == "message")
+                    {
+                        message = header.Value;
+                    }
+                    if (header.Key == "compression")
+                    {
+                        Enum.TryParse<CopyCompression>(header.Value, out compression);
+                    }
+                }
+                if (exception != null)
```
SergeyTeplyakov: nit: empty line before this block.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -83,74 +87,166 @@ public async Task<BoolResult> ShutdownAsync(Context context)
...
+            Debug.Assert(replyStream != null);
+
+            long chunks = 0L;
+            long bytes = 0L;
+            using (BufferedReadStream grpcStream = new BufferedReadStream(async () =>
             {
-                CopyFileResponse oneOfManyReply = replyStream.Current;
-                bytesReceived += oneOfManyReply.Content.Length;
-                oneOfManyReply.Content.WriteTo(fileStream);
+                if (await replyStream.MoveNext(ct).ConfigureAwait(false))
```
dannyvv: Would a cancellation token check make sense here?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -83,74 +87,166 @@ public async Task<BoolResult> ShutdownAsync(Context context)
...
+                        message = header.Value;
+                    }
+                    if (header.Key == "compression")
+                    {
+                        Enum.TryParse<CopyCompression>(header.Value, out compression);
+                    }
+                }
+                if (exception != null)
+                {
+                    Debug.Assert(message != null);
```
SergeyTeplyakov: Switch to Contract?

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedWriteStream.cs
dcwuser: Fixed

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -83,74 +87,166 @@ public async Task<BoolResult> ShutdownAsync(Context context)
...
             }
 
-            if (bytesReceived == 0)
-            {
-                return new CopyFileResult(CopyFileResult.ResultCode.SourcePathError, $"Received {bytesReceived} bytes for {contentHash}. Source file does not exist");
-            }
-            else if (fileSize >= 0 && bytesReceived != fileSize)
+        }
+
+        private async Task<(long, long)> StreamContentAsync(Stream targetStream, IAsyncStreamReader<CopyFileResponse> replyStream, CancellationToken ct = default(CancellationToken))
```
SergeyTeplyakov: nit: use names for tuples?

SergeyTeplyakov: >StreamContentAsync [](start = 41, length = 18)

nit: I would prefer not to use default args if don't need them.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dcwuser: Fixed

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dcwuser: Any exception that appears in the headers is part of the controlled logic in the service implementation (and happened before we started to stream). An uncaught exception will appear as an RpcException on the client. My preferred logic would be to just let that happen on the client, and log uncaught exceptions on the service.

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
SergeyTeplyakov: Github UI or codeflow messed up: I meant to use summary section here.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dcwuser: Fixed

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dcwuser: I assume s/pull/pool. Yes, eventually, but I didn't want to deal with that here.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dcwuser: Ack, sorry, fixed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dcwuser: Drive isn't used, but I didn't want to break calling contract by removing it. Will revisit in next iteration.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dcwuser: GRPC's MoveNext(ct) cancels if ct.IsCancellationRequested AND transmits that cancellation to the service, so I actually want the ct to get down into that method.

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
```diff
@@ -0,0 +1,110 @@
...
+                int copyCount = Math.Min(count - totalCount, _storage.Length - _readPointer);
+                Array.Copy(_storage, _readPointer, buffer, writePointer, copyCount);
+                _readPointer += copyCount;
+                writePointer += copyCount;
+                totalCount += copyCount;
+                _position += copyCount;
+
+                Debug.Assert(_readPointer <= _storage.Length);
+
+                if (_readPointer == _storage.Length)
```
SergeyTeplyakov: As we discussed offline, this code deserves some comments:)

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedWriteStream.cs
dcwuser: Fixed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dcwuser: Fixed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dcwuser: Fixed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dcwuser: To resolve this problem, went ahead and implemented next step of iterating through all stores, which naturally handled case of there being none.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
dcwuser: Fixed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dcwuser: Fixed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dcwuser: Fixed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dcwuser: Boo, hiss. I want a system that allows me to do asserts in a debug build that are not checked in retail.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dcwuser: Fixed.

Public/Src/Cache/ContentStore/Library/Service/Grpc/BufferedReadStream.cs
dcwuser: Added comments.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dcwuser: Went ahead and dealt with removing drive entirely in next iteration.

PR #100, opened 2019-04-05
Update dscript plugin to work with the latest VSCode

smera:
Fix the DScript VSCode plugin by moving to the latest packages and fixing some renaming issues


dannyvv:
Please do not check in buildxl.Png. Please add it to branding. The build script copies the png already to the out folder where the VSCode pacakge is constructed.

smera:
Removed it. The one coming from branding should be good enough then

---
In reply to: [480457436](https://github.com/Microsoft/BuildXL/pull/100#issuecomment-480457436) [](ancestors = 480457436)

PR #101, opened 2019-04-06
Bump minimum required version of MSBuild

smera:
Bump the minimum required version of MSBuild for BuildXL so it matches the latest API/cross targeting functionality


dannyvv:
@smera: Note you can now rerun from VSTS as well. It is under the tripple-dot in the top-right.

---
In reply to: [480458275](https://github.com/Microsoft/BuildXL/pull/101#issuecomment-480458275) [](ancestors = 480458275)

PR #102, opened 2019-04-08
Make purging on startup optional for quota keeper.

SergeyTeplyakov:
If the cache restarts during the build (which is a normal procedure), the cache may decide to start content eviction process at a startup when the cache is full.

Effectively, there is a race condition at startup when one part of the cache reloads the session state and tries to re-pin the content that was used by a previous service instance and another part of the cache tries to evict it. In some cases an important content can be evicted causing the build to fail.

This PR adds mitigation by exposing a configuration option that prevents purging on startup.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
+++ b/Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
@@ -1105,9 +1105,9 @@ namespace BuildXL.Cache.ContentStore.Stores
         /// <summary>
         ///     Complete all pending/background operations.
         /// </summary>
-        public async Task SyncAsync(Context context)
+        public async Task SyncAsync(Context context, bool purge = true)
         {
-            await _quotaKeeper.SyncAsync(context);
+            await _quotaKeeper.SyncAsync(context, purge);
 
             // Ensure there are no pending LRU updates.
             await ContentDirectory.SyncAsync();
--- a/Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
+++ b/Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
@@ -672,9 +672,8 @@ namespace BuildXL.Cache.ContentStore.Stores
             var quotaKeeperConfiguration = QuotaKeeperConfiguration.Create(
                 Configuration,
                 _distributedEvictionSettings,
-                size,
-                _settings.StartPurgingAtStartup,
-                _settings.UseLegacyQuotaKeeperImplementation);
+                _settings,
+                size);
             _quotaKeeper = QuotaKeeper.Create(
                 FileSystem,
                 _tracer,
--- a/Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
+++ b/Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
@@ -669,15 +669,18 @@ namespace BuildXL.Cache.ContentStore.Stores
                 GetPinnedSize,
                 _nagleQueue);
 
+            var quotaKeeperConfiguration = QuotaKeeperConfiguration.Create(
+                Configuration,
+                _distributedEvictionSettings,
+                size,
+                _settings.StartPurgingAtStartup,
+                _settings.UseLegacyQuotaKeeperImplementation);
             _quotaKeeper = QuotaKeeper.Create(
                 FileSystem,
                 _tracer,
-                Configuration,
-                size,
                 ShutdownStartedCancellationToken,
                 this,
-                _distributedEvictionSettings,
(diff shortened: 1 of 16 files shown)


Public/Src/Cache/ContentStore/Library/Stores/QuotaManagement/LegacyQuotaKeeper.cs
```diff
@@ -155,7 +157,7 @@ protected override void DisposeCore()
         public override async Task SyncAsync(Context context)
         {
             // Ensure there are no pending requests.
-            var emptyRequest = new Request<QuotaKeeperRequest, string>(QuotaKeeperRequest.Reserve(0));
+            var emptyRequest = new Request<QuotaKeeperRequest, string>(QuotaKeeperRequest.Synchronize());
```
JuanCarlosGI: Maybe change variable name?

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -669,15 +669,18 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
                 GetPinnedSize,
                 _nagleQueue);
 
+            var quotaKeeperConfiguration = QuotaKeeperConfiguration.Create(
+                Configuration,
+                _distributedEvictionSettings,
+                size,
+                _settings.StartPurgingAtStartup,
```
kewilli: >_settings [](start = 16, length = 9)

Why not just pass `_settings` here?

Public/Src/Cache/ContentStore/Library/Stores/QuotaManagement/LegacyQuotaKeeper.cs
```diff
@@ -80,9 +79,12 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
             _reserveTask = Task.Run(() => Reserve(new Context(context)));
             _taskTracker = new BackgroundTaskTracker(Component, new Context(context));
 
-            // Start purging immediately on startup to clear out residual content in the cache
-            // over the cache quota
-            StartPurging();
+            if (PurgeAtStartup)
```
kewilli: >PurgeAtStartup [](start = 16, length = 14)

Can we emit some diagnostic when we choose to not purge at startup?

kewilli: It seems like we would have sufficient logging for when we do purge at startup.

---
In reply to: [273233305](https://github.com/Microsoft/BuildXL/pull/102#discussion_r273233305) [](ancestors = 273233305)

SergeyTeplyakov: That makes a lot of sense! Will add.

---
In reply to: [273233305](https://github.com/Microsoft/BuildXL/pull/102#discussion_r273233305) [](ancestors = 273233305)

Public/Src/Cache/ContentStore/Library/Stores/QuotaManagement/QuotaKeeperConfiguration.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <see cref="ContentStoreConfiguration.InitialElasticSize"/>.
+        /// </summary>
+        public MaxSizeQuota InitialElasticSize { get; private set; }
+
+        /// <summary>
+        /// <see cref="ContentStoreConfiguration.HistoryWindowSize"/>.
+        /// </summary>
+        public int? HistoryWindowSize { get; private set; }
+
+        /// <nodoc />
```
kewilli: ><nodoc /> [](start = 12, length = 9)

`<see cref="DistributedEvictionSettings"/>` ?

Public/Src/Cache/ContentStore/Library/Stores/QuotaManagement/QuotaKeeperV2.cs
```diff
@@ -108,10 +107,13 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
                 () => ProcessReserveRequestsAsync(context.CreateNested()),
                 TaskCreationOptions.LongRunning).Unwrap();
 
-            // Start purging immediately on startup to clear out residual content in the cache
-            // over the cache quota
-            const string Operation = "PurgeRequest";
-            SendPurgeRequest(context, "Startup").FireAndForget(context, Operation);
+            if (PurgeAtStartup)
```
kewilli: >PurgeAtStartup [](start = 16, length = 14)

Ditto here

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
SergeyTeplyakov: God idea.

---
In reply to: [273230438](https://github.com/Microsoft/BuildXL/pull/102#discussion_r273230438) [](ancestors = 273230438)

PR #104, opened 2019-04-08
Retry adding file to cache if the operation fails

olkononenko:
There are some occasional (random?) build failures BXL cannot put a file into cache (another process has a lock). Adding retry logic, hopefully, it will address the build failures.

[AB#1462850](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1462850)


Public/Src/Utilities/Native/IO/Helpers.cs
```diff
@@ -135,12 +138,70 @@ public static class Helpers
...
+                isLastRetry =>
+                {
+                    possiblyResult = work(isLastRetry);
+                    return possiblyResult.Value.Succeeded;
+                },
+                logExceptions, rethrowExceptions, numberOfAttempts, initialTimeoutMs, postTimeoutMultiplier);
+
+            if (!possiblyResult.HasValue)
+            {
+                return new Failure<string>($"Exhausted all of the '{numberOfAttempts}' retry attempts.{(logExceptions ? " Look at the log for more details." : string.Empty)}");
```
SergeyTeplyakov: I think if possiblyResult is null then there is a bug in the code (like the number of attempts is 0.

I think we can enforce this by using `Contract.Assert(possiblyResult != null)`.

SergeyTeplyakov: Discussed offline.

---
In reply to: [273276304](https://github.com/Microsoft/BuildXL/pull/104#discussion_r273276304) [](ancestors = 273276304)

Public/Src/Utilities/Native/IO/Helpers.cs
```diff
@@ -135,12 +138,70 @@ public static class Helpers
...
                 }
             }
 
             return done;
         }
+
+        /// <summary>
+        /// A generic version of <see cref="RetryOnFailure"/>.
+        /// </summary>
+        public static Possible<TResult, Failure> RetryOnFailure<TResult>(
```
dannyvv: >RetryOnFailure [](start = 49, length = 14)

Should the RetryOnFilaure from above be implemented using this one?

dannyvv: never mind, this is not async....

---
In reply to: [273279329](https://github.com/Microsoft/BuildXL/pull/104#discussion_r273279329) [](ancestors = 273279329)

Public/Src/Utilities/Native/IO/Helpers.cs
```diff
@@ -63,6 +64,7 @@ public static class Helpers
             int postTimeoutMultiplier = DefaultPostTimeoutMultiplier)
         {
             Contract.Requires(work != null);
+            Contract.Requires(logExceptions || rethrowExceptions, "A caught exception must be either rethrown or logged.");
```
kewilli: >logExceptions || rethrowExceptions [](start = 30, length = 34)

`logExceptions ^ rethrowExceptions` ? Setting both to true does not log - it only throws.

Public/Src/Utilities/Native/IO/Helpers.cs
```diff
@@ -135,12 +138,76 @@ public static class Helpers
...
+                Contract.Assert(false, $"An operation resulted in an unhandled exception in all of the '{numberOfAttempts}' retry attempts");
+            }
+
+            return possiblyResult.Value;
+        }
+
+        /// <summary>
+        /// Async version of <see cref="RetryOnFailure"/>.
+        /// </summary>
+        public static async Task<Possible<TResult, Failure>> RetryOnFailureAsync<TResult>(
```
kewilli: >RetryOnFailureAsync [](start = 61, length = 19)

Could you add a unit test for this utility?

kewilli:
>         /// Default post timeout multiplier.

Could you expand on this doc? What does "post timeout" mean? What does it multiply?

---
Refers to: Public/Src/Utilities/Native/IO/Helpers.cs:29 in e80a371. [](commit_id = e80a371faea58aaca428d974265181436554cfc0, deletion_comment = False)

Public/Src/Utilities/Native/IO/Helpers.cs
```diff
@@ -135,12 +138,76 @@ public static class Helpers
...
+            if (!possiblyResult.HasValue)
+            {                
+                Contract.Assert(false, $"An operation resulted in an unhandled exception in all of the '{numberOfAttempts}' retry attempts");
+            }
+
+            return possiblyResult.Value;
+        }
+
+        /// <summary>
+        /// Async version of <see cref="RetryOnFailure"/>.
```
kewilli: >RetryOnFailure [](start = 40, length = 14)

There are two methods named `RetryOnFailure` here. Is this binding to the right one?

Public/Src/Utilities/Native/IO/Helpers.cs
kewilli: >RetryOnFailureAsync [](start = 61, length = 19)

`Contract.Requires(work != null)` ?

Public/Src/Utilities/Native/IO/Helpers.cs
```diff
@@ -135,12 +138,76 @@ public static class Helpers
...
+            // Since work returns Possible<TResult, Failure>, we expect it to gracefully handle any exception
+            // (i.e., return a failure instead of throwing an exception). However, there is no guarantee that
+            // it would behave this way. 
+            if (!possiblyResult.HasValue)
+            {                
+                Contract.Assert(false, $"An operation resulted in an unhandled exception in all of the '{numberOfAttempts}' retry attempts");
+            }
+
+            return possiblyResult.Value;
+        }
```
narasamdya: You don't need these two methods. See method `TryMaterializeAsync` in `Public\Src\Engine\Cache.Plugin.CacheCore\CacheCoreArtifactContentCache.cs`

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
```diff
@@ -767,9 +771,13 @@ public bool TrackAbsentPath(AbsolutePath trackedParentPath, AbsolutePath absentC
             }
             else
             {
-                Possible<ContentHash, Failure> possiblyStored = await cache.TryStoreAsync(
-                    fileRealizationModes,
-                    expandedPath);
+                Possible<ContentHash, Failure> possiblyStored = await Helpers.RetryOnFailureAsync(
+                    async _ =>
+                    {
+                        return await cache.TryStoreAsync(
```
narasamdya: >TryStoreAsync [](start = 43, length = 13)

Instead of doing it here, I think we should do it in `Public\Src\Engine\Cache.Plugin.CacheCore\CacheCoreArtifactContentCache.cs`.

PR #105, opened 2019-04-08
Update Kusto crash query functions

anniefu:
Update Kusto functions to remove deprecated "KnownBug" field and prioritize "Crash" bucket over "MissingDominoCompletionEvent"
alphabetically


mpysson:
>     "MissingDominoCompletionEvent")

This gives precedence to Crash over MissinDominoCompletionEvent. I guess the problem is when summarizing an entire build session.

I think we'll still get into trouble with your fix for internal error buckets that are alphabetically after MissingDominoCompletionEvent. Maybe add a prefix to Crash and MissingDomino.. to froce them to the head and tail of the sort order

---
Refers to: Shared/Scripts/kusto/functions/AnnotatedSessions.csl:81 in 165d303. [](commit_id = 165d303d89d9fe37161fbb671f98d0eb5da4e2ea, deletion_comment = False)

PR #106, opened 2019-04-08
Better unauthorized access

SergeyTeplyakov:
Today, if the CASaaS operation fails with `UnauthorizedAccessException` it is hard to tell which process is to blame.

Here is an example of the error:

```
b0f230db-0c02-4492-b7ac-4187db9caf45 Critical error occurred: DistributedContentSession.PutFileAsync stop 99.6744ms. result=[Error=[UnauthorizedAccessException: System.UnauthorizedAccessException: Unable to open a handle to 'd:\dbs\el\dF\Build\MetaBuild\oacrEnvironment\math_solverengine_x64_ship.json' as Open with Read. last error: [The process cannot access the file because it is being used by another process.] (error code 32)
```

This PR changes the code that throws `UnauthorizedAccessException` from `PassThroughFileSystem` members to also include a process that holds a handle to a file (if possible).

In this case the error message looks like this:

```
"Unable to open a handle to 'C:\Users\seteplia\AppData\Local\Temp\CloudStore\36aa81cc69ea\parent2\dst' as Create with Write. last error: [The process cannot access the file because it is being used by another process.] (error code 32). Handle was used by 'MyProcess' with PID '15852'.
```

Bug in AzDev: #1478669

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
+++ b/Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
@@ -146,7 +146,7 @@ namespace BuildXL.Cache.ContentStore.FileSystem
 
                 if (!createOrOpenResult.Succeeded)
                 {
-                    throw ThrowLastWin32Error($"Failed to create or open file {path.ToString()} to get its ID. Status: {createOrOpenResult.Status}");
+                    throw ThrowLastWin32Error(path.ToString(), $"Failed to create or open file {path} to get its ID. Status: {createOrOpenResult.Status}");
                 }
 
                 return (ulong)handle.DangerousGetHandle().ToInt64();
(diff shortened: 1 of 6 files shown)


Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
```diff
@@ -279,7 +279,11 @@ public void DeleteFile(AbsolutePath path)
                 // 0x20 is shared violation.
                 if (e.InnerException is NativeWin32Exception win32 && win32.ErrorCode == 0x20)
                 {
-                    throw new UnauthorizedAccessException(e.Message, e);
+                    var extraMessage = FileUtilities.TryFindOpenHandlesToFile(path.ToString(), out var info, printCurrentFilePath: false)
```
dannyvv: >TryFindOpenHandlesToFile [](start = 53, length = 24)

This is now the third instance I see this pattern. Do we need a single `FileUtilities.MakeUnauthoriazedAccessExceptionUsable(.....)`

Public/Src/Cache/ContentStore/Library/Stores/QuotaManagement/QuotaKeeper.cs
```diff
@@ -34,37 +33,38 @@ public enum QuotaKeeperCounters
     /// </summary>
     public abstract class QuotaKeeper : StartupShutdownBase
     {
+        /// <nodoc />
+        protected readonly bool PurgeAtStartup;
```
dannyvv: >PurgeAtStartup [](start = 32, length = 14)

Are this changes intended in this PR, They look like the same changes from a differnt PR...

SergeyTeplyakov: Removed them. Thanks.

PR #114, opened 2019-04-09
Fix XLG serialization crash when strong fingerprint computation fails

anniefu:
During cache lookup strong fingerprint computation, if ObservedInputProcessingStatus is not "Success", the strong fingerprint cannot be computed and a cache miss is forced.

In this case, make sure the strong fingerprint inputs retrieved from cache are logged to the XLG rather than some invalid defaults, even if the cache data could not be used to compute a strong fingerprint.

Fixes [AB#1504534](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1504534)

### DIFF
--- a/Public/Src/Engine/Scheduler/PipExecutor.cs
+++ b/Public/Src/Engine/Scheduler/PipExecutor.cs
@@ -1835,9 +1835,9 @@ namespace BuildXL.Scheduler
                                 }
 
                                 var pathSet = maybePathSet.Value;
-                                (ObservedInputProcessingResult observedInputProcessingResult, StrongContentFingerprint? strongContentFingerPrint, ObservedPathSet pathSetUsed, ContentHash pathSetHashUsed)
+                                (bool succeeded, ObservedInputProcessingResult observedInputProcessingResult, StrongContentFingerprint? strongContentFingerprint, ObservedPathSet pathSetUsed, ContentHash pathSetHashUsed)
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1835,9 +1835,9 @@ private static void ReportFileAccesses(ExecutionResult processExecutionResult, F
                                 }
 
                                 var pathSet = maybePathSet.Value;
-                                (ObservedInputProcessingResult observedInputProcessingResult, StrongContentFingerprint? strongContentFingerPrint, ObservedPathSet pathSetUsed, ContentHash pathSetHashUsed)
+                                (bool succeeded, ObservedInputProcessingResult observedInputProcessingResult, StrongContentFingerprint? strongContentFingerprint, ObservedPathSet pathSetUsed, ContentHash pathSetHashUsed)
```
dannyvv: >succeeded [](start = 38, length = 9)

I dont' see this variable used...

anniefu: It is sneakily hiding on the right side of line 1850

---
In reply to: [273725711](https://github.com/Microsoft/BuildXL/pull/114#discussion_r273725711) [](ancestors = 273725711)

PR #116, opened 2019-04-09
Workaround for MSBuild reporting duplicate edges

smera:
MSBuild static graph APIS are sometimes reporting duplicate edges. Temporarily make sure the MSBuild resolver dedups these.

### DIFF
--- a/Public/Src/FrontEnd/MsBuild.Serialization/ProjectWithPredictions.cs
+++ b/Public/Src/FrontEnd/MsBuild.Serialization/ProjectWithPredictions.cs
@@ -48,6 +48,7 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         public PredictedTargetsToExecute PredictedTargetsToExecute { get; }
 
         /// <nodoc/>
+        [JsonIgnore()]
         public IReadOnlyCollection<ProjectWithPredictions<TPathType>> ProjectReferences
         {
             get
--- a/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
+++ b/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
@@ -455,6 +455,11 @@ namespace BuildXL.FrontEnd.MsBuild
             processBuilder.AddOutputFile(logDirectory.Combine(PathTable, "msbuild.err"), FileExistence.Optional);
             processBuilder.AddOutputFile(logDirectory.Combine(PathTable, "msbuild.prf"), FileExistence.Optional);
 
+            if (m_resolverSettings?.EnableBinLogTracing == true)
+            {
+                processBuilder.AddOutputFile(logDirectory.Combine(PathTable, "msbuild.binlog"), FileExistence.Optional);
+            }
+
             // Unless the legacy non-isolated mode is explicitly specified, the project builds in isolation, and therefore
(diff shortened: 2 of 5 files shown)


Public/Src/FrontEnd/MsBuild.Serialization/ProjectWithPredictions.cs
```diff
@@ -48,6 +48,7 @@ public sealed class ProjectWithPredictions<TPathType>
         public PredictedTargetsToExecute PredictedTargetsToExecute { get; }
 
         /// <nodoc/>
+        [JsonIgnore()]
```
smera: >[JsonIgnore()] [](start = 8, length = 14)

Inconsequential, but without this attribute references are duplicated in the serialized format

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -455,6 +455,11 @@ private void SetProcessEnvironmentVariables(IReadOnlyDictionary<string, string>
             processBuilder.AddOutputFile(logDirectory.Combine(PathTable, "msbuild.err"), FileExistence.Optional);
             processBuilder.AddOutputFile(logDirectory.Combine(PathTable, "msbuild.prf"), FileExistence.Optional);
 
+            if (m_resolverSettings?.EnableBinLogTracing == true)
+            {
+                processBuilder.AddOutputFile(logDirectory.Combine(PathTable, "msbuild.binlog"), FileExistence.Optional);
```
smera: >"msbuild.binlog" [](start = 77, length = 16)

Inconsequential. It is always better to declare a file if we know it is going to be (uniquely) produced. Without this, the containing shared opaque takes ownership.

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -455,6 +455,11 @@ private void SetProcessEnvironmentVariables(IReadOnlyDictionary<string, string>
             processBuilder.AddOutputFile(logDirectory.Combine(PathTable, "msbuild.err"), FileExistence.Optional);
             processBuilder.AddOutputFile(logDirectory.Combine(PathTable, "msbuild.prf"), FileExistence.Optional);
 
+            if (m_resolverSettings?.EnableBinLogTracing == true)
```
jbayardo: Why use the null conditional operator? According to line 438, m_resolverSettings != null

dannyvv: :+1:

---
In reply to: [273752434](https://github.com/Microsoft/BuildXL/pull/116#discussion_r273752434) [](ancestors = 273752434)

smera: No real reasons, copy and paste :)

---
In reply to: [273753511](https://github.com/Microsoft/BuildXL/pull/116#discussion_r273753511) [](ancestors = 273753511,273752434)

smera:
> LGTM, though a test that fails when I fix this in msbuild would be nice, to know when to undo this change. It's probably a good idea to keep the bxl graph a 1-to-1 mapping to the msbuild graph as much as possible.

Done!

PR #120, opened 2019-04-10
Copy between separate Win machines

kewilli:
Open GRPC server to listen to all external requests.

[AB#1504491](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1504491)

When attempting to copy files between separate Windows machines, the GRPC client failed to connect to the GRPC server. The connection succeeded when using "localhost" as the target host, but failed when using the server's machine name or ip address. To investigate, a simple "Hello" verb was added to test client-server connections.

In the end, the error was in the construction of the ServerPort. It was previously only exposed to local communication. With this change, the server can respond to communication from external sources.

### DIFF
--- a/Public/Src/Cache/ContentStore/App/Hello.cs
+++ b/Public/Src/Cache/ContentStore/App/Hello.cs
@@ -1,0 +1,50 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Threading;
+using BuildXL.Cache.ContentStore.Exceptions;
+using BuildXL.Cache.ContentStore.Hashing;
+using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
+using BuildXL.Cache.ContentStore.Interfaces.Tracing;
+using BuildXL.Cache.ContentStore.Service;
+using BuildXL.Cache.ContentStore.Service.Grpc;
+using Microsoft.Practices.TransientFaultHandling;
+using CLAP;
+using BuildXL.Cache.ContentStore.Sessions;
+using Grpc.Core;
+using ContentStore.Grpc;
+using System.Linq;
+
+namespace BuildXL.Cache.ContentStore.App
+{
+    internal sealed partial class Application
+    {
+        public static ChannelOption[] DefaultChannelOptions = new ChannelOption[] { new ChannelOption(ChannelOptions.MaxSendMessageLength, -1), new ChannelOption(ChannelOptions.MaxReceiveMessageLength, -1) };
+
+        /// <summary>
+        /// Attempt to connect to a grpc port and send 'Hello'
+        /// </summary>
+        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
+        [Verb(Description = "Send 'Hello' to another CASaaS")]
+        internal void Hello(
+            string hash,
+            [Required, Description("Machine to copy from")] string host,
+            [Description("The GRPC port"), DefaultValue(0)] int grpcPort)
+        {
+            Initialize();
+
+            if (grpcPort <= 0)
+            {
+                throw new Exception("Must define grpc port greater than 0");
+            }
+
+            var _channel = new Channel(host, grpcPort, ChannelCredentials.Insecure, DefaultChannelOptions);
+            var _client = new ContentServer.ContentServerClient(_channel);
+            var helloResponse = _client.Hello(new HelloRequest(), new CallOptions(deadline: DateTime.UtcNow + TimeSpan.FromSeconds(2)));
+
+            Console.WriteLine(helloResponse.ToString());
+        }
+    }
+}
--- a/Public/Src/Cache/ContentStore/App/Hello.cs
+++ b/Public/Src/Cache/ContentStore/App/Hello.cs
@@ -21,8 +21,6 @@ namespace BuildXL.Cache.ContentStore.App
 {
     internal sealed partial class Application
     {
-        public static ChannelOption[] DefaultChannelOptions = new ChannelOption[] { new ChannelOption(ChannelOptions.MaxSendMessageLength, -1), new ChannelOption(ChannelOptions.MaxReceiveMessageLength, -1) };
-
         /// <summary>
         /// Attempt to connect to a grpc port and send 'Hello'
         /// </summary>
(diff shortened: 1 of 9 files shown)


Public/Src/Cache/ContentStore/App/Hello.cs
```diff
@@ -0,0 +1,50 @@
...
+    internal sealed partial class Application
+    {
+        public static ChannelOption[] DefaultChannelOptions = new ChannelOption[] { new ChannelOption(ChannelOptions.MaxSendMessageLength, -1), new ChannelOption(ChannelOptions.MaxReceiveMessageLength, -1) };
+
+        /// <summary>
+        /// Attempt to connect to a grpc port and send 'Hello'
+        /// </summary>
+        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
+        [Verb(Description = "Send 'Hello' to another CASaaS")]
+        internal void Hello(
```
dannyvv: >Hello [](start = 22, length = 5)

Are we sure we want to call this Hello? Sounds very old-school... `Ping` , `Initialzie`, `IsAlive`, `IsActive` seems more appropriate?

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
```diff
@@ -16,7 +18,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.Utilities
     /// </summary>
     public class GrpcDistributedPathTransformer : IAbsolutePathTransformer
     {
-        private static readonly string _localMachineName = Environment.MachineName;
+        private static readonly string _localIpAddress = Dns.GetHostEntry(Dns.GetHostName()).AddressList.FirstOrDefault(ip => ip.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork).ToString();
```
dannyvv: >FirstOrDefault [](start = 105, length = 14)

If I recall you might need to filter out loopback devices and I don't know about the connections in datacenters but there is a chance that there are multiple NIC's with different bandwidth. Should we sort by connection speed rather than picking the first?

Public/Src/Cache/ContentStore/App/Hello.cs
SergeyTeplyakov: I think this is done for consistency.

---
In reply to: [274110409](https://github.com/Microsoft/BuildXL/pull/120#discussion_r274110409) [](ancestors = 274110409)

Public/Src/Cache/ContentStore/App/Hello.cs
```diff
@@ -0,0 +1,50 @@
...
+            string hash,
+            [Required, Description("Machine to copy from")] string host,
```
SergeyTeplyakov: Machine to send Hello request to? Or something like this.

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
SergeyTeplyakov: I think this code deserves the comment to explain why Environment.MachineName is not good enough here.

SergeyTeplyakov: And this is especially "important" because we have a bunch of other places in the code where we still use Environment.MachineName. So if the new code is the right way to do that I think we should move this logic into a helper and use it everywhere. (or explain why this case is different).

---
In reply to: [274116211](https://github.com/Microsoft/BuildXL/pull/120#discussion_r274116211) [](ancestors = 274116211)

SergeyTeplyakov:
I think it make sense to have a short description for every PR with an explanation of what is going on.

I can see the description: "Open GRPC server to listen to all external requests."

But I can't see how this is implemented in the code... 

By looking into the change it seems that this PR creates a new verb (Hello) and contains a code cleanup...

Public/Src/Cache/ContentStore/App/Hello.cs
kewilli: Correct, this is for consistency. 

---
In reply to: [274115371](https://github.com/Microsoft/BuildXL/pull/120#discussion_r274115371) [](ancestors = 274115371,274110409)

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
kewilli: Both ways work, so I'll return to MachineName.

---
In reply to: [274116694](https://github.com/Microsoft/BuildXL/pull/120#discussion_r274116694) [](ancestors = 274116694,274116211)

kewilli: Moving back to MachineName

---
In reply to: [274111091](https://github.com/Microsoft/BuildXL/pull/120#discussion_r274111091) [](ancestors = 274111091)

kewilli:
Sure, I'll update the description with a short novella

---
In reply to: [481826342](https://github.com/Microsoft/BuildXL/pull/120#issuecomment-481826342) [](ancestors = 481826342)

Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
```diff
@@ -7,23 +7,24 @@
 using System.Diagnostics;
 using System.Diagnostics.ContractsLight;
 using System.Linq;
+using System.Net;
```
thpha: Do we need all these extra libraries now?

Public/Src/Cache/ContentStore/App/Hello.cs
```diff
@@ -0,0 +1,50 @@
...
+            if (grpcPort <= 0)
+            {
+                throw new Exception("Must define grpc port greater than 0");
+            }
+
+            var _channel = new Channel(host, grpcPort, ChannelCredentials.Insecure, DefaultChannelOptions);
+            var _client = new ContentServer.ContentServerClient(_channel);
+            var helloResponse = _client.Hello(new HelloRequest(), new CallOptions(deadline: DateTime.UtcNow + TimeSpan.FromSeconds(2)));
+
+            Console.WriteLine(helloResponse.ToString());
```
thpha: Change to _logger for consistency with others. Nit: Could print out a more intentional message here. Like "Hello was successful" or throw a CacheException when it fails.

Public/Src/Cache/ContentStore/App/Hello.cs
```diff
@@ -0,0 +1,50 @@
...
+using BuildXL.Cache.ContentStore.Sessions;
+using Grpc.Core;
+using ContentStore.Grpc;
+using System.Linq;
+
+namespace BuildXL.Cache.ContentStore.App
+{
+    internal sealed partial class Application
+    {
+        public static ChannelOption[] DefaultChannelOptions = new ChannelOption[] { new ChannelOption(ChannelOptions.MaxSendMessageLength, -1), new ChannelOption(ChannelOptions.MaxReceiveMessageLength, -1) };
```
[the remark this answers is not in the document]
kewilli: We do not. Removing!

---
In reply to: [274156509](https://github.com/Microsoft/BuildXL/pull/120#discussion_r274156509) [](ancestors = 274156509)

Public/Src/Cache/ContentStore/App/Hello.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// Attempt to connect to a grpc port and send 'Hello'
+        /// </summary>
+        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
+        [Verb(Description = "Send 'Hello' to another CASaaS")]
+        internal void Hello(
+            string hash,
+            [Required, Description("Machine to send Hello request to")] string host,
+            [Description("GRPC port on the target machine"), DefaultValue(0)] int grpcPort)
```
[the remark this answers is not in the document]
kewilli: `grpcPortFileName`? Can do.

---
In reply to: [274156826](https://github.com/Microsoft/BuildXL/pull/120#discussion_r274156826) [](ancestors = 274156826)

Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
kewilli: We only needed to add `System.Net`. I just sorted the others.

---
In reply to: [274158652](https://github.com/Microsoft/BuildXL/pull/120#discussion_r274158652) [](ancestors = 274158652)

thpha: Gotcha. Cool!

PR #122, opened 2019-04-10
Invoke scrubber for integration tests

almiliMSFT:
Scrubber is always invoked in regular builds, so it should be invoked by the integration test setup too.

[AB#1411172](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1411172)

### DIFF
--- a/Public/Src/Engine/Dll/EngineSchedule.cs
+++ b/Public/Src/Engine/Dll/EngineSchedule.cs
@@ -37,6 +37,8 @@ using BuildXL.Cache.ContentStore.Hashing;
 using static BuildXL.Utilities.FormattableStringEx;
 using Logger = BuildXL.Engine.Tracing.Logger;
 using SchedulerLogger = BuildXL.Scheduler.Tracing.Logger;
+using System.Threading;
+using JetBrains.Annotations;
 
 namespace BuildXL.Engine
 {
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -806,19 +810,95 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
...
-                    tempDirectoryCleaner: m_tempCleaner);
+                    tempDirectoryCleaner: tempCleaner);
 
                 Logger.Log.ScrubbingSharedOpaquesStarted(loggingContext);
-                scrubber.RemoveExtraneousFilesAndDirectories(Context.CancellationToken);
+                scrubber.RemoveExtraneousFilesAndDirectories(scheduler.Context.CancellationToken);
+            }
+        }
+
+        internal static IReadOnlyList<string> GetNonScrubbablePaths(
```
almiliMSFT: same as before, just moved from `Engine` to `EngineSchedule` so that it can be called from integration tests

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SchedulerIntegrationTestBase.cs
```diff
@@ -31,6 +31,7 @@
 using Xunit.Abstractions;
 using AssemblyHelper = BuildXL.Utilities.AssemblyHelper;
 using ProcessOutputs = BuildXL.Pips.Builders.ProcessOutputs;
+using System.Threading;
```
dannyvv: >stem.Threadi [](start = 8, length = 12)

nit: Didn't see used and System.* should be at the top of the usings according to our .formatting style file

almiliMSFT: fixed

PR #123, opened 2019-04-10
Add Default clause to SwitchExpression of DScript

dannyvv:


Public/Src/FrontEnd/TypeScript.Net/TypeScript.Net/Diagnostics/Errors.cs
```diff
@@ -228,8 +228,9 @@ public static class Errors
         public static readonly IDiagnosticMessage The_body_of_an_if_statement_cannot_be_the_empty_statement = new DiagnosticMessage { Code = 1313, Category = DiagnosticCategory.Error, Key = "The_body_of_an_if_statement_cannot_be_the_empty_statement_1313", Message = "The body of an 'if' statement cannot be the empty statement." };
         public static readonly IDiagnosticMessage Switch_expression_clause_expected = new DiagnosticMessage { Code = 1314, Category = DiagnosticCategory.Error, Key = "Switch_expression_clause_expected_1314", Message = "Switch expression clauses expected." };
         public static readonly IDiagnosticMessage Switch_expression_must_be_of_type_string = new DiagnosticMessage { Code = 1315, Category = DiagnosticCategory.Error, Key = "Switch_expression_must_be_of_type_string_1315", Message = "Switch expression only supports checking on string values. Type '{0}' is not supported." };
```
smera: >string [](start = 277, length = 6)

you are saying just 'string'. Shouldn't this be 'string or number'?

Public/Src/FrontEnd/TypeScript.Net/TypeScript.Net.UnitTests/TypeChecking/SwtichExpressionCheckerTests.cs
```diff
@@ -127,7 +127,37 @@ public void NoClauses()
...
+        }
+
+        [Fact]
+        public void OnlyDefaultClause()
+        {
+            string code =
+                @"
+let result = 'a' switch {default: 10};
+";
+            ExpectOneError(code, "Switch expression must have at least one clause that is not 'default'.");
```
smera: fine with this, but I don't think allowing this case will harm

Public/Src/FrontEnd/TypeScript.Net/TypeScript.Net/TypeChecking/Checker.cs
```diff
@@ -15925,13 +15925,23 @@ private IType CheckSwitchExpression(ISwitchExpression node, ITypeMapper contextu
                     TypeToString(expressionType));
             }
 
-            if (node.Clauses.Count == 0)
+            if (!node.Clauses.Any(clause => !clause.IsDefaultFallthrough))
```
smera: >!node.Clauses.Any(clause => !clause.IsDefaultFallthrough)) [](start = 16, length = 58)

reword in terms of .All to simplify negations?

Public/Src/FrontEnd/Script/Ast/Expressions/SwitchExpressionClause.cs
```diff
@@ -16,6 +16,9 @@ namespace BuildXL.FrontEnd.Script.Expressions
     /// </summary>
     public class SwitchExpressionClause : Expression
     {
+        /// <nodoc/>
+        public bool IsDefaultFallthrough { get; set; }
```
smera: >public bool IsDefaultFallthrough { get; set; } [](start = 8, length = 46)

nit: you could subclass and avoid the extra field on every regular clause. It will also simplify the serialization logic below.
All in all, no big deal.

Public/Src/FrontEnd/Script/Ast/Expressions/SwitchExpressionClause.cs
```diff
@@ -36,18 +39,40 @@ public class SwitchExpressionClause : Expression
             Expression = expression;
         }
 
+        /// <nodoc />
+        public SwitchExpressionClause(
```
smera: >SwitchExpressionClause [](start = 15, length = 22)

maybe explicitly passing IsDefaultFallThrough on construction? 
I find 'true' being the default slightly misleading. And the pattern of creating the clause and immediately setting 'false' for the other non-default case sort not super clean

Public/Src/FrontEnd/Script/Ast/Expressions/SwitchExpression.cs
```diff
@@ -90,16 +90,25 @@ protected override EvaluationResult DoEval(Context context, ModuleLiteral env, E
...
+                        return match;
+                    }
+
+                    if (expression.Equals(match))
+                    {
+                        return clause.Expression.Eval(context, env, frame);
+                    }
                 }
+
+
```
kewilli: Remove

Public/Src/FrontEnd/TypeScript.Net/TypeScript.Net/Types/Nodes/NodeInterfaces.cs
```diff
@@ -879,6 +879,9 @@ public interface ISwitchExpression : IExpression
     [NodeInfo(SyntaxKinds = new[] { SyntaxKind.SwitchExpressionClause }, NodeType = NodeType.Leaf)]
     public interface ISwitchExpressionClause : IExpression
     {
+        /// <nodoc/>
+        bool IsDefaultFallthrough { get; set; }
```
kewilli: >IsDefaultFallthrough [](start = 13, length = 20)

If anything in this PR deserves a bit more documentation, I think it would be this property :)

Public/Src/FrontEnd/TypeScript.Net/TypeScript.Net.UnitTests/TypeChecking/SwtichExpressionCheckerTests.cs
dannyvv: better stricter and relax later than the inverse :)

---
In reply to: [274186597](https://github.com/Microsoft/BuildXL/pull/123#discussion_r274186597) [](ancestors = 274186597)

Public/Src/FrontEnd/TypeScript.Net/TypeScript.Net/TypeChecking/Checker.cs
dannyvv: Open to suggestion for the expression. 
It is checking that there must exists at least one clause that is not a default... Not sure how to do with all...

---
In reply to: [274186828](https://github.com/Microsoft/BuildXL/pull/123#discussion_r274186828) [](ancestors = 274186828)

Public/Src/FrontEnd/Script/Ast/Expressions/SwitchExpressionClause.cs
dannyvv: thought about it at some point since the switch statement does that, but then I realized that it would have to expand all the visitors with another node and opted against.

---
In reply to: [274187446](https://github.com/Microsoft/BuildXL/pull/123#discussion_r274187446) [](ancestors = 274187446)

Public/Src/FrontEnd/Script/Ast/Expressions/SwitchExpression.cs
dannyvv: I miss our style checks...

---
In reply to: [274196949](https://github.com/Microsoft/BuildXL/pull/123#discussion_r274196949) [](ancestors = 274196949)

Public/Src/FrontEnd/Script/Ast/Expressions/SwitchExpressionClause.cs
dannyvv: Wouldn't that be weird, since it would have to fail on false... since false implies there must be a match expression

---
In reply to: [274188741](https://github.com/Microsoft/BuildXL/pull/123#discussion_r274188741) [](ancestors = 274188741)

smera: Oh, sorry, didn't notice there was another constructor with the match as a parameter. I think what set my off was that on the parser you are doing

node.IsDefaultFallthrough = match == null;

after constructing it, for every case

---
In reply to: [274201164](https://github.com/Microsoft/BuildXL/pull/123#discussion_r274201164) [](ancestors = 274201164,274188741)

PR #126, opened 2019-04-10
Ensure InputTracker tracks changing mount paths

anniefu:
When checking whether the build graph can be re-used, InputTracker should take into account any changes to build mount paths and automatically return a graph cache miss if there are any changes. 

Otherwise, a graph might be re-used with the wrong mount paths causing incorrect tracking and possibly Disallowed File Accesses for the new mount paths. For example, when different users share the same EngineCache and have different user profile mount paths.

### DIFF
--- a/Public/Src/Engine/Dll/InputTracker.cs
+++ b/Public/Src/Engine/Dll/InputTracker.cs
@@ -532,7 +532,7 @@ namespace BuildXL.Engine
             BinaryWriter writer,
             PathTable pathTable,
             IReadOnlyDictionary<string, string> buildParametersImpactingBuild,
-            IReadOnlyDictionary<string, IMount> mountsImpactingBuild,
+            [CanBeNull] IReadOnlyDictionary<string, IMount> mountsImpactingBuild,
             string changeTrackingStatePath)
         {
             Contract.Requires(writer != null);
@@ -584,7 +584,7 @@ namespace BuildXL.Engine
                     foreach (var kvp in mountsImpactingBuild)
                     {
                         writer.Write(kvp.Key);
-                        writer.Write(kvp.Value.Path.ToString(pathTable));
+                        writer.Write(kvp.Value == null ? null : kvp.Value.Path.ToString(pathTable));
                     }
                 }
             }
@@ -919,7 +919,7 @@ namespace BuildXL.Engine
                 {
                     // The previously used mount is not known
                     result.MissType = GraphCacheMissReason.MountChanged;
-                    result.FirstMissIdentifier = string.Format(CultureInfo.InvariantCulture, Strings.InputTracker_MountRemoved, mountName);
+                    result.FirstMissIdentifier = string.Format(CultureInfo.InvariantCulture, Strings.InputTracker_MountRemoved, mountName, previousPath);
                     Logger.Log.InputTrackerDetectedMountChanged(loggingContext, mountName, previousPath, string.Empty);
                     return result;
                 }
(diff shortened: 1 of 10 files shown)


Public/Src/Engine/Dll/InputTracker.cs
```diff
@@ -530,7 +530,9 @@ private void RegisterFile(string path, ContentHash hash)
         /// </summary>
         public void WriteToFile(
             BinaryWriter writer,
+            PathTable pathTable,
             IReadOnlyDictionary<string, string> buildParametersImpactingBuild,
+            IReadOnlyDictionary<string, IMount> mountsImpactingBuild,
```
dannyvv: Nit:perhaps annotate with [CanBeNull] so we get IDE help when calling this function.

Public/Src/Engine/Dll/InputTracker.cs
```diff
@@ -570,6 +572,23 @@ private void RegisterFile(string path, ContentHash hash)
                 }
             }
 
+            // Write the mounts that impact the build
+            if (mountsImpactingBuild == null)
+            {
+                writer.Write(0);
```
dannyvv: tiny nitpick: Upon deserialization we can't distinguish empty mounts from null mounts...

Public/Src/Engine/Dll/Strings.resx
```diff
@@ -216,4 +216,10 @@
   <data name="Layout_DefaultJunctionNameTotLogFolder" xml:space="preserve">
     <value>BuildXLCurrentLog</value>
   </data>
+  <data name="InputTracker_MountChanged" xml:space="preserve">
+    <value>Mount {0} was:'{1}' now:'{2}'</value>
+  </data>
+  <data name="InputTracker_MountRemoved" xml:space="preserve">
+    <value>{0} was previously defined as a mount but is now undefined</value>
```
dannyvv: would it be usefull to report what path it was previously defined with?

Public/Src/Engine/Dll/InputTracker.cs
anniefu: I don't think that's that big of a deal. But are you allowed to have a mount that has no name in the first place?

anniefu: I take that back, empty mount names don't make a difference here actually.

I'm just going to leave it as-is though to match the pattern that the environment variables dictionary above uses for serialization.

narasamdya:
>             if ((missReasonForCompatibleGraphFingerprint = previousFingerprint.CompareFingerprint(graphFingerprint.CompatibleFingerprint)) == GraphCacheMissReason.NoMiss)

Can we bump up the graph fingerprint version so that the new BuildXL will not throw exception (if possible) when trying to load old engine cache?

---
Refers to: Public/Src/Engine/Dll/InputTracker.cs:826 in 1021aee. [](commit_id = 1021aee7eb3f741acf49f9b8d6c58958815fad05, deletion_comment = False)

Public/Src/Engine/UnitTests/Engine/MiniBuild.cs
```diff
@@ -650,6 +650,7 @@ public void MiniBuildCachedGraphWithMountAndEnvironmentChanges()
             // Revert environment variable.
             Configuration.Startup.Properties[EnvVarName] = "Env1";
 
+            System.Diagnostics.Debugger.Launch();
```
narasamdya: >System.Diagnostics.Debugger.Launch(); [](start = 12, length = 37)

Remove :-)

Public/Src/Engine/Dll/InputTracker.cs
dannyvv: sounds good, was merely observigin the difference

---
In reply to: [274223325](https://github.com/Microsoft/BuildXL/pull/126#discussion_r274223325) [](ancestors = 274223325)

PR #130, opened 2019-04-11
Avoid scrubbing non-scrubbable mountpoints

jbayardo:
When working on the VcPkg integration, a few packages would fail to build due to BXL scrubbing internal CMake directories that were pre-made and empty by build-time (these were located inside a shared opaque).

The agreed-upon solution is to introduce an unsafe flag to instruct the shared opaque scrubber to specifically not remove these empty directories.

### DIFF
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -991,7 +991,10 @@ namespace BuildXL
                             CommandLineUtilities.ParseBoolEnumOption(opt, sign, PreserveOutputsMode.Enabled, PreserveOutputsMode.Disabled),
                             isUnsafe: true,
                             isEnabled: (() => sandboxConfiguration.UnsafeSandboxConfiguration.PreserveOutputs != PreserveOutputsMode.Disabled)),
-
+                        OptionHandlerFactory.CreateBoolOption(
+                            "unsafe_DisableSharedOpaqueEmptyDirectoryScrubbing",
+                            sign => schedulingConfiguration.UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing = sign,
+                            isUnsafe: true),
                         // TODO: Remove this!
                         OptionHandlerFactory.CreateBoolOption(
                             "unsafe_SourceFileCanBeInsideOutputDirectory",
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -889,6 +889,10 @@ namespace BuildXL
                             sign => schedulingConfiguration.UnsafeDisableGraphPostValidation = sign,
                             isUnsafe: true),
                         OptionHandlerFactory.CreateBoolOption(
+                            "unsafe_DisableSharedOpaqueEmptyDirectoryScrubbing",
+                            sign => schedulingConfiguration.UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing = sign,
+                            isUnsafe: true),
+                        OptionHandlerFactory.CreateBoolOption(
                             "unsafe_ExistingDirectoryProbesAsEnumerations",
                             sign => sandboxConfiguration.UnsafeSandboxConfigurationMutable.ExistingDirectoryProbesAsEnumerations = sign,
                             isUnsafe: true),
@@ -991,10 +995,6 @@ namespace BuildXL
                             CommandLineUtilities.ParseBoolEnumOption(opt, sign, PreserveOutputsMode.Enabled, PreserveOutputsMode.Disabled),
                             isUnsafe: true,
                             isEnabled: (() => sandboxConfiguration.UnsafeSandboxConfiguration.PreserveOutputs != PreserveOutputsMode.Disabled)),
-                        OptionHandlerFactory.CreateBoolOption(
-                            "unsafe_DisableSharedOpaqueEmptyDirectoryScrubbing",
-                            sign => schedulingConfiguration.UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing = sign,
-                            isUnsafe: true),
                         // TODO: Remove this!
                         OptionHandlerFactory.CreateBoolOption(
                             "unsafe_SourceFileCanBeInsideOutputDirectory",
--- a/Public/Src/Engine/Dll/DirectoryScrubber.cs
+++ b/Public/Src/Engine/Dll/DirectoryScrubber.cs
@@ -360,6 +360,7 @@ namespace BuildXL.Engine
                         {
                             try
                             {
+                                Tracing.Logger.Log.ScrubbingDirectory(pm.LoggingContext, directory);
                                 FileUtilities.DeleteDirectoryContents(directory, deleteRootDirectory: true, tempDirectoryCleaner: m_tempDirectoryCleaner);
                                 Interlocked.Increment(ref directoriesRemovedRecursively);
                             }
--- a/Public/Src/Engine/Dll/DirectoryScrubber.cs
+++ b/Public/Src/Engine/Dll/DirectoryScrubber.cs
@@ -360,7 +360,6 @@ namespace BuildXL.Engine
                         {
                             try
                             {
-                                Tracing.Logger.Log.ScrubbingDirectory(pm.LoggingContext, directory);
                                 FileUtilities.DeleteDirectoryContents(directory, deleteRootDirectory: true, tempDirectoryCleaner: m_tempDirectoryCleaner);
                                 Interlocked.Increment(ref directoriesRemovedRecursively);
                             }
--- a/Public/Src/Engine/Dll/Engine.cs
+++ b/Public/Src/Engine/Dll/Engine.cs
@@ -2173,7 +2173,8 @@ namespace BuildXL.Engine
                 { "unsafe_LazySymlinkCreation", Logger.Log.ConfigUnsafeLazySymlinkCreation },
                 { "unsafe_MonitorFileAccesses", Logger.Log.ConfigUnsafeDisabledFileAccessMonitoring },
                 { "unsafe_PreserveOutputs", Logger.Log.ConfigPreserveOutputs },
-                { "unsafe_SourceFileCanBeInsideOutputDirectory", loggingContext => { } },
+                { "unsafe_DisableSharedOpaqueEmptyDirectoryScrubbing", Logger.Log.ConfigUnsafeDisableSharedOpaqueEmptyDirectoryScrubbing },
+                { "unsafe_SourceFileCanBeInsideOutputDirectory", loggingContext => { } /* Special case: unsafe option we do not want logged */ },
                 { "unsafe_UnexpectedFileAccessesAreErrors", Logger.Log.ConfigUnsafeUnexpectedFileAccessesAsWarnings },
             };
         }
--- a/Public/Src/Engine/Dll/Engine.cs
+++ b/Public/Src/Engine/Dll/Engine.cs
@@ -2157,6 +2157,7 @@ namespace BuildXL.Engine
                 { "unsafe_DisableCycleDetection", Logger.Log.ConfigUnsafeDisableCycleDetection },
                 { "unsafe_DisableDetours", Logger.Log.ConfigDisableDetours },
                 { "unsafe_DisableGraphPostValidation", loggingContext => { } /* Special case: unsafe option we do not want logged */ },
+                { "unsafe_DisableSharedOpaqueEmptyDirectoryScrubbing", Logger.Log.ConfigUnsafeDisableSharedOpaqueEmptyDirectoryScrubbing },
                 { "unsafe_ExistingDirectoryProbesAsEnumerations", Logger.Log.ConfigUnsafeExistingDirectoryProbesAsEnumerations },
                 { "unsafe_ForceSkipDeps", Logger.Log.ForceSkipDependenciesEnabled },
                 { "unsafe_IgnoreGetFinalPathNameByHandle", Logger.Log.ConfigIgnoreGetFinalPathNameByHandle },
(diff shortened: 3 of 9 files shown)


Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -744,6 +744,12 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
             IEnumerable<string> nonScrubbablePaths,
             ITempDirectoryCleaner tempCleaner)
         {
+            if (mountPathExpander != null)
+            {
+                // We don't scrub mounts explicitly marked as non-scrubbable
+                nonScrubbablePaths = nonScrubbablePaths.Concat(mountPathExpander.GetNonScrubbableRoots().Select(p => p.ToString(scheduler.Context.PathTable)));
```
smera: >nonScrubbablePaths =  [](start = 16, length = 21)

This option is essentially unsafe for shared opaque dirs. A user may accidentally flag a mount as non-scrubbable and any shared opaque contained in that mount won't be cleaned up. This means non-deterministic behavior may occur during the build.
So I'd be ok with this change if we explicitly block shared opaques from being declared under such mounts.
If that's too restrictive, we can add an unsafe flag to disable the block, that users can opt into.

Public/Src/Engine/Dll/DirectoryScrubber.cs
```diff
@@ -360,6 +360,7 @@ public bool RemoveExtraneousFilesAndDirectories(CancellationToken cancellationTo
                         {
                             try
                             {
+                                Tracing.Logger.Log.ScrubbingDirectory(pm.LoggingContext, directory);
```
anniefu: >Tracing.Logger.Log.ScrubbingDirectory(pm.LoggingContext, directory); [](start = 32, length = 68)

Apparently we removed this logging to improve performance in this PR:
commit cc29aeea: Merged PR 389089: Directory Scrubber Perf Improvements ([https://dev.azure.com/mseng/_git/Domino/commit/cc29aeea3142e7aa2b8d2787c936458160eccb41?refName=refs%2Fheads%2Fmaster&_a=compare&path=%2FPublic%2FSrc%2FEngine%2FDomino.Engine%2FDirectoryScrubber.cs](https://dev.azure.com/mseng/_git/Domino/commit/cc29aeea3142e7aa2b8d2787c936458160eccb41?refName=refs%2Fheads%2Fmaster&_a=compare&path=%2FPublic%2FSrc%2FEngine%2FDomino.Engine%2FDirectoryScrubber.cs)) 

It's probably best to leave it out here to avoid perf regression.

Public/Src/Engine/Dll/EngineSchedule.cs
jbayardo: The reason I'm interested in this feature is: there is a shared opaque in say path ./A, but the folders ./A/B/*.dir should not be removed (the files within could be removed, but the directory must not be removed).

Thus, what Iman and I came up with is put a tracked r/w non-scrubbable mount in ./A/B, since glob mountpoints are no supported. If I understand correctly, this inherently introduces some non-determinism: initially non-needed files within the mountpoint may or may not be present in the first place.

I'm not sure if
> explicitly block shared opaques from being declared under such mounts.

Also means the intersection should not be allowed.

smera: Not scrubbing places where shared opaque outputs can be written to is unsafe. I think I understand the scenario you describe, but allowing that will also open the door to unsafe (unintended) behavior. So if there is a shared opaque A where a pip writes A/B/blah, and a non-scrubbable mount A/B, then blah won't be removed. This is unsafe unless you *really* understand what's going on.

I'd rather have a scrubber-specific configuration (under the schedule configuration) to tell the scrubber to not remove empty directories. That should also be flagged as unsafe though.

---
In reply to: [274675774](https://github.com/Microsoft/BuildXL/pull/130#discussion_r274675774) [](ancestors = 274675774)

anniefu:
For testing, a good place to add a test or two would be "SharedOpaqueDirectoryTests.cs" or "BaselineTests.cs" which uses our SchedulerIntegrationTestBase that runs a full BuildXL scheduler.

We have a OneNote guide that covers some of the basics of [how to write scheduler integration tests](https://microsoft.sharepoint.com/teams/toolsforeng/_layouts/15/WopiFrame.aspx?sourcedoc={8876c904-4346-41ba-a5ee-f26b404d16b7}&action=edit&wd=target%28Domino.one%7C10facd4f-4e15-4fe3-96ba-446f3fe6454e%2FHow%20to%20write%20scheduler%20integration%20tests%7Cbb922279-40a7-4e79-8b4e-438610c9c255%2F%29)

You'll probably need to add a non-scrubbable mount for your specific test. You can see an example of how to do that in the "BaseSetup" function of "PipTestBase.cs". You can add one at the beginning of just your test case instead of adding one to the entire testbase.

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -744,6 +744,12 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
             IEnumerable<string> nonScrubbablePaths,
             ITempDirectoryCleaner tempCleaner)
         {
+            if (mountPathExpander != null)
```
anniefu: >if (mountPathExpander != null) [](start = 12, length = 30)

Could you add an explanation of the motivation/reason for this change (like the one you mentioned in your other comment) to the PR description and the commit message when you eventually commit?

It's a nice to have thing in commits/PRs in general so we can understand why changes were made in the future.

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -744,6 +744,12 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
...
         {
+            if (mountPathExpander != null)
+            {
+                // We don't scrub mounts explicitly marked as non-scrubbable
+                nonScrubbablePaths = nonScrubbablePaths.Concat(mountPathExpander.GetNonScrubbableRoots().Select(p => p.ToString(scheduler.Context.PathTable)));
+            }
+
             var pathsToScrub = new List<string>();
             if (configuration.Engine.Scrub && mountPathExpander != null)
             {
```
mpysson: I think the configuration for scrubbing has gotten pretty complicated. 
1. There are scrubbable mounts, where anything except declared build outputs get deleted. But this only happens if the /scrub flag is passed
2. /scrubdirectory specifies some directories should be scrubbed. This happens regardless of whether the specified directory is under a scrubbable mount or not
3. Shared opaque directories use the directory scrubber, but there "scrub" is a bit different. It really means delete the previous build's outputs

All of these obeyed nonScrubbablePaths, which were special paths one should never scrub, like the cache directory.

Adding all mounts that aren't marked scrubbable to the nonScrubbablePaths modified the existing behavior of the above, in particular I don't think we want #3

narasamdya:
>             }

For the time being, I agree with Serge to add an unsafe flag to control the behavior of shared opaque scrubber, i.e., don't delete empty folder. 

We uncovered a kind of bug in BuildXL where if user says that the mount is not scurbbable, then BuildXL should not scrub it. But that non-scrubbable mounts should not be taken into account when scrubbing opaque directories.

---
Refers to: Public/Src/Engine/Dll/EngineSchedule.cs:757 in d83a4d0. [](commit_id = d83a4d089a8dac513b16004d3985b77177fe2b2c, deletion_comment = False)

narasamdya:
As others pointed out, we expect unit tests for changes. Please add some.

Public/Src/Engine/Dll/EngineSchedule.cs
jbayardo: Changed as reuqested. Please let me know if you have any comments

Public/Src/Engine/Dll/EngineSchedule.cs
jbayardo: Added context and the fix. Please let me know if you have any further comments.

jbayardo:
As requested, I introduced an unsafe flag that turns on the specific behavior needed and its corresponding test. I used it for the VcPkg integration I've been working on, and it is behaving as expected.

Please let me know if you have any further comments.

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -810,8 +810,7 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
                     // Everything that is not an output under a shared opaque is considered part of the build. 
                     isPathInBuild: path =>
                         // Scheduler.PipGraph.IsPathInBuild is used for extra safety.
-                        scheduler.PipGraph.IsPathInBuild(AbsolutePath.Create(scheduler.Context.PathTable, path)) ||
-                        !SharedOpaqueOutputHelper.IsSharedOpaqueOutput(path), 
+                        scheduler.PipGraph.IsPathInBuild(AbsolutePath.Create(scheduler.Context.PathTable, path)) || !SharedOpaqueOutputHelper.IsSharedOpaqueOutput(path) || (configuration.Schedule.UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing && Directory.Exists(path) && !Directory.EnumerateFileSystemEntries(path).Any()),
```
smera: >(configuration.Schedule.UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing && Directory.Exists(path) && !Directory.EnumerateFileSystemEntries(path).Any() [](start = 172, length = 151)

suggest to move this into its own method to simplify readability. E.g. 'ShouldRemoveEmptyDirectories'
And add some indenting

Public/Src/App/Bxl/Args.cs
```diff
@@ -991,7 +991,10 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                             CommandLineUtilities.ParseBoolEnumOption(opt, sign, PreserveOutputsMode.Enabled, PreserveOutputsMode.Disabled),
                             isUnsafe: true,
                             isEnabled: (() => sandboxConfiguration.UnsafeSandboxConfiguration.PreserveOutputs != PreserveOutputsMode.Disabled)),
-
+                        OptionHandlerFactory.CreateBoolOption(
+                            "unsafe_DisableSharedOpaqueEmptyDirectoryScrubbing",
```
narasamdya: >DisableSharedOpaqueEmptyDirectoryScrubbing" [](start = 36, length = 43)

Order it alphabetically.

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -2173,7 +2173,8 @@ public DirectoryDeletionLock CreateOutputDirectories(PathTable pathTable, Loggin
                 { "unsafe_LazySymlinkCreation", Logger.Log.ConfigUnsafeLazySymlinkCreation },
                 { "unsafe_MonitorFileAccesses", Logger.Log.ConfigUnsafeDisabledFileAccessMonitoring },
                 { "unsafe_PreserveOutputs", Logger.Log.ConfigPreserveOutputs },
-                { "unsafe_SourceFileCanBeInsideOutputDirectory", loggingContext => { } },
+                { "unsafe_DisableSharedOpaqueEmptyDirectoryScrubbing", Logger.Log.ConfigUnsafeDisableSharedOpaqueEmptyDirectoryScrubbing },
```
narasamdya: >_DisableSharedOpaqueEmptyDirectoryScrubbing [](start = 25, length = 43)

Order this alphabetically.

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -825,6 +826,11 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
             }
         }
 
+        private static bool ShouldRemoveEmptyDirectories(IConfiguration configuration, string path)
+        {
+            return configuration.Schedule.UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing && Directory.Exists(path) && !Directory.EnumerateFileSystemEntries(path).Any();
```
narasamdya: >UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing [](start = 42, length = 48)

Should it not be `!configuration.Scheduler.UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing` ? It's a bit confusing because the empty directory is removed if the flag is disabled.

narasamdya: >UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing [](start = 42, length = 48)

`Directory.EnumerateFileSystemEntries(path)` is painfully slow. If perf. matters, then we need to look for another way to check directory contents; perhaps using FindFirstFile. Otherwise, you can leave a remark there mentioning this issue.

Public/Src/Utilities/Configuration/IScheduleConfiguration.cs
```diff
@@ -286,5 +286,13 @@ public partial interface IScheduleConfiguration
         /// Skip hash source file pips during graph creation.
         /// </summary>
         bool SkipHashSourceFile { get; }
+
+        /// <summary>
+        /// Unsafe configuration that stops the shared opaque scrubber from deleting empty directories
+        /// </summary>
+        /// <remarks>
+        /// TODO: Remove this when https://gitlab.kitware.com/cmake/cmake/issues/19162 has reached mainstream versions
```
narasamdya: >TODO: Remove this when https://gitlab.kitware.com/cmake/cmake/issues/19162 has reached mainstream versions [](start = 12, length = 106)

Can you add some remarks (and example perhaps) why this flag is made unsafe?

Public/Src/Engine/Dll/EngineSchedule.cs
jbayardo: If the flag is false, then false is returned. The caller will do an or, and since false is the identity for or, whether the path is part of the build or not will be unchanged by this.

If the flag is true, then we first check if it is a directory (it could be a file!), and then check if it is empty, so true will only be returned if it is a empty directory.

Public/Src/Engine/UnitTests/Engine/SharedOpaqueEngineTests.cs
```diff
@@ -82,6 +82,35 @@ public void OutputsUnderSharedOpaqueAreSelectivelyScrubbed()
             Assert.True(File.Exists(Path.Combine(objDir, outOfBuildFile)));
         }
 
+        [Fact]
```
narasamdya: >[Fact] [](start = 8, length = 6)

Why don't you make it as [Theory] where the inline value is true for disabling the scrubbing and false for enabling the scrubbing.

Public/Src/Engine/Dll/EngineSchedule.cs
jbayardo: The current usage of this flag is to be run in an atypical scenario, where BXL is called exactly once, and all directories checked will be empty -as far as I have tested-, so it should not be a bottleneck.

It doesn't seem worth to break compatibility just for this. I have added a comment in case anyone ever uses it.

PR #132, opened 2019-04-11
Do not create checkpoint if no events were received.

SergeyTeplyakov:
Master service should not create checkpoints if no events were received.

The current logic was checking the result of `EventStore.GetLastProcessedSequencePoint` to `null` to check this case. But the sequence point can be non-null, but still contain no sequence numbers.

Fix for #AB1510624


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -541,7 +541,7 @@ internal async Task<BoolResult> CreateCheckpointAsync(OperationContext context)
         {
             // Need to obtain the sequence point first to avoid race between the sequence point and the database's state.
             EventSequencePoint currentSequencePoint = EventStore.GetLastProcessedSequencePoint();
-            if (currentSequencePoint == null)
+            if (currentSequencePoint == null || currentSequencePoint.SequenceNumber == null)
```
dannyvv: nit: elvis operator?

PR #137, opened 2019-04-12
Adjust grpc constants and measure the server handling time

semihokur:


Public/Src/Utilities/Configuration/EngineEnvironmentSettings.cs
```diff
@@ -135,6 +135,16 @@ public static class EngineEnvironmentSettings
         public static readonly Setting<TimeSpan> DistributionInactiveTimeout = CreateSetting("BuildXLDistribInactiveTimeoutMin", value => ParseTimeSpan(value, ts => TimeSpan.FromMinutes(ts)) ??
             TimeSpan.FromMinutes(30));
 
+        /// <summary>
+        /// The number of threads in the grpc thread pool.
+        /// </summary>
+        public static readonly Setting<int> GrpcThreadPoolSize = CreateSetting("BuildXLGrpcThreadPoolSize", value => ParseInt32(value) ?? 70);
```
mpysson: >70 [](start = 138, length = 2)

please add a comment on where 70 came from

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -275,10 +276,15 @@ private FileContentTable FileContentTable
 
             if (configuration.Distribution.IsGrpcEnabled)
             {
-                GrpcEnvironment.InitializeIfNeeded();
+                bool grpcHandlerInliningEnabled = GrpcConstants.HandlerInliningEnabled;
+
 #if FEATURE_CORECLR
-                global::Grpc.Core.GrpcEnvironment.SetHandlerInlining(false);
+                grpcHandlerInliningEnabled = false;
```
mpysson: >grpcHandlerInliningEnabled = false; [](start = 16, length = 35)

Why not enable? Not supported? Please add comment

Public/Src/Engine/Dll/Distribution/Grpc/ServerInterceptor.cs
```diff
@@ -53,31 +56,42 @@ private void InterceptCallContext(ServerCallContext context)
...
+        public async override Task<TResponse> UnaryServerHandler<TRequest, TResponse>(TRequest request, ServerCallContext context, UnaryServerMethod<TRequest, TResponse> continuation)
         {
-            InterceptCallContext(context);
-            return continuation(request, context);
+            (string, string) tuple = InterceptCallContext(context);
+            var watch = Stopwatch.StartNew();
+            var result = await continuation(request, context);
+            Logger.Log.GrpcTrace(m_loggingContext, string.Format(RespondedLogFormat, tuple.Item1, tuple.Item2, watch.ElapsedMilliseconds));
+            return result;
         }
```
dannyvv: Almost looks like it would warrant a helper function....

anniefu:
>     internal static class GrpcConstants

This is confusing, I expect a class called "GrpcConstants" to be only const values, not static. This seems like the correct class for the hardcoded 5 minutes, 70 threads, and true to be defined.

I would expect  EngineEnvironmentSettings to point to GrpcConstants and the Engine to be using EngineEnvironmentSettings to initialize Grpc, instead of GrpcConstants pointing to EngineEnvironmentSettings and the Engine calling into GrpcConstants.

---
Refers to: Public/Src/Engine/Dll/Distribution/Grpc/GrpcConstants.cs:9 in ec1e9fd. [](commit_id = ec1e9fd3efe57643d2c83c4829cdb4b23ae616b6, deletion_comment = False)

PR #140, opened 2019-04-12
Remove default args from FrontEndHostController

dannyvv:


Public/Src/FrontEnd/SdkTesting/Helper/TestRunner.cs
```diff
@@ -1,3 +1,4 @@
+
 // Copyright (c) Microsoft. All rights reserved.
```
smera: Extra new line

PR #143, opened 2019-04-13
Additional config file parse

norahuang:
Bring back the additional config file parsing feature.


Public/Src/FrontEnd/Script/ConfigurationProcessing/ConfigurationProcessor.cs
```diff
@@ -80,6 +80,12 @@ AbsolutePath IConfigurationProcessor.FindPrimaryConfiguration(IStartupConfigurat
                 return null;
             }
 
+            // Apply Additional configurations from the commandline
+            foreach (var additionalConfigurationFile in commandLineConfiguration.Startup.AdditionalConfigFiles)
+            {
+                configObjectLiteral = ParseAndInterpretConfigFile(additionalConfigurationFile, configObjectLiteral);
+            }
+
             // TODO: user override is not really working now. Fix me!
```
dannyvv: > TODO: user overrid [](start = 14, length = 19)

It would be nice if we could also restore support for config.user.dsc (not now, but in a future commit)

PR #144, opened 2019-04-13
AnnotatedSessions Perf Update

socat:
Ln 20, 52, 60, 93: Specifying cluster and database allows this to be portable.
Ln 21, 53, 61, 94: Ensure we're honoring the date range on all joined data sources.


dannyvv:
Thanks for your contribution!

PR #146, opened 2019-04-15
Temporary fix for dummy files in phony Ninja rules

jbayardo:
Ninja rules can produce dummy files through an special case of a phony rule (see https://ninja-build.org/manual.html#_the_literal_phony_literal_rule ). This is used in the case of CMake-generated Ninja files, specifically for the install functionality.

This PR is a temporary fix that just declares every output as Optional instead of Required, so that BuildXL won't kill the build when the rule is executed and the dummy output is not there.

A more permanent fix is more complicated, and requires changing the Ninjson utility that we use to generate the BuildXL graphs. The internal bug-track for this is [here](https://mseng.visualstudio.com/1ES/_workitems/edit/1512466).


Public/Src/FrontEnd/UnitTests/Ninja/Infrastructure/NinjaPipExecutionTestBase.cs
```diff
@@ -157,6 +157,35 @@ protected NinjaSpec CreateProjectThatCopiesResponseFile(string output, string re
             return new NinjaSpec(content, new[] { "all" });
         }
 
+        protected NinjaSpec CreateDummyFilePhonyProject(string phonyOutputFile1, string phonyOutputFile2, string dummyFile, string effectiveFile)
+        {
+            // Rule is built to mimick a Ninja install as generated by CMake
```
smera: >mimick [](start = 32, length = 6)

mimic

PR #148, opened 2019-04-16
GRPC file copies with paths, existence check, and minor clean up

thpha:
**1. Existence check now implemented**
Self-explanatory.

**2. GrpcDistributedPathTransformer updated to properly handle the two-drive scenario**
GRPC's PathTransformer left out code to replace the cache root with its junction if it's not the D: drive.

**3. Minor clean-up**
Removed nameByDrive logic leftover from previous implementation.

**4. Better testing for GRPC copies and added unit tests for existence check logic**
Self-explanatory.


dcwuser:
I don't like the change to absolute paths and don't see why it's needed.
 * CAS APIs all deal with only hashes and do not expose internal structure to callers. This violates that principal.
* Dealing with path logic here makes for one more place where things may need to be different for Windows/Mac/Linux. We have it working with hashes via code that has no dependence on path structure and I'd like to keep that good attribute.
* I think you mean StreamCas not LLS puts two caches on the same drive? That's true, but you'll notice the way we implemented StreamCas was to multi-plex to all Cas's, not to require callers to provide additional information about how to route the call. Let's repeat that here.

I feel pretty strongly about this, so blocking this check-in until we can have a discussion about this.

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
```diff
@@ -16,15 +18,22 @@ namespace BuildXL.Cache.ContentStore.Distributed.Utilities
...
     public class GrpcDistributedPathTransformer : IAbsolutePathTransformer
     {
+        private readonly IReadOnlyDictionary<string, string> _junctionsByDirectory;
         private static readonly string _localMachineName = Environment.MachineName;
         internal const string BlobFileExtension = ".blob";
 
+        /// <nodoc />
+        public GrpcDistributedPathTransformer(IReadOnlyDictionary<string, string> junctionsByDirectory = null)
+        {
+            _junctionsByDirectory = junctionsByDirectory ?? new Dictionary<string,string>();
```
dannyvv: Does the dictionary need a comparer based on OS? I.e. case insensitve on windows, case sensitive for linux and dynamic for mac?

dannyvv: See below that every path is just ToUppered anyway... perhaps a comment to the effect would be nice.
side-question: Do we have any assumptions on the network share supporting path transformations? If we have a unix box with an Samba server I recall that the system was mixed case sensitive depending on how you use it...

---
In reply to: [275859026](https://github.com/Microsoft/BuildXL/pull/148#discussion_r275859026) [](ancestors = 275859026)

thpha:
No, I don't mean StreamCas actually. LLS uses another DistributedContentStore which saves its content in Shared\dcs to avoid content churn.

---
In reply to: [483619080](https://github.com/Microsoft/BuildXL/pull/148#issuecomment-483619080) [](ancestors = 483619080)

thpha:
Discussed with Kevin and Sergey - leaving unaddressed for now. I turned off the LLS flag in CBTest that breaks the copier on Friday and most everything has been fine since then.
I'm guessing having it on is only important in production, due to scale. I filed a bug and we can return to this issue later.

---
In reply to: [483780460](https://github.com/Microsoft/BuildXL/pull/148#issuecomment-483780460) [](ancestors = 483780460,483619080)

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
```diff
@@ -167,6 +169,34 @@ message RemoveFromTrackerResponse {
     int64 FilesEvicted = 2;
 }
 
+// CopyFile
+message CopyFileRequest {
+    string traceId = 1;
+    string absolutePath = 2;
```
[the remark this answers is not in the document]
thpha: I'll do this in a future PR after I verify that changing the casing doesn't break anything.

PR #150, opened 2019-04-16
Improve performance for worker's reconciliation

SergeyTeplyakov:
TLDR; This pr speeds up reconciliation process that was causing builds to fail in MW_PS03 stamp 6 times.

Today some of the Windows builds are failing with materialization errors because reconciliation process takes too long.

Here is an example:

* Machine A starts the build
* During startup Machine A detects that reconciliation is required because local database is out of sync with the content directory.
* Machine A stops sending all events to the master and starts reconciliation process
* Content B is put to Machine A during the build:
** Content B is added to global redis but the master is not notified because reconciliation is still running
* The build takes longer than redis retention period (like 3 hours) but reconciliation is still not done.
* Machine C tries to pin the content B but it fails because content B is gone from redis but was still not added to LLS.

This change improves local database enumeration process and speeds up the reconciliation up to 6 times.

Here are some stats for a database with 15 million entries:

Before: 
Reconcile by 86492ms. Added: 0, removed: 39714
After
Reconcile by 12337ms. Added: 0, removed: 39714

The time went down 6 times and memory went down for about the same rate.

Some screenshots from a profiler:

## Before
Reconsile_Before

## After
Reconcile_After

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
@@ -341,16 +341,15 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             return true;
         }
 
-        private Task<BoolResult> RestoreFileAsync(OperationContext context, AbsolutePath checkpointTargetDirectory, string relativePath, string storageId, bool forceUseExistingFiles = false)
+        private Task<BoolResult> RestoreFileAsync(OperationContext context, AbsolutePath checkpointTargetDirectory, string relativePath, string storageId)
         {
             return context.PerformOperationAsync(
                 _tracer,
                 async () =>
                 {
                     var incrementalCheckpointFile = _incrementalCheckpointDirectory / relativePath;
-                    if ((forceUseExistingFiles || (
-                            _incrementalCheckpointInfo.TryGetValue(relativePath, out var fileStorageId)
-                            && storageId == fileStorageId))
+                    if ((_incrementalCheckpointInfo.TryGetValue(relativePath, out var fileStorageId)
+                            && storageId == fileStorageId)
                         && _database.IsImmutable(incrementalCheckpointFile)
                         && _fileSystem.FileExists(incrementalCheckpointFile))
                     {
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
@@ -7,9 +7,9 @@ using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Threading.Tasks;
-using BuildXL.Cache.ContentStore.Interfaces.Utils;
 using BuildXL.Cache.ContentStore.Distributed.NuCache.EventStreaming;
 using BuildXL.Cache.ContentStore.Distributed.Redis;
+using BuildXL.Cache.ContentStore.Extensions;
 using BuildXL.Cache.ContentStore.FileSystem;
 using BuildXL.Cache.ContentStore.Tracing;
 using BuildXL.Cache.ContentStore.Tracing.Internal;
@@ -45,7 +45,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 
         private CounterCollection<ContentLocationStoreCounters> Counters { get; }
 
-        private CheckpointConfiguration _configuration;
+        private readonly CheckpointConfiguration _configuration;
 
         /// <summary>
         /// Maps file name to storage id for the currently downloaded checkpoint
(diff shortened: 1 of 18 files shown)


SergeyTeplyakov:
BTW, there still a room for improvement here. The solution is still single threaded and based on the discussions at rocksdb discussion groups we may read the database in parallel to speed the reading process even more.

I can see two options here:

* "For now, the best you can get is to get all SST files' metadata through DB::GetColumnFamilyMetaData(), and use file boundaries as a hint to partition the data."
* Use different column families for different hashes. For instance, we can get the first character of the hash and compute family name based on that.
In this case, we can iterate the database in parallel and merge the sorted sequence after that.

Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
```diff
@@ -327,38 +329,48 @@ private async Task<bool> RestoreCheckpointIncrementalAsync(OperationContext cont
...
             }
-
+            
             // Finalize by adding the incremental checkpoint info file to the local incremental checkpoint directory
             await HardlinkWithFallBackAsync(context, checkpointFile, _incrementalCheckpointInfoFile);
             UpdateIncrementalCheckpointInfo(newCheckpointInfo);
             return true;
         }
+
+        private Task<BoolResult> RestoreFileAsync(OperationContext context, AbsolutePath checkpointTargetDirectory, string relativePath, string storageId, bool forceUseExistingFiles = false)
```
JuanCarlosGI: Maybe a simple comment explaining the motivation of forceUseExistingFiles? Specially since it is not used anywhere, as far as I can tell

Public/Src/Cache/ContentStore/DistributedTest/ContentLocation/LocalLocationStoreDistributedContentTests.cs
```diff
@@ -124,7 +127,7 @@ protected override void Dispose(bool disposing)
                 configuration.RedisGlobalStoreSecondaryConnectionString = _secondaryGlobalStoreDatabase.ConnectionString;
             }
 
-            configuration.DistributedCentralStore = new DistributedCentralStoreConfiguration(rootPath);
+            configuration.DistributedCentralStore = null;//new DistributedCentralStoreConfiguration(rootPath);
```
JuanCarlosGI: Forgot to delete comment?

Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
SergeyTeplyakov: It was used in one of my experiements. I removed it completely.

---
In reply to: [275919167](https://github.com/Microsoft/BuildXL/pull/150#discussion_r275919167) [](ancestors = 275919167)

Public/Src/Cache/ContentStore/Distributed/NuCache/MachineIdSets/ArrayMachineIdSetBase.cs
```diff
@@ -59,7 +36,7 @@ public override MachineIdSet SetExistence(IReadOnlyCollection<MachineId> machine
                 return this;
             }
 
-            var machineIds = new HashSet<ushort>(_machineIds);
+            var machineIds = new HashSet<ushort>(EnumerateRawMachineIds());
```
ljcollins25: >HashSet<ushort>(EnumerateRawMachineIds()); [](start = 33, length = 42)

Seems wasteful to create a hashset just to dedupe.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -176,6 +180,18 @@ public bool TryGetEntry(OperationContext context, ShortHash hash, out ContentLoc
         /// </summary>
         public abstract IEnumerable<ShortHash> EnumerateSortedKeys(CancellationToken token);
 
+        /// <summary>
+        /// Enumeration filter used by <see cref="ContentLocationDatabase.EnumerateEntriesWithSortedKeys"/> to filter out entries by raw value from a database.
+        /// </summary>
+        public delegate bool EnumerationFilter(byte[] value);
```
ljcollins25: >EnumerationFilter [](start = 29, length = 17)

Does this need a dedicated type?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStoreConfiguration.cs
```diff
@@ -307,6 +307,11 @@ public class CheckpointConfiguration
         /// </summary>
         public TimeSpan RestoreCheckpointInterval { get; set; } = TimeSpan.FromMinutes(10);
 
+        /// <summary>
+        /// Number of files downloaded in parallel during restroring process.
+        /// </summary>
+        public int RestoreCheckpointDegreeOfParallelism { get; set; } = 24;
```
ljcollins25: >RestoreCheckpointDegreeOfParallelism [](start = 19, length = 36)

Doesn't seem to be used.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
SergeyTeplyakov: I think so, because I would like to emphesize what is 'byte[]' argument is.

---
In reply to: [276376653](https://github.com/Microsoft/BuildXL/pull/150#discussion_r276376653) [](ancestors = 276376653)

PR #155, opened 2019-04-17
Cleanup Workspace resolvers

dannyvv:


Public/Src/FrontEnd/Script/Ast/Values/ObjectLiterals/ArrayLiteral.cs
```diff
@@ -191,7 +191,7 @@ public override bool TryProject(Context context, SymbolAtom name, ModuleLiteral
                 return true;
             }
 
-            var resolvedMember = context.PredefinedTypes.AmbientArray.ResolveMember(this, name);
+            var resolvedMember = ((ModuleRegistry)context.FrontEndHost.ModuleRegistry).PredefinedTypes.AmbientArray.ResolveMember(this, name);
```
smera: >(ModuleRegistry) [](start = 34, length = 16)

Maybe add TODOs in all these places were abstractions are broken

dannyvv: I was planning to search for all Casts when cleaning it up. Should be pretty clearn

Public/Src/FrontEnd/Script/Modules/ModuleRegistry.cs
```diff
@@ -23,6 +24,12 @@ public sealed class ModuleRegistry : IModuleRegistry
         /// </summary>
         public GlobalModuleLiteral GlobalLiteral { get; }
 
+        /// <nodoc />
+        public PrimitiveTypes PrimitiveTypes { get; }
+
```
smera: just checking. By doing this type of things you may be extending the lifespan of the containing instance. In the past the module registry was confined to AST/evaluation only. Is that still the case?

Public/Src/FrontEnd/Script/Ast/Values/ObjectLiterals/ArrayLiteral.cs
smera: Sure

---
In reply to: [276370271](https://github.com/Microsoft/BuildXL/pull/155#discussion_r276370271) [](ancestors = 276370271)

Public/Src/FrontEnd/Script/Modules/ModuleRegistry.cs
dannyvv: I checked for the app, if anything they globals and primitives are initialized later now. They should all be freed at the same time as before

PR #156, opened 2019-04-17
Don't use Path.Combine with a rooted path in macOS

kewilli:
Attempting to validate copy via GRPC on macOS lead to a single realization: `Path.Combine` drops every parameter before a rooted path. We were computing the local machine location (and the root of every grpc copy path) by combining the machine name with the network path root (e.g. `Path.Combine("myMac", "/Users/kevin/...")`. When updating Redis with the location of new content, this path was evaluated as `"/Users/kevin/..."`, omitting all mention of the machine on which the content resides. This makes it impossible for the target machine to find the source of the content.

To get around this issue, this PR does what one would expect `Path.Combine` would blindly do. This change is in `GrpcDistributedPathTransformer.GetLocalMachineLocation`.

Additionally, `IsUnc` doesn't work as expected in macOS, so it is no longer enforced by `ExtractHostHashFromAbsolutePath`.

[AB#1504491](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1504491)


Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
```diff
@@ -42,9 +43,18 @@ public byte[] GetLocalMachineLocation(AbsolutePath cacheRoot)
...
-                    Path.Combine(@"\\" + _localMachineName, networkPathRoot).ToUpperInvariant());
+            if (OperatingSystemHelper.IsWindowsOS)
+            {
+                // Only unify paths along casing if on Windows
+                networkPathRoot = Path.Combine(@"\\" + _localMachineName, networkPathRoot).ToUpperInvariant();
+            }
+            else
+            {
+                // Path.Combine ignores the first parameter if the second is a rooted path. To get the machine name before the rooted network path, the combination must be done manually.
+                networkPathRoot = @"\\" + _localMachineName + (networkPathRoot.StartsWith(Path.DirectorySeparatorChar.ToString()) ? string.Empty : Path.DirectorySeparatorChar.ToString()) + networkPathRoot;
```
kewilli: >StartsWith [](start = 79, length = 10)

A simpler option which is slightly better on the eyes is the following:

`Path.Combine(@"\\" + _localMachineName + Path.DirectorySeparatorChar, networkPathRoot.TrimLeft(Path.DirectorySeparatorChar.ToString()))`

dannyvv: >DirectorySeparatorChar [](start = 152, length = 22)

on non-windows this will be `/`
so we'll end up with paths like `\\MACHINE/share\folder`
is this expected and okay?

kewilli: Good point. I'll change this from`\\MACHINE/share/folder` to `/MACHINE/share/folder`.

---
In reply to: [276399282](https://github.com/Microsoft/BuildXL/pull/156#discussion_r276399282) [](ancestors = 276399282)

PR #157, opened 2019-04-17
Remove DScriptBase from most resolvers

dannyvv:


Public/Src/FrontEnd/MsBuild/MsBuildWorkspaceResolver.cs
```diff
@@ -130,7 +136,9 @@ public string DescribeExtent()
         {
             Contract.Requires(requestedQualifiers?.Length > 0);
 
-            InitializeInterpreter(host, context, configuration);
+            m_host = host;
+            m_context = context;
+            m_configuration = configuration;
```
smera: Contracts?

PR #159, opened 2019-04-17
Execute process pips externally via sandboxed process executor tool

narasamdya:
This change introduces a so-called sandboxed process executor tool that takes a sandboxed process info as an input and outputs a sandboxed process result containing details of file accesses. 

The tool will be used to run process pips that require admin privilege, and the tool will run inside a VM. Traditionally, BuildXL in SandboxedProcessPipExecutor will create a detoured child process and communicate with the child process by means of pipes. For process pips that require admin, SandboxedProcessPipExecutor will (1) serialize sandboxed process info, (2) launch the sandboxed process executor tool, and (3) deserialize sandboxed process result produce in (2).

The sandboxed process executor tool will either replace QuickBuild's Tracker.exe or be called by QuickBuild's Tracker.exe. The cutting layer allows the two scenarios to be done, but the latter is the easiest.

Some unit tests are still in progress, but the end-to-end pipeline is ready.

This change is a sum of the following work items.

[AB#1512600](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1512600)
[AB#1512602](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1512602)
[AB#1512603](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1512603)
[AB#1512604](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1512604)


narasamdya:
As a guide for reviewing, to save your time, you may want to start with:

- SandboxedProcessPipExecutor.cs 
- Tools/SandboxedProcessExecutor
- Newly added files.

Other changes are minor or are results of refactoring

Public/Src/App/Bxl/Strings.resx
```diff
@@ -997,6 +997,9 @@ Example: ad2d42d2ec5d2ca0c0b7ad65402d07c7ef40b91e</value>
   <data name="HelpText_DisplayHelp_RunInContainerAndAllowDoubleWrites" xml:space="preserve">
     <value>Configures the default for all scheduled pips to run in a container with output isolation, allowing double writes to occur. Individual pips can override this setting. This is an unsafe option.</value>
   </data>
+  <data name="HelpText_DisplayHelp_AdminRequiredProcessExecutionMode" xml:space="preserve">
+    <value>Mode for running processes that required admin privilege in VM. Allowed values are 'Internal', 'ExternalTool', 'ExternalVM'. Defaults to 'Internal'.</value>
```
almiliMSFT: Maybe add some high-level description of the 3 allowed values.

Public/Src/Engine/Processes/ExternalSandboxedProcess.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// </summary>
+        protected void ThrowBuildXLException(string message, Exception inner = null)
+        {
+            throw new BuildXLException($"[Pip{SandboxedProcessInfo.PipSemiStableHash:X16} -- {SandboxedProcessInfo.PipDescription}] {message}", inner);
+        }
+
+        /// <summary>
+        /// Gets the file to which sandboxed process info will be written.
+        /// </summary>
+        /// <returns></returns>
```
almiliMSFT: empty `<return>`

Public/Src/Engine/Processes/ExternalSandboxedProcess.cs
```diff
@@ -0,0 +1,134 @@
...
+            string directory = Path.GetDirectoryName(SandboxedProcessInfo.FileStorage.GetFileName(SandboxedProcessFile.StandardOutput));
+            string file = Path.Combine(directory, $"SandboxedProcessInfo-Pip{SandboxedProcessInfo.PipSemiStableHash:X16}");
+
+            return file;
+        }
+
+        /// <summary>
+        /// Gets the file to which sandboxed process result will be available.
+        /// </summary>
+        /// <returns></returns>
```
almiliMSFT: ditto

Public/Src/Engine/Processes/ExternalToolSandboxedProcess.cs
```diff
@@ -0,0 +1,239 @@
...
+using BuildXL.Interop;
+using BuildXL.Native.IO;
+using BuildXL.Utilities.Tasks;
+
+namespace BuildXL.Processes
+{
+    /// <summary>
+    /// Sanboxed process that will be executed by an external tool.
+    /// </summary>
+    public class ExternalToolSandboxedProcess : ExternalSandboxedProcess
```
almiliMSFT: This class looks very similar to `UnSandboxedProcess`.  Did you consider reusing `UnSandboxedProcess`?

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -504,17 +728,86 @@ public ArraySegment<byte> GetPayloadBytes(FileAccessSetup setup, MemoryStream st
...
                 WriteDllBlock(writer, setup);
-                WriteManifestTreeBlock(writer);
+                if (!IsManifestTreeBlockSealed)
+                {
+                    WriteManifestTreeBlock(writer);
+                }
+                else
+                {
+                    Contract.Assert(m_sealedManifestTreeBlock != null);
+                    writer.Write(m_sealedManifestTreeBlock);
```
almiliMSFT: are we going to have to update MacSandbox to support `m_sealedManifestTreeBlock`?  (i.e., is that ever going to be used on macOS?)

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -504,17 +728,86 @@ public ArraySegment<byte> GetPayloadBytes(FileAccessSetup setup, MemoryStream st
...
+                // The manifest tree block has to be serialized the last.
+                if (!IsManifestTreeBlockSealed)
+                {
+                    WriteManifestTreeBlock(writer);
+                }
+                else
+                {
+                    Contract.Assert(m_sealedManifestTreeBlock != null);
+                    writer.Write(m_sealedManifestTreeBlock);
+                }
```
almiliMSFT: You could move this whole `if (!IsManifestTreeBlockSealed) { ... } else { ... }` block to `WriteManifestTreeBlock` so that you don't have to repeat it both here and in `GetPayloadBytes()`

almiliMSFT: it looks like you've already done that, so maybe this `if` (and the one in `GetPayloadBytes`) is actually unnecessary

Public/Src/Engine/Processes/ExternalToolSandboxedProcess.cs
narasamdya: Yes, I was about to reuse UnSandboxedProcess, but there are a couple of items that need to be refactored. For example, in this class we're using StringBuilder, while in the other class we are using SandboxedProcessOutputBuilder. Also, this class needs the tool path, and is not callable from SandboxedProcessFactory. 

I'll keep them separate for now. But in the future, I may want to merge them; perhaps by creating a more general class that both this class and UnSandboxedProcess can use.

Another consideration is this is a subclass of `ExternalSandboxedProcess`. Another subclass that will be created soon is the one that send a command to VM to execute the sandboxed process executor tool.

---
In reply to: [276462691](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276462691) [](ancestors = 276462691)

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -281,21 +459,47 @@ private static void WriteChars(BinaryWriter writer, string str)
...
             writer.Write((uint)timeoutInMins);
         }
 
         [SuppressMessage("Microsoft.Globalization", "CA1308:NormalizeStringsToUppercase", Justification = "Architecture strings are USASCII")]
-        private static void WriteErrorDumpLocation(
-            BinaryWriter writer, string internalDetoursErrorNotificationFile)
+        private static void WriteErrorDumpLocation(BinaryWriter writer, string internalDetoursErrorNotificationFile)
         {
 #if DEBUG
             writer.Write((uint)0xABCDEF03); // "ABCDEF03"
```
mpysson: >0xABCDEF03 [](start = 31, length = 10)

Please extract these to a shared definition

Public/Src/Engine/UnitTests/Processes.Detours/SandboxedProcessInfoTest.cs
```diff
@@ -0,0 +1,140 @@
...
+                StandardObserverDescriptor = new SandboxedProcessInfo.ObserverDescriptor()
+                {
+                    WarningRegex = new SandboxedProcessInfo.RegexDescriptor("*warn", System.Text.RegularExpressions.RegexOptions.Compiled)
+                },
+            };
+
+            // Serialize and deserialize.
+            string file = Path.GetTempFileName();
+
+            using (FileStream fileStream = File.OpenWrite(file))
```
mpysson: >FileStream fileStream = File.OpenWrite(file) [](start = 19, length = 44)

Any reason to not just use a memorystream to save the I/O?

Public/Src/Utilities/Utilities/Tracing/Events.cs
```diff
@@ -294,13 +294,15 @@ public static class Tasks
 
             public const EventTask ExecutionAnalyzers = (EventTask)22;
 
+            public const EventTask SandboxedProcessExecutor = (EventTask)23;
```
mpysson: >SandboxedProcessExecutor [](start = 35, length = 24)

Nit - you could use slot 5 to compact our list

Public/Src/Tools/SandboxedProcessExecutor/Tracing/Log.cs
```diff
@@ -0,0 +1,48 @@
...
+        public abstract void SandboxedProcessExecutorInvoked(LoggingContext context, long runtimeMs, string commandLine);
+
+        [GeneratedEvent(
+            (int)LogEventId.SandboxedProcessExecutorCatastrophicFailure,
+            EventGenerators = EventGenerators.LocalAndTelemetry,
+            EventLevel = Level.Verbose,
+            Keywords = (int)Events.Keywords.UserMessage,
+            EventTask = (int)Events.Tasks.SandboxedProcessExecutor,
+            Message = "Catastrophic failure")]
+        public abstract void SandboxedProcessExecutorCatastrophicFailure(LoggingContext context, string exceptionMessage);
```
mpysson: >exceptionMessage [](start = 104, length = 16)

I suggest including this in the message so it will get into the DX0064 message BuildXL will log for the failed pip

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
```diff
@@ -0,0 +1,322 @@
...
+                m_telemetryStopwatch.Start();
+            }
+        }
+
+        private void TelemetryShutdown()
+        {
+            if (AriaV2StaticState.IsEnabled && m_configuration.EnableTelemetry)
+            {
+                m_telemetryStopwatch.Stop();
+                Logger.Log.SandboxedProcessExecutorInvoked(m_loggingContext, m_telemetryStopwatch.ElapsedMilliseconds, Environment.CommandLine);
```
mpysson: >Logger.Log.SandboxedProcessExecutorInvoked(m_loggingContext, m_telemetryStopwatch.ElapsedMilliseconds, Environment.CommandLine); [](start = 16, length = 128)

I'm worried about this scaling especially considering the full command line is being sent to telemetry. There's also connection setup/teardown overhead by hooking the process into telemetry. Do you think we can get away with just logging crashes to telemetry and relying on the calling build engines to log the fact that the pip was executed?

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        private void HandleUnhandledFailure(Exception exception)
+        {
+            // Show the exception to the user
+            ConsoleError(exception.ToString());
+            
+            // Log the exception to telemetry
+            if (AriaV2StaticState.IsEnabled)
+            {
+                Logger.Log.SandboxedProcessExecutorCatastrophicFailure(m_loggingContext, exception.ToString());
```
mpysson: >Logger.Log.SandboxedProcessExecutorCatastrophicFailure(m_loggingContext, exception.ToString()); [](start = 16, length = 95)

pull outside of if() to ensure you always log it to the console

Public/Src/Tools/SandboxedProcessExecutor/OutputErrorObserver.cs
```diff
@@ -0,0 +1,126 @@
...
+            }
+
+            if (m_logOutputToConsole)
+            {
+                Console.Out.WriteLine(outputLine);
+            }
+        }
+
+        /// <summary>
+        /// Observes error for warning.
```
mpysson: >warning [](start = 31, length = 7)

error

Public/Src/Tools/SandboxedProcessExecutor/OutputErrorObserver.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// </summary>
+        public void ObserveError(string errorLine)
+        {
+            if (IsWarning(errorLine))
+            {
+                Interlocked.Increment(ref m_warningCount);
```
mpysson: copy/paste? warning -> error?

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
```diff
@@ -0,0 +1,322 @@
...
+                }
+            }
+
+            if (info.GetCommandLine().Length > SandboxedProcessInfo.MaxCommandLineLength)
+            {
+                ConsoleError($"Process command line is longer than {SandboxedProcessInfo.MaxCommandLineLength} characters: {info.GetCommandLine().Length}");
+                return false;
+            }
+
+            m_outputErrorObserver = OutputErrorObserver.Create(info);
```
mpysson: >OutputErrorObserver [](start = 36, length = 19)

Do we need to be processing output differently for SandboxedProcessExecutor? Can we just pipe it directly to SandboxedProcessPipExecutor and just leverage the same code that's defined there?

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -273,6 +273,7 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process
             }
 
             fingerprinter.Add("DoubleWritePolicy", (byte)process.DoubleWritePolicy);
+            fingerprinter.Add("RequiresAdmin", process.RequiresAdmin ? 1 : 0);
```
mpysson: >fingerprinter.Add("RequiresAdmin", process.RequiresAdmin ? 1 : 0); [](start = 12, length = 66)

For sake of fingerprint compatibility, only add this if RequiresAdmin is true

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
```diff
@@ -114,6 +114,9 @@ namespace Transformer {
         /** Set outputs to remain writable */
         keepOutputsWritable?: boolean;
 
+        /** Whether this process requires admin privilege */
+        requiresAdmin?: boolean;
```
dannyvv: Can we make this an Enum for the future:

`adminElevation: "none" | "required" | "desired"`

At the moment we will run all child processes as the elevation level of the engine. We should go to a mode that even if the engine is elevated all child processes will run as not elevated to protect developers.
the future "desired" can decide to run the pip as admin when he engine is run as admin and unelevated if the engine is unelevated
"required" would mean that if the engine is unelevated this pip will fail unless your new feature is enabled on the commandline to remote the work.

Public/Src/Engine/Processes/ExternalToolSandboxedProcess.cs
```diff
@@ -0,0 +1,239 @@
...
+    {
+        /// <summary>
+        /// Relative path to the default tool.
+        /// </summary>
+        public const string DefaultToolRelativePath = @"tools\SandboxedProcessExecutor\SandboxedProcessExecutor.exe";
+
```
dannyvv: If you do this you must guard usage of the feature (on the commandline parser to be windows only and fail for the Mac)

dannyvv:
A lot of these changes could have gone in as seperate smaller improvements. In the future you might consider sending out smaller more frequent pr's, this will make it also a lot faster to review.

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -281,21 +459,47 @@ private static void WriteChars(BinaryWriter writer, string str)
             }
         }
 
+        private static string ReadChars(BinaryReader reader)
```
dannyvv: >ReadChars [](start = 30, length = 9)

This is so generic, shouldn't these be extension methods or use our BuildXLReader/Writer with their helpers?

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -281,21 +459,47 @@ private static void WriteChars(BinaryWriter writer, string str)
             }
         }
 
+        private static string ReadChars(BinaryReader reader)
+        {
+            uint length = reader.ReadUInt32();
+            if (length == 0)
+            {
+                return null;
```
dannyvv: >return null; [](start = 16, length = 12)

please add comment to the fact that there is an asymetry now between writing and reading.

`read(write(null)) == null`
`read(write("")) == null `

Ideally you use regular int with  -1 for null or write an extra bit.

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -281,21 +459,47 @@ private static void WriteChars(BinaryWriter writer, string str)
             }
         }
 
+        private static string ReadChars(BinaryReader reader)
+        {
+            uint length = reader.ReadUInt32();
```
dannyvv: >ReadUInt32 [](start = 33, length = 10)

Since this is used for paths, perhaps it is more effecient to use 7-bit encoded integers.

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -281,21 +459,47 @@ private static void WriteChars(BinaryWriter writer, string str)
...
 #if DEBUG
             writer.Write((uint)0xABCDEF03); // "ABCDEF03"
 #endif
             WriteChars(writer, internalDetoursErrorNotificationFile);
         }
 
+        private static string ReadErrorDumpLocation(BinaryReader reader)
+        {
+#if DEBUG
+            uint code = reader.ReadUInt32(); // "ABCDEF03"
```
dannyvv: >// "ABCDEF03" [](start = 45, length = 13)

obvious comment below.

Public/Src/Engine/Processes/FileAccessManifest.cs
dannyvv: :+1"

---
In reply to: [276463940](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276463940) [](ancestors = 276463940)

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -504,17 +728,86 @@ public ArraySegment<byte> GetPayloadBytes(FileAccessSetup setup, MemoryStream st
...
                 WriteTranslationPathStrings(writer, DirectoryTranslator);
                 WriteErrorDumpLocation(writer, InternalDetoursErrorNotificationFile);
-                WriteFlagsBlock(writer);
+                WriteFlagsBlock(writer, m_fileAccessManifestFlag);
                 WriteExtraFlagsBlock(writer, FileAccessManifestExtraFlag.None);
                 WritePipId(writer, PipId);
                 WriteReportBlock(writer, setup);
                 WriteDllBlock(writer, setup);
-                WriteManifestTreeBlock(writer);
+                if (!IsManifestTreeBlockSealed)
```
dannyvv: >IsManifestTreeBlockSealed [](start = 21, length = 25)

Very obfuscated that this is just a null-check for `m_sealedManifestTreeBlock` field. which MUST be read/written before. Using an indirection makes this code jump out as having a big bug, where it doesn't please juse use the `m_sealedManifestTreeBlock == null` check.

dannyvv: in both serialize and deserialize

---
In reply to: [276814592](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276814592) [](ancestors = 276814592)

Public/Src/Engine/Processes/FileAccessManifest.cs
dannyvv: agree.

---
In reply to: [276464305](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276464305) [](ancestors = 276464305)

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -504,17 +728,86 @@ public ArraySegment<byte> GetPayloadBytes(FileAccessSetup setup, MemoryStream st
...
+            using (var reader = new BinaryReader(stream, Encoding.Unicode, true))
+            {
+                DirectoryTranslator directoryTranslator = ReadTranslationPathStrings(reader);
+                string internalDetoursErrorNotificationFile = ReadErrorDumpLocation(reader);
+                FileAccessManifestFlag fileAccessManifestFlag = ReadFlagsBlock(reader);
+                long pipId = ReadPipId(reader);
+                string messageCountSemaphoreName = ReadChars(reader);
+
+                byte[] sealedManifestTreeBlock;
+                using (MemoryStream ms = new MemoryStream(4096))
```
dannyvv: >4096 [](start = 58, length = 4)

Probably wise to file a workitem to see about perf. a) If this is a good buffer size, b) to see if the buffers should be pooled, now they are just allocagted and thrown away

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -524,6 +817,12 @@ public ArraySegment<byte> GetPayloadBytes(FileAccessSetup setup, MemoryStream st
         [SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times")]
         public byte[] GetManifestTreeBytes()
         {
+            if (IsManifestTreeBlockSealed)
+            {
+                Contract.Assert(m_sealedManifestTreeBlock != null);
```
dannyvv: >t.Assert(m_sea [](start = 23, length = 14)

These asserts are crazy since they are teh same check as the boolean, so why have the boolean at all?

dannyvv: Clearly you don't trust the property :)

---
In reply to: [276815332](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276815332) [](ancestors = 276815332)

Public/Src/Engine/Processes/ProcessTimes.cs
```diff
@@ -124,5 +125,17 @@ public DateTime ExitTimeUtc
...
+        public void Serialize(BinaryWriter writer)
+        {
+            writer.Write(m_create);
+            writer.Write(m_exit);
+            writer.Write(m_kernel);
+            writer.Write(m_user);
+        }
+
+        /// <nodoc />
+        public static ProcessTimes Deserialize(BinaryReader reader) => new ProcessTimes(reader.ReadInt64(), reader.ReadInt64(), reader.ReadInt64(), reader.ReadInt64());
```
dannyvv: each on their own line with a named argument will make it easier to maintain parity bewteen read & write long-term.

dannyvv:
```
public static ProcessTimes Deserialize(BinaryReader reader) => new ProcessTimes(
    create: reader.ReadInt64(), 
    exit: reader.ReadInt64(), 
    kernel: reader.ReadInt64(), 
    user: reader.ReadInt64()
);
```

---
In reply to: [276815695](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276815695) [](ancestors = 276815695)

Public/Src/Engine/Processes/ReportedFileAccess.cs
```diff
@@ -622,6 +623,69 @@ public bool TryParseAbsolutePath(PipExecutionContext context, Process pip, out A
...
+
+        /// <nodoc />
+        public static ReportedFileAccess Deserialize(
+            BuildXLReader reader, 
+            IReadOnlyList<ReportedProcess> processes, 
+            Func<BuildXLReader, AbsolutePath> readPath)
+        {
+            return new ReportedFileAccess(
+                operation: (ReportedFileOperation)reader.ReadByte(),
+                process: processes != null ? processes[reader.ReadInt32Compact()] : ReportedProcess.Deserialize(reader),
```
dannyvv: >ReportedProcess [](start = 84, length = 15)

Why the asymetry between `Process.Serialize` and `ReportedProcess.Deserialize`?

Public/Src/Engine/Processes/ReportedFileAccess.cs
```diff
@@ -622,6 +623,69 @@ public bool TryParseAbsolutePath(PipExecutionContext context, Process pip, out A
...
+                requestedAccess: (RequestedAccess)reader.ReadInt32Compact(),
+                status: (FileAccessStatus)reader.ReadInt32Compact(),
+                explicitlyReported: reader.ReadBoolean(),
+                error: reader.ReadUInt32(),
+                usn: new Usn(reader.ReadUInt64()),
```
dannyvv: > usn: new Usn(reader.ReadUInt64() [](start = 15, length = 33)

Do USN's actually translate well accross VM's?

Public/Src/Engine/Processes/ReportedFileAccess.cs
```diff
@@ -622,6 +623,69 @@ public bool TryParseAbsolutePath(PipExecutionContext context, Process pip, out A
...
+            writer.Write(Error);
+            writer.Write(Usn.Value);
+            writer.Write((uint)DesiredAccess);
+            writer.Write((uint)ShareMode);
+            writer.Write((uint)CreationDisposition);
+            writer.Write((uint)FlagsAndAttributes);
+
+            if (writePath != null)
+            {
+                writePath.Invoke(writer, ManifestPath);
```
dannyvv: >Invoke [](start = 26, length = 6)

Why are you using `.Invoke` here? Please just use `writePath(writer, ManifestPath)`

Public/Src/Engine/Processes/ReportedProcess.cs
```diff
@@ -89,5 +90,35 @@ public ReportedProcess(uint processId, string path)
...
+            writer.Write(UserTime);
+            IOCounters.Serialize(writer);
+            writer.Write(ExitCode);
+            writer.Write(ParentProcessId);
+        }
+
+        /// <nodoc />
+        public static ReportedProcess Deserialize(BuildXLReader reader)
+        {
+            var reportedProcess = new ReportedProcess(reader.ReadUInt32(), reader.ReadString(), reader.ReadString());
```
dannyvv: >eader.Re [](start = 55, length = 8)

named arguments for ease of reviewing and future maintanance

Public/Src/Engine/Processes/ReportedProcess.cs
```diff
@@ -89,5 +90,35 @@ public ReportedProcess(uint processId, string path)
...
+            reportedProcess.CreationTime = reader.ReadDateTime();
+            reportedProcess.ExitTime = reader.ReadDateTime();
```
dannyvv: >ss.Exi [](start = 25, length = 6)

Cleaner to use object literal construction:
```
new X() {
  prop=value
}
```

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -364,5 +366,451 @@ public byte[] GetUnicodeEnvironmentBlock()
...
+        }
+
+        #endregion
+
+        #region Extra data structure for serialization
+
+        /// <summary>
+        /// Info about the source of standard input.
+        /// </summary>
+        public class StandardInputInfo : IEquatable<StandardInputInfo>
```
dannyvv: >StandardInputInfo [](start = 21, length = 17)

seperate file please

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -364,5 +366,451 @@ public byte[] GetUnicodeEnvironmentBlock()
...
+            }
+
+            /// <summary>
+            /// Deserializes an instance of <see cref="StandardInputInfo"/>
+            /// </summary>
+            public static StandardInputInfo Deserialize(BuildXLReader reader)
+            {
+                Contract.Requires(reader != null);
+
+                return new StandardInputInfo(reader.ReadNullableString(), reader.ReadNullableString());
```
dannyvv: named arguments

dannyvv:
```
tandardInputInfo(
    file: reader.ReadNullableString(), 
    data: reader.ReadNullableString()
);
```

---
In reply to: [276819630](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276819630) [](ancestors = 276819630)

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -364,5 +366,451 @@ public byte[] GetUnicodeEnvironmentBlock()
...
+                {
+                    return true;
+                }
+
+                if (info1 is null)
+                {
+                    return false;
+                }
+
+                return info1.Equals(info2);
```
dannyvv: Why composed this way and not have `Equals` use the == operator since that is the more geenric case?

Public/Src/FrontEnd/UnitTests/Libs/Sdk.Transformers.ts
```diff
@@ -168,6 +168,9 @@ namespace Transformer {
             /** Set outputs to remain writable */
             keepOutputsWritable?: boolean;
 
+            /** Whether this process requires admin privilege */
+            requiresAdmin?: boolean;
+
 			/** Whether this process should run in an isolated container (i.e. filesystem isolation)
```
dannyvv: >	 [](start = 1, length = 1)

tabs vs spaces

Public/Src/Tools/SandboxedProcessExecutor/Args.cs
```diff
@@ -0,0 +1,85 @@
...
+                else
+                {
+                    Console.Error.WriteLine($"Option '{option.Name}' is not recognized");
+                    return false;
+                }
+            }
+
+            if (!configuration.Validate(out string errorConfiguration))
+            {
+                Console.Error.WriteLine($"Configuration error: {errorConfiguration}");
```
dannyvv: >Configuration error: { [](start = 42, length = 22)

These strings should be in a ResX file for localization

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
dannyvv: isn't that done on the first line in this file?

---
In reply to: [276484932](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276484932) [](ancestors = 276484932)

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        /// <summary>
+        /// Creates an instance of <see cref="Executor"/>.
+        /// </summary>
+        public Executor(Configuration configuration)
+        {
+            Contract.Requires(configuration != null);
+
+            m_configuration = configuration;
+    }
```
dannyvv: NIT: indentation is off

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
```diff
@@ -0,0 +1,322 @@
...
+            Contract.Requires(result != null);
+
+            try
+            {
+                using (FileStream stream = File.OpenWrite(Path.GetFullPath(m_configuration.SandboxedProcessResultOutputFile)))
+                {
+                    result.Serialize(stream);
+                }
+            }
+            catch (IOException ioException)
```
dannyvv: >IOException [](start = 19, length = 11)

Wherever you catch IOException in this file you must also catch UnauthorizedAccessException

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
```diff
@@ -0,0 +1,322 @@
...
+                catch (BuildXLException ex)
+                {
+                    if (ex.LogEventErrorCode == NativeIOConstants.ErrorFileNotFound)
+                    {
+                        ConsoleError($"Failed to start process: '{info.FileName}' not found");
+                        return null;
+                    }
+                    else if (ex.LogEventErrorCode == NativeIOConstants.ErrorPartialCopy && (processRelaunchCount < ProcessRelauchCountMax))
+                    {
+                        ++processRelaunchCount;
```
dannyvv: >processRelaunchCount [](start = 26, length = 20)

`processRelaunchCount++`

dannyvv: >processRelaunchCount [](start = 26, length = 20)

Please log a message that we are going to launch the process again which should include the nr of times we will retry after this

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
```diff
@@ -0,0 +1,322 @@
...
+                return null;
+            }
+        }
+
+        private void ConsoleError(string errorMessage)
+        {
+            ConsoleColor original = Console.ForegroundColor;
+            Console.ForegroundColor = ConsoleColor.Red;
+            Console.Error.WriteLine(errorMessage);
+            Console.ForegroundColor = original;
```
dannyvv: This needs some locking in case you have multiple threads writing to the console...

Public/Src/Tools/SandboxedProcessExecutor/OutputErrorObserver.cs
```diff
@@ -0,0 +1,126 @@
...
+            }
+
+            if (m_logErrorToConsole)
+            {
+                Console.Error.WriteLine(errorLine);
```
dannyvv: >e.Error.WriteLine(e [](start = 22, length = 19)

Shouldn't this be colored as well like in Executor?

Public/Src/Tools/SandboxedProcessExecutor/SandboxedProcessExecutor.dsc
```diff
@@ -0,0 +1,36 @@
...
+
+    export declare const qualifier: BuildXLSdk.DefaultQualifier;
+
+    @@public
+    export const exe = BuildXLSdk.executable({
+        generateLogs: true,
+        allowUnsafeBlocks: true,
+        assemblyName: "SandboxedProcessExecutor",
+        rootNamespace: "BuildXL.SandboxedProcessExecutor",
+        skipDocumentationGeneration: true,
```
dannyvv: >skipDocumentationGeneration [](start = 8, length = 27)

This should be false

Public/Src/Utilities/Configuration/Mutable/SandboxConfiguration.cs
```diff
@@ -213,5 +215,8 @@ public UnsafeSandboxConfiguration UnsafeSandboxConfigurationMutable
 
         /// <inheritdoc/>
         ISandboxContainerConfiguration ISandboxConfiguration.ContainerConfiguration => ContainerConfiguration;
+
+        /// <inheritdoc />
+        public AdminRequiredProcessExecutionMode AdminRequiredProcessExecutionMode { get; set; }
```
dannyvv: >AdminRequiredProcessExecutionMode [](start = 49, length = 33)

More proof that in the DScript side it should be an enum instead of a simple boolean

Public/Src/Utilities/Utilities/BuildXLReaderExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace BuildXL.Utilities
+{
+    /// <summary>
+    /// Extensions for <see cref="BuildXLReader"/>.
+    /// </summary>
+    public static class BuildXLReaderExtensions
```
dannyvv: You should add your char read & write methods here

Public/Src/Engine/Processes/FileAccessManifest.cs
narasamdya: I don't know if we're ever going to use it on MacOS. But for sure, to use it on Mac, we need to ensure that the manifest tree is not modified once we create `ISandboxedProcess`. Currently it's not the case for Mac. are we going to have to update MacSandbox to support `m_sealedManifestTreeBlock`?  

---
In reply to: [276463940](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276463940) [](ancestors = 276463940)

Public/Src/Engine/Processes/FileAccessManifest.cs
narasamdya: Unfortunately not. GetManifestTreeBytes does serialize the manifest if manifest is not sealed. For testing, this method is appropriate, but it doesn't seem we can use that for serialization and deserialization.

---
In reply to: [276464697](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276464697) [](ancestors = 276464697)

Public/Src/Engine/UnitTests/Processes.Detours/SandboxedProcessInfoTest.cs
narasamdya: No particular reason. I'm changing it to memory stream.

---
In reply to: [276482723](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276482723) [](ancestors = 276482723)

Public/Src/Utilities/Utilities/Tracing/Events.cs
narasamdya: I thought it was reserved for something else, but if not, I'll use it.

---
In reply to: [276482913](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276482913) [](ancestors = 276482913)

Public/Src/Tools/SandboxedProcessExecutor/Tracing/Log.cs
narasamdya: I'm going to make all of event logs for the tool to be telemetry only.

---
In reply to: [276483642](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276483642) [](ancestors = 276483642)

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
narasamdya: As discussed, I'll change the default enableTelemetry to false, and make it false if it runs under a build engine.

---
In reply to: [276484358](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276484358) [](ancestors = 276484358)

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
narasamdya: Yes, like other tools, I want to start first with Console.WriteLine.

---
In reply to: [276822890](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276822890) [](ancestors = 276822890,276484932)

Public/Src/Tools/SandboxedProcessExecutor/OutputErrorObserver.cs
narasamdya: No we observed standard error for warning.

---
In reply to: [276485208](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276485208) [](ancestors = 276485208)

Public/Src/Tools/SandboxedProcessExecutor/OutputErrorObserver.cs
narasamdya: No, we're observing warning from std error.

---
In reply to: [276485314](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276485314) [](ancestors = 276485314)

Public/Src/Tools/SandboxedProcessExecutor/OutputErrorObserver.cs
narasamdya: Yes. Danny also spotted that.

---
In reply to: [276826980](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276826980) [](ancestors = 276826980)

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
narasamdya: This is what I'm doing. This observer is essentially used to observe warning lines, and increment the counter. We do the same thing in  `SanboxedProcessPipExecutor.cs`. If we want to use the code in `SanboxedProcessPipExecutor.cs`, then I need to establish a pipe between the sandboxed executor tool and `SandboxedProcessPipExecutor.cs`. Currently, we don't want to go into that direction because establishing pipes leads us to implementing client/server, which is a bit error prone, and moreover, we need to understand the security policy for those pipes in VM.

---
In reply to: [276486637](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276486637) [](ancestors = 276486637)

narasamdya:
Thanks for the feedback. I realized this before sending the PR. The meat of changes here is not big. However, to reuse code, I need to do some refactoring, and thus many files got touched. That's why I gave a guide on how to effectively review this PR.

---
In reply to: [484666763](https://github.com/Microsoft/BuildXL/pull/159#issuecomment-484666763) [](ancestors = 484666763)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
narasamdya: Mike and I discuss the run unelevated case, and we doubt that the scenario will appear. I'll start simple with boolean. This simplicity allows me to use Process.Options (no changes to Process.cs and ProcessBuilder.cs). 

Also the enforcement of the required-admin flag will be done using the runexternal feature. If internal or external tool, then it runs with the elevation level of the engine. If VM, then it runs with VM admin privilege.

For dev build, the default will be internal (we don't expect people to always have VM ready). In this case, you can think that we run the required-admin pip best effort as admin. In CB, the default will be VM.

For tests that require admin, if we run in VM, then the tests should fail if they are not run under admin in VM. This last feature still need to be done.

---
In reply to: [276810187](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276810187) [](ancestors = 276810187)

Public/Src/Engine/Processes/FileAccessManifest.cs
narasamdya: BinaryWriter/Reader have already methods for writing/reading string.

What I'm doing here is reuse WriteChars method that is used to serialize manifest to Detours code. By reusing, I introduced ReadChars that "matches" WriteChars. Yes, I'm aware of the asymmetry, and will add comment about that. But, changing WriteChars entails changing the code in Detours as well. I'm not doing that right now.

---
In reply to: [276812324](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276812324) [](ancestors = 276812324)

Public/Src/Engine/Processes/FileAccessManifest.cs
narasamdya: See previous comment about Detours.

---
In reply to: [276813401](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276813401) [](ancestors = 276813401)

Public/Src/Engine/Processes/FileAccessManifest.cs
narasamdya: Yes, I'm aware of this, and will add a comment about that.

---
In reply to: [276813025](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276813025) [](ancestors = 276813025)

Public/Src/Engine/Processes/FileAccessManifest.cs
narasamdya: I introduced property as a  name for that condition. Somehow, it's a reflex when I added that contract :-)

---
In reply to: [276815368](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276815368) [](ancestors = 276815368,276815332)

Public/Src/Engine/Processes/ReportedFileAccess.cs
narasamdya: I think that the methods look asymmetric because of additional arguments, `processMap` and `processes`. These two additional arguments are used for compacting the serialization, i.e., we don't need to repeatedly serialize the same `ReportedProcess`.

---
In reply to: [276816570](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276816570) [](ancestors = 276816570)

Public/Src/Engine/Processes/ReportedFileAccess.cs
narasamdya: I made it optional at first, so I was writing something like writePath?.Invoke(..). But then I changed it, and forget to remove the invoke.

---
In reply to: [276817065](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276817065) [](ancestors = 276817065)

Public/Src/Engine/Processes/ReportedFileAccess.cs
narasamdya: In general, they don't. But note that we're going to map host volume in the VM, and thus the USN's can still be used.

---
In reply to: [276816840](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276816840) [](ancestors = 276816840)

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
narasamdya: I thought they were small at first, but they grew large, and yes, we need to place them in separate files.

---
In reply to: [276819364](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276819364) [](ancestors = 276819364)

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
narasamdya: There are a few ways of writing equalities. I can use == in Equals, but it will only save me `ReferenceEquals(this, standardInputInfo)` from Equals.

---
In reply to: [276820203](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276820203) [](ancestors = 276820203)

Public/Src/FrontEnd/UnitTests/Libs/Sdk.Transformers.ts
narasamdya: I think mine is spaces (consistent with the previous fields). The container fields are all tabs.

---
In reply to: [276821383](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276821383) [](ancestors = 276821383)

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
narasamdya: Haha, style :-)

---
In reply to: [276824008](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276824008) [](ancestors = 276824008)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
dannyvv: Interally you can keep a boolean, on the flags no problem!
I just want our API to be future-proof. Also just in case we get multiple ways to do admin, maybe no the mac it will be different and one way will work or not.

Let's just feature-proof it.
I'm happy to do the work for you if that is the problem, just let me know and I'll pusht he change to your branch.

"Waiting" for a response on that for now.



---
In reply to: [277062597](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277062597) [](ancestors = 277062597,276810187)

Public/Src/Engine/Processes/ReportedFileAccess.cs
dannyvv: Please add a comment and assert to the effect that it happens to be a magic VM by cloudbuild that happens to run on this box. 
There is nothing that precludes someone in the future to change the VM and put it on single machine inthe machine pool to lock down futher and how we are conflating usn's. I pitty the person that will have to debug it then.

---
In reply to: [277070365](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277070365) [](ancestors = 277070365,276816840)

Public/Src/Tools/SandboxedProcessExecutor/Executor.cs
dannyvv: it was indeed our running joke :D

---
In reply to: [277081280](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277081280) [](ancestors = 277081280,276824008)

Public/Src/Tools/SandboxedProcessExecutor/OutputErrorObserver.cs
dannyvv: Cool, didn't know I split in two :)

---
In reply to: [276874292](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276874292) [](ancestors = 276874292,276826980)

Public/Src/Utilities/Configuration/Mutable/SandboxConfiguration.cs
narasamdya: See the discussion about enum in the previous comment.

---
In reply to: [276827371](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276827371) [](ancestors = 276827371)

Public/Src/Engine/Processes/ExternalToolSandboxedProcess.cs
narasamdya: I decided to refactor the common part into AsyncProcessExecutor.

---
In reply to: [276465387](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276465387) [](ancestors = 276465387,276462691)

Public/Src/Engine/Processes/FileAccessManifest.cs
narasamdya: Mike doesn't seem to be fond of introducing work items. I'll add TODO instead.

---
In reply to: [276815177](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276815177) [](ancestors = 276815177)

Public/Src/Engine/Processes/ReportedFileAccess.cs
narasamdya: I'll add it in the next iteration.

---
In reply to: [277084423](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277084423) [](ancestors = 277084423,277070365,276816840)

Public/Src/Tools/SandboxedProcessExecutor/SandboxedProcessExecutor.dsc
narasamdya: Will be removed in the next iteration.

---
In reply to: [276827074](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276827074) [](ancestors = 276827074)

Public/Src/Utilities/Utilities/BuildXLReaderExtensions.cs
narasamdya: See comment in the `FileAccessManifest.cs`. 

---
In reply to: [276827527](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276827527) [](ancestors = 276827527)

Public/Src/App/Bxl/Args.cs
```diff
@@ -203,6 +203,9 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                             "additionalConfigFile",
                             "ac",
                             opt => ParsePathOption(opt, pathTable, startupConfiguration.AdditionalConfigFiles)),
+                        OptionHandlerFactory.CreateOption(
```
almiliMSFT: For command line switches that make sense for macOS only we currently use `#if PLATFORM_OSX`.

That said, maybe it would make more sense to use `#if !PLATFORM_OSX` here instead of checking and throwing below (line 1224)

Public/Src/Engine/Processes/AsyncProcessExecutor.cs
```diff
@@ -0,0 +1,235 @@
...
+            m_timeout = timeout;
+            m_sandboxedProcessInfo = sandboxedProcessInfo;
+        }
+
+        /// <summary>
+        /// Starts process.
+        /// </summary>
+        public void Start()
+        {
+            if (!FileUtilities.FileExistsNoFollow(Process.StartInfo.FileName))
```
almiliMSFT: Any particular reason to check `FileExistsNoFollow` here?  Do you intentionally want to forbid running a process through a symlink?

Public/Src/Engine/Processes/AsyncProcessExecutor.cs
```diff
@@ -0,0 +1,235 @@
...
+using BuildXL.Native.IO;
+using BuildXL.Utilities;
+using BuildXL.Utilities.Tasks;
+
+namespace BuildXL.Processes
+{
+    /// <summary>
+    /// Executes process asynchronously.
+    /// </summary>
+    public class AsyncProcessExecutor : IDisposable
```
almiliMSFT: 👍

Public/Src/Engine/Processes/ExternalSandboxedProcess.cs
```diff
@@ -0,0 +1,157 @@
...
+                {
+                    return SandboxedProcessResult.Deserialize(stream);
+                }
+            }
+            catch (IOException ioException)
+            {
+                ThrowBuildXLException($"Failed to deserialize sandboxed process result '{file}'", ioException);
+            }
+
+            return null;
```
almiliMSFT: nit: maybe move inside the `catch` (to make it clearer that this is unreachable)

Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -210,37 +186,21 @@ public virtual async Task<SandboxedProcessResult> GetResultAsync()
...
-            m_exitTime = DateTime.UtcNow;
-
-            // 2: kill process if timed out
-            var timedOut = finishedTask != WhenExited;
-            if (timedOut)
-            {
-                LogProcessState($"Process timed out after {ProcessInfo.Timeout.Value}; it will be forcefully killed.");
-                await KillAsync();
-            }
+            await m_processExecutor.WaitForExitAsync();
```
almiliMSFT: Although I like it better how it looks now, I am not 100% sure that the new semantics of this method is going to always be correct.

Before, we executed steps 1, 2, 3, 4 in order; now the order changed to 1, 2, 4, 3 (i.e., wait for stdout/stderr before waiting for process reports).  In most cases, this order doesn't matter.  But, consider the following scenario:
  - process `A` starts
  - process `A` spawns a child process `a` which is declared as "allowed to survive nested process"
  - `A` exits; since `a` is still running (and will never finish), stdout/stderr of `A` may still be opened (because child processes often inherit them)
It is `GetReportsAsync()` that checks if any child processes are still running and, in particular, handles allowed surviving child processes.  With this change, that code may never kick in. 

Adding @Kr1sso.  It would probably make sense to get some feedback from him on this too.

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -364,5 +366,163 @@ public byte[] GetUnicodeEnvironmentBlock()
...
+                    PipSemiStableHash = pipSemiStableHash,
+                    TimeoutDumpDirectory = timeoutDumpDirectory,
+                    SandboxKind = sandboxKind,
+                    PipDescription = pipDescription,
+                    SandboxedProcessStandardFiles = sandboxedProcessStandardFiles,
+                    StandardInputSourceInfo = standardInputSourceInfo,
+                    StandardObserverDescriptor = standardObserverDescriptor
+                };
+
+                Encoding GetEncoding(BuildXLReader r)
```
dannyvv: Nit:`DeserializeEncoding` since you are using a reader

Public/Src/Engine/Processes/SandboxedProcessOutput.cs
```diff
@@ -63,6 +63,59 @@ public sealed class SandboxedProcessOutput
...
+            string value = reader.ReadNullableString();
+            string fileName = reader.ReadNullableString();
+
+            int codePage = reader.ReadInt32();
+
+#if DISABLE_FEATURE_EXTENDED_ENCODING
+            Encoding encoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+#else
+            Encoding encoding = Encoding.GetEncoding(codePage);
+#endif
```
dannyvv: Why not user the uelper you added to SandboxProcessInfo?

Public/Src/Utilities/Utilities/BuildXLReaderExtensions.cs
ljcollins25: >BuildXLReaderExtensions [](start = 24, length = 23)

Is this needed? Can't you just add methods to the BuildXLReader/Writer directly.

Public/Src/App/Bxl/Args.cs
narasamdya: This may potentially be used in `PLATFORM_OSX`. For example, we may want to run pip with super user privilege. Currently, the execution mode name comes under the same umbrella called 'admin'. 

For the time being, we'll keep the option as is without enclosing it in `#if !PLATFORM_OSX`.

---
In reply to: [277332947](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277332947) [](ancestors = 277332947)

Public/Src/Engine/Processes/AsyncProcessExecutor.cs
narasamdya: There's no particular reason. It's a kind a reflex after working on symlinks for some time :-)

---
In reply to: [277334192](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277334192) [](ancestors = 277334192)

Public/Src/Engine/Processes/AsyncProcessExecutor.cs
narasamdya: Thanks for the thumb. How did you produce it?

---
In reply to: [277335432](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277335432) [](ancestors = 277335432)

Public/Src/Engine/Processes/UnSandboxedProcess.cs
narasamdya: I'll pass an argument to `WaitForExitAsync` to get the reports in the middle.

---
In reply to: [277343477](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277343477) [](ancestors = 277343477)

Public/Src/Utilities/Utilities/BuildXLReaderExtensions.cs
narasamdya: I'm thinking that all the derivative methods should go to the extension. Or, do you have any concern?

---
In reply to: [277360478](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277360478) [](ancestors = 277360478)

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
```diff
@@ -114,6 +114,9 @@ namespace Transformer {
         /** Set outputs to remain writable */
         keepOutputsWritable?: boolean;
 
+        /** Privilege level requires by this process to execute. */
```
Kr1sso: Nit: required

Public/Src/Deployment/buildXL.dsc
```diff
@@ -8,6 +8,9 @@ namespace BuildXL {
 
     export declare const qualifier: BuildXLSdk.DefaultQualifier;
 
+    const genVSEnvVar = "[Sdk.BuildXL]GenerateVSSolution";
```
Kr1sso: Extract this into the BuildXL SDK to avoid duplication?

Public/Src/Engine/Processes/AsyncProcessExecutor.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        /// <summary>
+        /// Task that completes once this process dies.
+        /// </summary>
+        private Task WhenExited => m_processExitedTcs.Task;
+
+        /// <summary>
+        /// Checks if process exit is completed.
+        /// </summary>
+        public bool ExitCompleted => m_processExitedTcs.Task.IsCompleted;
```
Kr1sso: `WhenExited.IsCompleted` instead?

Public/Src/Engine/Processes/AsyncProcessExecutor.cs
```diff
@@ -0,0 +1,259 @@
...
+            m_timeout = timeout;
+            m_sandboxedProcessInfo = sandboxedProcessInfo;
+        }
+
+        /// <summary>
+        /// Starts process.
+        /// </summary>
+        public void Start()
+        {
+
```
Kr1sso: Nit: A lot of whitespace

Public/Src/FrontEnd/UnitTests/Libs/Sdk.Transformers.ts
```diff
@@ -168,25 +168,28 @@ namespace Transformer {
...
-			 * TODO: input isolation is not implemented
-			 */
-			containerIsolationLevel?: ContainerIsolationLevel;
-
-			/**
-			 * The policy to apply when a double write occurs.
-			 * Default is globally controlled by the sandbox configuration
-			 */
-			doubleWritePolicy?: DoubleWritePolicy;
+            /** Privilege level requires by this process to execute. */
```
Kr1sso: Nit: required

Public/Src/Tools/SandboxedProcessExecutor/Args.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        private void WriteHelp()
+        {
+            HelpWriter writer = new HelpWriter();
+            writer.WriteBanner("Tool for executing and monitoring process in a sandbox");
+
+            writer.WriteLine("");
+            writer.WriteLine("/sandboxedProcessInfo:<file>   -- Sandboxed process info input file [short: /i]");
+            writer.WriteLine("/sandboxedProcessResult:<file> -- Sandboxed process result output file [short: /r]");
+            writer.WriteLine("/sandboxedProcessResult:<file> -- Sandboxed process result output file [short: /r]");
```
Kr1sso: Duplicate of line 79?

Public/Src/Tools/SandboxedProcessExecutor/Program.cs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using BuildXL.Processes;
+using BuildXL.ToolSupport;
+
+namespace BuildXL.SandboxedProcessExecutor
+{
+    internal sealed class Program : ToolProgram<Args>
```
Kr1sso: Looking at it, seems to me like we should merge this with SandboxExec at some point and have only one tool.

Public/Src/Engine/Processes/UnSandboxedProcess.cs
Kr1sso: 👍 Thanks.

Public/Src/Tools/SandboxedProcessExecutor/Program.cs
narasamdya: Agree. At the moment I'm a bit hesitant to change SandboxExec because things would look ugly. I created AB#1516956 to track the issue.


---
In reply to: [277769688](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277769688) [](ancestors = 277769688)

Public/Src/Engine/Processes/AsyncProcessExecutor.cs
almiliMSFT: from GitHub UI: I just press type `:` and wait for an automatic completion to offer the thumbs up emoji

Public/Src/Engine/UnitTests/Processes.Detours/SandboxedProcessInfoTest.cs
mpysson: Still a FileStream. Reactivating in case you forgot

---
In reply to: [276867353](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276867353) [](ancestors = 276867353,276482723)

Public/Src/Tools/SandboxedProcessExecutor/OutputErrorObserver.cs
mpysson: you're right

---
In reply to: [276870633](https://github.com/Microsoft/BuildXL/pull/159#discussion_r276870633) [](ancestors = 276870633,276485208)

Public/Src/App/Bxl/Args.cs
```diff
@@ -1215,7 +1218,15 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
...
                 }
-                
+
+                if (OperatingSystemHelper.IsUnixOS)
+                {
+                    // TODO: Non Windows OS doesn't support admin-required process external execution mode.
+                    if (sandboxConfiguration.AdminRequiredProcessExecutionMode != AdminRequiredProcessExecutionMode.Internal)
+                    {
+                        throw CommandLineUtilities.Error(Strings.Args_AdminRequiredProcessExecutionMode_NotSupportedOnNonWindows, sandboxConfiguration.AdminRequiredProcessExecutionMode.ToString());
+                    }
```
mpysson: What's going to happen for our test pips that require admin on Windows but don't on macOS? For instance the integration test assembly runs incremental scheduling tests on Windows and needs admin but those test cases don't apply to macOS

narasamdya: When running on macOS, currently one should not change the execution mode, and so pips will be run as they are today. Yes, some tests are only applied to windows. For inc. scheduling tests, all of them are marked `requiresJournalScan: true` which make them get filtered out when running on macOS. Later, if we need to address low-privilege on macOS, we can enrich the `privilegeLevel` enum in the pip specification. For example: `standard | adminWin | suUnix`.

---
In reply to: [277877974](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277877974) [](ancestors = 277877974)

Public/Src/Engine/UnitTests/Processes.Detours/SandboxedProcessInfoTest.cs
narasamdya: I fixed one in `FileAccessManifest.cs`, but apparently your comment here is in a different file :-)

---
In reply to: [277861318](https://github.com/Microsoft/BuildXL/pull/159#discussion_r277861318) [](ancestors = 277861318,276867353,276482723)

PR #166, opened 2019-04-18
Sever dependency from BuildXL.FrontEnd.Script on BuildXL.FrontEnd.Core

dannyvv:


Public/Src/FrontEnd/Script/BuildXL.FrontEnd.Script.dsc
```diff
@@ -17,7 +17,7 @@ namespace Script {
                 importFrom("System.Collections.Immutable").pkg,
             ]),
 
-            Core.dll,
+            //Core.dll,
```
smera: >//Core.dll, [](start = 12, length = 11)

remove?

PR #169, opened 2019-04-19
fix path issue under macOS

Jamesits:
macOS filesystems are case-sensitive by default.


Jamesits:
@dannyvv I found mono is case insensitive while searching for files…

dannyvv:
@jamesits This looks like a good change anyways. I'd like to keep it, unless you disagree.

Jamesits:
@dannyvv I’m fine to keep it

dannyvv:
Thanks @Jamesits!

PR #172, opened 2019-04-19
Use exponential back off with key vault.

SergeyTeplyakov:
Instead of accessing a key vault instance once, the code starts using exponential backoff with some random seed to prevent key vault from being overwhelmed too quickly.

### DIFF
--- a/Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
+++ b/Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
@@ -24,6 +24,7 @@ using BuildXL.Cache.ContentStore.Interfaces.Stores;
 using BuildXL.Cache.ContentStore.Interfaces.Time;
 using BuildXL.Cache.Host.Configuration;
 using BuildXL.Cache.ContentStore.Utils;
+using Microsoft.Practices.TransientFaultHandling;
 
 namespace BuildXL.Cache.Host.Service.Internal
 {
@@ -377,24 +378,20 @@ namespace BuildXL.Cache.Host.Service.Internal
             if (_distributedSettings.EventHubSecretName != null &&
                 _distributedSettings.GlobalRedisSecretName != null)
             {
-                try
-                {
-                    return await _arguments.Host.RetrieveKeyVaultSecretsAsync(
-                        new List<string>(storageSecretNames)
+                var retryPolicy = CreateKeyVaultRetryPolicy();
+                return await retryPolicy.ExecuteAsync(
+                    async () =>
+                    {
+                        return await _arguments.Host.RetrieveKeyVaultSecretsAsync(
+                            new List<string>(storageSecretNames)
                             {
                                 _distributedSettings.EventHubSecretName,
                                 _distributedSettings.GlobalRedisSecretName,
                                 _distributedSettings.SecondaryGlobalRedisSecretName
                             }.Where(s => s != null).ToList(),
-                        token);
-                }
-                // In some cases, KeyVault provider may fail with HttpRequestException with an inner exception like 'The remote name could not be resolved: 'login.windows.net'.
-                // Theoretically, this should be handled by the host, but to make error handling simple and consistent (this method throws one exception type) the handling is happening here.
-                catch (Exception e) when (e.Message.Contains("The remote name could not be resolved"))
-                {
-                    errorBuilder.Append($"Unable to get key vault settings because the key vault server is unavailable. Exception: {e}. ");
-                    return null;
-                }
+                            token);
+                    },
+                    token);
             }
 
             return null;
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -451,5 +448,43 @@ public KeyVaultConfigurationException(string message, Exception innerException)
...
+        private static RetryPolicy CreateKeyVaultRetryPolicy()
+        {
+            return new RetryPolicy(
+                new KeyVaultRetryPolicy(),
+                new ExponentialBackoff(
+                    name: "KeyVaultExponentialBackoff",
+                    retryCount: 5,
+                    minBackoff: TimeSpan.FromSeconds(10),
+                    maxBackoff: TimeSpan.FromSeconds(60),
+                    deltaBackoff: TimeSpan.FromSeconds(10),
```
dannyvv: >deltaBackoff [](start = 20, length = 12)

NIT: pr is good, just a generic comment that api makes it a bit hard to read what is really happening from the callsite without knowing the implementation.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -451,5 +448,43 @@ public KeyVaultConfigurationException(string message, Exception innerException)
...
             }
         }
+
+        private static RetryPolicy CreateKeyVaultRetryPolicy()
+        {
+            return new RetryPolicy(
+                new KeyVaultRetryPolicy(),
+                new ExponentialBackoff(
+                    name: "KeyVaultExponentialBackoff",
+                    retryCount: 5,
```
kewilli: >5 [](start = 32, length = 1)

Should this, minimum backoff, maximum backoff, and delta backoff be configurable?

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -451,5 +448,43 @@ public KeyVaultConfigurationException(string message, Exception innerException)
...
+
+                if (message.Contains("The remote name could not be resolved"))
+                {
+                    // In some cases, KeyVault provider may fail with HttpRequestException with an inner exception like 'The remote name could not be resolved: 'login.windows.net'.
+                    // Theoretically, this should be handled by the host, but to make error handling simple and consistent (this method throws one exception type) the handling is happening here.
+                    // This is a recoverable error.
+                    return true;
+                }
+
+                if (message.Contains("429"))
```
kewilli: >message.Contains("429") [](start = 20, length = 23)

This seems incredibly generic. Do we know the expected exception type? Or anything else within the body of the error message?

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
SergeyTeplyakov: It may, but this is a temporary solution before we move to the secret store. And we're not planning to use key vault in a way that will cause issues even without the backoff algorithm. So I'm not sure the configuration is worth it.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
SergeyTeplyakov: It is KeyVaultErrorException but the exception type is internal.

The full message is: Operation returned an invalid status code '429'. And I can look for it.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -451,5 +448,43 @@ public KeyVaultConfigurationException(string message, Exception innerException)
...
+
+        private static RetryPolicy CreateKeyVaultRetryPolicy()
+        {
+            return new RetryPolicy(
+                new KeyVaultRetryPolicy(),
+                new ExponentialBackoff(
+                    name: "KeyVaultExponentialBackoff",
+                    retryCount: 5,
+                    minBackoff: TimeSpan.FromSeconds(10),
+                    maxBackoff: TimeSpan.FromSeconds(60),
```
ljcollins25: >60 [](start = 53, length = 2)

Make this configurable.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
ljcollins25: >5 [](start = 32, length = 1)

Make this configurable.

PR #178, opened 2019-04-19
Use historic cpu usage info as weight

semihokur:
In the historic running timetable, we do record the cpu usage per pip. That is calculated as 100*(usertime+kerneltime)/wallclocktime. That information can be used as weight. This is mainly for WDG builds.

### DIFF
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -1026,6 +1026,9 @@ namespace BuildXL
                             "useHardlinks",
                             sign => engineConfiguration.UseHardlinks = sign),
                         OptionHandlerFactory.CreateBoolOption(
+                            "useHistoricalCpuUsageInfo",
+                            sign => schedulingConfiguration.UseHistoricalCpuUsageInfo = sign),
+                        OptionHandlerFactory.CreateBoolOption(
                             "useHistoricalRamUsageInfo",
                             sign => schedulingConfiguration.UseHistoricalRamUsageInfo = sign),
                         OptionHandlerFactory.CreateOption(
--- a/Public/Src/Engine/Scheduler/ProcessRunnablePip.cs
+++ b/Public/Src/Engine/Scheduler/ProcessRunnablePip.cs
@@ -21,6 +21,16 @@ namespace BuildXL.Scheduler
         /// <nodoc/>
         public Process Process => (Process)Pip;
 
+       
+        /// <summary>
+        /// Process weight
+        /// </summary>
+        /// <remarks>
+        /// If process weight is defined as greater than 1 in the specs, use it. 
+        /// Otherwise, use the weight based on historic cpu usage.
+        /// </remarks>
+        public int Weight => Process.Weight > 1 ? Process.Weight : m_weightBasedOnHistoricCpuUsage;
+
         /// <nodoc/>
         public RunnableFromCacheResult CacheResult { get; private set; }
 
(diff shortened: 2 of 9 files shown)


Public/Src/Engine/Scheduler/ProcessRunnablePip.cs
```diff
@@ -47,15 +57,27 @@ public sealed class ProcessRunnablePip : RunnablePip
...
             int priority,
             Func<RunnablePip, Task> executionFunc,
             IPipExecutionEnvironment environment,
+            ushort cpuUsageInPercents = 0,
             Pip pip = null)
             : base(phaseLoggingContext, pipId, PipType.Process, priority, executionFunc, environment, pip)
         {
+            if (cpuUsageInPercents > 100)
+            {
+                m_weightBasedOnHistoricCpuUsage = (int)Math.Ceiling(cpuUsageInPercents / 100.0);
```
dannyvv: >(int)Math.Ceiling(cpuUsageInPercents / 100.0); [](start = 50, length = 46)

Isn't this loosing all resolution? It is now either 0 or 1...

dannyvv: ah, understood. Perhaps a copy of the comment on `ProcessorsInPercents` field would be usefull to remind future readers

---
In reply to: [277112078](https://github.com/Microsoft/BuildXL/pull/178#discussion_r277112078) [](ancestors = 277112078)

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2713,7 +2713,18 @@ private Task SchedulePip(PipId pipId, PipType pipType, RunnablePipObserver obser
             // Offload the execution of the pip to one of the queues in the PipQueue.
             // If it is a meta or SealDirectory pip and the PipQueue has started draining, then the execution will be inlined here!
             // Because it is not worth to enqueue the fast operations such as the execution of meta and SealDirectory pips.
-            var runnablePip = RunnablePip.Create(m_executePhaseLoggingContext, this, pipId, pipType, priority ?? GetPipPriority(pipId), m_executePipFunc);
+
+            ushort cpuUsageInPercent = m_scheduleConfiguration.UseHistoricalCpuUsageInfo ? RunningTimeTable[m_pipTable.GetPipSemiStableHash(pipId)].ProcessorsInPercents : (ushort)0;
```
mpysson: >ProcessorsInPercents [](start = 148, length = 20)

Amazing we had this already calculated but never actually utilized it...

Public/Src/App/Bxl/Args.cs
```diff
@@ -1018,6 +1018,9 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                         OptionHandlerFactory.CreateBoolOption(
                             "useHardlinks",
                             sign => engineConfiguration.UseHardlinks = sign),
+                        OptionHandlerFactory.CreateBoolOption(
+                            "useHistoricalCpuUsageInfo",
```
mpysson: >useHistoricalCpuUsageInfo [](start = 29, length = 25)

Now that this is dynamic, it might be good to include it in the execution log for ease of postmortem datamining. Food for thought, no need to do that in this PR

Public/Src/Engine/Scheduler/Scheduler.cs
semihokur: exactly. I started doing the changes in running time table, and then I noticed that we already have what we need there.

Public/Src/Engine/Scheduler/ProcessRunnablePip.cs
semihokur: i do not think that we need that resolution at this moment. There are several process pips launching thousands of child processes and using around 600% cpu. The purpose of the changes is to make those pips acquire 6 slots. If there is one using 520% cpu, we do not currently mind that it uses 6 slots.

Public/Src/Engine/Scheduler/ProcessRunnablePip.cs
```diff
@@ -21,6 +21,16 @@ public sealed class ProcessRunnablePip : RunnablePip
...
 
+       
+        /// <summary>
+        /// Process weight
+        /// </summary>
+        /// <remarks>
+        /// If process weight is defined as greater than 1 in the specs, use it. 
+        /// Otherwise, use the weight based on historic cpu usage.
+        /// </remarks>
+        public int Weight => Process.Weight > 1 ? Process.Weight : m_weightBasedOnHistoricCpuUsage;
```
anniefu: > 1 [](start = 45, length = 2)

Can you put this in a "MIN_PROCESS_WEIGHT" or similar const somewhere that is also shared with the definition of Weight in Process.cs?

anniefu: Not going to block PR on this, but we should definitely do it :) 

---
In reply to: [277380578](https://github.com/Microsoft/BuildXL/pull/178#discussion_r277380578) [](ancestors = 277380578)

PR #179, opened 2019-04-19
Expose weight through to QTest SDK

mpysson:
Exposing the process weight configuration through the QTest SDK and also through to the Managed SDK. Only enabled for QTest pips.

### DIFF
--- a/Public/Sdk/Public/Managed/testing.dsc
+++ b/Public/Sdk/Public/Managed/testing.dsc
@@ -11,7 +11,7 @@ import * as Xml from "Sdk.Xml";
  * uses the test frameworks runtest function to execute the test.
  */
 @@public
-export function test(args: TestArguments) : TestResult {
+export function test(args: TestArguments) : TestResult {   
     let testFramework = args.testFramework;
     if (!testFramework)
     {
@@ -266,6 +266,9 @@ export interface TestRunArguments {
      * Allows test runs to be tagged.
      */
     tags?: string[];
+
+    /** Optionally override to increase the weight of test pips that require more machine resources */
+    weight?: number;
 }
 
 @@public
--- a/Public/Sdk/Public/Managed/testing.dsc
+++ b/Public/Sdk/Public/Managed/testing.dsc
@@ -11,7 +11,7 @@ import * as Xml from "Sdk.Xml";
  * uses the test frameworks runtest function to execute the test.
  */
 @@public
-export function test(args: TestArguments) : TestResult {   
+export function test(args: TestArguments) : TestResult {
     let testFramework = args.testFramework;
     if (!testFramework)
     {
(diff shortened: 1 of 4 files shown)


Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -258,6 +258,8 @@ export interface QTestArguments extends Transformer.RunnerArguments {
     qTestAdditionalOptions?: string;
     /** Path to runsettings file that will be passed on to vstest.console.exe. */
     vstestSettingsFile?: File;
+    /** Optionally override to increase the weight of test pips that require more machine resources */
+    weight?: number;
```
dannyvv: If we expect this to be truely common, perhaps it is worth adding it to Transformer.RunnerArguments?

mpysson: I was debating that. My concern was that it would be misleading if it were available to specify on everything that derives from RunnerArguments but only a small number of things implement it. I figured we could always promote if as we make it more ubiquitous in our SDK. What do you think?

Public/Sdk/Public/Managed/testing.dsc
```diff
@@ -11,7 +11,7 @@ import * as Xml from "Sdk.Xml";
  * uses the test frameworks runtest function to execute the test.
  */
 @@public
-export function test(args: TestArguments) : TestResult {
+export function test(args: TestArguments) : TestResult {
```
almiliMSFT: nit: trailing spaces
```suggestion
export function test(args: TestArguments) : TestResult {
```

Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -164,7 +164,7 @@ export function runQTest(args: QTestArguments): Result {
         consoleOutput: consolePath,
         workingDirectory: sandboxDir,
         tempDirectory: tempDirectory,
-        // TODO - hook this up for Heisenbug queues (runs tests a bunch under qtest process)
+        weight : args.weight,
```
almiliMSFT:
```suggestion
weight: args.weight,
```

Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
norahuang: Can we foresee any other usage of the "weight" feature in the near future? If not, I think it would be fine to keep it only in the QTestArguments. We can move it to Transformer.RunnerArguments later when the second consumer other than qtest pip emerge.

PR #180, opened 2019-04-19
Downgrade detours heap size logging to diagnostic

mpysson:
This log message was added over a year ago to debug a detours bug. The bug's long been fixed and the message is pretty chatty. Downgrade to to a diagnostic event.


dannyvv:
/azp run  
Part B BuildXL PR Validation

PR #184, opened 2019-04-22
Don't preemptively set u+x permissions when executing a process

almiliMSFT:
[AB#1488150](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1488150)

[AB#1516419](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1516419)

### DIFF
--- a/Public/Src/Engine/Processes/UnSandboxedProcess.cs
+++ b/Public/Src/Engine/Processes/UnSandboxedProcess.cs
@@ -306,13 +306,6 @@ namespace BuildXL.Processes
                 ThrowCouldNotStartProcess(I($"File '{ProcessInfo.FileName}' not found"), new Win32Exception(0x2));
             }
 
-#if PLATFORM_OSX
-            // TODO: TASK 1488150
-            // When targeting macOS, we make sure the 'execute bit' has been set on the binary about to be started,
-            // especially running on VSTS VMs currently has issues around file permission preservance
-            SetFilePermissionsForFilePath(ProcessInfo.FileName, FilePermissions.S_IXUSR);
-#endif
-
             Process = new Process();
             Process.StartInfo = new ProcessStartInfo
             {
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -301,16 +301,19 @@ protected Process CreateAndSetUpProcess()
         {
             Contract.Requires(Process == null);
 
-            if (!FileUtilities.FileExistsNoFollow(ProcessInfo.FileName))
+            var mode = GetFilePermissionsForFilePath(ProcessInfo.FileName, followSymlink: false);
```
mpysson: >GetFilePermissionsForFilePath [](start = 23, length = 29)

Does this need to be within the #if PLATFORM_OSX?

Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -301,16 +301,19 @@ protected Process CreateAndSetUpProcess()
         {
             Contract.Requires(Process == null);
 
-            if (!FileUtilities.FileExistsNoFollow(ProcessInfo.FileName))
```
mpysson: >if (!FileUtilities.FileExistsNoFollow(ProcessInfo.FileName)) [](start = 12, length = 60)

I don't think the new behavior is equivalent in the windows case

Public/Src/Engine/Processes/UnSandboxedProcess.cs
almiliMSFT: You are right about both, will fix it...

Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -301,16 +301,24 @@ protected Process CreateAndSetUpProcess()
         {
             Contract.Requires(Process == null);
 
+#if !PLATFORM_OSX
```
dannyvv: W
Shouldn't we keep the exe does not exist check for windows?

almiliMSFT: It is kept (here it's `! PLATFORM_OSX` so this executes on Windows)

PR #185, opened 2019-04-22
Add factory for GRPC P2P cache

kewilli:
New factory for GRPC peer-to-peer cache. Here's a rough sketch of the types:

MemoizationStoreCacheFactory : ICacheFactory
    - MemoizationStoreAdapterCache : ICache
        - LocalCache : OneLevelCache
            - DualServerClientContentStore : IContentStore
                - LocalContentServer
                    - DistributedContentStore : IContentStore
                        - FileSystemContentStore : IContentStore
                - ServiceClientContentStore : IContentStore
            - SqlLiteMemoizationStore

### DIFF
--- a/Public/Src/Cache/ContentStore/App/Hello.cs
+++ b/Public/Src/Cache/ContentStore/App/Hello.cs
@@ -32,18 +32,25 @@ namespace BuildXL.Cache.ContentStore.App
             [Description("GRPC port on the target machine"), DefaultValue(ServiceConfiguration.GrpcDisabledPort)] int grpcPort,
             [Description("Name of the memory mapped file used to share GRPC port. 'CASaaS GRPC port' if not specified.")] string grpcPortFileName)
         {
-            Initialize();
-
-            if (grpcPort == 0)
+            try
             {
-                grpcPort = Helpers.GetGrpcPortFromFile(_logger, grpcPortFileName);
+                Initialize();
+
+                if (grpcPort == 0)
+                {
+                    grpcPort = Helpers.GetGrpcPortFromFile(_logger, grpcPortFileName);
+                }
+
+                var _channel = new Channel(host, grpcPort, ChannelCredentials.Insecure);
+                var _client = new ContentServer.ContentServerClient(_channel);
+                var helloResponse = _client.Hello(new HelloRequest(), new CallOptions(deadline: DateTime.UtcNow + TimeSpan.FromSeconds(2)));
+
+                _logger.Always("Hello response {0}: {1}", helloResponse.Success ? "succeeded" : "failed", helloResponse.ToString());
             }
-
-            var _channel = new Channel(host, grpcPort, ChannelCredentials.Insecure);
-            var _client = new ContentServer.ContentServerClient(_channel);
-            var helloResponse = _client.Hello(new HelloRequest(), new CallOptions(deadline: DateTime.UtcNow + TimeSpan.FromSeconds(2)));
-
-            _logger.Always($"Hello response {(helloResponse.Success ? "succeeded" : "failed")}: {helloResponse.ToString()}");
+            catch (Exception e)
+            {
+                _logger.Always(e.ToString());
+            }
         }
     }
 }
--- a/Public/Src/Cache/ContentStore/Interfaces/FileSystem/AbsolutePath.cs
+++ b/Public/Src/Cache/ContentStore/Interfaces/FileSystem/AbsolutePath.cs
@@ -37,6 +37,26 @@ namespace BuildXL.Cache.ContentStore.Interfaces.FileSystem
         public static int MinimumPathLength = OperatingSystemHelper.IsWindowsOS ? 2 : 1;
 
         /// <summary>
+        /// Create an AbsolutePath for the root of a drive.
+        /// </summary>
+        public static AbsolutePath RootPath(char driveLetter = default(char))
+        {
+            if (OperatingSystemHelper.IsWindowsOS)
+            {
+                if (driveLetter == default(char))
+                {
+                    throw new ArgumentException(nameof(driveLetter));
+                }
+
+                return new AbsolutePath($"{driveLetter}:{System.IO.Path.DirectorySeparatorChar}{System.IO.Path.DirectorySeparatorChar}");
+            }
+            else
+            {
+                return new AbsolutePath(System.IO.Path.DirectorySeparatorChar.ToString());
+            }
+        }
+
+        /// <summary>
         ///     Initializes a new instance of the <see cref="AbsolutePath" /> class.
         /// </summary>
         public AbsolutePath(string path)
--- a/Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
+++ b/Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
@@ -1,0 +1,172 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.ContractsLight;
+using System.Threading;
+using System.Threading.Tasks;
+using BuildXL.Cache.ContentStore.Distributed.Redis.Credentials;
+using BuildXL.Cache.ContentStore.Interfaces.Distributed;
+using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
+using BuildXL.Cache.ContentStore.Interfaces.Logging;
+using BuildXL.Cache.ContentStore.Interfaces.Results;
+using BuildXL.Cache.ContentStore.Interfaces.Sessions;
+using BuildXL.Cache.ContentStore.Interfaces.Stores;
+using BuildXL.Cache.ContentStore.Interfaces.Tracing;
+using BuildXL.Cache.ContentStore.Service;
+using BuildXL.Cache.ContentStore.Sessions;
+using BuildXL.Cache.ContentStore.Stores;
+using BuildXL.Cache.Host.Configuration;
+
+namespace BuildXL.Cache.Host.Service.Internal
+{
+    /// <summary>
+    /// Content store containing both a distributed <see cref="LocalContentStore"/> and <see cref="ServiceClientContentStore"/>.
+    /// </summary>
+    public class DualServerClientContentStore : IContentStore
+    {
+        private LocalContentServer _localContentServer;
+        private ServiceClientContentStore _serviceClientContentStore;
+
+        /// <inheritdoc />
+        public bool StartupCompleted => (_localContentServer?.StartupCompleted ?? false) && (_serviceClientContentStore?.StartupCompleted ?? false);
+
+        /// <inheritdoc />
+        public bool StartupStarted => (_localContentServer?.StartupStarted ?? false) && (_serviceClientContentStore?.StartupStarted ?? false);
+
+        /// <inheritdoc />
+        public bool ShutdownCompleted => (_localContentServer?.ShutdownCompleted ?? false) && (_serviceClientContentStore?.ShutdownCompleted ?? false);
+
+        /// <inheritdoc />
+        public bool ShutdownStarted => (_localContentServer?.ShutdownStarted ?? false) && (_serviceClientContentStore?.ShutdownStarted ?? false);
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="fileSystem">File system object used by the client.</param>
+        /// <param name="logger">Logger.</param>
+        /// <param name="copier">File copier used by the content server.</param>
+        /// <param name="pathTransformer">File path transformer used by the content server.</param>
+        /// <param name="maxQuotaMB">Maximum size (in MB) of the cache.</param>
+        /// <param name="cacheName">Name of the cache.</param>
+        /// <param name="grpcPort">GRPC port which the server listens to and the client transmits to.</param>
+        /// <param name="stampId">Redis stamp identifier.</param>
+        /// <param name="ringId">Redis ring identifier.</param>
+        /// <param name="scenarioName">Unique name for the CAS. Allows multiple CAS services to coexist.</param>
+        /// <param name="dataRootPath">Path to the root of CAS content.</param>
+        public DualServerClientContentStore(
+            IAbsFileSystem fileSystem,
+            ILogger logger,
+            IAbsolutePathFileCopier copier,
+            IAbsolutePathTransformer pathTransformer,
+            int maxQuotaMB,
+            string cacheName,
+            int grpcPort,
+            string stampId,
+            string ringId,
+            string scenarioName,
+            AbsolutePath dataRootPath)
+        {
+            Contract.Assert(grpcPort >= 0);
+            Contract.Assert(maxQuotaMB >= 0);
+            Contract.Assert(!string.IsNullOrWhiteSpace(cacheName));
+            Contract.Assert(!string.IsNullOrWhiteSpace(ringId));
+            Contract.Assert(!string.IsNullOrWhiteSpace(stampId));
+
+            var redisConnectionString = Environment.GetEnvironmentVariable(EnvironmentConnectionStringProvider.RedisConnectionStringEnvironmentVariable);
+
+            var localCasSettings = LocalCasSettings.Default(maxQuotaMB, dataRootPath.Path, cacheName, (uint)grpcPort);
+            localCasSettings.ServiceSettings.ScenarioName = scenarioName;
+
+            if (dataRootPath.IsLocal)
+            {
+                localCasSettings.DrivePreferenceOrder.Add(AbsolutePath.RootPath(dataRootPath.DriveLetter).Path);
+            }
+
+            var contentServerFactory = new ContentServerFactory(new DistributedCacheServiceArguments(
+                logger,
+                copier,
+                pathTransformer,
+                new TestDistributedCacheServiceHost(),
+                new HostInfo(stampId, ringId, new List<string>()),
+                default(CancellationToken),
+                dataRootPath.Path,
+                new DistributedCacheServiceConfiguration(
+                    localCasSettings,
+                    DistributedContentSettings.CreateEnabled(new Dictionary<string, string>() { { stampId, redisConnectionString } }, true)),
+                string.Empty));
+            _localContentServer = contentServerFactory.Create();
+
+            _serviceClientContentStore = new ServiceClientContentStore(
+                logger,
+                fileSystem,
+                new ServiceClientContentStoreConfiguration(cacheName, new ServiceClientRpcConfiguration(grpcPort), scenarioName));
+        }
+
+        /// <inheritdoc />
+        public CreateSessionResult<IReadOnlyContentSession> CreateReadOnlySession(Context context, string name, ImplicitPin implicitPin)
+        {
+            return _serviceClientContentStore.CreateReadOnlySession(context, name, implicitPin);
+        }
+
+        /// <inheritdoc />
+        public CreateSessionResult<IContentSession> CreateSession(Context context, string name, ImplicitPin implicitPin)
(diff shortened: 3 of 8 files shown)


Public/Src/Cache/ContentStore/App/Hello.cs
```diff
@@ -32,18 +32,25 @@ internal sealed partial class Application
...
+                var _channel = new Channel(host, grpcPort, ChannelCredentials.Insecure);
+                var _client = new ContentServer.ContentServerClient(_channel);
+                var helloResponse = _client.Hello(new HelloRequest(), new CallOptions(deadline: DateTime.UtcNow + TimeSpan.FromSeconds(2)));
 
-            _logger.Always($"Hello response {(helloResponse.Success ? "succeeded" : "failed")}: {helloResponse.ToString()}");
+                _logger.Always("Hello response {0}: {1}", helloResponse.Success ? "succeeded" : "failed", helloResponse.ToString());
+            }
+            catch (Exception e)
+            {
+                _logger.Always(e.ToString());
```
dannyvv: shouldn't this throw to ensure a failure is escalated?

Public/Src/Cache/ContentStore/Interfaces/FileSystem/AbsolutePath.cs
```diff
@@ -36,6 +36,26 @@ public class AbsolutePath : PathBase, IEquatable<AbsolutePath>
...
+        public static AbsolutePath RootPath(char driveLetter = default(char))
+        {
+            if (OperatingSystemHelper.IsWindowsOS)
+            {
+                if (driveLetter == default(char))
+                {
+                    throw new ArgumentException(nameof(driveLetter));
+                }
+
+                return new AbsolutePath($"{driveLetter}:{System.IO.Path.DirectorySeparatorChar}{System.IO.Path.DirectorySeparatorChar}");
```
dannyvv: >Char}{System.IO.Path.DirectorySeparatorChar} [](start = 90, length = 44)

Why double path separator like:
`c:\\`

Public/Src/Cache/MemoizationStore/Library/Sessions/LocalCache.cs
```diff
@@ -109,15 +124,34 @@ public class LocalCache : OneLevelCache
...
-            bool checkLocalFiles,
-            bool emptyFileHashShortcutEnabled)
+            ContentStoreSettings contentStoreSettings,
+            bool createServer,
+            int grpcPort = 0,
+            int maxQuotaMB = 0,
+            string cacheName = null,
+            string stampId = null,
+            string ringId = null,
+            string scenarioName = null)
```
dannyvv: Should we add a LocalCacheSettings class or struct to handle all these default values and `base(` and `this(` calls and make it a bit more maintainable for new values going forwrad

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+            string scenarioName,
+            AbsolutePath dataRootPath)
+        {
+            Contract.Assert(grpcPort >= 0);
+            Contract.Assert(maxQuotaMB >= 0);
+            Contract.Assert(!string.IsNullOrWhiteSpace(cacheName));
+            Contract.Assert(!string.IsNullOrWhiteSpace(ringId));
+            Contract.Assert(!string.IsNullOrWhiteSpace(stampId));
+
+            var redisConnectionString = Environment.GetEnvironmentVariable(EnvironmentConnectionStringProvider.RedisConnectionStringEnvironmentVariable);
```
dannyvv: >redisConnectionString [](start = 16, length = 21)

Do we need some validation for this environment variable, or is it totally optional? Perhaps some logging that we picked it up from an environment variable?

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+        public Task<GetStatsResult> GetStatsAsync(Context context)
+        {
+            return _serviceClientContentStore.GetStatsAsync(context);
+        }
+
+        /// <inheritdoc />
+        public async Task<BoolResult> ShutdownAsync(Context context)
+        {
+            await _serviceClientContentStore.ShutdownAsync(context).ThrowIfFailure();
+            return await _localContentServer.ShutdownAsync(context).ThrowIfFailure();
```
dannyvv: Would it be beneificial for when we have two async calls in this class to do them in parralel rather than sequetial?

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        /// <inheritdoc />
+        public async Task<BoolResult> StartupAsync(Context context)
```
dannyvv: >StartupAsync [](start = 38, length = 12)

Same as `ShutdownAsync`

Public/Src/Cache/ContentStore/App/Hello.cs
SergeyTeplyakov: I can't remember the semantics of this method: will it be visible to the user in interactive mode?

Public/Src/Cache/ContentStore/Interfaces/FileSystem/AbsolutePath.cs
```diff
@@ -36,6 +36,26 @@ public class AbsolutePath : PathBase, IEquatable<AbsolutePath>
...
+            }
+            else
+            {
+                return new AbsolutePath(System.IO.Path.DirectorySeparatorChar.ToString());
```
SergeyTeplyakov: nit: I would extract this into a static property like DirectorySeparator for simplicities sake.

Kr1sso: Please use `VolumeSeparatorChar` to indicate the root volume only.

Public/Src/Cache/ContentStore/Interfaces/FileSystem/AbsolutePath.cs
```diff
@@ -36,6 +36,26 @@ public class AbsolutePath : PathBase, IEquatable<AbsolutePath>
...
+        /// <summary>
+        /// Create an AbsolutePath for the root of a drive.
+        /// </summary>
+        public static AbsolutePath RootPath(char driveLetter = default(char))
+        {
+            if (OperatingSystemHelper.IsWindowsOS)
+            {
+                if (driveLetter == default(char))
+                {
+                    throw new ArgumentException(nameof(driveLetter));
```
SergeyTeplyakov: Should it be more descriptive error message in this case?

And I'm not sure about using default arg here... Maybe two overloads are better when one of them is not supported by one platform and another is not supported by another.

For instance, right now for unix the argument is just ignored, but I would say it should be similar exception.

SergeyTeplyakov: I would vote for a more descriptive error message here.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <inheritdoc />
+        public bool ShutdownCompleted => (_localContentServer?.ShutdownCompleted ?? false) && (_serviceClientContentStore?.ShutdownCompleted ?? false);
+
+        /// <inheritdoc />
+        public bool ShutdownStarted => (_localContentServer?.ShutdownStarted ?? false) && (_serviceClientContentStore?.ShutdownStarted ?? false);
+
+        /// <summary>
+        /// 
+        /// </summary>
```
SergeyTeplyakov: nodoc is fine here, I think.
Or fill the summary.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <param name="stampId">Redis stamp identifier.</param>
+        /// <param name="ringId">Redis ring identifier.</param>
+        /// <param name="scenarioName">Unique name for the CAS. Allows multiple CAS services to coexist.</param>
+        /// <param name="dataRootPath">Path to the root of CAS content.</param>
+        public DualServerClientContentStore(
+            IAbsFileSystem fileSystem,
+            ILogger logger,
+            IAbsolutePathFileCopier copier,
+            IAbsolutePathTransformer pathTransformer,
+            int maxQuotaMB,
```
SergeyTeplyakov: Voting for configuration class.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
SergeyTeplyakov: As I suggested already, I think it make sense to create a configuration type and put this value directly there instead of reading environment variables in the constructor.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
SergeyTeplyakov: >_localContentServer [](start = 25, length = 19)

It seems that both fields should be readonly.

SergeyTeplyakov: >_localContentServer [](start = 25, length = 19)

I can't understand the nullability semantics here. I can see that some members are using null-safe access but some others not.

And it seems that these fields should not be null and all ?. accesses are redundant and confusing...

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+using BuildXL.Cache.ContentStore.Interfaces.Tracing;
+using BuildXL.Cache.ContentStore.Service;
+using BuildXL.Cache.ContentStore.Sessions;
+using BuildXL.Cache.ContentStore.Stores;
+using BuildXL.Cache.Host.Configuration;
+
+namespace BuildXL.Cache.Host.Service.Internal
+{
+    /// <summary>
+    /// Content store containing both a distributed <see cref="LocalContentStore"/> and <see cref="ServiceClientContentStore"/>.
```
SergeyTeplyakov: The comment tells about the inner structure of the class but doesn't explain why we need this inner structure.

Could you please explain why we have to combine service client and localcontentserver in one place? I think it should be valuable.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+                logger,
+                copier,
+                pathTransformer,
+                new TestDistributedCacheServiceHost(),
+                new HostInfo(stampId, ringId, new List<string>()),
+                default(CancellationToken),
+                dataRootPath.Path,
+                new DistributedCacheServiceConfiguration(
+                    localCasSettings,
+                    DistributedContentSettings.CreateEnabled(new Dictionary<string, string>() { { stampId, redisConnectionString } }, true)),
```
SergeyTeplyakov: Please use named argument here.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+        public void Dispose()
+        {
+            _serviceClientContentStore?.Dispose();
+            _localContentServer?.Dispose();
+        }
+
+        /// <inheritdoc />
+        public Task<GetStatsResult> GetStatsAsync(Context context)
+        {
+            return _serviceClientContentStore.GetStatsAsync(context);
```
SergeyTeplyakov: Should the stats be combined from two fields?

Public/Src/Cache/MemoizationStore/Library/Sessions/LocalCache.cs
```diff
@@ -35,16 +37,29 @@ public class LocalCache : OneLevelCache
             ConfigurationModel configurationModel = null,
             IClock clock = null,
             bool checkLocalFiles = true,
-            bool emptyFileHashShortcutEnabled = false)
+            bool emptyFileHashShortcutEnabled = false,
+            bool createServer = false,
```
SergeyTeplyakov: Need a settings class IMO. Or if you'll add one that I've suggested already, maybe we can just pass it here.

Public/Src/Cache/MemoizationStore/Library/Sessions/LocalCache.cs
SergeyTeplyakov: +100 500!!!!!!

---
In reply to: [277486273](https://github.com/Microsoft/BuildXL/pull/185#discussion_r277486273) [](ancestors = 277486273)

Public/Src/Cache/ContentStore/App/Hello.cs
SergeyTeplyakov: I think we should definitely change the output message here to something: "Unhandled error occurred in Hello" something... I've seen some cases in Kusto when it is not clear what was the error. And I think it make sense to print only message to the console and log the entire exception.

---
In reply to: [277806629](https://github.com/Microsoft/BuildXL/pull/185#discussion_r277806629) [](ancestors = 277806629)

kewilli: This should never be run in any official capacity. This is a toy verb to help debug an existing issue.

---
In reply to: [277485681](https://github.com/Microsoft/BuildXL/pull/185#discussion_r277485681) [](ancestors = 277485681)

kewilli: This error will never surface in Kusto, because it will never be run in any official capacity. This is a toy verb like HardlinkBenchmark and CopyFile.

---
In reply to: [277811685](https://github.com/Microsoft/BuildXL/pull/185#discussion_r277811685) [](ancestors = 277811685,277806629)

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
kewilli: Whoops! Missed this one.

---
In reply to: [277808204](https://github.com/Microsoft/BuildXL/pull/185#discussion_r277808204) [](ancestors = 277808204)

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
kewilli: Lance told me that the order here was important, so I'm trusting that they should not be done in parallel.

---
In reply to: [277486705](https://github.com/Microsoft/BuildXL/pull/185#discussion_r277486705) [](ancestors = 277486705)

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
kewilli: But since we're moving to use configurations, documentation would be wasted here.

---
In reply to: [277815496](https://github.com/Microsoft/BuildXL/pull/185#discussion_r277815496) [](ancestors = 277815496,277808204)

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
dannyvv: K. Sounds good

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+using BuildXL.Cache.ContentStore.Sessions;
+using BuildXL.Cache.ContentStore.Stores;
+using BuildXL.Cache.Host.Configuration;
+
+namespace BuildXL.Cache.Host.Service.Internal
+{
+    /// <summary>
+    /// Content store containing both a distributed <see cref="LocalContentStore"/> and <see cref="ServiceClientContentStore"/>.
+    /// </summary>
+    public class DualServerClientContentStore : IContentStore
```
ljcollins25: >IContentStore [](start = 48, length = 13)

Derive from StartupShutdownBase

Public/Src/Cache/VerticalStore/MemoizationStoreAdapter/MemoizationStoreCacheFactory.cs
```diff
@@ -121,6 +128,57 @@ public sealed class Config : CasConfig
...
+            /// identifies a particular CAS instance.
+            /// </summary>
+            [DefaultValue(null)]
+            public string ScenarioName { get; set; }
+
+            /// <summary>
+            /// The stamp identifier for this service instance.
+            /// </summary>
+            [DefaultValue(null)]
+            public string StampId { get; set; }
```
ljcollins25: >StampId [](start = 26, length = 7)

Missing this above in the config description comment.

Public/Src/Cache/VerticalStore/MemoizationStoreAdapter/MemoizationStoreCacheFactory.cs
```diff
@@ -121,6 +128,57 @@ public sealed class Config : CasConfig
...
+
+            /// <summary>
+            /// The ring identifier for this service instance.
+            /// </summary>
+            [DefaultValue(null)]
+            public string RingId { get; set; }
```
ljcollins25: >RingId [](start = 26, length = 6)

Missing this above in the config description comment.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
ljcollins25: >DistributedContentSettings.CreateEnabled(new Dictionary<string, string>() { { stampId, redisConnectionString } }, true)) [](start = 20, length = 120)

I was thinking a path to a settings file should be specified as a part of the config data (or embedded therein). Are you doing this to simply things for initial testing, or is this supposed to be the final state?

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        /// <inheritdoc />
+        public async Task<BoolResult> StartupAsync(Context context)
+        {
+            await _localContentServer.StartupAsync(context).ThrowIfFailure();
+            return await _serviceClientContentStore.StartupAsync(context).ThrowIfFailure();
+        }
+
+        private sealed class TestDistributedCacheServiceHost : IDistributedCacheServiceHost
```
kewilli: >TestDistributedCacheServiceHost [](start = 29, length = 31)

Don't call it test

Public/Src/Cache/DistributedCache.Host/Service/Internal/DualServerClientContentStore.cs
```diff
@@ -0,0 +1,172 @@
...
+                return key;
+            }
+
+            public void OnStartedService()
+            {
+            }
+
+            public Task OnStartingServiceAsync()
+            {
+                return Task.Run(() => { });
```
kewilli: >Run [](start = 28, length = 3)

CompletedTask

kewilli:
Restarting this PR to expect a persistent CAS instance it can connect to.

PR #190, opened 2019-04-23
Cleanup: Usings and header in tools folder

dannyvv:


Public/Src/Tools/Execution.Analyzer/Tracing/Log.cs
```diff
@@ -5,9 +5,8 @@
 using BuildXL.Tracing;
 using BuildXL.Utilities.Instrumentation.Common;
 using BuildXL.Utilities.Tracing;
-
 #if !FEATURE_MICROSOFT_DIAGNOSTICS_TRACING
```
Kr1sso: Left an empty if/def.

dannyvv: Will look for this in a future PR, there might be more

PR #193, opened 2019-04-23
Make StoreOutputToCache- compatible with rewritten targets of copy and write file

narasamdya:
If the targets of the copy/write file pips are rewritten, then we need to store them to cache, although /storeoutputtocache- is used.

[AB#1511558](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1511558)

### DIFF
--- a/Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/StoreNoOutputsToCacheTests.cs
+++ b/Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/StoreNoOutputsToCacheTests.cs
@@ -181,6 +181,64 @@ namespace IntegrationTest.BuildXL.Scheduler
         }
 
         [Fact]
+        public void RewriteCopyFileTarget()
+        {
+            const string Content = "Test";
+            var copySource = CreateSourceFile();
+            var copyTarget = CreateOutputFileArtifact();
+
+            File.WriteAllText(ArtifactToString(copySource), Content);
+            var copy = CreateAndScheduleCopyFile(copySource, copyTarget);
+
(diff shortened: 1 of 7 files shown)


Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/StoreNoOutputsToCacheTests.cs
```diff
@@ -180,6 +180,64 @@ public void RewriteProcessStoreOutputsToCacheTest()
...
+
+            File.WriteAllText(ArtifactToString(copySource), Content);
+            var copy = CreateAndScheduleCopyFile(copySource, copyTarget);
+
+            var rewrittenCopyTarget = copyTarget.CreateNextWrittenVersion();
+
+            var processAndOutputs = CreateAndSchedulePipBuilder(new Operation[]
+            {
+                Operation.ReadFile(copyTarget),
+                Operation.WriteFile(rewrittenCopyTarget, Content)
```
ljcollins25: >Content [](start = 57, length = 7)

Change the content of the file here (similarly for the rewriting process of write file below). Otherwise, your test could work if the process pip stored its outputs.

narasamdya: When you specify the content in `Operation.WriteFile`, the behavior is append. So I don't need to change anything here.

---
In reply to: [277890624](https://github.com/Microsoft/BuildXL/pull/193#discussion_r277890624) [](ancestors = 277890624)

PR #198, opened 2019-04-24
Fix nuget spec gen to ensure roslyn 3.0 package is not treated like a…

dannyvv:
… managed package


almiliMSFT:
Update the title to avoid `...`

PR #202, opened 2019-04-25
Fix pip materialization issues on macOS

Kr1sso:
This PR is the first attempt to fix pip materialization issues we are seeing on macOS builds. After a detailed analysis it seems that inside of `LocalDiskContentStore.TryMaterializeAsync(...)` the materialization step sometimes fails after the cache has successfully placed a file on disk. Thus, we add more telemetry and a retry logic around `TryOpenAndTrackPathAsync(...)` as the issue seems to originate from there.

[AB#1514088](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1514088)


Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
```diff
@@ -207,7 +207,6 @@ public sealed class LocalDiskContentStore : ILocalDiskFileSystemView
                         : CreateSymlinkIfNotExistsOrTargetMismatch(expandedPath.Path, reparsePointInfo.Value.GetReparsePointTarget());
                 }
 
-                // TODO:58494: We shouldn't need to open a new handle here. In fact, that is a race, and why there isn't a FileContentTable.RecordContentHashAsync taking a path.
```
almiliMSFT: We still open a new handle here, right?  So this comment still seems relevant

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
```diff
@@ -926,50 +925,58 @@ private void TryFlushPageCacheToFileSystem(string path)
...
-                        // in that case we still need to call TrackChangesToFile to ensure the tracker latches to disabled.
-                        return TrackChangesToFile(handle, path, identityAndContentInfo.Value, knownFileName: fileName);
-                    });
-            }
-            catch (BuildXLException ex)
-            {
-                return new RecoverableExceptionFailure(ex);
-            }
+                        Tracing.Logger.Log.RetryTryOpenAndTrackPathAsync(m_loggingContext, path.ExpandedPath);
+                        return new Possible<TrackedFileContentInfo, Failure>(new Failure<string>($"TryOpenAndTrackPathAsync() failed to establish identity and track file: {path.ExpandedPath}"));
```
almiliMSFT: nit: instead of `new Possible(new Failure())` you can just return `new Failure<string>("...")`

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
Kr1sso: The comment is present inside the actual call inside `TryOpenAndTrackPathAsync`

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
```diff
@@ -926,50 +925,58 @@ private void TryFlushPageCacheToFileSystem(string path)
...
-                    path.ExpandedPath,
-                    FileDesiredAccess.GenericRead,
-                    FileShare.Read | FileShare.Delete,
-                    FileMode.Open,
-                    FileFlagsAndAttributes.FileFlagOverlapped | FileFlagsAndAttributes.FileFlagOpenReparsePoint,
-                    (handle, length) =>
+
+            // TODO:58494: We shouldn't need to open a new handle here. In fact, that is a race, and why there isn't a RecordContentHashAsync taking a path.
+            // Somebody could change the file since we finished copying from the cache. The cache should instead be enlightened to use the file content table.
+            return Helpers.RetryOnFailure(
```
almiliMSFT: @narasamdya : can you confirm that `TryOpenAndTrackPathAsync` gets called only when we believe that the file exists on disk?

Public/Src/Engine/Cache/Log.cs
```diff
@@ -129,5 +129,14 @@ internal abstract partial class Logger
             EventTask = (ushort)Events.Tasks.Storage,
             Message = "Retry on loading and deserializing metadata: Succeeded: {succeeded} | Retry count: {retryCount}")]
         internal abstract void RetryOnLoadingAndDeserializingMetadata(LoggingContext loggingContext, bool succeeded, int retryCount);
+
+        [GeneratedEvent(
+            (int)EventId.RetryTryOpenAndTrackPathAsync,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Warning,
```
narasamdya: >Level.Warning, [](start = 25, length = 14)

Should this be a warning? WDG doesn't like warning because they treat warning as error. I suggest we change it to verbose.

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
```diff
@@ -926,50 +925,58 @@ private void TryFlushPageCacheToFileSystem(string path)
...
-                        // Note that the identity kind may be Anonymous if we couldn't establish an identity for the target file;
-                        // in that case we still need to call TrackChangesToFile to ensure the tracker latches to disabled.
-                        return TrackChangesToFile(handle, path, identityAndContentInfo.Value, knownFileName: fileName);
-                    });
-            }
-            catch (BuildXLException ex)
-            {
-                return new RecoverableExceptionFailure(ex);
-            }
+                        Tracing.Logger.Log.RetryTryOpenAndTrackPathAsync(m_loggingContext, path.ExpandedPath);
```
narasamdya: >RetryTryOpenAndTrackPathAsync [](start = 43, length = 29)

As this log goes to "localOnly", I suggest that you should use the log mechanism that the Helpers.RetryOnFailure provides. Yes, that amounts to throwing an exception, but I believe that the retry should be rare.

Public/Src/Utilities/Native/IO/Unix/FileSystem.Unix.cs
```diff
@@ -1188,35 +1187,24 @@ private bool SupportCopyOnWrite()
         /// </summary>
         public int GetFilePermission(string path, bool followSymlink = false, bool throwOnFailure = true)
         {
-            int result = GetFilePermissionsForFilePath(path, followSymlink);
```
narasamdya: >GetFilePermissionsForFilePath [](start = 25, length = 29)

Should we also remove this method from the C++ side?

narasamdya:
"the materialization step sometimes fails after the cache has successfully placed a file on disk" -- I don't quite follow the statement. I mean that the materialization step should only be considered success if the cache has successfully placed the file on disk.

I'm just trying to understand the issue.

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
narasamdya: Yes, `TryOpenAndTrackPathAsync`  implicitly assume that the file exists. For example, it's only called if file materialization from cache succeeds, which means the file gets materialized on disk.

I'm happy to discuss the issue offline in case I misunderstand something. Let's discuss this first (waiting on this)

---
In reply to: [278650019](https://github.com/Microsoft/BuildXL/pull/202#discussion_r278650019) [](ancestors = 278650019)

Public/Src/Engine/Cache/Log.cs
Kr1sso: Removed.

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
Kr1sso: Done, removed the added log statement.

Public/Src/Utilities/Native/IO/Unix/FileSystem.Unix.cs
Kr1sso: It still is being referenced elsewhere.

PR #204, opened 2019-04-25
Fix EventCount telemetry for macOS

almiliMSFT:
The current EventCount table has columns whose names start with a digit; the AriaV2 static library for macOS doesn't seem to support this.

To avoid having to alter the exiting EventCount table (and breaking existing queries), when running on macOS, use a different (new) table, namely `EventCountMacOS`, and prefix its column names with a constant `e`.

[AB#1511704](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1511704)

### DIFF
--- a/Public/Src/Utilities/Instrumentation/LogGen/Generators/AriaV2.cs
+++ b/Public/Src/Utilities/Instrumentation/LogGen/Generators/AriaV2.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using BuildXL.Utilities;
 using Microsoft.CodeAnalysis;
 
 namespace BuildXL.LogGen.Generators
@@ -15,12 +16,18 @@ namespace BuildXL.LogGen.Generators
         private const string GlobalInstrumentationNamespaceCommon = "global::BuildXL.Utilities.Instrumentation.Common";
         private const string GlobalTracing = "global::BuildXL.Tracing";
 
+        private const string EventCountMethodName = "EventCount";
+
         public override void GenerateLogMethodBody(LoggingSite site, Func<string> getMessageExpression)
         {
             m_codeGenerator.Ln("if ({0}.AriaV2StaticState.IsEnabled)", GlobalInstrumentationNamespaceCommon);
             using (m_codeGenerator.Br)
             {
-                m_codeGenerator.Lns("var eventData = new {0}.AriaEvent(\"{1}\", \"{2}\", \"{3}\")", GlobalInstrumentationNamespaceCommon, site.Method.Name, m_targetFramework, m_targetRuntime);
+                var tableName = OperatingSystemHelper.IsUnixOS && site.Method.Name == EventCountMethodName
+                    ? site.Method.Name + "MacOS"
+                    : site.Method.Name;
+
+                m_codeGenerator.Lns("var eventData = new {0}.AriaEvent(\"{1}\", \"{2}\", \"{3}\")", GlobalInstrumentationNamespaceCommon, tableName, m_targetFramework, m_targetRuntime);
 
                 // Save context fields that all events save
                 m_codeGenerator.Lns("eventData.SetProperty(\"Environment\", {0}.Session.Environment)", site.LoggingContextParameterName);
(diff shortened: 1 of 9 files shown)


Public/Src/Utilities/Instrumentation/LogGen/Generators/AriaV2.cs
```diff
@@ -15,12 +16,18 @@ internal sealed class AriaV2 : GeneratorBase
...
+        private const string EventCountMethodName = "EventCount";
+
         public override void GenerateLogMethodBody(LoggingSite site, Func<string> getMessageExpression)
         {
             m_codeGenerator.Ln("if ({0}.AriaV2StaticState.IsEnabled)", GlobalInstrumentationNamespaceCommon);
             using (m_codeGenerator.Br)
             {
-                m_codeGenerator.Lns("var eventData = new {0}.AriaEvent(\"{1}\", \"{2}\", \"{3}\")", GlobalInstrumentationNamespaceCommon, site.Method.Name, m_targetFramework, m_targetRuntime);
+                var tableName = OperatingSystemHelper.IsUnixOS && site.Method.Name == EventCountMethodName
+                    ? site.Method.Name + "MacOS"
```
dannyvv: >MacOS [](start = 42, length = 5)

A comment on why we need to 'patch' the table name vs creating two events and #if for mac/windows vs joining the host os in telemetry would be nice.

Public/Src/Utilities/Native/IO/Unix/FileUtilities.Unix.cs
```diff
@@ -484,21 +484,18 @@ public void CloneFile(string source, string destination, bool followSymlink)
 
             if (!openResult.Succeeded)
             {
-                openResult.CreateFailureForError().Throw();
+                openResult
+                    .CreateFailureForError()
+                    .Annotate($"{nameof(s_fileSystem.TryCreateOrOpenFile)} failed in {nameof(UsingFileHandleAndFileLength)}")
```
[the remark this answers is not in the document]
almiliMSFT: I don't see it in your PR so I'll keep it unless I get merge conflicts

Public/Src/Utilities/Native/IO/Unix/FileSystem.Unix.cs
```diff
@@ -871,10 +871,15 @@ public void GetChainOfReparsePoints(SafeFileHandle handle, string sourcePath, IL
...
+        {
+            return TryGetReparsePointTarget(sourcePath);
+        }
+
+        internal Possible<string> TryGetReparsePointTarget(string sourcePath)
         {
             try
             {
-                var maxPathLength = BuildXL.Native.IO.NativeIOConstants.MaxPath + 1;
+                var maxPathLength = NativeIOConstants.MaxPath + 1;
```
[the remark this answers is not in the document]
almiliMSFT: ok

Public/Src/Utilities/Instrumentation/LogGen/Generators/AriaV2.cs
almiliMSFT: done

Public/Src/Utilities/Instrumentation/LogGen/Generators/AriaV2.cs
```diff
@@ -94,13 +110,14 @@ private void WritePayload(LoggingSite site, LoggingSite.AddressedType item)
                     m_codeGenerator.Ln("foreach (var item in {0})", item.Address);
                     using (m_codeGenerator.Br)
                     {
+                        m_codeGenerator.Ln("var scrubbedPropertyName = {0}.AriaV2StaticState.ScrubEventProperty(item.Key);", GlobalInstrumentationNamespaceCommon);
```
mpysson: >ScrubEventProperty [](start = 93, length = 18)

Who knows if we have other fields that start with a number. What do you think about embedding the "don't start with a number" rule directly into ScrubEventProperty? That way we could keep the schema of our telemetry the same across platforms. Downside would be a one time breaking change.

almiliMSFT: If you think this breaking change is no big deal, your suggestion would be preferred

almiliMSFT: changed

PR #206, opened 2019-04-25
Add lazy directory materialization for IPC pips

olkononenko:
The change is mainly needed so drop pips would not materialize all the files from inside drop artifacts. For example, if customers want to upload only small subset of files from a directory, in the past we would materialize all the files (even though we did not need them) and then upload only those files that match a filter. With this change, we would not materialize directory artifacts at all; instead, dropdaemon will only materialize  those files that match the filter.

[AB#1487719](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1487719)

### DIFF
--- a/Public/Sdk/Public/Transformers/Transformer.Ipc.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Ipc.dsc
@@ -42,7 +42,7 @@ namespace Transformer {
          * moniker) just before the files are needed.  This makes sense for pips that expect that often
          * times they will not have to access the actual files on disk.  
          */
-        lazilyMaterializedDependencies?: Input[];        
+        lazilyMaterializedDependencies?: Input[];
 
         /**
          * Whether this pip must execute on the master node in a distributed build.  Defaults to false.
--- a/Public/Sdk/Public/Transformers/Transformer.Ipc.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Ipc.dsc
@@ -35,12 +35,16 @@ namespace Transformer {
         connectRetryDelayMillis?: number;
 
         /** 
-         * Files not to materialize eagerly.  
+         * Artifact not to materialize eagerly.  
          * 
-         * IPC pips may want to use this option when they will explicitly request file materialization
+         * IPC pips may want to use this option when they will explicitly request artifact materialization
          * from BuildXL (via a BuildXL service identified by the Transformer.getIpcServerMoniker()
-         * moniker) just before the files are needed.  This makes sense for pips that expect that often
-         * times they will not have to access the actual files on disk.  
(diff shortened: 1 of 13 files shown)


Public/Src/FrontEnd/Script/Ambients/Transformers/AmbientTransformer.Ipc.cs
```diff
@@ -159,17 +159,23 @@ private bool TryScheduleIpcPip(Context context, ObjectLiteral obj, bool allowUnd
...
-                    skipMaterializationFiles[i] = Converter.ExpectFile(
+                    Converter.ExpectFileOrStaticDirectory(
                         skipMaterializationLiteral[i],
-                        strict: true,
+                        out var fileArtifact,
+                        out var staticDirectory,
                         context: new ConversionContext(pos: i, objectCtx: skipMaterializationLiteral));
+
+                    // either fileArtifact or staticDirectory is valid at this point (otherwise, ExpectFileOrStaticDirectory would have thrown an exception) 
+                    skipMaterializationArtifacts[i] = fileArtifact.IsValid
```
dannyvv: perhaps an assert to ensure we don't get unexpected behavior when this assumption is no longer valid int he future...

narasamdya:
Don't you have to update `addArtifactsToDrop` in `Public\Src\Tools\DropDaemon\Tool.DropDaemonRunner.dsc` ?

narasamdya:
Add a test in `Public\Sdk\UnitTests\Transformers\Test.ipcSend.dsc` ensuring that the lazy field accept a directory.

narasamdya:
>          * Files not to materialize eagerly.  

Update comment.

---
Refers to: Public/Sdk/Public/Transformers/Transformer.Ipc.dsc:38 in 98fc149. [](commit_id = 98fc14933f2c3bc055dd48ebc85bef3834c2a1be, deletion_comment = False)

Public/Sdk/Public/Transformers/Transformer.Ipc.dsc
```diff
@@ -42,7 +42,7 @@ namespace Transformer {
          * moniker) just before the files are needed.  This makes sense for pips that expect that often
          * times they will not have to access the actual files on disk.
```
narasamdya: Include some description from this PR's description into this remarks.

Public/Src/Tools/DropDaemon/Tool.DropDaemonRunner.dsc
```diff
@@ -278,7 +278,10 @@ export namespace DropDaemonRunner {
                     ...directoryMessageBody,
```
narasamdya: >    [](start = 16, length = 3)

nit: remove extra line

PR #208, opened 2019-04-26
Handle 'rename directory' operation on macOS

almiliMSFT:
Special handling for directory rename.

Scenario: a pip writes a bunch of files into a directory (e.g., 'out.tmp') and then renames that directory (e.g., to 'out').  Once 'out.tmp' is renamed to 'out', we need to explicitly update all previously reported paths under 'out.tmp'.  Since we cannot change the past and directly mutate previously reported paths, we simply enumerate the content of the renamed directory and report all the files in there as writes.  This is exactly how this is done on Windows, except that it's implemented in the Detours layer.               

[AB#1362637](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1362637)


Examples/DotNetCoreBuild/test/TestDirectories/testDirectories.dsc
```diff
@@ -68,6 +68,25 @@ export const writeHardLinkInSharedOpaqueDirectoryIsAllowed = !Context.isWindowsO
     linkFileIntoDirectory(f`module.config.dsc`, sod, true);
 })();
 
+@@public
+export const moveDirectoryInsideSOD = !Context.isWindowsOS() && (() => {
```
dannyvv: Please use context.getcurrenthost this property should be deprecated since it doesn't scale to other os'

almiliMSFT: will change throughout...

Public/Src/Engine/Processes/SandboxedProcess.MacKext.cs
```diff
@@ -519,6 +519,27 @@ private void HandleKextReport(AccessReport report)
...
+                // special handling for directory rename:
+                //   - scenario: a pip writes a bunch of files into a directory (e.g., 'out.tmp') and then renames that directory (e.g., to 'out')
+                //   - up to this point we know about the writes into the 'out.tmp' directory
+                //   - once 'out.tmp' is renamed to 'out', we need to explicitly update all previously reported paths under 'out.tmp'
+                //       - since we cannot rewrite the past and directly mutate previously reported paths, we simply enumerate
+                //         the content of the renamed directory and report all the files in there as writes
+                //       - (this is exactly how this is done on Windows, except that it's implemented in the Detours layer)
+                else if (report.Operation == FileOperation.OpKAuthMoveDest && FileUtilities.DirectoryExistsNoFollow(reportPath))
+                {
+                    FileUtilities.EnumerateFiles(
```
smera: >EnumerateFiles [](start = 34, length = 14)

A caveat here: this will miss the case of a file that gets created and later deleted. This case is handled today with shared opaques, and even if the file is not there in the end, it is considered for violation analysis.
This is not the case for exclusive opaques, where we still rely on enumerations. But the plan was to eventually converge both types of opaque so they all use detours.

Not sure if this case is important to you. Just pointing it out: you can miss DFAs in this way that otherwise you'll get.

almiliMSFT: Thanks for pointing this out.  Indeed, there is a short window for the process to delete a file right after renaming the directory and before BuildXL receives the report that the directory was renamed.

With detours, this directory enumeration is done synchronously (right after (or before) the directory is renamed); on macOS with the kext, we cannot do it synchronously, so this is the best we can do. 

I don't know how important it is to catch the scenario you described.  In that case, the only thing we miss to detect is that a process created and then deleted a file; I don't think it can happen that we don't detect an actual output of the process though, which is the more important case.

PR #209, opened 2019-04-26
More descriptive error message when finalizer fails for FileStream.

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Library/FileSystem/TrackingFileStream.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            Interlocked.Increment(ref ProperlyClosed);
+            GC.SuppressFinalize(this);
+        }
+
+        ~TrackingFileStream()
+        {
+            Interlocked.Increment(ref Leaked);
+
+            try
```
dcwuser: Do we want to somehow log (maybe by leaving a file in a well-known-directory) that the finalizer ran at all?

dcwuser: Never mind, I see you do this via statistics.

Public/Src/Cache/ContentStore/Library/FileSystem/TrackingFileStream.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Threading;
+using JetBrains.Annotations;
+using Microsoft.Win32.SafeHandles;
+
+namespace BuildXL.Cache.ContentStore.FileSystem
+{
+    /// <summary>
+    /// Special tracking file stream that fails with more readable error message if unhandled error occurs in the finalizer of the instance.
+    /// </summary>
+    internal class TrackingFileStream : FileStream
```
dannyvv: Is it worth considering a more generic IDisposableWarpper?

`class LeakDetection<T> : T, IDisposable`

PR #220, opened 2019-04-29
Grpc p2p cache factory

kewilli:
This cache definition should be run with a persistant CAS instance, so other machines and other builds can consume the cached bits. This PR removes the in-proc CAS in lieu of a client which connects to an existing CAS instance. This persistent CAS can be started via the 'distributedService' verb on ContentStoreApp.exe, and can consume a `DistributedContentSettings` json object.

This has been tested to work on Windows. This has not yet been tested on Mac.


Public/Src/Cache/ContentStore/App/Application.cs
```diff
@@ -468,9 +468,8 @@ private void VerifyCachePathOrNameProvided(string name, string path)
             localCasSettings.ServiceSettings = new LocalCasServiceSettings(60, scenarioName: _scenario, grpcPort: grpcPort, grpcPortFileName: _scenario, bufferSizeForGrpcCopies: bufferSizeForGrpcCopies);
 
             var redisConnectionString = Environment.GetEnvironmentVariable(EnvironmentConnectionStringProvider.RedisConnectionStringEnvironmentVariable);
```
kewilli: Remove

Public/Src/Cache/ContentStore/Interfaces/FileSystem/AbsolutePathExtensions.cs
```diff
@@ -20,5 +21,20 @@ public static string GetFileName(this AbsolutePath path)
...
+        {
+            if (OperatingSystemHelper.IsWindowsOS)
+            {
+                return new AbsolutePath($"{path.DriveLetter}:{Path.DirectorySeparatorChar}");
+            }
+            else
+            {
+                return new AbsolutePath(Path.VolumeSeparatorChar.ToString());
+            }
+        }
```
dannyvv: Can you add this to PathTable?

I beleive I have seen this code before in another place...

kewilli: This was in a PR which was abandoned. It doesn't look like it's anywhere else in BuildXL.

---
In reply to: [279586982](https://github.com/Microsoft/BuildXL/pull/220#discussion_r279586982) [](ancestors = 279586982)

Public/Src/Cache/ContentStore/App/Hello.cs
```diff
@@ -23,18 +23,25 @@ internal sealed partial class Application
...
+                var _channel = new Channel(host, grpcPort, ChannelCredentials.Insecure);
+                var _client = new ContentServer.ContentServerClient(_channel);
+                var helloResponse = _client.Hello(new HelloRequest(), new CallOptions(deadline: DateTime.UtcNow + TimeSpan.FromSeconds(2)));
 
-            _logger.Always("Hello response {0}: {1}", helloResponse.Success ? "succeeded" : "failed", helloResponse.ToString());
+                _logger.Always("Hello response {0}: {1}", helloResponse.Success ? "succeeded" : "failed", helloResponse.ToString());
+            }
+            catch (Exception e)
+            {
+                _logger.Always(e.ToString());
```
dcwuser: This is good, but (1) was there some specific thing that happened that this was added to address and (2) do you need to add additional text to identify site of failure in log, so does the logger add enough for you to see where it occurred already?

kewilli: 1) A failed connection will throw an exception, though I don't remember which kind of exception. That exception flowed through and was not logged.

2) This verb is not intended to ever be used in a production environment, outside testing that one can connect to an existing deployment. By default, this logs the entirety of the exception, rather than just the exception's message.

---
In reply to: [279877617](https://github.com/Microsoft/BuildXL/pull/220#discussion_r279877617) [](ancestors = 279877617)

Public/Src/Cache/ContentStore/Interfaces/FileSystem/AbsolutePathExtensions.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Diagnostics.ContractsLight;
 using System.IO;
+using BuildXL.Cache.ContentStore.Interfaces.Utils;
```
SergeyTeplyakov: Still needed?

kewilli: Nope. Good catch!

---
In reply to: [280151698](https://github.com/Microsoft/BuildXL/pull/220#discussion_r280151698) [](ancestors = 280151698)

Public/Src/Cache/MemoizationStore/Library/Sessions/LocalCacheConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        /// <nodoc />
+        public uint RetryCount { get; }
+
+        /// <summary>
+        /// Create a local cache configuration which has distributed features.
+        /// </summary>
+        public static LocalCacheConfiguration CreateServerEnabled(int grpcPort, string cacheName, string scenarioName, uint retryIntervalSeconds, uint retryCount)
+        {
+            Contract.Assert(grpcPort > 0, $"Local server must have a positive GRPC port. Found {grpcPort}.");
```
SergeyTeplyakov: nit: can be requires, because it's responsibility of the client to provide correct data.

Public/Src/Cache/MemoizationStore/Library/Sessions/LocalCacheConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            return new LocalCacheConfiguration(true, grpcPort, cacheName, scenarioName, retryIntervalSeconds, retryCount);
+        }
+
+        /// <summary>
+        /// Create a local cache configuration which does not have any distributed features.
+        /// </summary>
+        public static LocalCacheConfiguration CreateServerDisabled()
+        {
+            return new LocalCacheConfiguration(false);
```
SergeyTeplyakov: nit: I would suggest to use name for true/false, like LocalCacheCOnfiguration(enableContentServer: false)

Public/Src/Cache/MemoizationStore/Library/Sessions/LocalCacheConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <summary>
+        /// Name of the custom scenario that the CAS services.
+        /// Allows multiple CAS services to coexist in a machine
+        /// since this factors into the cache root and the event that
+        /// identifies a particular CAS instance.
+        /// </summary>
+        public string ScenarioName { get; }
+
+        /// <nodoc />
+        public uint RetryIntervalSeconds { get; }
```
SergeyTeplyakov: nit: should we always use int instead of uint?

kewilli: Sure! But I won't update any of our public surface from `uint` to `int`.

---
In reply to: [280152542](https://github.com/Microsoft/BuildXL/pull/220#discussion_r280152542) [](ancestors = 280152542)

SergeyTeplyakov: I thought that this is the new stuff, so it means that it should not be a problem. Right?

kewilli: I should have been more specific. I won't update the constructor for `ServiceClientContentStore` from `uint` to `int`, because it is used by AzDev.

PR #221, opened 2019-04-29
Empty file optimizations

dcwuser:
Change empty file optimization to no have empty files in cache directory at all. This should address situation with 1000s of empty file placements are slow due to (1) replica logic for hard-links to file in cache and (2) per-hash lock.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
+++ b/Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
@@ -350,7 +350,7 @@ namespace BuildXL.Cache.ContentStore.FileSystem
 
             using (await ConcurrentAccess.WaitToken())
             {
-                return await OpenInsideSemaphoreAsync(path, fileAccess, fileMode, share, options, bufferSize);
+                return OpenInternal(path, fileAccess, fileMode, share, options, bufferSize);
             }
         }
 
@@ -416,9 +416,9 @@ namespace BuildXL.Cache.ContentStore.FileSystem
 
         private async Task CopyFileInsideSemaphoreAsync(AbsolutePath sourcePath, AbsolutePath destinationPath, bool replaceExisting)
         {
-            // It is very important to call OpenInsideSemaphoreAsync and not to call OpenAsync method that will re-acquire the semaphore once again.
+            // It is very important to call OpenInternal and not to call OpenAsync method that will re-acquire the semaphore once again.
             // Violating this rule may cause a deadlock.
-            using (var readStream = await OpenInsideSemaphoreAsync(
+            using (var readStream = OpenInternal(
                 sourcePath, FileAccess.Read, FileMode.Open, FileShare.Read | FileShare.Delete, FileOptions.None, AbsFileSystemExtension.DefaultFileStreamBufferSize))
             {
                 if (readStream == null)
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -2202,17 +2210,16 @@ public async Task<EvictResult> EvictAsync(Context context, ContentHashWithLastAc
...
-                    var putResult = await PutStreamAsync(context, _emptyFileStream, contentHash, pinRequest);
-                    if (!putResult.Succeeded)
-                    {
-                        // Because this is an optimization for the empty file, we shouldn't fail here.
-                        // Instead, let the normal path have a try.
-                        context.Debug($"Failed to place empty file by putting the empty stream: {putResult}");
-                    }
+                    // TODO: Respect I/O semaphore, set file permissions
+                    FileSystem.CreateDirectory(contentHashWithPath.Path.Parent);
+                    using (File.Create(contentHashWithPath.Path.Path)) { }
```
SergeyTeplyakov: I think we still should use FileSystem here, something like
`using(await FileSystem.CreateFile())`

where CreateFile is an extension method that calls OpenAsync with the right attributes.

I can see why this code will work, but I think its not a huge amount of work to use file system abstraction here.

dannyvv:
@ Please ensure you have a reasonable description before merging. I updated the PR title, but Github is not as nice as VSO you'll have to edit the commit each time, it won't use the data from the PR :(

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
kewilli: I don't think such a method exists on `IAbsFileSystem`, but it does exist on `BuildXL.Native.IO.FileUtilites` as `WriteAllBytesAsync` or `TryCreateOrOpenFile`.

---
In reply to: [279584408](https://github.com/Microsoft/BuildXL/pull/221#discussion_r279584408) [](ancestors = 279584408)

SergeyTeplyakov: I know, the idea is to add such an extension method that will do this.

---
In reply to: [279866313](https://github.com/Microsoft/BuildXL/pull/221#discussion_r279866313) [](ancestors = 279866313,279584408)

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
```diff
@@ -416,9 +416,9 @@ private static bool GetSequentialScanOnOpenStreamThresholdEnvVariable(out long r
 
         private async Task CopyFileInsideSemaphoreAsync(AbsolutePath sourcePath, AbsolutePath destinationPath, bool replaceExisting)
         {
-            // It is very important to call OpenInsideSemaphoreAsync and not to call OpenAsync method that will re-acquire the semaphore once again.
+            // It is very important to call OpenInternal and not to call OpenAsync method that will re-acquire the semaphore once again.
             // Violating this rule may cause a deadlock.
-            using (var readStream = await OpenInsideSemaphoreAsync(
+            using (var readStream = OpenInternal(
```
SergeyTeplyakov: >OpenInternal [](start = 36, length = 12)

OpenInsideSemaphoreInternal?

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
dcwuser: You have convinced me. However, it's been such a pain getting checks to pass, I'm going to go ahead and merge now that they are passing, and will do a follow-up PR for this tomorrow.

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
dcwuser: I got rid of the "InsideSemphore" part on purpose, because to me that implies that the method provides the semaphore, not that the method is intended to be called from another method that provides the semaphore.

PR #226, opened 2019-04-30
Bump fingerprint version

olkononenko:
Some cache entries contain corrupted data. Instead of tracking it down and deleting those entries - let's bump the FP version.


Public/Src/Engine/Scheduler/Fingerprints/PipFingerprintingVersion.cs
```diff
@@ -32,7 +32,8 @@ public enum PipFingerprintingVersion
         /// 56: Added NeedsToRunInContainer, ContainerIsolationLevel and DoubleWritePolicy
         /// 57: Fixed enumeration in StoreContentForProcessAndCreateCacheEntryAsync
         /// 58: Added RequiresAdmin field into the process pip.
+        /// 59: Bumping version due to corrupted data in the cache.
```
dannyvv: GitHub Test

olkononenko:
No longer needed.

PR #228, opened 2019-04-30
Use extra retry strategy for Event Hub client

SergeyTeplyakov:
Even though Event Hub client has it's own retry policy we have to add another one on top to handle a few missing cases, namely `SocketException` and `TimeoutException`.


Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -43,6 +46,8 @@ public sealed class EventHubContentLocationEventStore : ContentLocationEventStor
 
         private readonly ActionBlock<ProcessEventsInput>[] _eventProcessingBlocks;
 
+        private readonly Microsoft.Practices.TransientFaultHandling.RetryPolicy _extraEventHubClientRetryPolicy;
```
dcwuser: Why both using definition and full name here?

SergeyTeplyakov: Removed. Had a conflict but the conflict was resolved by `using RetryPolicy=...` at the top of the file.

PR #233, opened 2019-05-01
Telemetry improvements for EH client and FileSystemContentStore

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -17,6 +17,7 @@
 using Microsoft.Azure.EventHubs;
 using Microsoft.Practices.TransientFaultHandling;
 using static BuildXL.Cache.ContentStore.Distributed.NuCache.EventStreaming.ContentLocationEventStoreCounters;
+using Exception = System.Exception;
```
dcwuser: Someone named a type Exception?! Or is there some other reason this is required?

SergeyTeplyakov: Removed. It seems that R# inserted this, but for no good reason.

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <nodoc />
+        public static TResult WithLockAcquisitionDuration<TResult, TLockKey>(this TResult result, in LockSet<TLockKey>.LockHandle handle)
+            where TResult : ResultBase
+            where TLockKey : IEquatable<TLockKey>
+        {
+            if (handle.LockAcquisitionDuration != null)
+            {
+                var message = $", lock acquired by {(long)handle.LockAcquisitionDuration.Value.TotalMilliseconds}ms, ";
+                result.ExtraDiagnosticsMessage = message;
```
dcwuser: Is this the only place you are using the new ExtraDiagnosticsMessage property? Wouldn it be better to just add a LockAcquisitionDuration property to Result?

SergeyTeplyakov: This is the only place, but this aspect is added into 3 different (and unrelated) result types: OpenStreamResult, PlaceFileResult and PutFIleResult.
My first idea was to add `LockDuration` property, but once I realized that the types are unrelated I decided to go with a more general approach and add an ability to put extra message into the result.

Public/Src/Cache/ContentStore/Interfaces/Results/ResultsExtensions.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Runtime.CompilerServices;
```
kewilli: Remove?

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -2275,7 +2266,7 @@ public async Task<EvictResult> EvictAsync(Context context, ContentHashWithLastAc
 
                     if (code != PlaceFileResult.ResultCode.Unknown)
                     {
-                        return new PlaceFileResult(code, contentSize);
+                        return new PlaceFileResult(code, contentSize).WithLockAcquisitionDuration(contentHashHandle);
```
dannyvv: >.W [](start = 69, length = 2)

nit: Slight style-improvement to put this indented on the next line to signify an operation is happening on the result.

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            return result;
+        }
+
+        /// <nodoc />
+        public static string GetExtraDiagnosticsMessageForTracing(this ResultBase result)
+        {
+            if (!string.IsNullOrEmpty(result.ExtraDiagnosticsMessage))
+            {
+                return " " + result.ExtraDiagnosticsMessage;
```
kewilli: >return " " + result.ExtraDiagnosticsMessage; [](start = 16, length = 44)

Is it true that an example message would read:
`Blah.OpenStream stop 42ms , lock acquired by 2ms,  result=[Success]` ?

Can we pare this down? Perhaps remove the two commas, the extra space, and " acquired by".

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
dannyvv: >ExtraDiagnosticsMessage [](start = 23, length = 23)

Should we have an assert that it is not alredy set and being overwritten?

SergeyTeplyakov: Sounds like a good idea.

Public/Src/Cache/ContentStore/Interfaces/Results/ResultBase.cs
```diff
@@ -111,6 +111,11 @@ protected ResultBase(ResultBase other, string message)
         /// </summary>
         public TimeSpan Duration { get; set; }
 
+        /// <summary>
+        /// Gets or sets additional information that can be used for tracing purposes.
+        /// </summary>
+        public string ExtraDiagnosticsMessage { get; set; }
```
kewilli: >ExtraDiagnosticsMessage [](start = 22, length = 23)

`Extra` tells us nothing about the property. Would it be helpful to rename it for its actual purpose (e.g. `LockDurationMs`)?

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
SergeyTeplyakov: Yep, I've messed up. The output right now is `104ms , lock acquired by 0ms,  result=`.  I'll change it.

Public/Src/Cache/ContentStore/Library/Tracing/ContentSessionTracer.cs
```diff
@@ -233,7 +233,7 @@ public virtual void PlaceFileStop(Context context, ContentHash input, PlaceFileR
         {
             if (context.IsEnabled)
             {
-                TracerOperationFinished(context, result, $"{Name}.{PlaceFileCallName} stop {result.DurationMs}ms input=[{input}] result=[{result}]");
+                TracerOperationFinished(context, result, $"{Name}.{PlaceFileCallName} stop {result.DurationMs}ms{result.GetExtraDiagnosticsMessageForTracing()} input=[{input}] result=[{result}]");
```
ljcollins25: >{result.GetExtraDiagnosticsMessageForTracing()} [](start = 112, length = 47)

I think this should go after the result for two reasons: 1. consistency with old message format. 2. result success is very important information and I wouldn't want that pushed aside by supplemental info.

Public/Src/Cache/ContentStore/Interfaces/Results/ResultBase.cs
ljcollins25: I think it makes sense to have it be more general since we can use it for other purposes. Also having lock duration in ResultBase is a bit weird.

---
In reply to: [280252679](https://github.com/Microsoft/BuildXL/pull/233#discussion_r280252679) [](ancestors = 280252679)

ljcollins25:
>         public string Diagnostics;

Can we overload this rather than adding a new property?

---
Refers to: Public/Src/Cache/ContentStore/Interfaces/Results/ResultBase.cs:107 in 2af5ffb. [](commit_id = 2af5ffb862f60676eedf6cb153c0db63f9210be8, deletion_comment = False)

Public/Src/Cache/ContentStore/Interfaces/Results/ResultBase.cs
SergeyTeplyakov: Repurposed Diagnostics field for this.

---
In reply to: [280482874](https://github.com/Microsoft/BuildXL/pull/233#discussion_r280482874) [](ancestors = 280482874,280252679)

Public/Src/Cache/ContentStore/Library/Tracing/ContentSessionTracer.cs
SergeyTeplyakov: I decided to move this into ToString implementation.

---
In reply to: [280481495](https://github.com/Microsoft/BuildXL/pull/233#discussion_r280481495) [](ancestors = 280481495)

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace BuildXL.Cache.ContentStore.Utils
+{
+    internal static class ResultsExtensions
+    {
+        /// <nodoc />
+        public static TResult WithLockAcquisitionDuration<TResult, TLockKey>(this TResult result, in LockSet<TLockKey>.LockHandle handle)
+            where TResult : ResultBase
+            where TLockKey : IEquatable<TLockKey>
+        {
+            Contract.Requires(!result.Succeeded || result.Diagnostics == null, $"Diagnostics property can be set only once.");
```
kewilli: >$ [](start = 79, length = 1)

Unnecessary formatted string

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+            if (result.Succeeded && handle.LockAcquisitionDuration != null)
+            {
+                var message = $", LockWait={(long)handle.LockAcquisitionDuration.Value.TotalMilliseconds}ms";
```
kewilli: >message [](start = 20, length = 7)

This local variable is unnecessary

kewilli: >, [](start = 32, length = 1)

I don't think that this comma fits with any of our existing telemetry. Could this be removed?

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+                result.Diagnostics = message;
```
kewilli: >Diagnostics [](start = 23, length = 11)

We should not replace an existing value in `Diagnostics`. Please consider appending instead.

Public/Src/Cache/ContentStore/Interfaces/Results/ResultsExtensions.cs
```diff
@@ -12,6 +12,12 @@ namespace BuildXL.Cache.ContentStore.Interfaces.Results
     /// </summary>
     public static class ResultsExtensions
     {
+        /// <nodoc />
+        public static string GetDiagnosticsMessageForTracing(this ResultBase result)
```
kewilli: >GetDiagnosticsMessageForTracing [](start = 29, length = 31)

Would it be simpler to add a special getter for `Diagnostics` instead? Then we don't need to remember to use this method in the future. I can guarantee that I won't remember.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -2215,9 +2213,7 @@ public async Task<EvictResult> EvictAsync(Context context, ContentHashWithLastAc
                 // operations in the in-memory representation of the cache.
                 if (_settings.UseEmptyFileHashShortcut && contentHashWithPath.Hash.IsEmptyHash())
                 {
-                    FileSystem.CreateDirectory(contentHashWithPath.Path.Parent);
-                    using (await FileSystem.OpenAsync(contentHashWithPath.Path, FileAccess.Write, FileMode.Create, FileShare.None, FileOptions.None, 1).ConfigureAwait(false)) { }
-
+                    await FileSystem.CreateEmptyFileAsync(contentHashWithPath.Path);
```
kewilli: >CreateEmptyFileAsync [](start = 37, length = 20)

Will this be called anywhere else? If not, then let's revert.

SergeyTeplyakov: I actually really prefer to have higher level methods like this that explain immediately the purpose of the code. So I actually would encourage us to have more code like that even when it is used only once.

Public/Src/Cache/ContentStore/Interfaces/Results/ResultsExtensions.cs
SergeyTeplyakov: I'm afraid of a breaking change. Currently `Diagnositcs` is a field and I was not sure if it's safe to make it a property. On the other hand, this extension method is special and gives you an ability to see all the other cases in one place: everyone who calls this method uses `Diagnostics` property to get extra information for successful case (because before this change, `Diagnostics` property was only used for a failure case).

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
SergeyTeplyakov: The comma actually perfectly fits into existing telemetry because this method is called from `PutResult.ToString()` method that uses the following approach: `$"Success Hash={ContentHash}, Size={ContentSize}"`. With this change it'll become: `Success Hash=hash, Size=42, LockWait=42ms`.

SergeyTeplyakov: The comma actually perfectly fits into existing telemetry because this method is called from PutResult.ToString() method that uses the following approach: $"Success Hash={ContentHash}, Size={ContentSize}". With this change it'll become: Success Hash=hash, Size=42, LockWait=42ms.

---
In reply to: [280619318](https://github.com/Microsoft/BuildXL/pull/233#discussion_r280619318) [](ancestors = 280619318)

Public/Src/Cache/ContentStore/Interfaces/Results/ResultsExtensions.cs
SergeyTeplyakov: I'm afraid of a breaking change. Currently Diagnositcs is a field and I was not sure if it's safe to make it a property. On the other hand, this extension method is special and gives you an ability to see all the other cases in one place: everyone who calls this method uses Diagnostics property to get extra information for successful case (because before this change, Diagnostics property was only used for a failure case).

---
In reply to: [280620200](https://github.com/Microsoft/BuildXL/pull/233#discussion_r280620200) [](ancestors = 280620200)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
SergeyTeplyakov: I actually really prefer to have higher level methods like this that explain immediately the purpose of the code. So I actually would encourage us to have more code like that even when it is used only once.

---
In reply to: [280620653](https://github.com/Microsoft/BuildXL/pull/233#discussion_r280620653) [](ancestors = 280620653)

kewilli:
>         protected string GetErrorString()

Will this now return `Error=[Something went wrong] Diagnostics=[, LockWait=42ms]`?

---
Refers to: Public/Src/Cache/ContentStore/Interfaces/Results/ResultBase.cs:157 in 419fc8c. [](commit_id = 419fc8ce01f383432b5128b7f64819ba04d703fa, deletion_comment = False)

Public/Src/Cache/ContentStore/Library/Utils/ResultsExtensions.cs
kewilli: You're right, I was incorrect to suggest that it wouldn't fit with any of our existing telemetry. While it does fit with PutResult, it doesn't fit with OpenStreamResult and PlaceFileResult.

---
In reply to: [280638720](https://github.com/Microsoft/BuildXL/pull/233#discussion_r280638720) [](ancestors = 280638720,280619318)

SergeyTeplyakov: > > ```
> >     protected string GetErrorString()
> > ```
> 
> Will this now return `Error=[Something went wrong] Diagnostics=[, LockWait=42ms]`?
> 
> Refers to: Public/Src/Cache/ContentStore/Interfaces/Results/ResultBase.cs:157 in [419fc8c](https://github.com/Microsoft/BuildXL/commit/419fc8ce01f383432b5128b7f64819ba04d703fa). [](commit_id = [419fc8c](https://github.com/Microsoft/BuildXL/commit/419fc8ce01f383432b5128b7f64819ba04d703fa), deletion_comment = False)

It won't because you can set lock duration only for successful case.

PR #236, opened 2019-05-02
Use different BundleIDs for debug and release kexts

almiliMSFT:
This change improves the typical dev loop by allowing 2 kexts to be loaded at the same time:
  1. one that came with the LKG (which one is compiled as `release`), and
  2. one compiled as `debug` from the current sources

The first one is typically used to compile current sources, and the second one is then typically used for running tests.


Kr1sso:
Don't forget to adjust unload on the VSTS definitions

PR #237, opened 2019-05-02
[WIP] Try to force net core qualifiers to analyze Mojave build breaks

Kr1sso:
An attempt to fix build breaks on Azure VMs running on latest macOS Mojave. Feel free to ignore this.

### DIFF
--- a/Public/Src/Cache/VerticalStore/Deployment.dsc
+++ b/Public/Src/Cache/VerticalStore/Deployment.dsc
@@ -8,6 +8,7 @@ namespace Deployment {
     @@public
     export const deployment: Deployment.Definition = {
         contents: [
+            Analyzer.exe,
             InMemory.dll,
             Interfaces.dll,
             BasicFilesystem.dll,
@@ -16,7 +17,6 @@ namespace Deployment {
             MemoizationStoreAdapter.dll,
             VerticalAggregator.dll,
         ...addIf(!BuildXLSdk.isDotNetCoreBuild,
-                Analyzer.exe,
                 Compositing.dll,
                 InputListFilter.dll
             )
--- a/Public/Src/Cache/VerticalStore/Deployment.dsc
+++ b/Public/Src/Cache/VerticalStore/Deployment.dsc
@@ -4,34 +4,22 @@
 import * as Deployment from "Sdk.Deployment";
 import * as Managed from "Sdk.Managed";
 
-namespace Deployment{
-
-
-export declare const qualifier: BuildXLSdk.DefaultQualifier;
-
-@@public
-export const deployment: Deployment.Definition = {
-    contents: BuildXLSdk.isDotNetCoreBuild ?
-    [
-        BasicFilesystem.dll,
-        BuildCacheAdapter.dll,
-        InMemory.dll,
-        Interfaces.dll,
-        ImplementationSupport.dll,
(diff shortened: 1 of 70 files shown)


Public/Src/Cache/VerticalStore/Deployment.dsc
```diff
@@ -4,34 +4,22 @@
...
-};
-
+namespace Deployment {
+    @@public
+    export const deployment: Deployment.Definition = {
+        contents: [
+            InMemory.dll,
+            Interfaces.dll,
+            BasicFilesystem.dll,
+            BuildCacheAdapter.dll,
```
dannyvv: You might want to check with David & Lance: They were checking if we could remove these..

Public/Src/Deployment/buildXL.dsc
```diff
@@ -21,9 +21,9 @@ namespace BuildXL {
             // analyzers
             importFrom("BuildXL.Tools").Execution.Analyzer.exe,
             importFrom("BuildXL.Tools").BxlScriptAnalyzer.exe,
-            ...(qualifier.targetFramework === "netcoreapp2.2" ? [] : [
+            ...addIf(qualifier.targetFramework !== "netcoreapp2.2",
```
dannyvv: >ifier.targetFramework !== "netcoreapp2.2 [](start = 25, length = 40)

If we will never support this on .net core (i.e. not even on 3.0 we should use `addIf(BuildXLSdk.isFullFramework,`

Public/Src/Cache/VerticalStore/Deployment.dsc
Kr1sso: This is semantically still the same, just without duplication. I'll let them decide once they refactor.

Public/Src/Deployment/buildXL.dsc
Kr1sso: Speaking of, I don't know a single reason why this should not run on macOS?

Public/Src/Deployment/NugetPackages.dsc
```diff
@@ -8,17 +8,18 @@ import * as Managed from "Sdk.Managed.Shared";
...
         ? "BuildXL"
         : "Microsoft.BuildXL";
+
     const packageTargetFolder = BuildXLSdk.Flags.isMicrosoftInternal
         ? r`${qualifier.configuration}/pkgs`
         : r`${qualifier.configuration}/public/pkgs`;
 
-
-    const net472 = pack({
+    const net472 = Context.getCurrentHost().os === "macOS" ? undefined : pack({
```
dannyvv: >t.getCurrentHost().os === "macO [](start = 25, length = 31)

perhaps add a local boolean shouldProduceAllNugetPacakges. Reason is that the check here is NOT the inverse of the check on line 113. If we add a third OS we'll end up with a funny split. so this should be always `=== "win"` and `!== "win"` or  `"macOS"`, but not `=== "win"` on one and `=== "macOs"` on the other :D

Kr1sso: Yeah, will refactor this, its a WIP in the PR title 💃

Public/Src/Deployment/buildXL.dsc
Kr1sso: Adding it.

PR #239, opened 2019-05-02
Add support for pip priority field and schedule according to pip priority as well as historic data.

erickulcyk:
In order to make mularchy builds feasible, the order of pips has to be priority according to similar memory access patters (eg x64 pips near other x64 pips, x86 pips near other x86 pips).  To facilitate this, a priority field has been added to process pips which will group them together in the scheduler.  The scheduler now divides the entire integer range in 100 pieces, one for each value of the priority you can specify, and within that range will be encoded the original priority value based upon critical path and historic run time.


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5046,7 +5046,11 @@ private void PrioritizeAndSchedule(LoggingContext loggingContext, IEnumerable<No
                                 }
                             }
 
-                            pipRuntimeInfo.Priority = (criticalPath < 0 || criticalPath > MaxInitialPipPriority) ? MaxInitialPipPriority : unchecked((int)criticalPath);
```
dannyvv: >pipRuntimeInfo [](start = 28, length = 14)

Signing off, but please have Kevin, Lance or Semih double-check this logic. how priority integer mixes with the critical path weight.

narasamdya:
Can you add description about the work? Some of us, like me, don't have the context about what you're doing.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5046,7 +5046,11 @@ private void PrioritizeAndSchedule(LoggingContext loggingContext, IEnumerable<No
                                 }
                             }
 
-                            pipRuntimeInfo.Priority = (criticalPath < 0 || criticalPath > MaxInitialPipPriority) ? MaxInitialPipPriority : unchecked((int)criticalPath);
+                            int priorityIncrements = (int.MaxValue / (Process.MaxPriority + 1));
+                            int priorityBase = m_pipTable.GetPipPriority(pipId) * priorityIncrements;
+                            int criticalPathPriority = (criticalPath < 0 || criticalPath > MaxInitialPipPriority) ? MaxInitialPipPriority : unchecked((int)criticalPath);
+                            criticalPathPriority = Math.Min(criticalPathPriority, priorityIncrements - 1);
+                            pipRuntimeInfo.Priority = priorityBase + criticalPathPriority;
```
narasamdya: Some of us, like me, don't have the context of the work. So it's a bit hard to assess the work. Can you provide one?

Public/Src/Pips/Dll/MutablePipState.cs
```diff
@@ -37,14 +37,17 @@ internal class MutablePipState
 
         internal readonly PipType PipType;
 
+        internal readonly int Priority;
```
ljcollins25: >Priority [](start = 30, length = 8)

Seems like this only applies to process pips. Should probably go in ProcessMutablePipState.

Public/Src/Engine/Scheduler/Scheduler.cs
erickulcyk: Added a description to the pr

---
In reply to: [280630658](https://github.com/Microsoft/BuildXL/pull/239#discussion_r280630658) [](ancestors = 280630658)

dannyvv: Note to the person merging this pr. GitHub is different than vso. You'll have to do manual work to copy paste the or details into the merge commit as GitHub chose to append all commit messages in the or and totally ignore the pr info....

erickulcyk: good to know, thanks!

---
In reply to: [280865393](https://github.com/Microsoft/BuildXL/pull/239#discussion_r280865393) [](ancestors = 280865393)

Public/Src/Pips/Dll/MutablePipState.cs
erickulcyk: Do you think it should still be accessed through piptable or be a public property of processmutablepipstate

---
In reply to: [280853322](https://github.com/Microsoft/BuildXL/pull/239#discussion_r280853322) [](ancestors = 280853322)

Public/Src/Engine/Scheduler/Scheduler.cs
ljcollins25: I would recommend a different approach. Reserve the top bits (say 7 + 1 for the unused sign bit) for the priority and XOR that with the critical path priority (which should be adjusted so that its maximum value is int.MaxValue >> 7)

erickulcyk: That is basically what I am doing right?  except for instead of 128 I am using 100.

---
In reply to: [280869710](https://github.com/Microsoft/BuildXL/pull/239#discussion_r280869710) [](ancestors = 280869710)

erickulcyk: going to assume you meant OR

---
In reply to: [280885317](https://github.com/Microsoft/BuildXL/pull/239#discussion_r280885317) [](ancestors = 280885317,280869710)

PR #244, opened 2019-05-02
Test querying the journal on getting the journal accessor, even in elevated mode

narasamdya:
In the past in elevated mode, we simply return the journal accessor without checking if we can query the journal. This causes a problem because then tests that require journal scan will run although the journal cannot be queried, e.g., due to failure in opening volume handle.


dannyvv:
just a note: We might have to check the osgtools repo. They have WarnignsAsErrors turned on. If we expect this to be printed in their CI environment we might have to exclude this warning from warnaserror.

PR #245, opened 2019-05-03
Avoid using and killing VBCSCompiler.exe

jbayardo:
MSBuild spawns the VBCSCompiler.exe in order to avoid repeated calls to compilers and lessen the cost of initializing them.

The problem is that the instance is a singleton, and thus many different pips may use the same instance; hence, if one pip kills it as part of its cleanup process, then a different pip may error because it was using the instance.

The solution here is to instruct MSBuild not to _use_ VBCSCompiler.exe and stop BuildXL from killing it as part of the cleanup; notably, it may still be launched even if it is instructed not to use VBCSCompiler.

### DIFF
--- a/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
+++ b/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
@@ -536,17 +536,6 @@ namespace BuildXL.FrontEnd.MsBuild
                 AddMsBuildProperty(pipDataBuilder, kvp.Key, kvp.Value);
             }
 
-            // Compilation uses VBCSCompiler.exe as a cache; this is a process that lives for longer than
-            // the pip itself does, so we will kill it as part of the clean-up. By forcefully adding the
-            // property below, we ensure that the cache is not used by any of the processes, thus making
-            // VBCSCompiler.exe safe to kill.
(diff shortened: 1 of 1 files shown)


Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -503,7 +503,8 @@ private void SetProcessEnvironmentVariables(IReadOnlyDictionary<string, string>
             processBuilder.AllowedSurvivingChildProcessNames = ReadOnlyArray<PathAtom>.FromWithoutCopy(
                 PathAtom.Create(m_context.StringTable, "mspdbsrv.exe"),
                 PathAtom.Create(m_context.StringTable, "vctip.exe"),
-                PathAtom.Create(m_context.StringTable, "conhost.exe"));
+                PathAtom.Create(m_context.StringTable, "conhost.exe"),
+                PathAtom.Create(m_context.StringTable, "VBCSCompiler.exe"));
```
dannyvv: >VBCSCompiler [](start = 56, length = 12)

Small suggestion. It seems that this method is called for each and every pip.

It would be more performant if the the array would be created only once for the lifetime of the frontend and then reused.

The same holds true for a few other things like the arguments for logging etc.
Not needed in this PR. Just sharing to improve perf in the future.

PR #246, opened 2019-05-03
[MacSandbox] skip unnecessarily creating trie nodes on lookups

almiliMSFT:
Before this PR, on every trie operation a leaf node corresponding to a given key is retrieved, creating any intermediate nodes along the way. 

This is unnecessary for lookup operations.  When the key does not exist in the trie, instead of traversing the trie until a leaf node is found and creating all intermediate nodes along the way, the lookup should fail as soon as an intermediate node is not found.

As a consequence of the previous implementation, a process could connect to the kext and then immediately disconnect, which would cause the size of the [`connectedClients_`](https://github.com/Microsoft/BuildXL/blob/2054bd1bf5e790119dfe979d8897e0aeca493045/Public/Src/Sandbox/MacOs/Sandbox/Src/BuildXLSandbox.hpp#L48) dictionary to monotonically grow.  This can become a problem because those nodes are released only when the kext is unloaded.

### DIFF
--- a/Public/Src/Sandbox/MacOs/Interop/Interop.xcodeproj/project.pbxproj
+++ b/Public/Src/Sandbox/MacOs/Interop/Interop.xcodeproj/project.pbxproj
@@ -87,7 +87,7 @@
 		F58E9204220B5B3C0083C57E /* utf8proc.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = utf8proc.c; sourceTree = "<group>"; };
 		F58E9205220B5B3C0083C57E /* utf8proc_data.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = utf8proc_data.c; sourceTree = "<group>"; };
 		F58E9206220B5B3C0083C57E /* utf8proc.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = utf8proc.h; sourceTree = "<group>"; };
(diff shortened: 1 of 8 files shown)


Public/Src/Sandbox/MacOs/Interop/Interop.xcodeproj/project.pbxproj
```diff
@@ -87,7 +87,7 @@
 		F58E9204220B5B3C0083C57E /* utf8proc.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = utf8proc.c; sourceTree = "<group>"; };
 		F58E9205220B5B3C0083C57E /* utf8proc_data.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = utf8proc_data.c; sourceTree = "<group>"; };
 		F58E9206220B5B3C0083C57E /* utf8proc.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = utf8proc.h; sourceTree = "<group>"; };
-		F598838E22527E8700A7A2D9 /* BundleInfoTest.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; name = BundleInfoTest.xcconfig; path = ../BundleInfoTest.xcconfig; sourceTree = "<group>"; };
+		F598838E22527E8700A7A2D9 /* BundleInfoDebug.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; name = BundleInfoDebug.xcconfig; path = ../BundleInfoDebug.xcconfig; sourceTree = "<group>"; };
```
almiliMSFT: I should have done this back in the PR where I renamed `BundleInfoTest.xcconfig` to `BundleInfoDebug.xcconfig`

PR #248, opened 2019-05-03
ExecutionLog events are now processed asynchronously on master

semihokur:
Processing some executionlog events on master can take more than 5 minutes. Because we do not send an ACK to the worker before the master completes processing the event, the worker assumes that the message failed to send, so the worker attempts to re-send the message. With changes below, master sends an ACK before processing the events.

### DIFF
--- a/Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
+++ b/Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
@@ -81,7 +81,7 @@ namespace BuildXL.Engine.Distribution
         private BinaryLogReader m_executionLogBinaryReader;
         private MemoryStream m_executionLogBufferStream;
         private ExecutionLogFileReader m_executionLogReader;
-        private readonly object m_logBlobLock = new object();
+        private readonly SemaphoreSlim m_logBlobMutex = TaskUtilities.CreateMutex();
         private readonly object m_hashListLock = new object();
 
         private int m_lastBlobSeqNumber = -1;
@@ -230,7 +230,7 @@ namespace BuildXL.Engine.Distribution
             await Task.Yield();
 
             // Execution log events cannot be logged by multiple threads concurrently since they must be ordered
-            lock (m_logBlobLock)
+            using (await m_logBlobMutex.AcquireAsync())
             {
                 // We need to dequeue and process the blobs in order. 
                 // Here, we do not necessarily process the blob that is just added to the queue above.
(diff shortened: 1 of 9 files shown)


Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -214,11 +219,30 @@ private void SendBuildRequests()
...
         {
+            Contract.Requires(notification.ExecutionLogData != null || notification.ExecutionLogData.Count != 0)
+
+            m_executionBlobQueue.Add(notification);
+
+            // After we put the executionBlob in a queue, we can unblock the caller and give an ACK to the worker.
+            await Task.Yield();
+
             // Execution log events cannot be logged by multiple threads concurrently since they must be ordered
             lock (m_logBlobLock)
```
ljcollins25: >lock (m_logBlobLock) [](start = 12, length = 20)

Use a semaphore here. TaskUtilities.CreateMutex

dannyvv:
Adding Erik and Kevin as FYI for grpc pattern with ack.

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -214,11 +219,30 @@ private void SendBuildRequests()
             }
         }
 
-        public void LogExecutionBlob(ArraySegment<byte> executionLogBlob, int blobSequenceNumber)
+        [SuppressMessage("AsyncUsage", "AsyncFixer02:awaitinsteadofwait")]
```
olkononenko: >SuppressMessage [](start = 9, length = 15)

In other places we use a fully qualified name (ie.., System.Diagnostics.CodeAnalysis.SuppressMessage), should we make it consistent?

Public/Src/Engine/Scheduler/Distribution/Worker.cs
```diff
@@ -262,10 +262,12 @@ public virtual void Start()
         /// <summary>
         /// Signals that build is finished and that worker should exit
         /// </summary>
-        public virtual void Finish(string buildFailure)
+#pragma warning disable 1998
```
olkononenko: Can we use SuppressMessage here instead? If not, can you add a comment describing the meaning of '1998'?

semihokur: suppressmessage does not work here. I added a comment though

PR #249, opened 2019-05-03
Start implementation of BuildXL VFS cas process.

ljcollins25:
Start implementation of BuildXL VFS cas process. Currently just proxies to a backing content server. Next step is to actually add virtualization.


Public/Src/Cache/ContentStore/Vfs/VfsCasConfiguration.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </summary>
+        public AbsolutePath VfsRootPath { get; }
+
+        /// <summary>
+        /// The cache name
+        /// </summary>
+        public string CacheName { get; }
+
+        /// <summary>
+        /// What does this even mean?
```
kewilli: >What does this even mean? [](start = 12, length = 25)

:|

Public/Src/Cache/ContentStore/Vfs/VfsProgram.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        public override int Run(VfsCasConfiguration arguments)
+        {
+            var runner = new VfsCasRunner();
+            runner.RunAsync(arguments).GetAwaiter().GetResult();
+            return 0;
+        }
+
+        public override bool TryParse(string[] rawArgs, out VfsCasConfiguration arguments)
```
kewilli: >TryParse [](start = 29, length = 8)

Why not use CLAP?

Public/Src/Cache/ContentStore/Vfs/VirtualizedContentSession.cs
```diff
@@ -0,0 +1,152 @@
...
+using BuildXL.Utilities.Tasks;
+using BuildXL.Utilities.Tracing;
+
+namespace BuildXL.Cache.ContentStore.Vfs
+{
+    /// <todoc />
+    public class VirtualizedContentSession : ContentSessionBase
+    {
+        private IContentSession InnerSession { get; }
+        //private VirtualizationRegistry VirtualizationRegistry { get; }
```
kewilli: >//private VirtualizationRegistry VirtualizationRegistry { get; } [](start = 8, length = 64)

Remove?

Public/Src/Cache/ContentStore/Vfs/VirtualizedContentSession.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        /// <inheritdoc />
+        protected override Task<PutResult> PutStreamCoreAsync(OperationContext operationContext, HashType hashType, Stream stream, UrgencyHint urgencyHint, Counter retryCounter)
+        {
+            return InnerSession.PutStreamAsync(operationContext, hashType, stream, operationContext.Token, urgencyHint);
+        }
+
+        //protected override async Task<PlaceFileResult> PlaceFileCoreAsync(
+        //    OperationContext operationContext,
```
kewilli: Remove?

Public/Src/Cache/ContentStore/Vfs/VfsProgram.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Threading.Tasks;
+using BuildXL.ToolSupport;
+
+namespace BuildXL.Cache.ContentStore.Vfs
+{
+    using System.Diagnostics.ContractsLight;
+    using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
+    using static CommandLineUtilities;
+
+    public class VfsProgram : ToolProgram<VfsCasConfiguration>
```
kewilli: >VfsProgram [](start = 17, length = 10)

Should this be inside ContentStoreApp.exe instead?

Public/Src/Cache/ContentStore/Vfs/VfsProgram.cs
ljcollins25: I'll probably end up using that just because I don't want to write a help text infrastructure. The reason I avoided it was because CLAP doesn't support key value pair and full path parsing.

---
In reply to: [280869902](https://github.com/Microsoft/BuildXL/pull/249#discussion_r280869902) [](ancestors = 280869902)

Public/Src/Cache/ContentStore/Vfs/VirtualizedContentSession.cs
ljcollins25: Still debating that with myself. Will discuss with team.

---
In reply to: [280870341](https://github.com/Microsoft/BuildXL/pull/249#discussion_r280870341) [](ancestors = 280870341)

Public/Src/Cache/ContentStore/Vfs/VfsProgram.cs
ljcollins25: Still debating that with myself. Will discuss with team.

---
In reply to: [280871326](https://github.com/Microsoft/BuildXL/pull/249#discussion_r280871326) [](ancestors = 280871326)

Public/Src/Cache/ContentStore/Vfs/VfsCasRunner.cs
```diff
@@ -0,0 +1,127 @@
...
+using BuildXL.Cache.ContentStore.Sessions;
+using BuildXL.Cache.ContentStore.Stores;
+
+namespace BuildXL.Cache.ContentStore.Vfs
+{
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Threading.Tasks;
+    using BuildXL.Cache.ContentStore.Interfaces.Extensions;
+    using BuildXL.Cache.ContentStore.Interfaces.Logging;
```
SergeyTeplyakov: Move it to the top? Or move everything here?

Public/Src/Cache/ContentStore/Vfs/VfsCasRunner.cs
```diff
@@ -0,0 +1,127 @@
...
+                        { configuration.CacheName, configuration.ServerRootPath }
+                        },
+                        configuration.ServerGrpcPort)))
+                {
+
+                    await server.StartupAsync(context).ThrowIfFailure();
+
+                    await WaitForTerminationAsync(context);
+
+                    await server.ShutdownAsync(context).ThrowIfFailure();
```
SergeyTeplyakov: Just curious, should we shutdown in a finally block?

Public/Src/Cache/ContentStore/Vfs/VirtualizedContentStore.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            return InnerStore.GetStatsAsync(context);
+        }
+
+        /// <inheritdoc />
+        protected override async Task<BoolResult> StartupCoreAsync(OperationContext context)
+        {
+            await InnerStore.StartupAsync(context).ThrowIfFailure();
+
+            return await base.StartupCoreAsync(context);
```
SergeyTeplyakov: nit: I think initialization and de-initialization should happen in a symmetric fasion: initialize the base version and then inner , and de-initialize the Inner and then the base.

It means that I think the base calls here should be the first...

Public/Src/Cache/ContentStore/Vfs/VirtualizedContentSession.cs
```diff
@@ -0,0 +1,109 @@
...
+using BuildXL.Utilities.Tasks;
+using BuildXL.Utilities.Tracing;
+
+namespace BuildXL.Cache.ContentStore.Vfs
+{
+    /// <todoc />
+    public class VirtualizedContentSession : ContentSessionBase
+    {
+        private IContentSession InnerSession { get; }
+        private VirtualizedContentStore Store { get; }
```
SergeyTeplyakov: Maybe a field instead of get-only private readonly property?

SergeyTeplyakov:
> using System;

nit: license?

---
Refers to: Public/Src/Cache/ContentStore/Vfs/VirtualizedContentSession.cs:1 in f5984ad. [](commit_id = f5984ad2ddf75e72f549718dcfff1d7d7d8e660a, deletion_comment = False)

SergeyTeplyakov:
Just to understand the current state: it seems that this is just an infra to glue pieces together, there is no actual logic there yet. Right?

Public/Src/Cache/ContentStore/Vfs/VfsCasRunner.cs
ljcollins25: I don't think it makes sense to shutdown if you fail to startup.

---
In reply to: [281273745](https://github.com/Microsoft/BuildXL/pull/249#discussion_r281273745) [](ancestors = 281273745)

ljcollins25:
Right.

---
In reply to: [489699591](https://github.com/Microsoft/BuildXL/pull/249#issuecomment-489699591) [](ancestors = 489699591)

PR #252, opened 2019-05-04
Increase WorkerAttachTimeout to 45min from 30min

semihokur:


semihokur:
TTFP is more than 30 minutes for CNMFR branches, so workers failed to attach the master if the graph is constructed.

PR #253, opened 2019-05-06
Enable distributed copies on BuildXL builds on macOS

kewilli:
Validated by building the same content on the same macOS box twice, each time using a separate local cache instance. The first build had a 0% cache hit rate, as expected. The second build also had a 0% cache hit rate, but copied ~4k files. It appears that copies are functioning as expected, but more investigation is necessary to understand the low hit rate.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using BuildXL.Cache.ContentStore.Extensions;
 using BuildXL.Cache.ContentStore.Hashing;
 using BuildXL.Cache.ContentStore.Interfaces.Distributed;
 using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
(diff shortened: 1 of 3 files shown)


Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
```diff
@@ -50,31 +51,30 @@ public byte[] GetLocalMachineLocation(AbsolutePath cacheRoot)
...
             }
 
-            var cacheRootString = cacheRoot.Path.ToUpperInvariant();
+            var cacheRootString = cacheRoot.Path.PathToUpperInvariant();
 
             // Determine if cacheRoot needs to be accessed through its directory junction
             var directories = _junctionsByDirectory.Keys;
             var directoryToReplace = directories.SingleOrDefault(directory =>
-                                        cacheRootString.StartsWith(directory, StringComparison.OrdinalIgnoreCase));
+                                        cacheRootString.StartsWith(directory, OperatingSystemHelper.IsWindowsOS ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal));
```
almiliMSFT: In all our current macOS builds the file system is configured to be case insensitive.

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
```diff
@@ -50,31 +51,30 @@ public byte[] GetLocalMachineLocation(AbsolutePath cacheRoot)
                 cacheRoot = cacheRoot / Constants.SharedDirectoryName;
             }
 
-            var cacheRootString = cacheRoot.Path.ToUpperInvariant();
+            var cacheRootString = cacheRoot.Path.PathToUpperInvariant();
```
SergeyTeplyakov: I'm curious, why not to add a similar extension method to AbsolutePath itself, instead of adding yet another extension method to string?

I think it'll look better and the solution won't pollute an intellisense:

`cacheRoot.ToUpperInvariant()`?

Public/Src/Cache/ContentStore/Library/Extensions/StringExtensions.cs
```diff
@@ -57,5 +58,18 @@ public static string ReplaceIgnoreCase(this string str, string oldValue, string
 
             return str;
         }
+
+        /// <summary>
+        /// Returns the upper invariant when paths are case-insensitive
+        /// </summary>
+        public static string PathToUpperInvariant(this string str)
```
SergeyTeplyakov: As I mentioned already, I would suggest to add an extension method to AbsolutePath or PathBase instead.

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
dcwuser: Why not an actual method? Adding methods isn't breaking.

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
kewilli: That's great! I've updated the PR based on this.

---
In reply to: [281299637](https://github.com/Microsoft/BuildXL/pull/253#discussion_r281299637) [](ancestors = 281299637)

Public/Src/Cache/ContentStore/Library/Extensions/StringExtensions.cs
kewilli: Removed!

---
In reply to: [281329075](https://github.com/Microsoft/BuildXL/pull/253#discussion_r281329075) [](ancestors = 281329075)

Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcFileCopier.cs
```diff
@@ -76,13 +76,13 @@ public async Task<CopyFileResult> CopyFileAsync(AbsolutePath sourcePath, Absolut
...
             var hashLiteral = segments.Last();
-            if (hashLiteral.EndsWith(GrpcDistributedPathTransformer.BlobFileExtension))
+            if (hashLiteral.EndsWith(GrpcDistributedPathTransformer.BlobFileExtension, StringComparison.OrdinalIgnoreCase))
             {
                 hashLiteral = hashLiteral.Substring(0, hashLiteral.Length - GrpcDistributedPathTransformer.BlobFileExtension.Length);
             }
             var hashTypeLiteral = segments.ElementAt(segments.Count - 1 - 2);
 
-            if (!Enum.TryParse(hashTypeLiteral, out HashType hashType))
+            if (!Enum.TryParse(hashTypeLiteral, true, out HashType hashType))
```
SergeyTeplyakov: nit: can you put a name to this argument, please?

PR #256, opened 2019-05-06
Remove unspecified TRP warnings for projects without ProjectReferences

jbayardo:
When an MSBuild project does not have any ProjectReferences, no warnings due to a missing Target Reference Protocol should be emitted.

### DIFF
--- a/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
+++ b/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
@@ -560,13 +560,25 @@ namespace BuildXL.FrontEnd.MsBuild
                     pipDataBuilder.Add(PipDataAtom.FromString($"/t:{target}"));
                 }
             }
-            else if (project.ProjectReferences.Count > 0)
+            else
             {
-                // The prediction for the targets to execute is not available. Just log this as a warning for now, defaults targets will be used.
-                Tracing.Logger.Log.ProjectIsNotSpecifyingTheProjectReferenceProtocol(
-                    m_context.LoggingContext,
-                    Location.FromFile(project.FullPath.ToString(PathTable)),
-                    project.FullPath.GetName(m_context.PathTable).ToString(m_context.StringTable));
+                // The prediction for the targets to execute is not available. We need to log this.
+                if (project.ProjectReferences.Count > 0)
+                {
+                    Tracing.Logger.Log.ProjectIsNotSpecifyingTheProjectReferenceProtocol(
+                        m_context.LoggingContext,
+                        Location.FromFile(project.FullPath.ToString(PathTable)),
+                        project.FullPath.GetName(m_context.PathTable).ToString(m_context.StringTable));
+                }
+                else
+                {
+                    // If the project has no references, then we log it as an informational message rather than a warning.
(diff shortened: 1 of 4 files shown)


smera:
@cdmihai is added to the review. #Closed

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -560,7 +560,7 @@ private bool TryAddMsBuildArguments(ProjectWithPredictions project, PipDataBuild
                     pipDataBuilder.Add(PipDataAtom.FromString($"/t:{target}"));
                 }
             }
-            else
+            else if (project.ProjectReferences.Count > 0)
```
smera: >if (project.ProjectReferences.Count > 0) [](start = 17, length = 40)

Not sure about this. Shouldn't all projects comply to the project reference protocol in this new world of static graph MSBuild?
I understand that when projects are leaves the protocol is inconsequential, but a project not specifying the protocol sounds fishy enough to me to do something about that (we warn today, but we may decide to error in the future). 
@micodoba@microsoft.com , thoughts?

jbayardo: Why is it fishy? In CloudBuild, for example, there are many leaf projects that are just moving files around, or strong-signing NuGet dependencies; moreover, there are roughly 40 of these.

In cases like this, warning about missing the TRP for leaves is just adding noise, as there may be many other projects which genuinely need attention. Furthermore, the warnings are misleading: there are no references, so what should the TRP be?

cdmihai: Is there an option to downgrade the warning to informational logs for leaf nodes?

jbayardo: I didn't see any way to downgrade the warning, so I just added a new informational event specific for the leaves.

smera: In an ideal world all SDKs implement the protocol. This protocol specification goes beyond the concrete references a node may have. So even if a node doesn't have any references, the protocol will say what happens if references are added.

But after discussing this with Mihai, he convinced me that in our current not-ideal world, it is ok to let that slide. So I'm good.

---
In reply to: [281338766](https://github.com/Microsoft/BuildXL/pull/256#discussion_r281338766) [](ancestors = 281338766)

smera:
>             EventLevel = Level.Warning,

Talking with Mihai about this, we also agreed that this should probably be turned into an error. Things are almost bound to fail anyway, and it is better if we start with a more strict set of constraints and relax them later than the other way around.
@julian@bayardo.info , maybe you want to do that as port of this PR? Sounds close enough to the changes you are already proposing. 
Otherwise I'll send a separate PR with that change, just let me know.

---
Refers to: Public/Src/FrontEnd/MsBuild/Tracing/Log.cs:228 in 865749b. [](commit_id = 865749b0247731e1588bd6939eaea1e28781d7d4, deletion_comment = False)

jbayardo:
As we spoke, although we definitely should turn it into an error, that is not something we can do right now due to constraints from the CB repo, which is on-going work.

PR #264, opened 2019-05-07
Don't use external feeds for internal builds

dannyvv:


config.dsc
```diff
@@ -44,19 +44,19 @@ config({
                 hash: "17E8C8C0CDCCA3A6D1EE49836847148C4623ACEA5E6E36E10B691DA7FDC4C39200",
             },
 
-            repositories: (importFile(f`config.microsoftInternal.dsc`).isMicrosoftInternal
+            // Note: From a compliance point of view it is important that MicrosoftInternal has a single feed.
```
almiliMSFT: nit: moving this comment right below `"BuildXL.Selfhost"` might be more effective

dannyvv: good call! thanks

dannyvv: Note I'll wait 15 minutes to push that change so that VSO has time to propagate the upstream

---
In reply to: [281806682](https://github.com/microsoft/BuildXL/pull/264#discussion_r281806682) [](ancestors = 281806682)

PR #265, opened 2019-05-07
Adding qTestContextInfo for qtest to upload test result to VSTS

norahuang:
[AB#1526581](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1526581)


Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -2,7 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 import {Artifact, Cmd, Transformer, Tool} from "Sdk.Transformers";
-const root = Environment.hasVariable("[Sdk.BuildXL]qtestDeploymentPath") ? d`${Environment.getFileValue("[Sdk.BuildXL]qtestDeploymentPath")}` : d`.`;
+
+const root = Environment.hasVariable("QTEST_DEPLOYMENT_PATH") ? d`${Environment.getFileValue("QTEST_DEPLOYMENT_PATH")}` : d`.`;
```
dannyvv: >QTEST_DEPLOYMENT_PATH [](start = 38, length = 21)

Sad to see the pattern with module name grouping being broken, but I understand we can't control other environments and if those names have been used we have to handle it :(

norahuang: It is the environment variable passed from GBR. I actually decided to get rid of the group name since I wanted to follow the convention in GBR side as they set the environment variable. But for all other env var that passed down through /p: I tried my best to keep the [Sdk.BuildXL].

PR #270, opened 2019-05-08
Temporarily set Test.BuildXL.Utilities to requires admin for testing purpose

narasamdya:
I want to build the change in this draft PR in CB with my custom BuildXL.

Currently, the easiest way, IMO, to do so is to create a PR, and get the commit id, and submit a build to CB with that commit id.


jwdonahue:
Why would you want to run any part a build as admin?

narasamdya:
> Why would you want to run any part a build as admin?

We're transitioning to low-privilege build environment, but some of our tests require admin privilege.  Those tests will be marked as "require admin", and will be run in VM.

PR #271, opened 2019-05-08
Add multiple proto file support and a toggle for grpc to the proto/gr…

dannyvv:
…pc sdk

### DIFF
--- a/Public/Sdk/Public/Protocols/Grpc/protoc.dsc
+++ b/Public/Sdk/Public/Protocols/Grpc/protoc.dsc
@@ -23,45 +23,65 @@ export const tool: Transformer.ToolDefinition = {
  */
 @@public
 export function generate(args: Arguments) : Result {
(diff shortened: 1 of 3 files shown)


Public/Sdk/Public/Protocols/Grpc/protoc.dsc
```diff
@@ -24,44 +24,80 @@ export const tool: Transformer.ToolDefinition = {
...
+                result.getOutputFile(grpcCsFile)
+            ),
         ],
     };
 }
 
 @@public
 export interface Arguments extends Transformer.RunnerArguments{
-    proto: File,
+    proto: File | File[],
```
smera: Is this File | File[] actually needed? Just makes the logic more complicated. Can't we live with File[] only?

dannyvv: sure

PR #276, opened 2019-05-08
Preparation work for logging enhancements.

SergeyTeplyakov:
This is a prerequisite for the changes on CB side In order to enhance logging for CASaas.

### DIFF
--- a/Public/Src/Cache/ContentStore/Interfaces/Logging/IOperationLogger.cs
+++ b/Public/Src/Cache/ContentStore/Interfaces/Logging/IOperationLogger.cs
@@ -3,6 +3,12 @@
 
 using System;
 using System.Diagnostics.ContractsLight;
+using System.IO;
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Interfaces/Logging/IOperationLogger.cs
```diff
@@ -92,8 +103,27 @@ public enum OperationKind
         /// <nodoc />
         public Exception Exception { get; }
 
+        /// <summary>
+        /// Tracing severity of the result.
+        /// </summary>
+        public Severity Severity { get; }
```
dannyvv: Presuming this is for something in the future? I didn't see any usage of these two new fields.

SergeyTeplyakov: This will be used by CB.

PR #279, opened 2019-05-09
Allow injection of a shim process in lieu of all or some child processes

erikma:
For Windows only, allow Detouring of a top-level process to decide to run a shim program instead of any or all of its child processes, passing the original command line on the shim's command line, and implicitly passing the shim the current working directory and the process environment of the original child process.

This facilitates remoting execution of the child process to a different VM, machine, container, or cloud service - the AnyBuild prototype.

Managed code configuration can decide to shim all processes except an opt-out list (typical case: run all children of top-level Gulp process remotely except cmd.exe and Gulp children), or to shim specific child processes (typical case: run most processes locally but select cl.exe and link.exe to run remotely).

### DIFF
--- a/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
+++ b/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.ContractsLight;
+using BuildXL.Utilities;
+
+namespace BuildXL.Processes
+{
+    /// <summary>
+    /// Configuration for shim execution in place of child processes of build engines
+    /// executing in an AnyBuild context.
+    /// </summary>
+    public sealed class AnyBuildShimInfo
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        public AnyBuildShimInfo(AbsolutePath substituteProcessExecutionShimPath)
+        {
+            Contract.Requires(substituteProcessExecutionShimPath.IsValid);
+            SubstituteProcessExecutionShimPath = substituteProcessExecutionShimPath;
+        }
+        
+        /// <summary>
+        /// Children of the Detoured process are not directly executed,
+        /// but instead this substitute shim process is injected instead, with
+        /// the original process's command line appended.
+        /// </summary>
+        public AbsolutePath SubstituteProcessExecutionShimPath { get; set; }
+    }
+}
--- a/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
+++ b/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
@@ -1,24 +1,27 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
 using BuildXL.Utilities;
 
 namespace BuildXL.Processes
 {
     /// <summary>
-    /// Configuration for shim execution in place of child processes of build engines
-    /// executing in an AnyBuild context.
+    /// Configuration for shim execution in place of child processes executing in a build sandbox.
     /// </summary>
     public sealed class AnyBuildShimInfo
     {
         /// <summary>
         /// Constructor.
         /// </summary>
-        public AnyBuildShimInfo(AbsolutePath substituteProcessExecutionShimPath)
+        public AnyBuildShimInfo(AbsolutePath substituteProcessExecutionShimPath, bool shimAllProcesses, IReadOnlyCollection<ShimProcessMatch> processMatches)
         {
             Contract.Requires(substituteProcessExecutionShimPath.IsValid);
+
             SubstituteProcessExecutionShimPath = substituteProcessExecutionShimPath;
+            ShimAllProcesses = shimAllProcesses;
+            ShimProcessMatches = processMatches;
         }
         
         /// <summary>
@@ -27,5 +30,48 @@ namespace BuildXL.Processes
         /// the original process's command line appended.
         /// </summary>
         public AbsolutePath SubstituteProcessExecutionShimPath { get; set; }
+
+        /// <summary>
+        /// Specifies the shim injection mode. When true, <see cref="ShimProcessMatches"/>
+        /// specifies the processes that should not be shimmed. When false, ShimProcessMatches
+        /// specifies the processes that should be shimmed.
+        /// </summary>
+        public bool ShimAllProcesses { get; }
+
+        /// <summary>
+        /// Processes that should or should not be be shimmed, according to the shim injection mode specified
+        /// by <see cref="ShimAllProcesses"/>.
+        /// </summary>
+        public IReadOnlyCollection<ShimProcessMatch> ShimProcessMatches { get; }
+    }
+
+    /// <summary>
+    /// Process matching information used for including or excluding processes in <see cref="AnyBuildShimInfo"/>.
+    /// </summary>
+    /// <remarks>In unmanaged code this is decoded into class ShimProcessMatch in DetoursHelpers.cpp.</remarks>
+    public sealed class ShimProcessMatch
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        public ShimProcessMatch(PathAtom processName, PathAtom argumentMatch)
+        {
+            Contract.Requires(processName.IsValid);
+
+            ProcessName = processName;
+            ArgumentMatch = argumentMatch;
+        }
+
+        /// <summary>
+        /// A process name to match, including extension. E.g. "MSBuild.exe".
+        /// </summary>
+        public PathAtom ProcessName { get; }
+
+        /// <summary>
+        /// An optional string to match in the argument of the process to further refine the match.
+        /// Can be used for example with ProcessName="node.exe" ArgumentMatch="gulp.js" to match the
+        /// Gulp build engine process.
+        /// </summary>
+        public PathAtom ArgumentMatch { get; }
     }
 }
--- a/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
+++ b/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
@@ -48,7 +48,7 @@ namespace BuildXL.Processes
     /// <summary>
     /// Process matching information used for including or excluding processes in <see cref="AnyBuildShimInfo"/>.
     /// </summary>
-    /// <remarks>In unmanaged code this is decoded into class ShimProcessMatch in DetoursHelpers.cpp.</remarks>
+    /// <remarks>In unmanaged code this is decoded into class ShimProcessMatchInfo in DetoursHelpers.cpp.</remarks>
     public sealed class ShimProcessMatch
     {
         /// <summary>
--- a/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
+++ b/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
@@ -48,7 +48,7 @@ namespace BuildXL.Processes
     /// <summary>
     /// Process matching information used for including or excluding processes in <see cref="AnyBuildShimInfo"/>.
     /// </summary>
-    /// <remarks>In unmanaged code this is decoded into class ShimProcessMatchInfo in DetoursHelpers.cpp.</remarks>
+    /// <remarks>In unmanaged code this is decoded into class ShimProcessMatch in DetoursHelpers.cpp.</remarks>
     public sealed class ShimProcessMatch
     {
         /// <summary>
--- a/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
+++ b/Public/Src/Engine/Processes/AnyBuildShimInfo.cs
@@ -18,6 +18,7 @@ namespace BuildXL.Processes
         public AnyBuildShimInfo(AbsolutePath substituteProcessExecutionShimPath, bool shimAllProcesses, IReadOnlyCollection<ShimProcessMatch> processMatches)
         {
             Contract.Requires(substituteProcessExecutionShimPath.IsValid);
+            Contract.Requires(processMatches != null);
 
             SubstituteProcessExecutionShimPath = substituteProcessExecutionShimPath;
             ShimAllProcesses = shimAllProcesses;
--- a/Public/Src/Engine/Processes/FileAccessManifest.cs
+++ b/Public/Src/Engine/Processes/FileAccessManifest.cs
@@ -468,13 +468,15 @@ namespace BuildXL.Processes
 
             if (SubstituteProcessExecutionInfo == null)
             {
-                // Emit just a zero-length substituteProcessExecShimPath when substitution is turned off.
+                writer.Write(0);  // ShimAllProcesses false value.
+
+                // Emit a zero-length substituteProcessExecShimPath when substitution is turned off.
                 WriteChars(writer, null);
                 return;
             }
 
-            WriteChars(writer, SubstituteProcessExecutionInfo.SubstituteProcessExecutionShimPath.ToString(m_pathTable));
             writer.Write(SubstituteProcessExecutionInfo.ShimAllProcesses ? (uint)1 : (uint)0);
+            WriteChars(writer, SubstituteProcessExecutionInfo.SubstituteProcessExecutionShimPath.ToString(m_pathTable));
             writer.Write((uint)SubstituteProcessExecutionInfo.ShimProcessMatches.Count);
 
             if (SubstituteProcessExecutionInfo.ShimProcessMatches.Count > 0)
--- a/Public/Src/Engine/Processes/FileAccessManifest.cs
+++ b/Public/Src/Engine/Processes/FileAccessManifest.cs
@@ -467,10 +467,10 @@ namespace BuildXL.Processes
             string substituteProcessExecShimPath;
             uint shimAllProcessesFlagValue;
             uint numShimProcessMatches;
-            if (AnyBuildShimInfo != null && AnyBuildShimInfo.shimAllProcesses)
+            if (AnyBuildShimInfo != null)
             {
                 substituteProcessExecShimPath = AnyBuildShimInfo.SubstituteProcessExecutionShimPath.ToString(m_pathTable);
-                shimAllProcessesFlagValue = 1;
+                shimAllProcessesFlagValue = AnyBuildShimInfo.ShimAllProcesses ? 1 : 0;
                 numShimProcessMatches = (uint)AnyBuildShimInfo.ShimProcessMatches.Count;
             }
             else
--- a/Public/Src/Engine/Processes/FileAccessManifest.cs
+++ b/Public/Src/Engine/Processes/FileAccessManifest.cs
@@ -465,6 +465,20 @@ namespace BuildXL.Processes
 #endif
             WriteChars(writer,
                 AnyBuildShimInfo?.SubstituteProcessExecutionShimPath.ToString(m_pathTable) ?? null);
+
+            writer.Write((AnyBuildShimInfo?.ShimAllProcesses ?? false) ? 1 : 0);
+
+            uint numShimProcessMatches = (uint)(AnyBuildShimInfo?.ShimProcessMatches.Count ?? 0);
+            writer.Write(numShimProcessMatches);
+
+            if (numShimProcessMatches > 0)
+            {
+                foreach (ShimProcessMatch match in AnyBuildShimInfo.ShimProcessMatches)
+                {
+                    WriteChars(writer, match.ProcessName.ToString(m_pathTable.StringTable));
+                    WriteChars(writer, match.ArgumentMatch.IsValid ? match.ArgumentMatch.ToString(m_pathTable.StringTable) : null);
+                }
+            }
         }
         
         // See unmanaged decoder at DetoursHelpers.cpp :: CreateStringFromWriteChars()
--- a/Public/Src/Engine/Processes/FileAccessManifest.cs
+++ b/Public/Src/Engine/Processes/FileAccessManifest.cs
@@ -422,13 +422,13 @@ namespace BuildXL.Processes
         }
 
         /// <summary>
-        /// Optional AnyBuild shim execution information.
+        /// Optional child substitute shim execution information.
         /// </summary>
         /// <remarks>
         /// Internal since this is set as a side effect of initializing a FileAccessManifest
         /// in <see cref="SandboxedProcessInfo"/>.
         /// </remarks>
-        internal AnyBuildShimInfo AnyBuildShimInfo { get; set; }
+        internal SubstituteProcessExecutionInfo SubstituteProcessExecutionInfo { get; set; }
 
         /// <summary>
         /// Adds a policy to an entire scope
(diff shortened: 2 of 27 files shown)


Public/Src/Engine/Processes/AnyBuildShimInfo.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    /// <summary>
+    /// Configuration for shim execution in place of child processes executing in a build sandbox.
+    /// </summary>
+    public sealed class AnyBuildShimInfo
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        public AnyBuildShimInfo(AbsolutePath substituteProcessExecutionShimPath, bool shimAllProcesses, IReadOnlyCollection<ShimProcessMatch> processMatches)
```
narasamdya: Do you allow `processMatches` to be null?

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -448,7 +457,31 @@ public void AddPath(AbsolutePath path, FileAccessPolicy mask, FileAccessPolicy v
 
             m_rootNode.AddNodeWithScope(this, path, new FileAccessScope(mask, values), expectedUsn ?? ReportedFileAccess.NoUsn);
         }
+
+        private void WriteAnyBuildShimBlock(BinaryWriter writer)
+        {
+#if DEBUG
+            writer.Write(0xABCDEF04); // "ABCDEF04"
+#endif
+            WriteChars(writer,
```
narasamdya: >W [](start = 12, length = 1)

Why not have the following condition?

```
if (AnyBuildShimInfo != null) { writer.Write(1); /* Do the stuff you have there */ }
else { writer.Write(0); }
```

At least it reduces the occurrences of `?.` and `??` operators.

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -101,11 +102,17 @@ public sealed class SandboxedProcessInfo
             bool testRetries = false,
             LoggingContext loggingContext = null,
             IDetoursEventListener detoursEventListener = null,
-            IKextConnection sandboxedKextConnection = null)
+            IKextConnection sandboxedKextConnection = null,
+            AnyBuildShimInfo shimInfo = null)
```
narasamdya: >AnyBuildShimInfo shimInfo = null [](start = 12, length = 32)

Why do we need to pass `shimInfo` to the constructor? I don't see it being used in this class besides assigning it to `fileAccessManifest`, which I think is better to do that outside of this class.

Public/Src/Sandbox/Windows/DetoursServices/DetoursServices.h
```diff
@@ -129,3 +129,29 @@ class TranslatePathTuple
         return fromPath;
     }
 };
+
+// CODESYNC: AnyBuildShimInfo.cs :: ShimProcessMatch class
+class ShimProcessMatch
+{
+public:
+    std::unique_ptr<wchar_t> ProcessName;
```
narasamdya: >std::unique_ptr<wchar_t> ProcessName; [](start = 4, length = 37)

Any reason why not just use std::wstring ?

Public/Src/Sandbox/Windows/DetoursServices/DetoursHelpers.cpp
```diff
@@ -687,33 +711,15 @@ bool ParseFileAccessManifest(
...
-        ZeroMemory((void*)g_internalDetoursErrorNotificationFile, sizeof(wchar_t) * (manifestInternalDetoursErrorNotificationFileSize + 1));
-        memcpy_s((void*)g_internalDetoursErrorNotificationFile, sizeof(wchar_t) * (manifestInternalDetoursErrorNotificationFileSize + 1), (wchar_t*)(&payloadBytes[offset]), sizeof(wchar_t) * manifestInternalDetoursErrorNotificationFileSize);
-    }
-    
-    if (g_internalDetoursErrorNotificationFile != nullptr && g_internalDetoursErrorNotificationFile[0] == L'\0')
-    {
-        manifestInternalDetoursErrorNotificationFileSize = 0;
-        delete[] g_internalDetoursErrorNotificationFile;
-        g_internalDetoursErrorNotificationFile = nullptr;
-    }
```
narasamdya: Thanks for refactoring this.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1778,6 +1778,235 @@ NTSTATUS NTAPI Detoured_ZwSetInformationFile(
...
+/// Table entries used below that match process names and optional argument matches
+/// to processes that should or should not be shimmed.
+typedef struct
+{
+    const wchar_t *const ProcessName;
+
+    // Optional, if present a case-sensitive string match is performed against the command arguments.
+    // When present, this is an AND operation, i.e. the ProcessName and this search must both be found.
+    const wchar_t *const ArgMatch;
+} ShimProcessMatchInfo;
```
narasamdya: > ShimProcessMatchInfo; [](start = 1, length = 22)

Does this get used somewhere?

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1778,6 +1778,235 @@ NTSTATUS NTAPI Detoured_ZwSetInformationFile(
...
+{
+    return std::wstring(trim_start(buffer), trim_end(buffer + len));
+}
+inline void trim_inplace(std::wstring& str)
+{
+    str.assign(trim_start(str.c_str()),
+        trim_end(str.c_str() + str.length()));
+}
+
+// Returns new value for lpCommandLine pointing to the remainder of the string
```
narasamdya: >// Returns  [](start = 0, length = 11)

I feel that this shim logic needs to go to a separate class or util class.

Public/Src/Engine/Processes/AnyBuildShimInfo.cs
erikma: No, will add an assert on that

---
In reply to: [282685088](https://github.com/microsoft/BuildXL/pull/279#discussion_r282685088) [](ancestors = 282685088)

narasamdya:
Can we have a (unit) test on this shim logic with a dummy test shim program?

Public/Src/Sandbox/Windows/DetoursServices/DetoursHelpers.cpp
erikma: yw. I'll follow up after this PR with a refactor to convert this code to be more like the MacOS code, using a byte *cursor and simplify a lot of this code. Didn't want to mingle two major changes.

---
In reply to: [282689460](https://github.com/microsoft/BuildXL/pull/279#discussion_r282689460) [](ancestors = 282689460)

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: Nope, left over from refactoring/reimplementing. Removing.

---
In reply to: [282692819](https://github.com/microsoft/BuildXL/pull/279#discussion_r282692819) [](ancestors = 282692819)

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
erikma: Purely convenience - the calling code in another repo looks like:

            var shimInfo = new AnyBuildShimInfo(AbsolutePath.Create(_pathTable, shimPath));

            var sandboxedProcessInfo = new SandboxedProcessInfo(
                _pathTable,
                new LocalSandboxedFileStorage(),
                executablePath,
                disableConHostSharing: true,
                loggingContext: loggingContext,
                shimInfo: shimInfo)

The ctor being used implicitly creates a FileManifest, so wiring up this way allowed that shorter constructor to work properly.
I can remove this approach... what is your preference?


---
In reply to: [282687412](https://github.com/microsoft/BuildXL/pull/279#discussion_r282687412) [](ancestors = 282687412)

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -421,6 +421,15 @@ public void UnsetMessageCountSemaphore()
...
         }
 
+        /// <summary>
+        /// Optional AnyBuild shim execution information.
+        /// </summary>
+        /// <remarks>
+        /// Internal since this is set as a side effect of initializing a FileAccessManifest
+        /// in <see cref="SandboxedProcessInfo"/>.
+        /// </remarks>
+        internal AnyBuildShimInfo AnyBuildShimInfo { get; set; }
```
dannyvv: `[CanBeNull]` attribute untill we use the C#'s [https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/) feature

dannyvv: >AnyBuildS [](start = 17, length = 9)

Isn't this more generically applicable? Should we brand it?

dannyvv: Espcially since Cpp side uses a different naming: `ShimProcess`

---
In reply to: [282950720](https://github.com/microsoft/BuildXL/pull/279#discussion_r282950720) [](ancestors = 282950720)

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -448,7 +457,44 @@ public void AddPath(AbsolutePath path, FileAccessPolicy mask, FileAccessPolicy v
...
+            {
+                substituteProcessExecShimPath = null;
+                shimAllProcessesFlagValue = 0;
+                numShimProcessMatches = 0;
+            }
+
+            WriteChars(writer, substituteProcessExecShimPath);
+            writer.Write(shimAllProcessesFlagValue);
+            writer.Write(numShimProcessMatches);
+
```
dannyvv: Shouldn't we just emit a single bit to indicate if we are going to write this tripple? (or overload null for shimPath and then skip the extra fields? Now every single process that ever gets executed under our detours is duplicating this extra bytes when there are no shims...

dannyvv:
Note: @almili Doesn't the mac code rely on the same binary manifest?

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1778,6 +1778,224 @@ NTSTATUS NTAPI Detoured_ZwSetInformationFile(
...
+    return rv;
+}
+
+// https://stackoverflow.com/questions/216823/whats-the-best-way-to-trim-stdstring
+inline const wchar_t *trim_start(const wchar_t *str)
+{
+    while (wmemchr(L" \t\n\r", *str, 4))  ++str;
+    return str;
+}
+inline const wchar_t *trim_end(const wchar_t *end)
```
dannyvv: whileline spacing between the methods

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1791,6 +2009,40 @@ BOOL WINAPI Detoured_CreateProcessW(
...
     _Out_       LPPROCESS_INFORMATION lpProcessInformation)
 {
+    if (g_substituteProcessExecutionShimPath != nullptr)
+    {
+        wstring command;
+        wstring commandArgs;
+        const wchar_t *lpCommandArgs;
+        if (lpApplicationName == nullptr)
+        {
+            FindApplicationNameFromCommandLine(lpCommandLine, command, commandArgs);
```
dannyvv: Maybe this is new Cpp but it reads werid to me that the out parameter are not marked with `&`

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1778,6 +1778,224 @@ NTSTATUS NTAPI Detoured_ZwSetInformationFile(
...
+}
+inline void trim_inplace(std::wstring& str)
+{
+    str.assign(trim_start(str.c_str()),
+        trim_end(str.c_str() + str.length()));
+}
+
+// Returns new value for lpCommandLine pointing to the remainder of the string
+// to use as the command line parameters.
+static const void FindApplicationNameFromCommandLine(const wchar_t *lpCommandLine, _Out_ wstring &command, _Out_ wstring &commandArgs)
```
dannyvv: >FindApplicationNameFromCommandLine [](start = 18, length = 34)

Shouldn't we use the CRT library for this?

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1778,6 +1778,224 @@ NTSTATUS NTAPI Detoured_ZwSetInformationFile(
         FileInformationClass);
 }
 
+/// Runs an injected AnyBuild shim instead of a child process, passing the
+/// original command and arguments to the shim along with, implicitly,
```
dannyvv: Maybe put these functions in a seperate file?

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1778,6 +1778,224 @@ NTSTATUS NTAPI Detoured_ZwSetInformationFile(
...
+    {
+        Dbg(L"Failure running AnyBuild shim process - failed to allocate buffer of size %d.", fullCmdLineSizeInChars * sizeof(WCHAR));
+        SetLastError(ERROR_OUTOFMEMORY);
+        return FALSE;
+    }
+
+    fullCommandLine[0] = L'\0';
+    if (lpApplicationName != nullptr)
+    {
+        fullCommandLine[0] = L'"';
```
dannyvv: >ne[0] = L [](start = 21, length = 9)

shouldn't we only do this if there is a space?
Some tools actually print the entire commandline and now this is changing behavior

dannyvv: The more I think about it is that it seems the shim tool will need separation between the cmd and args fields and don't try to reconstrcuct a full commandline. You need it for the matching but when executing it would probably be safest to maintain compability

---
In reply to: [282958577](https://github.com/microsoft/BuildXL/pull/279#discussion_r282958577) [](ancestors = 282958577)

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1778,6 +1778,224 @@ NTSTATUS NTAPI Detoured_ZwSetInformationFile(
...
+        return true;
+    }
+
+    return wcsstr(commandArgs, argMatch) != nullptr;
+}
+
+static bool ShouldShim(const wstring &command, const wchar_t *commandArgs)
+{
+    // Easy cases.
+    if (g_pShimProcessMatches == nullptr || g_pShimProcessMatches->empty())
```
dannyvv: >himProcessMatches == nullptr || g_pShimProcessMatch [](start = 12, length = 51)

Shouldn't this be an assert case since this method only gets called when `g_substituteProcessExecutionShimPath` is not null. Why would the other side set g_substituteProcessExecutionShimPath and not the others, that should be an error case there too

dannyvv: Ah you are overloading null with wildcards.

---
In reply to: [282960153](https://github.com/microsoft/BuildXL/pull/279#discussion_r282960153) [](ancestors = 282960153)

dannyvv: No, this is still weird. 
If nothing is to be shimmed, it shouldn't set `substituteProcessExecutionShimPath`
if everything is to be shimmed: `g_ProcessExecutionShimAllProcesses` should be set to true.

Else it should be an error if there are zero matches.


---
In reply to: [282960313](https://github.com/microsoft/BuildXL/pull/279#discussion_r282960313) [](ancestors = 282960313,282960153)

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1778,6 +1778,224 @@ NTSTATUS NTAPI Detoured_ZwSetInformationFile(
...
+    bool foundMatch = false;
+
+    for (auto it = g_pShimProcessMatches->begin(); it != g_pShimProcessMatches->end(); ++it)
+    {
+        ShimProcessMatch *pMatch = *it;
+
+        const wchar_t *processName = pMatch->ProcessName.get();
+        size_t processLen = wcslen(processName);
+
+        // lpAppName is longer than e.g. "cmd.exe", see if lpAppName ends with e.g. "\cmd.exe"
```
dannyvv: >\cmd.exe [](start = 85, length = 8)

Just checking if the PATHEXT logic happens in the shell or if it also happens at this layer. if so you probably need to do more matching as `cmd` would look in order or `pathext` environment for existing files.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: Added new SubstituteProcessExecution.cpp, .h

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: Added new SubstituteProcessExecution.cpp, .h

Public/Src/Sandbox/Windows/DetoursServices/DetoursServices.h
erikma: Moved to wstring

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
narasamdya: My preference is not to modify `FileAccessManifest` inside `SandboxedProcessInfo`. For the case you mentioned above, you can always modify it to
```
          var shimInfo = new AnyBuildShimInfo(AbsolutePath.Create(_pathTable, shimPath));
          var fam = new FileAccessManifest(pathTable);
          fam.AnyBuildShimInfo = shimInfo;
           var sandboxedProcessInfo = new SandboxedProcessInfo(
                _pathTable,
                new LocalSandboxedFileStorage(),
                executablePath,
                disableConHostSharing: true,
                loggingContext: loggingContext,
                shimInfo: shimInfo)
```

---
In reply to: [282945198](https://github.com/microsoft/BuildXL/pull/279#discussion_r282945198) [](ancestors = 282945198,282687412)

erikma: Removed.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: Is there a library/function specific for this? It would be the case for *nix but in Windows it's less clear. In this case I need to pull out the (optional path +)command to be able to run checks against its name for matching.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: The logic does in fact do the command pull-out for analysis but in the end when shimming is replaces the lpCommandName with the shim, and prepends the lpCommandLine with the original command (if not null) surrounded in quotes, to make it easier to parse in the shim.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: Needs unit testing BTW, as part of resolving Iman's comment about the need for a test for this logic.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: Added assert

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: The shim I have (not visible in this PR) actually has to run the PATHEXT logic itself since it's not using ShellExecute.

Public/Src/Engine/Processes/FileAccessManifest.cs
erikma: Updated to predicate the full body on whether the shim path is non-empty.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
dannyvv: Hmmm can't find it....

---
In reply to: [283048828](https://github.com/microsoft/BuildXL/pull/279#discussion_r283048828) [](ancestors = 283048828)

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -12,6 +12,7 @@
 using BuildXL.Native.IO;
 using BuildXL.Native.Processes;
 using BuildXL.Utilities;
+using JetBrains.Annotations;
 
 namespace BuildXL.Processes
```
dmgonch: test comment

Public/Src/Sandbox/Windows/DetoursServices/DataTypes.h
```diff
@@ -539,6 +539,27 @@ typedef struct ManifestDllBlock_t
 } ManifestDllBlock;
 typedef const ManifestDllBlock * PCManifestDllBlock;
 
+// ==========================================================================
+// == ManifestSubstituteProcessExecutionShim
+// ==========================================================================
+typedef struct ManifestSubstituteProcessExecutionShim_t
```
almiliMSFT: Is this going to be an empty struct when running in Release configuration?  If so, `sizeof(ManifestSubstituteProcessExecutionShim_t)` is technically undefined by the C language standard.  On Windows this seems to always evaluate to 0, in which case everything works just fine; on Mac, we've seen cases where `sizeof()` can be non-zero, in which case things will not work.

I see that all the validation passed for this PR.  Unfortunately though, I don't think we have anything that validates the release configuration on Mac.

erikma: Didn't know that. I can rearrange to place one of the optional int32 values into the struct itself with a comment that it's needed to anchor the release version of the struct at size > 0

almiliMSFT: I just tried it on my laptop, and I got errors:
```
[0:10] error DX0011: [Pip89B70B5516562FBA, bash, HelloWorld, scriptWithChildProcesses, {}] Process start failed with error code 80131500: [Pip-8523331309700501574 -- Pip89B70B5516562FBA, bash, HelloWorld, scriptWithChildProcesses, {}] Process creation failed: Failed to notify kernel extension about process start, make sure the extension is loaded
```
Please don't merge this yet.  I'll update our CI validation to include some validation for the release bits

almiliMSFT: Your suggestion makes sense.

erikma: Please pull latest commit with update

almiliMSFT: Works on my laptop!

I just updated CI validation, but feel free to merge without re-running Mac validation

erikma: Thanks for the extra validation, glad not to break Mac :)

almiliMSFT: Sorry, I was wrong: it only worked because I had all cache hits.

I'm still getting the same errors when it executes processes.

almiliMSFT: I updated CI validations, so if they pass you're good to go.

erikma: Managed code side of Detours logic was not aligned with the previous commit, new one should be good. Again thanks!

PR #259, opened 2019-05-07
Pass low privilege account and password to VM

norahuang:
This PR consists two items:
1) Pass low privilege account and password to VM. 
2) Enable propagation of privilege level from test sdk to transformer.execute.

PR #281, opened 2019-05-09
Nuget Analyzer + Spec Generator Refactoring

Kr1sso:
Since we've introduced `.NETCore` and `.NETStandard` support a while back, most of the functionality for all new target framework monikers (tfm) has been patched on top of our existing full framework structure. 

This PR refactors the Nuget analyze and spec generation steps to support all tfms from `.NET 1.0` to `.NETCore 3.0` and even the newer `.NETStandard2.1` versions. Full framework specs are clearly separated from .NETCore ones - this was a conscious decision as we are phasing out full framework support and won't add new dependencies from this point on. BuildXL will be transitioning to using `.NETCoreApp3.0` on all platforms soon.


Private/InternalSdk/VstsArtifacts/Microsoft.VisualStudio.Services.InteractiveClient.dsc
```diff
@@ -16,7 +16,7 @@ namespace Contents {
 
 @@public
 export const pkg: Managed.ManagedNugetPackage =
-    Managed.Factory.createNugetPackge(
+    Managed.Factory.createNugetPackage(
```
almiliMSFT: missed opportunity for a separate PR :)

Public/Sdk/Public/Deployment/operations.dsc
```diff
@@ -114,16 +114,14 @@ function flattenItem(
...
         return flattenStaticDirectory(item, targetFolder, handleDuplicateFile, currentResult, provenance);
-    } 
+    }
 
     if (isFile(item)) {
         return flattenFile(item, targetFolder.combine(item.name), handleDuplicateFile, currentResult, provenance);
     }
 
     if (isDeployable(item)) {
-        // TODO: TEMPORARY HACK: ['deploy'] just a little hack until I get an LKG using the NugetSpecGenerator.cs (ie, next build)
```
almiliMSFT: I guess this "next build" took quite a while to produce :)

Public/Sdk/Public/Managed/Frameworks/netcoreapp2.2/netcoreapp2.2.dsc
```diff
@@ -16,8 +16,9 @@ function macOSRuntimeExtensions(file: File): boolean {
 
 function ignoredAssembly(file: File): boolean {
     let f = file.name;
-    return f.equals(a`System.Security.AccessControl.dll`)
-    || f.equals(a`System.Security.Principal.Windows.dll`);
+    // We skip deploying those files from the .NET Core package as we need those very assemblies from their dedicated package
+    // to compile our platform abstraction layer, which depends on datatypes present only in the dedicated packages
+    return f.equals(a`System.Security.AccessControl.dll`) || f.equals(a`System.Security.Principal.Windows.dll`);
```
almiliMSFT: Do you know why use `.equals()` here and not `===`?  I'd expect `===` to work just the same.

Can you also rename `f` to `fileName`?

Public/Src/FrontEnd/Nuget/NugetFrameworkMonikers.cs
```diff
@@ -111,65 +120,46 @@ public NugetFrameworkMonikers(StringTable stringTable)
...
+            Net451 = Register(stringTable, "net451", ".NETFramework4.5.1", ref FullFrameworkVersionHistory);
+            Net452 = Register(stringTable, "net452", ".NETFramework4.5.2", ref FullFrameworkVersionHistory);
+            Net46  = Register(stringTable, "net46",  ".NETFramework4.6", ref FullFrameworkVersionHistory);
+            Net461 = Register(stringTable, "net461", ".NETFramework4.6.1", ref FullFrameworkVersionHistory);
+            Net462 = Register(stringTable, "net462", ".NETFramework4.6.2", ref FullFrameworkVersionHistory);
+            Net472 = Register(stringTable, "net472", ".NETFramework4.7.2", ref FullFrameworkVersionHistory);
         }
 
-        private PathAtom Register(StringTable stringTable, string smallMoniker, string largeMoniker)
+        private PathAtom Register(StringTable stringTable, string smallMoniker, string largeMoniker, ref List<PathAtom> versions)
```
almiliMSFT: You don't need `ref` here.  I kind of like that you did use `ref` because it draws attention to this list and makes it clear that it will be mutated; but on the other hand, it's not very idiomatic to pass a list by reference when you just want to add elements to the list.  I don't mind it either way.

Public/Sdk/Public/Deployment/operations.dsc
dannyvv: yeah, I dropped the ball and forgot about it. Thanks for catching & fixing Kristijan!

---
In reply to: [282554031](https://github.com/microsoft/BuildXL/pull/281#discussion_r282554031) [](ancestors = 282554031)

Public/Sdk/Public/Managed/Tools/ResGen.Lite/resgen.lite.dsc
```diff
@@ -23,17 +23,17 @@ namespace Tool {
                 importFrom("System.Threading.Tasks.Extensions").pkg,
 
                 ] : []),
-            importFrom("Microsoft.CodeAnalysis.Common").pkg,
-            importFrom("Microsoft.CodeAnalysis.CSharp").pkg,
-            importFrom("Microsoft.CodeAnalysis.CSharp.Workspaces").pkg,
-            importFrom("Microsoft.CodeAnalysis.Workspaces.Common").pkg,
+            importFrom("Microsoft.CodeAnalysis.Common").withQualifier({ targetFramework: "netstandard2.0" }).pkg,
```
dannyvv: >ier({ targetFramework [](start = 66, length = 21)

Just curious why we are forcing netstandard here rather than let nuget decide the 'best' one...
Would you mind adding a little comment why?

Public/Sdk/Public/Managed/Tools/ResGen.Lite/resgen.lite.dsc
```diff
@@ -23,17 +23,17 @@ namespace Tool {
...
-            importFrom("Microsoft.CodeAnalysis.Common").pkg,
-            importFrom("Microsoft.CodeAnalysis.CSharp").pkg,
-            importFrom("Microsoft.CodeAnalysis.CSharp.Workspaces").pkg,
-            importFrom("Microsoft.CodeAnalysis.Workspaces.Common").pkg,
+            importFrom("Microsoft.CodeAnalysis.Common").withQualifier({ targetFramework: "netstandard2.0" }).pkg,
+            importFrom("Microsoft.CodeAnalysis.CSharp").withQualifier({ targetFramework: "netstandard2.0" }).pkg,
+            importFrom("Microsoft.CodeAnalysis.CSharp.Workspaces").withQualifier({ targetFramework: "netstandard2.0" }).pkg,
+            importFrom("Microsoft.CodeAnalysis.Workspaces.Common").withQualifier({ targetFramework: "netstandard2.0" }).pkg,
 
             // There is a gap in our nuget integration so temporarily adding the dependencies manually
```
dannyvv: >e is a gap in our nuget integration so temporarily ad [](start = 19, length = 53)

This comment is actually wrong :)
By design we do NOT pass the transitive nuget references where msbuild does do the transitive. In my view those are training wheels that work fine for small codebases but in larger codebases wiht say > 100 projects thins bloat and escalate fast. It also means that a nuget packages dependencies are part of their public signature so a high chance changes here will be breaking changes. Asking consumers to manually specify the dependencies gives a more robust experience with less version issues and less breaks

I agree that for some packages it would be super useful to have the compiler take the entire transitive dependencies, but I was thinking of having a modifier for that on the reference. We just haven't gotten to that.

Public/Sdk/SelfHost/RuntimeContracts/RuntimeContracts.dsc
```diff
@@ -40,7 +40,7 @@ export function withRuntimeContracts(args: Managed.Arguments, contractsLevel?: C
     return args.merge<Managed.Arguments>({
         defineConstants: getContractsSymbols(contractsLevel || ContractLevel.full, isDebug),
         references: [
-            importFrom("RuntimeContracts").pkg
+            importFrom("RuntimeContracts").withQualifier({targetFramework: 'netstandard2.0'}).pkg
```
dannyvv: >").withQualifier({targetFramewor [](start = 40, length = 32)

same question... why force netstandard?

Public/Src/FrontEnd/Core/PackageHashFile.cs
```diff
@@ -21,17 +21,17 @@ namespace BuildXL.FrontEnd.Core
...
         // If the file format changes, we have to ignore the files on disk,
         // buf if the file format is the same and only generated specs format has changed,
         // then we can reuse files from disk and force specs regeneration.
 
-        // The file format change will force specs regeneration. 
+        // The file format change will force specs regeneration.
         // Change the version if the nuget spec generation has changed in a backward incompatible way.
         private const string HashFileFormatVersion = "8";
-        private const string GeneratedSpecsVersion = "13";
+        private const string GeneratedSpecsVersion = "14";
```
dannyvv: >"14" [](start = 53, length = 4)

Note we have a bug somehwere that in some cases it won't properly generate, you might want to send email to the team after this is in that if they see issues to run `rmdir /s/q out\frontend\nuget\specs`

Public/Src/FrontEnd/Nuget/NugetAnalyzedPackage.cs
```diff
@@ -532,44 +371,38 @@ private bool TryParseDependenciesFromNuSpec()
...
-                    }
+                        // If there is at least one valid dependency for a known framework, then the package is defined as managed
+                        IsManagedPackage = true;
 
-                    // If the package has a pinned tfm and the groups tfm does not match, skip the groups dependency resolution
-                    if (!string.IsNullOrEmpty(this.Tfm) && NugetFrameworkMonikers.TargetFrameworkNameToMoniker.TryGetValue(this.Tfm, out Moniker pinnedTfm) && !PathAtom.Equals(pinnedTfm, targetFramework))
-                    {
-                        continue;
-                    }
+                        // Only add the group dependency target framework if the nuget package itself also contains specific assemblies of the same version
```
dannyvv: > group dependency t [](start = 39, length = 19)

Thanks, this is what Nora ran into with her PR:
[https://github.com/microsoft/BuildXL/pull/259/files](https://github.com/microsoft/BuildXL/pull/259/files) file `BuildXL.Utilities.dsc`

Public/Src/FrontEnd/Nuget/NugetSpecGenerator.cs
```diff
@@ -115,68 +119,80 @@ public ISourceFile CreatePackageConfig()
...
-                                PropertyAccess("Managed", "Factory", "createNugetPackge"),
-                                new LiteralExpression(analyzedPackage.Id),
-                                new LiteralExpression(analyzedPackage.Version),
-                                PropertyAccess("Contents", "all"),
-                                Array(compile),
-                                Array(runtime),
-                                Array(dependencies)
+                    cases.Add(
+                        new CaseClause(
+                            new LiteralExpression(monikers.Last().ToString(m_pathTable.StringTable)),
```
dannyvv: >Last [](start = 59, length = 4)

shouldn't this be first as well to be symetrical with the runtime and the dependencies?

dannyvv: >Last [](start = 59, length = 4)

It might be nicer to use 'first' here and when adding the other cases jsut do `.Skip(1)`

Public/Src/FrontEnd/Nuget/NugetSpecGenerator.cs
```diff
@@ -115,68 +119,80 @@ public ISourceFile CreatePackageConfig()
...
+                            new ReturnStatement(
+                                new CallExpression(
+                                    PropertyAccess("Managed", "Factory", "createNugetPackage"),
+                                    new LiteralExpression(analyzedPackage.Id),
+                                    new LiteralExpression(analyzedPackage.Version),
+                                    PropertyAccess("Contents", "all"),
+                                    Array(compile),
+                                    Array(runtime),
+                                    Array(new CallExpression(new Identifier("...addIfLazy"),
```
dannyvv: >("... [](start = 75, length = 5)

WOW, This shouldn't work... This technically should be an interpollationexpression, we might run into issues if we add more validation later since this is not a valid identifier

dannyvv: >("... [](start = 75, length = 5)

I fully don't understand why we generate the dependnecies only for the first one.
Is this to emulate that we are still generating the code wrong.
i.e. the compatiblity fallback logic should technically be done on each individual aspect:
* runtime
* compile
* dependencies

Public/Src/FrontEnd/Nuget/NugetSpecGenerator.cs
```diff
@@ -13,6 +13,7 @@
 using TypeScript.Net.Types;
 using static TypeScript.Net.DScript.SyntaxFactory;
 using static BuildXL.FrontEnd.Nuget.SyntaxFactoryEx;
+using System;
```
almiliMSFT: nit: move to the top

Public/Src/FrontEnd/Nuget/NugetSpecGenerator.cs
```diff
@@ -275,13 +280,46 @@ private IStatement CreatePackageContents()
             );
         }
 
+        internal static void FindAllCompatibleFrameworkMonikers(NugetAnalyzedPackage analyzedPackage, Action<List<PathAtom>> callback, params List<PathAtom>[] tfmHistory)
+        {
+            if (analyzedPackage.TargetFrameworks.Count > 0)
+            {
+                foreach (var versionHistory in tfmHistory)
+                {
+                    var indices = analyzedPackage.TargetFrameworks.Select(moniker => versionHistory.IndexOf(moniker)).Where(idx => idx != -1).OrderBy(x => x).ToList();
```
almiliMSFT: nit: can you put `.Select`, `Where` and `OrderBy` on separate lines

config.dsc
```diff
@@ -169,7 +169,6 @@ config({
                 { id: "Microsoft.Data.Edm", version: "5.8.2" },
 
                 // xUnit
-                { id: "xunit", version: "2.4.1" },
```
almiliMSFT: is this intentional?

Private/InternalSdk/VstsArtifacts/Microsoft.VisualStudio.Services.InteractiveClient.dsc
Kr1sso: 😄

Public/Sdk/Public/Managed/Frameworks/netcoreapp2.2/netcoreapp2.2.dsc
Kr1sso: I guess `.equals()` just because of jumping between DScript and C# - renamed it.

Public/Src/FrontEnd/Nuget/NugetFrameworkMonikers.cs
Kr1sso: Removed it.

Public/Sdk/Public/Managed/Tools/ResGen.Lite/resgen.lite.dsc
Kr1sso: With the change (as described in the PR introduction) we deliberately split full framework and `.NETCore` / `.NETStandard` assemblies in the spec generation. There is no more "we build for 472, so choose `.NETStandard` 1.0 - 2.0 assemblies". The CodeAnalysis packages all come with .NETStandard assemblies only, hence we have to force the qualifier here to use the .NETStandard assemblies from the package. I think it is cleaner this way as it communicates intent by the author of the DScript file and indicates what assemblies to use. Anther consideration is we are switching to .NETCoreApp only soon, at that point the qualifier enforcement can be removed all together. I'll leave a comment too.

Public/Sdk/Public/Managed/Tools/ResGen.Lite/resgen.lite.dsc
Kr1sso: Good thought, once we switch to Nuget assemblies and remove our hand written parser, this is a feature we should add!

Public/Sdk/SelfHost/RuntimeContracts/RuntimeContracts.dsc
Kr1sso: Same answer as above, package has `.NETStandard` assemblies only, hence wouldn't coerce to full framework qualifiers and we explicitly choose what to use.

Public/Src/FrontEnd/Core/PackageHashFile.cs
Kr1sso: Will do!

config.dsc
Kr1sso: It's intentional, as this is just a facade package that has no content and is not referenced by the transitive closure at all.

Public/Src/FrontEnd/Nuget/NugetSpecGenerator.cs
Kr1sso: We don't generate only for the first one, but for every assembly in the package up to all the versions higher (compatible) or the next assembly with a higher tfm version. Here is an example of the a generated spec:

```
@@public
export const pkg: Managed.ManagedNugetPackage = (() => {
    switch (qualifier.targetFramework) {
        case "net20":
        case "net35":
            return Managed.Factory.createNugetPackage(
                "System.Data.SQLite.Linq",
                "1.0.102.0",
                Contents.all,
                [Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net20/System.Data.SQLite.Linq.dll`)],
                [
                    Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net20/System.Data.SQLite.Linq.dll`),
                ],
                [...addIfLazy(qualifier.targetFramework === "net20", () => [])]
            );
        case "net40":
            return Managed.Factory.createNugetPackage(
                "System.Data.SQLite.Linq",
                "1.0.102.0",
                Contents.all,
                [Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net40/System.Data.SQLite.Linq.dll`)],
                [
                    Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net40/System.Data.SQLite.Linq.dll`),
                ],
                [...addIfLazy(qualifier.targetFramework === "net40", () => [])]
            );
        case "net45":
            return Managed.Factory.createNugetPackage(
                "System.Data.SQLite.Linq",
                "1.0.102.0",
                Contents.all,
                [Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net45/System.Data.SQLite.Linq.dll`)],
                [
                    Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net45/System.Data.SQLite.Linq.dll`),
                ],
                [...addIfLazy(qualifier.targetFramework === "net45", () => [])]
            );
        case "net451":
        case "net452":
            return Managed.Factory.createNugetPackage(
                "System.Data.SQLite.Linq",
                "1.0.102.0",
                Contents.all,
                [Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net451/System.Data.SQLite.Linq.dll`)],
                [
                    Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net451/System.Data.SQLite.Linq.dll`),
                ],
                [...addIfLazy(qualifier.targetFramework === "net451", () => [])]
            );
        case "net46":
        case "net461":
        case "net462":
        case "net472":
            return Managed.Factory.createNugetPackage(
                "System.Data.SQLite.Linq",
                "1.0.102.0",
                Contents.all,
                [Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net46/System.Data.SQLite.Linq.dll`)],
                [
                    Managed.Factory.createBinaryFromFiles(f`${packageRoot}/lib/net46/System.Data.SQLite.Linq.dll`),
                ],
                [...addIfLazy(qualifier.targetFramework === "net46", () => [])]
            );
        default:
            Contract.fail("Unsupported target framework");
    };
}
)();
```

PR #286, opened 2019-05-09
Add priority field to transformer dsc files.

erickulcyk:
Add priority field to transformer dsc files.  This is part of the change to support mularchy and was missed in the previous change

### DIFF
--- a/Public/Sdk/Public/Prelude/Prelude.OfficeBackCompatHacks.dsc
+++ b/Public/Sdk/Public/Prelude/Prelude.OfficeBackCompatHacks.dsc
@@ -132,6 +132,8 @@ namespace Transformer {
         isLight?: boolean;
         /** Obsolete: You must use import {Transformer} from "Sdk.Transformers" */
         keepOutputsWritable?: boolean;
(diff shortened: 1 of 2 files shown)


Public/Sdk/Public/Prelude/Prelude.OfficeBackCompatHacks.dsc
```diff
@@ -132,6 +132,8 @@ namespace Transformer {
         isLight?: boolean;
         /** Obsolete: You must use import {Transformer} from "Sdk.Transformers" */
         keepOutputsWritable?: boolean;
+        /** Obsolete: You must use import {Transformer} from "Sdk.Transformers" */
+        priority?: number;
```
dannyvv: > priority?: number; [](start = 7, length = 19)

Since you need ot update codegen to use it. Would you mind removing this and adding `import {Transformer} from "Sdk.Transformers"` at the top of the generated file?

erickulcyk: Which file do you mean is generated?  Are you saying somewhere in Office code we should add this?

PR #287, opened 2019-05-09
[macOS] Fix a crash that happens when a process times out

almiliMSFT:
[AB#1528677](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1528677)


Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -212,6 +212,12 @@ public virtual async Task<SandboxedProcessResult> GetResultAsync()
             SandboxedProcessReports reports = null;
 
             await m_processExecutor.WaitForExitAsync();
+            if (m_processExecutor.Killed)
+            {
+                // call here this.KillAsync() because a subclass may override it and a
```
almiliMSFT:
```suggestion
                // call here this.KillAsync() because a subclass may override it and the
```

Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -212,6 +212,12 @@ public virtual async Task<SandboxedProcessResult> GetResultAsync()
             SandboxedProcessReports reports = null;
 
             await m_processExecutor.WaitForExitAsync();
+            if (m_processExecutor.Killed)
+            {
+                // call here this.KillAsync() because a subclass may override it and the
```
almiliMSFT:
```suggestion
                // call here this.KillAsync() because a subclass may override it to
```

Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -212,6 +212,12 @@ public virtual async Task<SandboxedProcessResult> GetResultAsync()
             SandboxedProcessReports reports = null;
 
             await m_processExecutor.WaitForExitAsync();
+            if (m_processExecutor.Killed)
+            {
+                // call here this.KillAsync() because a subclass may override it and the
+                // subclass may want to do extra processing when a process is killed
```
almiliMSFT:
```suggestion
                // to do some extra processing when a process is killed
```

PR #288, opened 2019-05-09
Add an option to help support vsjit debugging with qtest

anirudhsanthiar:


Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -294,6 +295,8 @@ export interface QTestArguments extends Transformer.RunnerArguments {
     zipSandbox? : boolean;
     /** Privilege level required by this process to execute. */
     privilegeLevel?: "standard" | "admin";
+    /** Applicable only when QTestType is MsTest. When enabled, uses VsJitDebugger to launch a popup to automatically attach the debugger */
+    enableVsJitDebugger?: boolean;
```
norahuang: No need to add the argument here. It is not used anywhere. Your currently get the flag directly from environment variables but not passing this flag from build definition. Same comment apply to "qCodeCoverageEnumType" and "zipSandbox". If you decided to change to take the flag from build definition, you can add it with later.

PR #289, opened 2019-05-10
Add consistency check for cache

SergeyTeplyakov:
This is WIP code review with self check capabilities added to `FileSystemContentStoreInternal`.

What's left:
- tests
- configuration propagation.

[AB#1530246](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1530246)


Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+            var stopwatch = Stopwatch.StartNew();
+            var contentHashes = ReadContentHashesFromDisk(context);
+            _tracer.Debug(context, $"SelfCheck: Enumerated {contentHashes.Count} entries by {stopwatch.ElapsedMilliseconds}ms.");
+
+            stopwatch.Restart();
+
+            int invalidEntries = 0;
+
+            foreach (var (contentHash, fileInfo) in contentHashes)
+            {
```
ljcollins25: >{ [](start = 12, length = 1)

Maybe add a periodic log message for progress.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+            }
+
+            return reconcileTime.Value.IsRecent(Clock.UtcNow, _settings.SelfCheckFrequency);
+        }
+
+        private void MarkSelfChecked(bool success)
+        {
+            if (success)
+            {
+                FileSystem.WriteAllText(_selfCheckFilePath, $"{_settings.SelfCheckEpoch}|{Clock.UtcNow.ToReadableString()}");
```
ljcollins25: >WriteAllText [](start = 27, length = 12)

Include list of paths which were scanned to skip when resuming check.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -673,14 +867,14 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
...
                 FileSystem,
                 _tracer,
                 ShutdownStartedCancellationToken,
                 this,
                 quotaKeeperConfiguration);
 
-            var result = await _quotaKeeper.StartupAsync(context);
+            var result = await QuotaKeeper.StartupAsync(context);
 
             _taskTracker = new BackgroundTaskTracker(Component, new Context(context));
```
ljcollins25: >_taskTracker [](start = 12, length = 12)

Call SelfCheck as background task.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+                return false;
+            }
+
+            var reconcileTime = DateTimeUtilities.FromReadableTimestamp(parts[1]);
+            if (reconcileTime == null)
+            {
+                return false;
+            }
+
+            return reconcileTime.Value.IsRecent(Clock.UtcNow, _settings.SelfCheckFrequency);
```
ljcollins25: >reconcileTime.Value.IsRecent(Clock.UtcNow, _settings.SelfCheckFrequency); [](start = 19, length = 73)

Given that CASaaS is long lived, this will only make sense if SelfCheck is called inside a timer. Otherwise, the interval might be reached sometime after the call to SelfCheck.

ljcollins25:
Add to ContentStoreApp to allow local testing.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+
+        private async Task RemoveInvalidContentAsync(OperationContext context, ContentHash contentHash)
+        {
+            // In order to remove the content we have to do the following things:
+            // Remove file from disk
+            // Update memory content directory
+            // Update quota keeper
+            // Notify distributed store that the content is gone from this machine
+            // The first 3 things are happening in the first call.
+            await EvictAsync(
```
ljcollins25: Add and pass the force flag to evict the content even if it is pinned.

Public/Src/Cache/ContentStore/Library/Stores/ContentStoreSettings.cs
```diff
@@ -38,6 +40,16 @@ public sealed class ContentStoreSettings
         /// </summary>
         public bool StartPurgingAtStartup { get; set; } = true;
 
+        /// <summary>
```
ljcollins25: Add boolean flag as to whether SelfCheck is enabled at all.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+                var actualHashAndSize = await TryHashFileAsync(context, fileInfo.FullPath, expectedHash.HashType);
+                if (actualHashAndSize != null && actualHashAndSize.Value.Hash != expectedHash)
+                {
+                    // Don't need to add an expected hash into the error string because the client code will always put it into the final error message.
+                    return (isValid: false, error: $"Hash mismatch. Actual hash is {actualHashAndSize.Value.Hash}.");
+                }
+
+                return (isValid: true, error: string.Empty);
+            }
+            catch (IOException e)
```
ljcollins25: >IOException e [](start = 19, length = 13)

Catch all exceptions

ljcollins25:
>     public sealed class ContentStoreSettings

Add setting for maxconcurrency

---
Refers to: Public/Src/Cache/ContentStore/Library/Stores/ContentStoreSettings.cs:11 in 589dc1e. [](commit_id = 589dc1e35f4e78b69ffb2fc01550b50ba130c2f8, deletion_comment = False)

Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
```diff
@@ -16,7 +15,9 @@
 using BuildXL.Utilities.Tasks;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
+using DateTimeUtilities = BuildXL.Cache.ContentStore.Utils.DateTimeUtilities;
```
dcwuser: Why?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -28,6 +28,8 @@
 using BuildXL.Native.IO;
 using BuildXL.Utilities.Collections;
 using BuildXL.Utilities.Tracing;
+using DateTimeUtilities = BuildXL.Cache.ContentStore.Utils.DateTimeUtilities;
```
dcwuser: Why? Please no changes to using blocks of random files with no other changes. And in general let's only define namespace aliases if forced to by ambiguity. Is this one of the many bad things that re-sharper does?

Public/Src/Cache/ContentStore/Library/Stores/MemoryContentDirectory.cs
```diff
@@ -620,13 +620,11 @@ public Task SyncAsync()
...
         /// </summary>
-        public static async Task TransformFile
-            (
+        public static async Task TransformFile(
             Context context,
             IAbsFileSystem fileSystem,
             AbsolutePath directoryPath,
-            Func<KeyValuePair<ContentHash, ContentFileInfo>, KeyValuePair<ContentHash, ContentFileInfo>> transformer
-            )
+            Func<KeyValuePair<ContentHash, ContentFileInfo>, KeyValuePair<ContentHash, ContentFileInfo>> transformer)
```
dcwuser: Your new formatting is preferred and more consistent, but this change is unrelated to feature and adds a whole new file to PR.

Public/Src/Cache/ContentStore/Library/Stores/QuotaManagement/LegacyQuotaKeeper.cs
```diff
@@ -192,10 +192,11 @@ public override async Task<ReserveTransaction> ReserveAsync(long contentSize)
                 throw new CacheException($"Failed to reserve space for content size=[{contentSize}], result=[{result}]");
             }
 
-            return new ReserveTransaction(contentSize, Evicted);
+            return new ReserveTransaction(contentSize, OnContentEvicted);
         }
 
-        private void Evicted(long size)
+        /// <inheritdoc />
+        public override void OnContentEvicted(long size)
```
dcwuser: And here, pure name changing unrelated to feature adds an entire file to PR. Am I missing how this is related?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
SergeyTeplyakov: I moved some functionality from distributed layer into the library into the file with the same type.

I can move the entire DateTImeUtilities into LIbrary though. In this case the only change will be the namespace change.

---
In reply to: [283070517](https://github.com/microsoft/BuildXL/pull/289#discussion_r283070517) [](ancestors = 283070517)

Public/Src/Cache/ContentStore/Library/Stores/MemoryContentDirectory.cs
SergeyTeplyakov: will revert.

---
In reply to: [283070929](https://github.com/microsoft/BuildXL/pull/289#discussion_r283070929) [](ancestors = 283070929)

Public/Src/Cache/ContentStore/Library/Stores/QuotaManagement/LegacyQuotaKeeper.cs
SergeyTeplyakov: This is not "unrelated". I've made the method public and virtual because FileSystemContentSTore now can notify the keeper that the size has changed.

---
In reply to: [283071106](https://github.com/microsoft/BuildXL/pull/289#discussion_r283071106) [](ancestors = 283071106)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+                    _tracer.Warning(context, $"Found invalid entry in cache. Hash={contentHash}. {error}. Evicting the file...");
+                    await RemoveInvalidContentAsync(context, contentHash);
+                    invalidEntries++;
+                }
+            }
+
+            return Result.Success(new SelfCheckResult(invalidHashes: invalidEntries, totalFilesScanned: contentHashes.Count));
+        }
+
+        private async Task RemoveInvalidContentAsync(OperationContext context, ContentHash contentHash)
```
dcwuser: May not be easiest now, but shouldn't this be consolidated into a Delete method at the IContentStore level?

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+                (InvalidFiles, TotalFilesScanned) = (invalidHashes, totalFilesScanned);
+
+            /// <nodoc />
+            public static SelfCheckResult UpToDate() => new SelfCheckResult(invalidHashes: 0, totalFilesScanned: -1);
+
+            /// <inheritdoc />
+            public override string ToString() => $"InvalidHashes={InvalidFiles}, TotalFilesScanned={TotalFilesScanned}";
+        }
+
+        private async Task<Result<SelfCheckResult>> SelfCheckCoreAsync(OperationContext context)
```
dcwuser: Maybe choose method names that clearly indicate whether method act on one file, one cycle, etc.

dcwuser: More comments.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+        {
+            var stopwatch = Stopwatch.StartNew();
+            var contentHashes = ReadContentHashesFromDisk(context);
```
dcwuser: Prefer to use in-memory enumeration. Reduces I/O, and if a file isn't in the in-memory representation, it won't be pinned/placed anyway. Consider this issue of higher important than others.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -349,6 +354,195 @@ private Task Upgrade2To3(Context context)
...
+                var (isValid, error) = await ValidateFileAsync(context, contentHash, fileInfo);
+                if (!isValid)
+                {
+                    _tracer.Warning(context, $"Found invalid entry in cache. Hash={contentHash}. {error}. Evicting the file...");
+                    await RemoveInvalidContentAsync(context, contentHash);
+                    invalidEntries++;
+                }
+            }
+
+            return Result.Success(new SelfCheckResult(invalidHashes: invalidEntries, totalFilesScanned: contentHashes.Count));
```
dcwuser: contentHashes.Count is incorrect if we break from cancelation. Better to keep running total in local variable.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
dcwuser: Agree with Lance. Maybe I am missing reason for this approach, though. Why not just create repeating Timer with specified interval? Within callback, simply exit if last callback is still running as determined by Interlocked.CompareExchange. Also turn off when shutdown/canceled. This seems simpler, AND it handles unusual case of empty cache without entering a tight loop.

---
In reply to: [282938941](https://github.com/microsoft/BuildXL/pull/289#discussion_r282938941) [](ancestors = 282938941)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
SergeyTeplyakov: Renamed to SeflCheckStoreCoreAsync and SelfCheckStoreAsync respectively.

---
In reply to: [283072883](https://github.com/microsoft/BuildXL/pull/289#discussion_r283072883) [](ancestors = 283072883)

Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
SergeyTeplyakov: Because the helper's functionality has moved from one type to another.

---
In reply to: [283070386](https://github.com/microsoft/BuildXL/pull/289#discussion_r283070386) [](ancestors = 283070386)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
SergeyTeplyakov: As discussed, will only trace the last hash.

---
In reply to: [282936721](https://github.com/microsoft/BuildXL/pull/289#discussion_r282936721) [](ancestors = 282936721)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
SergeyTeplyakov: Let's discuss offline. Should be easier.

---
In reply to: [283074338](https://github.com/microsoft/BuildXL/pull/289#discussion_r283074338) [](ancestors = 283074338,282938941)

SergeyTeplyakov:
>     public sealed class ContentStoreSettings

I think this should be in a separate iteration.

---
In reply to: [491354901](https://github.com/microsoft/BuildXL/pull/289#issuecomment-491354901) [](ancestors = 491354901)

---
Refers to: Public/Src/Cache/ContentStore/Library/Stores/ContentStoreSettings.cs:11 in 589dc1e. [](commit_id = 589dc1e35f4e78b69ffb2fc01550b50ba130c2f8, deletion_comment = False)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
```diff
@@ -0,0 +1,465 @@
...
+                var hashInfo = contentHashes[index];
+                processedEntries++;
+
+                var (isValid, error) = await ValidateFileAsync(context, hashInfo.hash, hashInfo.fileInfo);
+                if (!isValid)
+                {
+                    _tracer.Warning(context, $"SelfCheck: Found invalid entry in cache. Hash={hashInfo.hash}. {error}. Evicting the file...");
+                    await _contentStoreInternal.RemoveInvalidContentAsync(context, hashInfo.hash);
+                    invalidEntries++;
+                }
```
ljcollins25: Do we not have to worry about cancellation exceptions? Can we wrap all this in a PerformOperationAsync (the entire for loop) so that we can capture exceptions (in particular cancellation) and still write the state.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
```diff
@@ -0,0 +1,465 @@
...
+
+                // Tracking the progress if needed.
+                traceProgressIfNeeded(hashInfo.hash);
+
```
ljcollins25: Only call this if the hash has changed from the last hash. Otherwise, we can save that we have processed the hash when we've only processed one of the replicas.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
```diff
@@ -0,0 +1,465 @@
...
+            /// </summary>
+            UpToDate,
+        }
+
+        /// <summary>
+        /// Self check state serialized/deserialized from disk.
+        /// </summary>
+        internal readonly struct SelfCheckState : IEquatable<SelfCheckState>
+        {
+            private const string EmptyHash = "NO_HASH";
```
ljcollins25: >EmptyHash [](start = 33, length = 9)

FinishedHash = "FINISHED_NOT_A_HASH"

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
```diff
@@ -0,0 +1,465 @@
...
+                }
+
+                return targetIndex;
+            }
+
+            void traceProgressIfNeeded(ContentHash currentHash)
+            {
+                if (progressReportingTask.IsCompleted)
+                {
+                    progressReportingTask = Task.Delay(_settings.SelfCheckProgressReportingInterval);
```
ljcollins25: >progressReportingTask = Task.Delay(_settings.SelfCheckProgressReportingInterval); [](start = 20, length = 81)

This a bit of a weird pattern. Can we just use a timestamp?

PR #290, opened 2019-05-10
Qtest untrack more

norahuang:


Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -117,18 +117,24 @@ export function runQTest(args: QTestArguments): Result {
...
+        qTestContextInfoPath = qTestContextInfoFile.path;
+        untrackingCBPaths =  {
+            unsafe: {
+                untrackedPaths: [
+                    qTestContextInfoFile,
+                ],
+                untrackedScopes: [
+                    d`d:/data`,
+                    d`d:/app`,
+                    Environment.hasVariable("QAUTHMATERIALROOT") ? d`${Environment.getDirectoryValue("QAUTHMATERIALROOT")}` : undefined,
```
dannyvv: >Environment [](start = 20, length = 11)

I suspect this mgith fail if we have undefined in the list. Better would be:
```
...addIf(Environment.hasVariable("QAUTHMATERIALROOT"),
    Environment.getDirectoryValue("QAUTHMATERIALROOT")
);
```

dannyvv: >d`$ [](start = 67, length = 3)

Since getDirectoryValue returns a directory there shouldn't be a need for wrapping it in d`...`. If you do need it please file a bug on me to figure out why :)

almiliMSFT: I just tried adding `undefined` as an untracked scope and it worked. 

While I agree with Danny's comment wholeheartedly, I suggest making this change in a separate PR.  The reason is that we still don't have a release candidate for next week, and we need this change to pass our canary validation, so I'd like to merge it as soon as we can.

norahuang: As Aleks suggest, I will open an other PR to address this comment. It is so difficult to get all validation finally passed.

PR #292, opened 2019-05-10
Fix long path problem for SetFileAccessControl

norahuang:
Fixes [AB#1519076](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1519076)


narasamdya:
Please give a more detailed title in the commit. E.g., Fix long path problem for SetFileAccessControl.

narasamdya:
>         public bool HasWritableAccessControl(string path)

Please ensure that we have unit tests for both methods below where we test the long path.

---
Refers to: Public/Src/Utilities/Native/IO/Windows/FileUtilities.Win.cs:1492 in 9fc68c2. [](commit_id = 9fc68c2473980b7ae0212f144c1e89296bf277c3, deletion_comment = False)

PR #293, opened 2019-05-10
Proper handling of MSBuild projects which don't implement the target protocol

smera:
This PR implements the following behavior regarding projects not implementing the target protocol:
* Leaf projects are always allowed. Only a verbose message is logged. Rationale: a leaf project doesn't need a target protocol.
* Non-leaf projects are blocked by default. A configuration option is added to opt-in to allow projects to not implement the protocol.
* If the user opted in to allow projects to not implement the protocol, a warning is issued for each project found in this scenario. Default targets are appended to the statically predicted list of targets of the reference as a heuristic to guess what targets the dependency needs. A warning is also issued for each project where non-empty default targets are appended due to this.

### DIFF
--- a/Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
+++ b/Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
@@ -57,10 +57,10 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         public IReadOnlyCollection<string> EntryPointTargets { get; }
 
         /// <summary>
-        /// Whether projects are allows to not implement the target protocol
+        /// Whether projects are allowed to not implement the target protocol
         /// </summary>
         /// <remarks>
-        /// If true, default targets are used as a heuristics
+        /// If true, default targets are used as heuristics
         /// </remarks>
         public bool AllowProjectsWithoutTargetProtocol { get; }
 
--- a/Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
+++ b/Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
@@ -56,6 +56,14 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         /// </remarks>
         public IReadOnlyCollection<string> EntryPointTargets { get; }
 
+        /// <summary>
+        /// Whether projects are allows to not implement the target protocol
+        /// </summary>
+        /// <remarks>
+        /// If true, default targets are used as a heuristics
+        /// </remarks>
+        public bool AllowProjectsWithoutTargetProtocol { get; }
+
         /// <nodoc/>
         public MSBuildGraphBuilderArguments(
             string enlistmentRoot,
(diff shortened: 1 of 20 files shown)


Public/Src/FrontEnd/MsBuild/Tracing/Log.cs
```diff
@@ -226,20 +226,30 @@ internal Logger()
...
             EventTask = (ushort)Events.Tasks.Engine,
             EventOpcode = (byte)Events.Tasks.Parser,
             Keywords = (int)(Events.Keywords.UserMessage | Events.Keywords.Diagnostics))]
         public abstract void ProjectIsNotSpecifyingTheProjectReferenceProtocol(LoggingContext context, Location location, string projectName);
 
+        [GeneratedEvent(
+            (ushort)LogEventId.ProjectPredictedTargetsAlsoContainDefaultTargets,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Warning,
+            Message = Events.LabeledProvenancePrefix + "Default targets were appended to the predicted target of project '{projectName}'. " +
```
erikma: >Default targets [](start = 56, length = 15)

Should list the default targets added in this message (DRI reduction)

Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
```diff
@@ -56,6 +56,14 @@ public sealed class MSBuildGraphBuilderArguments
         /// </remarks>
         public IReadOnlyCollection<string> EntryPointTargets { get; }
 
+        /// <summary>
+        /// Whether projects are allows to not implement the target protocol
```
erikma: >allows [](start = 33, length = 6)

allowed

Public/Src/FrontEnd/MsBuild.Serialization/MSBuildGraphBuilderArguments.cs
```diff
@@ -56,6 +56,14 @@ public sealed class MSBuildGraphBuilderArguments
         /// </remarks>
         public IReadOnlyCollection<string> EntryPointTargets { get; }
 
+        /// <summary>
+        /// Whether projects are allows to not implement the target protocol
+        /// </summary>
+        /// <remarks>
+        /// If true, default targets are used as a heuristics
```
erikma: >a heuristics [](start = 49, length = 12)

a heuristics -> heuristics

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -570,40 +600,14 @@ private bool TryAddMsBuildArguments(ProjectWithPredictions project, PipDataBuild
...
-            {
-                var targets = project.PredictedTargetsToExecute.Targets;
-                Contract.Assert(targets.Count > 0);
-                foreach (string target in targets)
-                {
-                    pipDataBuilder.Add(PipDataAtom.FromString($"/t:{target}"));
-                }
-            }
-            else
+            var targets = project.PredictedTargetsToExecute.Targets;
```
cdmihai: Where are the default targets being added if any of the referencing projects do not specify the protocol? Can't find it.

cdmihai: Ah, it's happening here: https://github.com/microsoft/BuildXL/pull/293/files#diff-033a479d4db36673974a933ddc616049R453

Public/Src/FrontEnd/MsBuild/Tracing/Log.cs
smera: Yeap, good point. Added.

---
In reply to: [283011368](https://github.com/microsoft/BuildXL/pull/293#discussion_r283011368) [](ancestors = 283011368)

PR #295, opened 2019-05-10
Skip caching symlink/junction directory on Windows

norahuang:
Windows currently does not support caching symlink/junction directories in output directory.
However, we don't want to fail the pip only because of this reason. 
The changes in this PR will allow the pip success, but the symlink/junction directories in the output directories will not be cached. Meanwhile, the pip that contains symlink/junction directories in the output directories will not be cached as well.

Fixes [AB#1519070](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1519070)


Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -202,9 +202,10 @@ private IEnumerable<Operation> GetLayoutProducingOperations(string rootDir)
                 .Split(new[] { '\n', '\r' }, StringSplitOptions.RemoveEmptyEntries)
                 .Select(spec => $"{rootDir}/{spec.Trim()}")
                 .Select(spec =>
-                    spec.Contains("->") ? OpCreateSym(X(spec)) :
-                    spec.EndsWith("/")  ? OpCreateDir(X(spec)) :
-                                          OpWriteFile(X(spec)));
+                    spec.Contains("->") ? (spec.EndsWith("/") ? OpCreateSym(X(spec), Operation.SymbolicLinkFlag.DIRECTORY ) : 
+                                                               OpCreateSym(X(spec), Operation.SymbolicLinkFlag.FILE)) :
+                                          (spec.EndsWith("/")  ? OpCreateDir(X(spec)) :
```
almiliMSFT: nit: weird formatting.  Can you please unindent this line and the next one to where they were before?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -202,9 +202,10 @@ private IEnumerable<Operation> GetLayoutProducingOperations(string rootDir)
                 .Split(new[] { '\n', '\r' }, StringSplitOptions.RemoveEmptyEntries)
                 .Select(spec => $"{rootDir}/{spec.Trim()}")
                 .Select(spec =>
-                    spec.Contains("->") ? OpCreateSym(X(spec)) :
-                    spec.EndsWith("/")  ? OpCreateDir(X(spec)) :
-                                          OpWriteFile(X(spec)));
+                    spec.Contains("->") ? (spec.EndsWith("/") ? OpCreateSym(X(spec), Operation.SymbolicLinkFlag.DIRECTORY ) :
```
almiliMSFT: can you please move this `spec.EndsWith("/")` conditional inside the `OpCreateSym` method invocation
```suggestion
                    spec.Contains("->") ? OpCreateSym(X(spec), spec.EndsWith("/") ?Operation.SymbolicLinkFlag.DIRECTORY : Operation.SymbolicFlag.FILE) : 
```

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -202,9 +202,10 @@ private IEnumerable<Operation> GetLayoutProducingOperations(string rootDir)
                 .Split(new[] { '\n', '\r' }, StringSplitOptions.RemoveEmptyEntries)
                 .Select(spec => $"{rootDir}/{spec.Trim()}")
                 .Select(spec =>
-                    spec.Contains("->") ? OpCreateSym(X(spec)) :
-                    spec.EndsWith("/")  ? OpCreateDir(X(spec)) :
-                                          OpWriteFile(X(spec)));
+                    spec.Contains("->") ? (spec.EndsWith("/") ? OpCreateSym(X(spec), Operation.SymbolicLinkFlag.DIRECTORY ) : 
+                                                               OpCreateSym(X(spec), Operation.SymbolicLinkFlag.FILE)) :
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -349,6 +350,30 @@ void ValidateProducedLayout()
...
         }
 
+        [Feature(Features.Symlink)]
+        [Feature(Features.OpaqueDirectory)]
+        [TheoryIfSupported(requiresSymlinkPermission: true)]
+        public void DirectorySymlinksInOutputDirectoryShouldSuccess()
+        {    
+            AbsolutePath rootDirAbsPath = CreateUniqueObjPath($"{SealDirectoryKind.SharedOpaque}.framework");
+            string rootDir = rootDirAbsPath.ToString(Context.PathTable);
+            System.Diagnostics.Debugger.Launch();
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -349,6 +350,30 @@ void ValidateProducedLayout()
             }
         }
 
+        [Feature(Features.Symlink)]
+        [Feature(Features.OpaqueDirectory)]
+        [TheoryIfSupported(requiresSymlinkPermission: true)]
+        public void DirectorySymlinksInOutputDirectoryShouldSuccess()
```
almiliMSFT:
```suggestion
        public void DirectorySymlinksInOutputDirectoryShouldSucceed()
```

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -349,6 +350,30 @@ void ValidateProducedLayout()
...
+            producerPipBuilder.ToolDescription = StringId.Create(Context.StringTable, "producer");
+
+            // if dirKind is an opaque output, then just declare the root directory as an opaque output;
+            // otherwise, add all outputs explicitly and schedule a seal directory pip to seal them.
+            Process producerPip;
+            DirectoryArtifact outputDirArtifact;
+            producerPipBuilder.AddOutputDirectory(rootDirAbsPath, kind: SealDirectoryKind.SharedOpaque);
+            producerPip = SchedulePipBuilder(producerPipBuilder).Process;
+            outputDirArtifact = producerPip.DirectoryOutputs.First();
+            ScheduleRunResult firstRunResult = RunScheduler().AssertSuccess();
```
almiliMSFT: unused variables: `firstRunResult`, `outputDirArtifact`.

In addition to expecting success, we should also specify what else should happen.  For example:
  - what happens when you clean the output folder and re-run the same pip?  Cache hit will happen but the directory symlink will not be restored on Windows.
  - also what happens when you re-run the pip in terms of scrubbing?  Is the directory symlink supposed to be scrubbed?  

All that said, I think it might be better to enable the existing `DirectorySymlinksInOutputDirectory` unit tests for Windows and then in that unit test have different assertions based on what operating system the test is running on

norahuang: Thanks again. Will try to reuse the test case.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -3659,6 +3660,18 @@ internal static void UpdateFileData(Dictionary<AbsolutePath, FileOutputData> map
                                 continue;
                             }
 
+                            if (!OperatingSystemHelper.IsUnixOS)
+                            {
+                                var pathstring = access.ToString(pathTable);
+                                if (Directory.Exists(pathstring))
```
almiliMSFT: One problem with `Directory.Exists()` is that a pip can produce a directory symlink that points to a non-existent file.  In that case, `Directory.Exists()` will (presumably) return false.  That's why I'd prefer to use `GetFileAttributes` instead and then check that both `ReparsePoint` and `Directory` flags are set

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
norahuang: Tried to run DirectorySymlinksInOutputDirectory in windows. None of the condition has any luck to succeed. Checkback the test cases in DirectorySymlinkTests.cs, all test cases under this file are designed with requiresUnixBasedOperatingSystem=true. So I am wondering the history of the feature. Is it supposed to be different in windows and mscOs? I will spent a bit more time to understand the symlink feature and then back to the unit test here.

almiliMSFT: It's not really that it is "supposed" to be different, it just currently is different because directory symlinks have not yet been fully implemented for Windows.  So my suggestion was to reuse the same test but add `if (OperatingSystemHelper.IsUnixOS)` around all those checks that don't pass on Windows so that we have one place where we can see what works and what doesn't work on Windows.  @narasamdya, what do you think?  Having a separate unit test is also fine with me.

norahuang: By running the existing test case for MAC, found out the behavior of the consuming directory symlink was not as expected. Without specified input of files in the output share opaque directory, BuildXL report read access violation. After offline check with @almiliMSFT, we agreed to limit the test only on the producer with testing the cache behavior.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
narasamdya: Remove launching debugger.

---
In reply to: [283067785](https://github.com/microsoft/BuildXL/pull/295#discussion_r283067785) [](ancestors = 283067785)

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -3659,6 +3660,18 @@ internal static void UpdateFileData(Dictionary<AbsolutePath, FileOutputData> map
...
 
+                            if (!OperatingSystemHelper.IsUnixOS)
+                            {
+                                var pathstring = access.ToString(pathTable);
+                                if (Directory.Exists(pathstring))
+                                {
+                                    // We run into here becuase maybeResult.Result != PathExistence.ExistsAsDirectory
+                                    // But the path is actually a Directory, so it is a dir symlink or junction.
+                                    // We don't support storing directory symlinks/junctions to the cache in Windows right now.
+                                    continue;
```
narasamdya: >continue; [](start = 36, length = 9)

There are two choices here:
- We log error, stating that we don't support directory symlink production.
- We log warning, but make the pip uncacheable.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -349,6 +350,30 @@ void ValidateProducedLayout()
...
+            // otherwise, add all outputs explicitly and schedule a seal directory pip to seal them.
+            Process producerPip;
+            DirectoryArtifact outputDirArtifact;
+            producerPipBuilder.AddOutputDirectory(rootDirAbsPath, kind: SealDirectoryKind.SharedOpaque);
+            producerPip = SchedulePipBuilder(producerPipBuilder).Process;
+            outputDirArtifact = producerPip.DirectoryOutputs.First();
+            ScheduleRunResult firstRunResult = RunScheduler().AssertSuccess();         
+        }
+
         [Feature(Features.Symlink)]
```
narasamdya: I think the bug is not about making directory symlink work on Windows. The fix for the bug is simply to avoid catastrophic failure. Otherwise this will not be a bug, but rather a feature request.

As I mentioned in the other comment, we have two choices:
- Fail the pip and log error if the pip produces a directory symlink.
- Warn, but make the pip uncacheable. 
I prefer the latter.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -3659,6 +3660,18 @@ internal static void UpdateFileData(Dictionary<AbsolutePath, FileOutputData> map
                                 continue;
                             }
 
+                            if (!OperatingSystemHelper.IsUnixOS)
```
ljcollins25: >if [](start = 28, length = 2)

How is this case handled for normal output directories? Seems like it should be the same between the two.

norahuang: For situations other than SharedOpaque, the content in the output directory is scan and listed as files. So I think we won't encounter the symlink directory problem.

ljcollins25: Seems like we treat them as files for normal opaque directories (see comment here (line 614): [http://index/?query=EnumerateDynamicOutputDirectory&rightProject=BuildXL.Scheduler&file=Artifacts%5CFileContentManager.cs&line=614](http://index/?query=EnumerateDynamicOutputDirectory&rightProject=BuildXL.Scheduler&file=Artifacts%5CFileContentManager.cs&line=614)). To be consistent, seems like we do that same here.

---
In reply to: [284017124](https://github.com/microsoft/BuildXL/pull/295#discussion_r284017124) [](ancestors = 284017124)

norahuang: Thanks. The latest change achieve the consistence by disable output cache for the symlink dir and pip cache for all output directories.

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -609,6 +609,15 @@ internal static string PipWriteFileFailedMessage(string pipDescription, string p
             Message = "Cleaning output directory '{destinationPath}' for pip {pipDescription} resulted in error: {errorMessage}")]
         public abstract void StorageCacheCleanDirectoryOutputError(LoggingContext loggingContext, string pipDescription, string destinationPath, string errorMessage);
 
+        [GeneratedEvent(
+            (ushort)LogEventId.StorageSymlinkDirInOutputDirectoryWarning,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Warning,
+            Keywords = (int)Events.Keywords.UserMessage,
+            EventTask = (ushort)Events.Tasks.Storage,
+            Message = "[{pipDescription}] Cache pip with symlink directory '{symlinkPath}' in output directory resulted in error")]
```
almiliMSFT: I wouldn't say "resulted in **error**" when this is a warning message.

Instead, I'd say something like "Detected junctions or directory symlink output '{symlinkPath}'. The pip will not be stored to cache."

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -94,9 +96,19 @@ protected override void Dispose(bool disposing)
...
 Versions/A/file
-Versions/A/sym-loop -> ../sym-A
-Versions/sym-A -> A
-Versions/sym-sym-A -> sym-A
+Versions/A/sym-loop -> ../sym-A/
+Versions/sym-A -> A/
+Versions/sym-sym-A -> sym-A/
+";
+
+        private const string DirectoryLayoutWinTest = @"
```
almiliMSFT: how's this layout different from the one above?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -204,13 +216,14 @@ protected override void Dispose(bool disposing)
...
+            var layout = OperatingSystemHelper.IsUnixOS ? DirectoryLayout : DirectoryLayoutWinTest;
+            return layout
                 .Split(new[] { '\n', '\r' }, StringSplitOptions.RemoveEmptyEntries)
                 .Select(spec => $"{rootDir}/{spec.Trim()}")
                 .Select(spec =>
-                    spec.Contains("->") ? OpCreateSym(X(spec)) :
+                    spec.Contains("->") ? OpCreateSym(X(spec), spec.EndsWith("/") ? Operation.SymbolicLinkFlag.DIRECTORY : Operation.SymbolicLinkFlag.FILE) :
                     spec.EndsWith("/")  ? OpCreateDir(X(spec)) :
-                                          OpWriteFile(X(spec)));
+                                          OpWriteFile(X(spec)));
```
almiliMSFT:
```suggestion
                                          OpWriteFile(X(spec)));
```

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -250,6 +263,52 @@ public void TestEnumerateDirectory()
...
+                producerPip = SchedulePipBuilder(producerPipBuilder).Process;
+                outputDirArtifact = SealDirectory(rootDirAbsPath, dirKind, dao.Outputs.ToArray());
+            }
+
+            // first run, expect cache miss
+            RunScheduler().AssertSuccess().AssertCacheMiss(producerPip.PipId);
+
+            // rerun, check cache miss, symlink directory in output can't be cached
+            RunScheduler().AssertSuccess().AssertCacheMiss(producerPip.PipId);
+            AssertWarningEventLogged(LogEventId.StorageSymlinkDirInOutputDirectoryWarning, count: 4);
```
almiliMSFT: Is it true that this warning happens after the first run too?  I'd say yes, so assert it after the first run as well.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
norahuang: since the window native can not deal with symlink loop

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
norahuang: sure, I will separate them.

Public/Src/Engine/Scheduler/Tracing/Log.cs
norahuang: Thanks

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -1635,6 +1636,15 @@ private void PopulateArtifactsToMaterialize(PipArtifactsState state, bool? allow
                             state.MaterializedDirectoryContents.Add(file);
                         }
 
+                        if (!OperatingSystemHelper.IsUnixOS)
+                        {
+                            if (FileSystemWin.IsPathReparseDirectory(file.Path.ToString(Context.PathTable)))
```
ljcollins25: >IsPathReparseDirectory [](start = 46, length = 22)

This should not be here. The file may not even be on disk at this point. What is the intent of this line?

Public/Src/Utilities/Native/IO/Windows/FileSystem.Win.cs
```diff
@@ -2911,6 +2911,17 @@ private bool TryGetFileAttributesViaFindFirstFile(string path, out FileAttribute
             }
         }
 
+        public static bool IsPathReparseDirectory(string path)
```
narasamdya: >public static bool IsPathReparseDirectory(string path) [](start = 8, length = 54)

No need to change anything here. We have in `FileUtilities` a function called `IsDirectorySymlinkOrJunction`.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -35,6 +35,7 @@
 using BuildXL.Utilities.Tracing;
 using static BuildXL.Utilities.FormattableStringEx;
 using static BuildXL.Scheduler.FileMonitoringViolationAnalyzer;
+using BuildXL.Native.IO.Windows;
```
narasamdya: >BuildXL.Native.IO.Windows; [](start = 6, length = 26)

I don't think you need this.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -3492,6 +3493,27 @@ internal static void UpdateFileData(Dictionary<AbsolutePath, FileOutputData> map
             }
         }
 
+        private static bool IsWindowsOutputSymlinkDirectoryOrJunction(AbsolutePath outputPath, OperationContext operationContext, string description, PathTable pathTable, ExecutionResult processExecutionResult)
+        {
+            if (!OperatingSystemHelper.IsUnixOS)
+            {
+                var pathstring = outputPath.ToString(pathTable);
+                if (FileSystemWin.IsPathReparseDirectory(pathstring))
```
narasamdya: >FileSystemWin [](start = 20, length = 13)

We should only use, as much as possible, `FileUtilities`.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -3492,6 +3493,27 @@ internal static void UpdateFileData(Dictionary<AbsolutePath, FileOutputData> map
...
+        private static bool IsWindowsOutputSymlinkDirectoryOrJunction(AbsolutePath outputPath, OperationContext operationContext, string description, PathTable pathTable, ExecutionResult processExecutionResult)
+        {
+            if (!OperatingSystemHelper.IsUnixOS)
+            {
+                var pathstring = outputPath.ToString(pathTable);
+                if (FileSystemWin.IsPathReparseDirectory(pathstring))
+                {
+                    // We don't support storing directory symlinks/junctions to the cache in Windows right now.
+                    // We won't fail the pip
+                    // We won't cache it neither
```
narasamdya: >neither [](start = 41, length = 7)

either

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -3492,6 +3493,27 @@ internal static void UpdateFileData(Dictionary<AbsolutePath, FileOutputData> map
...
+                    Logger.Log.StorageSymlinkDirInOutputDirectoryWarning(
+                        operationContext,
+                        description,
+                        pathstring);
+                    processExecutionResult.MustBeConsideredPerpetuallyDirty = true;
```
narasamdya: >processExecutionResult.MustBeConsideredPerpetuallyDirty = true; [](start = 20, length = 63)

As you discovered last time, this will get overriden.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -3492,6 +3493,27 @@ internal static void UpdateFileData(Dictionary<AbsolutePath, FileOutputData> map
             }
         }
 
+        private static bool IsWindowsOutputSymlinkDirectoryOrJunction(AbsolutePath outputPath, OperationContext operationContext, string description, PathTable pathTable, ExecutionResult processExecutionResult)
```
narasamdya: >IsWindowsOutputSymlinkDirectoryOrJunction [](start = 28, length = 41)

```
private static bool CheckForAllowedDirectorySymlinkOrJunctionProduction(...)
{
    if (OperatingSystemHelper.IsUnixOS)
    {
        return true;
    }

    if (FileUtilities.IsDirectorySymlinkOrJunction(path))
    {
        // log warning.
        return false;
    }

    return true;
}
```

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -13,6 +13,7 @@
 using BuildXL.Cache.ContentStore.Hashing;
 using BuildXL.Engine.Cache.Artifacts;
 using BuildXL.Native.IO;
+using BuildXL.Native.IO.Windows;
```
narasamdya: >BuildXL.Native.IO.Windows; [](start = 6, length = 26)

As @  pointed out, no need change here.

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -609,6 +609,15 @@ internal static string PipWriteFileFailedMessage(string pipDescription, string p
             Message = "Cleaning output directory '{destinationPath}' for pip {pipDescription} resulted in error: {errorMessage}")]
         public abstract void StorageCacheCleanDirectoryOutputError(LoggingContext loggingContext, string pipDescription, string destinationPath, string errorMessage);
 
+        [GeneratedEvent(
+            (ushort)LogEventId.StorageSymlinkDirInOutputDirectoryWarning,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Warning,
+            Keywords = (int)Events.Keywords.UserMessage,
+            EventTask = (ushort)Events.Tasks.Storage,
+            Message = "[{pipDescription}] Pip produced a directory symlink '{symlinkPath}', which is not supported. The pip will hence not be stored to cache.")]
```
narasamdya: >The pip will hence not be stored to cache [](start = 116, length = 41)

The pip will not be cached.

narasamdya: >a directory symlink [](start = 55, length = 19)

"a directory symlink or junction"

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -250,6 +263,56 @@ public void TestEnumerateDirectory()
...
+            else
+            {
+                var dao = InferIOFromOperations(operations, force: true);
+                foreach (var output in dao.Outputs) producerPipBuilder.AddOutputFile(output.Path);
+                foreach (var input in dao.Dependencies) producerPipBuilder.AddInputFile(input);
+                producerPip = SchedulePipBuilder(producerPipBuilder).Process;
+                outputDirArtifact = SealDirectory(rootDirAbsPath, dirKind, dao.Outputs.ToArray());
+            }
+
+            System.Diagnostics.Debugger.Launch();
```
narasamdya: >System.Diagnostics.Debugger.Launch(); [](start = 12, length = 37)

Remove debugger.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -250,6 +263,56 @@ public void TestEnumerateDirectory()
...
+        [Feature(Features.OpaqueDirectory)]
+        [TheoryIfSupported(requiresWindowsBasedOperatingSystem: true, requiresSymlinkPermission: true)]
+        [InlineData(SealDirectoryKind.Full)]
+        [InlineData(SealDirectoryKind.Partial)]
+        [InlineData(SealDirectoryKind.Opaque)]
+        [InlineData(SealDirectoryKind.SharedOpaque)]
+        public void DirectorySymlinksInOutputDirectoryProducerOnlyOnWindows(SealDirectoryKind dirKind)
+        {
+            XAssert.IsFalse(dirKind.IsSourceSeal());
+
```
almiliMSFT: nit: two empty lines
```suggestion
```

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
norahuang: Removed. This might be needed if we consider the consumer pip who might seal the directory after the create symlink pip generate it. But at this point since we don't care the functionality of the symlink directory, I will remove it.

PR #300, opened 2019-05-14
Make sure shared opaques are flagged as such on graceful cancellation

smera:
On graceful cancellation, pips may not go through the normal post process step, which causes shared opaque outputs to not be flagged as such. This can be problematic for subsequent builds.

This PR makes sure opaque flagging happens upon cancellation. It also fixes a small bug in the scheduler where the cancellation step was never executed when the next pip step was not inlined (e.g. postprocess).

There doesn't seem to be any test infrastructure around graceful cancellation. I can add tests for this particular case upon request, but it may imply a bigger work item.

### DIFF
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -2803,6 +2803,13 @@ namespace BuildXL.Scheduler
 
             bool inline = false;
 
+            // If the pip should be cancelled, make sure we inline the next step. The pip queue may be also flagged as cancelled and won't dequeue the pip otherwise.
+            // The check for cancellation will then happen on ExecutePipStep and the pip will be transitioned to PipExecutionStep.Cancel
+            if (ShouldCancelPip(runnablePip))
+            {
+                inline = true;
+            }
+
             // If the next queue is none or the same as the previous one, do not change the current queue and inline execution here.
             // However, when choosing worker, we should enqueue again even though the next queue is chooseworker again.
             if (nextQueue == DispatcherKind.None)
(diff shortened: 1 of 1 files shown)


almiliMSFT:
👍

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3092,7 +3108,7 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
 
             // If schedule is terminating (e.g., StopOnFirstFailure), cancel the pip
             // as long as (i) 'start' step has been executed, (ii) the pip is in running state, and (iii) the pip has not been cancelled before.
-            if (m_scheduleTerminating && step != PipExecutionStep.Start && GetPipRuntimeInfo(pipId).State == PipState.Running && !runnablePip.IsCancelled)
+            if (ShouldCancelPip(runnablePip))
             {
                 return runnablePip.Cancel();
             }
```
semihokur: I'd like to move this stuff to ExecuteAsyncOrEnqueue as we discussed; so that we will have one place for the cancellation-related transition.

smera: Tried that. The problem is that if I did that we'd have to duplicate the logic to transition a pip to the next state in ExecuteAsyncOrEnqueue(...) (so it will be there AND in ExecutePip(...)). Otherwise the original next state will still be the next one.

So I think the change I'm proposing here is cleaner: we just make sure that a cancelled pip reaches ExecutePipStep(...). Then ExecutePip(...) will make sure the pips gets transitioned to Cancel, and then, in the next cycle, the actual cancellation step will be executed.

---
In reply to: [283903593](https://github.com/microsoft/BuildXL/pull/300#discussion_r283903593) [](ancestors = 283903593)

PR #301, opened 2019-05-14
Fixed underbuild caused by enumerating directory symlinks

narasamdya:
Directory symlinks are classified as files, and thus in observed input processor, computing directory fingerprint for their enumerations is skipped. Thus, when one add/remove file from the directory, the pip doesn't rebuild.

An alternative solution (which looks like the proper one) would be at Detours level. That is, in the same way as treating file symlinks, we follow the chain of directory symlinks. For each intermediate links, we report as file reads, and thus we will hash their paths. For the final directory, we report as enumeration, and thus the directory fingerprint will be computed.

However, doing the above solution would require Office to start specifying intermediate directory symlinks as dependencies. Since Office has already been using directory symlinks/junctions for both dev builds and CloudBuild, this option is, currently, difficult to realize. 

One can see this solution from a different perspective. Currently, we don't require users to specify directory dependencies for directory enumeration. Similarly, here we don't require users to specify directory symlink dependencies when the enumeration is done via directory symlinks.

[AB#1530590](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1530590)


Public/Src/Deployment/NugetPackages.dsc
```diff
@@ -122,9 +122,11 @@ namespace NugetPackages {
                 cacheHashing,
             ]),
             sdks,
-            osxX64,
+            ...addIf(!BuildXLSdk.Flags.genVSSolution,
+                    osxX64,
```
Kr1sso: Nit: indentation of the whole block

Kr1sso:
Unit tests are fixed, macOS should be up and running again once you get master bits :)

Public/Src/Engine/Scheduler/DirectoryEnumerationResult.cs
```diff
@@ -43,7 +43,7 @@ namespace BuildXL.Scheduler
         public DirectoryEnumerationResult(PathExistence existence, IReadOnlyList<(AbsolutePath, string)> members)
         {
             Contract.Requires(members != null);
-            Contract.Requires(members.Count == 0 || existence == PathExistence.ExistsAsDirectory);
+            Contract.Requires(members.Count == 0 || existence != PathExistence.Nonexistent);
```
almiliMSFT: It seems strange that a `DirectoryEnumerationResult` can be provided for a path that `ExistsAsFile`

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -547,6 +547,32 @@ public void ConcurrentAccessToHardlinksPointingToSameFileIntertwinedWithOtherRea
...
+                targetDirectory.ToString(Context.PathTable), 
+                isTargetFile: false).Succeeded);
+
+            ProcessWithOutputs processWithOutputs = CreateAndSchedulePipBuilder(new[]
+            {
+                Operation.EnumerateDir(DirectoryArtifact.CreateWithZeroPartialSealId(directorySymlink)),
+                Operation.WriteFile(CreateOutputFileArtifact())
+            });
+
+            RunScheduler().AssertSuccess();
```
almiliMSFT: On macOS you have to allow access to `directorySymlink` file, otherwise you'll get:
```
[0:04] Error DX0500: [PipEB9A5D1F51C0C26E, Test.BuildXL.Executables.TestProcess, TestModule, testValue, {}] - Disallowed file accesses were detected (R = read, W = write):
        Disallowed file accesses performed by: /usr/bin/time
         R .../TestProj/Out/Objects.noindex/a/x/o3qjkq7e5t4nqzoz8wqko3e1g64rte/t_10/DirectorySym4A8FD4B8/0/obj/readonly/Symlink_3
```

This should do the trick:
```sharp
var pipBuilder = CreatePipBuilder(new[]
            {
                Operation.EnumerateDir(DirectoryArtifact.CreateWithZeroPartialSealId(directorySymlink)),
                Operation.WriteFile(CreateOutputFileArtifact())
            });
            pipBuilder.AddInputFile(directorySymlink);
            ProcessWithOutputs processWithOutputs = SchedulePipBuilder(pipBuilder);
```

Public/Src/Engine/Scheduler/Fingerprints/ObservedInputProcessor.cs
```diff
@@ -1603,20 +1616,20 @@ public Possible<PathExistence> TryProbeAndTrackForExistence(AbsolutePath path, C
                 var fullGraphExistExistence = FileSystemView.GetExistence(path, FileSystemViewMode.FullGraph);
                 if (fullGraphExistExistence.Result == PathExistence.ExistsAsDirectory)
                 {
-                    return PathExistence.ExistsAsDirectory;
+                    existence = PathExistence.ExistsAsDirectory;
```
ljcollins25: >existence = PathExistence.ExistsAsDirectory; [](start = 20, length = 44)

I prefer these as return statements since it makes it clear that there is no further processing to be done.

ljcollins25:
>         private enum FileSystemEntryFlags

Add flags for IsDirectorySymlink and NotDirectorySymlink to enable caching of this check.

---
Refers to: Public/Src/Engine/Scheduler/FileSystem/FileSystemView.cs:543 in e41a91c. [](commit_id = e41a91c92c1897737ffa00afb624a0261e2557b6, deletion_comment = False)

Public/Src/Engine/Scheduler/FileSystem/FileSystemView.cs
```diff
@@ -151,7 +151,9 @@ public Possible<PathExistence> TryEnumerateDirectory(AbsolutePath path, FileSyst
             PathExistence existence;
             if (PathExistenceCache.TryGetValue(path, out entry) && entry.TryGetExistence(mode, out existence))
             {
-                if (existence != PathExistence.ExistsAsDirectory)
+                if (existence == PathExistence.Nonexistent
+                    || (existence == PathExistence.ExistsAsFile 
+                        && !FileUtilities.IsDirectorySymlinkOrJunction(path.ToString(PathTable))))
                 {
```
ljcollins25: >{ [](start = 16, length = 1)

Update flags IsDirectorySymlink and CheckedIsDirectorySymlink to enable caching of this check.

PathExistenceCache.AddOrUpdate(path, false,
                            (key, data) => { throw Contract.AssertFailure("Entry should already be added for path"); },
                            (key, data, oldValue) => oldValue.SetFlag(FileSystemEntryFlags.CheckedIsDirectorySymlink | (isDirectorySymlink ?  FileSystemEntryFlags.IsDirectorySymlink : FileSystemEntryFlags.None)));

Public/Src/Engine/Scheduler/FileSystem/FileSystemView.cs
```diff
@@ -151,7 +151,9 @@ public Possible<PathExistence> TryEnumerateDirectory(AbsolutePath path, FileSyst
             PathExistence existence;
             if (PathExistenceCache.TryGetValue(path, out entry) && entry.TryGetExistence(mode, out existence))
             {
-                if (existence != PathExistence.ExistsAsDirectory)
+                if (existence == PathExistence.Nonexistent
```
ljcollins25: >Nonexistent [](start = 47, length = 11)

Why do we do this for Nonexistent as well?

narasamdya:
Thanks!

---
In reply to: [492293469](https://github.com/microsoft/BuildXL/pull/301#issuecomment-492293469) [](ancestors = 492293469)

Public/Src/Deployment/NugetPackages.dsc
```diff
@@ -122,9 +122,11 @@ namespace NugetPackages {
                 cacheHashing,
             ]),
             sdks,
-            osxX64,
+            ...addIf(!BuildXLSdk.Flags.genVSSolution,
```
dannyvv: >addIf [](start = 15, length = 5)

seperate pr?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
narasamdya: Thanks for testing it on Mac. I was relying on the PR build, but the first time it builds, some weird things happened.

---
In reply to: [283882342](https://github.com/microsoft/BuildXL/pull/301#discussion_r283882342) [](ancestors = 283882342)

Public/Src/Engine/Scheduler/Fingerprints/ObservedInputProcessor.cs
narasamdya: This is leftover from my debugging -- I tried to check the path existence at the end of the method. I'll revert it.

---
In reply to: [283894583](https://github.com/microsoft/BuildXL/pull/301#discussion_r283894583) [](ancestors = 283894583)

Public/Src/Engine/Scheduler/FileSystem/FileSystemView.cs
narasamdya: I'm preserving the old behavior where if it is Nonexistent, then we return it immediately.

---
In reply to: [283897412](https://github.com/microsoft/BuildXL/pull/301#discussion_r283897412) [](ancestors = 283897412)

Public/Src/Deployment/NugetPackages.dsc
narasamdya: Sure.

---
In reply to: [283919613](https://github.com/microsoft/BuildXL/pull/301#discussion_r283919613) [](ancestors = 283919613)

PR #302, opened 2019-05-14
Set IsCancelled to true if cancellation was requested

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
```diff
@@ -133,7 +133,7 @@ protected override async Task<BoolResult> TryGetFileCoreAsync(OperationContext c
                     Tracer.Debug(context, $@"Downloading blob '{_configuration.ContainerName}\{blobName}' failed with recoverable exception: {e}.");
                 }
 
-                return AttemptResult.FromException(isRecoverable, e);
+                return AttemptResult.FromException(isRecoverable, e, context.Token);
```
dannyvv: Should we rename this to CancelationToken to ensure we don't confuse this with AuthToken and other kinds of Tokens.

SergeyTeplyakov: Maybe, but I'm not sure we have any other tokens so widely used... I'll discuss this with other cache members to get their opinion.

PR #303, opened 2019-05-14
[WIP] Use .NET Core as the default when building on Windows

almiliMSFT:


dannyvv:
Note, Just as FYI, You likely know but for the full switch we also need to:
* update build definitions (to set the buildxl toolset to use in cloudbuild) 
* update publish LKG step to use win-x64

PR #305, opened 2019-05-14
Stop using native file handle during file open

SergeyTeplyakov:
[AB#1530241](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1530241)

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
+++ b/Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
@@ -452,8 +452,10 @@ namespace BuildXL.Cache.ContentStore.FileSystem
             {
                 return TryOpenFileUnix(path, accessMode, mode, share, options, bufferSize);
             }
-
-            return TryOpenFileWin(path, accessMode, mode, share, options, bufferSize);
+            else
+            {
+                return TryOpenFileWin(path, accessMode, mode, share, options, bufferSize);
+            }
         }
 
         private Stream TryOpenFileUnix(AbsolutePath path, FileAccess accessMode, FileMode mode, FileShare share, FileOptions options, int bufferSize)
@@ -470,7 +472,16 @@ namespace BuildXL.Cache.ContentStore.FileSystem
 
             return new FileStream(path.Path, mode, accessMode, share, bufferSize, options);
         }
-        
+
+        /// <summary>
+        /// Tries opening a file with a given <paramref name="path"/>.
+        /// </summary>
+        /// <return>
+        /// Method returns null if file or directory is not found.
+        /// </return>
+        /// <remarks>
+        /// The method throws similar exception that <see cref="FileStream"/> constructor.
+        /// </remarks>
         private Stream TryOpenFileWin(AbsolutePath path, FileAccess accessMode, FileMode mode, FileShare share, FileOptions options, int bufferSize)
         {
             options = GetOptions(path, options);
--- a/Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
+++ b/Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
@@ -20,6 +20,7 @@ using BuildXL.Utilities;
 using Microsoft.Win32.SafeHandles;
 using AbsolutePath = BuildXL.Cache.ContentStore.Interfaces.FileSystem.AbsolutePath;
 using FileInfo = BuildXL.Cache.ContentStore.Interfaces.FileSystem.FileInfo;
+using static BuildXL.Cache.ContentStore.FileSystem.NativeMethods;
 
 namespace BuildXL.Cache.ContentStore.FileSystem
 {
@@ -350,7 +351,7 @@ namespace BuildXL.Cache.ContentStore.FileSystem
 
             using (await ConcurrentAccess.WaitToken())
             {
-                return OpenInternal(path, fileAccess, fileMode, share, options, bufferSize);
+                return TryOpenFile(path, fileAccess, fileMode, share, options, bufferSize);
             }
         }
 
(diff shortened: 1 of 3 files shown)


Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
```diff
@@ -445,89 +446,86 @@ private async Task CopyFileInsideSemaphoreAsync(AbsolutePath sourcePath, Absolut
...
 
-                void openHandle()
-                {
-                    handle = NativeMethods.CreateFile(path.Path, access, share, IntPtr.Zero, mode, options, IntPtr.Zero);
-                }
+            var (result, exception) = tryOpenFile();
 
-                openHandle();
+            if (exception != null && mode == FileMode.Create &&
+                (exception.NativeErrorCode() == ERROR_ACCESS_DENIED || exception.NativeErrorCode() == ERROR_SHARING_VIOLATION))
```
dannyvv: >exception.NativeErrorCode() [](start = 17, length = 27)

Is this computation cached? should we create a local to avoid double computation?

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
```diff
@@ -445,89 +446,86 @@ private async Task CopyFileInsideSemaphoreAsync(AbsolutePath sourcePath, Absolut
...
+                                mode,
+                                accessMode),
+                            exception.NativeErrorCode());
                 }
+            }
 
-                var needToDisposeHandle = true;
+            return result;
+
+            (FileStream fileStream, Exception Exception) tryOpenFile()
```
dannyvv: >(FileStr [](start = 12, length = 8)

Would Either<T,U> make more sense?

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
SergeyTeplyakov: this is a property getter and the bitmask and it is called for error case only. I was thinking of using a field here, but I think an additional local makes code harder to read.

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
SergeyTeplyakov: yep, but without proper union types this is effectively the same thing.

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
```diff
@@ -445,89 +446,86 @@ private async Task CopyFileInsideSemaphoreAsync(AbsolutePath sourcePath, Absolut
...
+            if (OperatingSystemHelper.IsUnixOS)
+            {
+                return TryOpenFileUnix(path, accessMode, mode, share, options, bufferSize);
+            }
 
-                    if (mode == FileMode.Open && !FileExists(path))
-                    {
-                        return null;
-                    }
+            return TryOpenFileWin(path, accessMode, mode, share, options, bufferSize);
```
dcwuser: Very minor: I'm not opposed to early returns, but for two equal options of equal complexity, I think having an else block is more natural.

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
dcwuser: >tryOpenFile [](start = 57, length = 11)

Just make a full private method.

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
```diff
@@ -445,89 +446,86 @@ private async Task CopyFileInsideSemaphoreAsync(AbsolutePath sourcePath, Absolut
...
+            if (mode == FileMode.Open && !FileExists(path))
+            {
+                return null;
+            }
 
-                SafeFileHandle handle;
+            return new FileStream(path.Path, mode, accessMode, share, bufferSize, options);
+        }
+        
+        private Stream TryOpenFileWin(AbsolutePath path, FileAccess accessMode, FileMode mode, FileShare share, FileOptions options, int bufferSize)
```
dcwuser: Add more comments describing what method is accomplishing and how, which errors are handled (and which not) etc.

dcwuser:
>         public async Task<Stream> OpenAsync(AbsolutePath path, FileAccess fileAccess, FileMode fileMode, FileShare share, FileOptions options, int bufferSize)

Can we add comments detailing contract while they are still fresh in your head? I.e. precisely what is returned in what circumstances, including exception cases.

---
Refers to: Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs:343 in e4f18fe. [](commit_id = e4f18fe777d0efc057fd040b01c875794faf6f76, deletion_comment = False)

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
kewilli: Shouldn't this be implemented at the BuildXL.Native.IO.FileUtilities level so we can share any changes across all of BXL?

SergeyTeplyakov: This API is very weird and returns null in some cases. The behavior is very unusual and I doubt that it should be the same in Bxl.

So in general I agree that this class should be unified with bxl but I'm not sure that this is feasible right now.

---
In reply to: [285679929](https://github.com/microsoft/BuildXL/pull/305#discussion_r285679929) [](ancestors = 285679929)

SergeyTeplyakov:
>         public async Task<Stream> OpenAsync(AbsolutePath path, FileAccess fileAccess, FileMode fileMode, FileShare share, FileOptions options, int bufferSize)

I think this is pretty much already described in the interface declaration, but I've enhanced the comment a bit there.

---
In reply to: [492758896](https://github.com/microsoft/BuildXL/pull/305#issuecomment-492758896) [](ancestors = 492758896)

---
Refers to: Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs:343 in e4f18fe. [](commit_id = e4f18fe777d0efc057fd040b01c875794faf6f76, deletion_comment = False)

PR #306, opened 2019-05-15
Don't use '-parallel all' when running XUnit tests

almiliMSFT:
Using `-parallel all` only causes races and flakiness in some of our tests.  Even the XUnit team seems to be discouraging people from using this.


almiliMSFT:
Using -parallel all only causes races and flakiness in some of our tests.

PR #315, opened 2019-05-16
Fail if IFileCopier returned the result with the wrong size.

SergeyTeplyakov:
This PR adds an ability for `DistributedContentCopier` to check the size returned by `IFileCopier` and fail if the size doesn't match the input.

The PR also adds a test class `DistributedContentCopier` that should be enhanced in the future to cover other business rules.

AP#1530234

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
@@ -236,6 +236,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.Stores
             // before each retry, clear the list of bad locations so we can retry them all.
             // this helps isolate transient network errors.
             badContentLocations.Clear();
+            string lastErrorMessage = null;
 
             foreach (MachineLocation location in hashInfo.Locations)
             {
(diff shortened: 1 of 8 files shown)


Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -359,35 +360,50 @@ void traceCopyFailed(Context c)
...
-                                if (putResult.ContentHash == hashInfo.ContentHash)
-                                {
-                                    // Don't delete the temporary file! It no longer exists after the Put moved it into the cache
-                                    deleteTempFile = false;
-                                    return (putResult, false);
-                                }
-                                else
+                                if (putResult.ContentHash != hashInfo.ContentHash)
                                 {
+                                    lastErrorMessage =
```
dannyvv: >lastErrorMessage [](start = 36, length = 16)

Since you are looping over a set of locations  would it be helfpull if the location that has an issue was part of the message? Also perhaps this should be a telemetry counter so we can keep an eye out?

SergeyTeplyakov: When the issue occurs we trace the location there. Let me think about whether it's useful to have the location in the final result. (I think you're right).

dcwuser:
Wow, this is a very different and much bigger approach than I had in mind. My version of this was much smaller. Within the distributed copy logic, I just never completed the put if the file size didn't match. It look like this version is tied up with repudiation of content that is in the cache. We do want that eventually but I don't understand why it is here for this little check. Can we talk about this offline?

Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -397,11 +413,11 @@ void traceCopyFailed(Context c)
                             {
                                 // Nothing is known about the put's failure. Give up on all locations, do not retry.
                                 // An example of a failure requiring this: Failed to reserve space for content
-                                var errorMessage = $"Put file for content hash {hashInfo.ContentHash} failed";
+                                var errorMessage = $"Put file for content hash {hashInfo.ContentHash} failed with error {putResult.ErrorMessage} ";
```
kewilli: >  [](start = 144, length = 1)

Please remove the trailing space

Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -359,35 +360,50 @@ void traceCopyFailed(Context c)
...
                                 break;
                             default:
-                                return (new ErrorResult(copyFileResult, $"{AttemptTracePrefix(attemptCount)} File copier result code {copyFileResult.Code} is not recognized").AsResult<PutResult>(), true);
+                                return (result: new ErrorResult(copyFileResult, $"{AttemptTracePrefix(attemptCount)} File copier result code {copyFileResult.Code} is not recognized").AsResult<PutResult>(), retry: true);
                         }
 
                         if (copyFileResult.Succeeded)
                         {
+                            // The copy succeeded, but it is possible that the resulting size doesn't match an expected one.
+                            if (hashInfo.Size != -1 && copyFileResult.Size != null && hashInfo.Size != copyFileResult.Size.Value)
```
dcwuser: Given the weird size values we have seen in the past, I would be slightly more careful in this test. Suggest running only if hashInfo.Size > 0.

dcwuser:
Read more carefully. I was wrong in my first understanding of what you were doing. Approach looks right. Just one substantive comment left.

---
In reply to: [493172094](https://github.com/microsoft/BuildXL/pull/315#issuecomment-493172094) [](ancestors = 493172094)

PR #316, opened 2019-05-16
Deny write access to inputs that happen to fall under shared opaques

smera:
Shared opaques define an 'allow-write' cone underneath them. However, input files may fall under a shared opaque cone. That means those input files will be write-allowed when those write accesses should be blocked instead. 

A manifestation of this is: a pip that declares a dependency on an input file and a shared opaque that contains that input is allowed to write to that file without any violations being flagged.

This PR changes the detours manifest to make sure this case is considered. In a similar line, dynamic inputs (directory dependencies to shared opaques) follow the same treatment.

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -1911,8 +1911,11 @@ namespace BuildXL.Processes
                 // The file dependency may be under the cone of a shared opaque, which will give write access
                 // to it. Explicitly block this, since we want inputs to not be written. Observe we already know 
                 // this is not a rewrite.
-                mask: m_excludeReportAccessMask & ~FileAccessPolicy.AllowRealInputTimestamps 
-                    & (pathIsUnderSharedOpaque ? ~FileAccessPolicy.AllowWrite: FileAccessPolicy.MaskNothing)); 
+                mask: m_excludeReportAccessMask &
+                      ~FileAccessPolicy.AllowRealInputTimestamps &
+                      (pathIsUnderSharedOpaque ? 
+                          ~FileAccessPolicy.AllowWrite: 
+                          FileAccessPolicy.MaskNothing)); 
 
             allInputPaths.Add(path);
 
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -1815,7 +1815,11 @@ namespace BuildXL.Processes
             m_fileAccessManifest.AddPath(
                 path,
                 values: FileAccessPolicy.AllowRead | FileAccessPolicy.AllowReadIfNonexistent | FileAccessPolicy.ReportAccess,
-                mask: ~FileAccessPolicy.AllowRealInputTimestamps);
+                // The file dependency may be under the cone of a shared opaque, which will give write access
+                // to it. Explicitly block this (no need to check if this is under a shared opaque, since otherwise
+                // it didn't have write access to begin with). Observe we already know this is not a rewrite since dynamic rewrites
+                // are not allowed by construction under shared opaques.
+                mask: ~FileAccessPolicy.AllowRealInputTimestamps & ~FileAccessPolicy.AllowWrite);
 
             allInputPathsUnderSharedOpaques.Add(path);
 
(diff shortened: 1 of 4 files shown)


Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs
```diff
@@ -976,23 +1041,24 @@ public void CacheConvergenceCanTriggerAdditionalOutputContentAwareViolations()
...
             ProcessWithOutputs originalProducerPipResult = SchedulePipBuilder(originalProducerPipBuilder);
 
             // doubleWriteProducerPip writes the same artifact in a shared opaque directory
-            ProcessBuilder doubleWriteProducerPipBuilder = CreatePipBuilder(writeOperation);
+            ProcessBuilder doubleWriteProducerPipBuilder = CreatePipBuilder(writeOperations);
             doubleWriteProducerPipBuilder.AddOutputDirectory(sharedOpaqueDirPath, SealDirectoryKind.SharedOpaque);
 
-            // Let's make doubleWriteProducerPip depend on originalProducerPip so we avoid potential file locks on the double write
-            doubleWriteProducerPipBuilder.AddInputDirectory(originalProducerPipResult.Process.DirectoryOutputs.Single());
+            // Let's make doubleWriteProducerPip depends on originalProducerPip so we avoid potential file locks on the double write
```
jbayardo: depend

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1897,18 +1901,26 @@ private void AddStaticFileDependencyToFileAccessManifest(FileArtifact dependency
...
             m_fileAccessManifest.AddPath(
                 path,
                 values: FileAccessPolicy.AllowRead | FileAccessPolicy.AllowReadIfNonexistent,
-                mask: m_excludeReportAccessMask & ~FileAccessPolicy.AllowRealInputTimestamps); // Make sure we fake the input timestamp
+                // Make sure we fake the input timestamp
+                // The file dependency may be under the cone of a shared opaque, which will give write access
+                // to it. Explicitly block this, since we want inputs to not be written. Observe we already know 
+                // this is not a rewrite.
+                mask: m_excludeReportAccessMask & ~FileAccessPolicy.AllowRealInputTimestamps 
+                    & (pathIsUnderSharedOpaque ? ~FileAccessPolicy.AllowWrite: FileAccessPolicy.MaskNothing));
```
olkononenko: nit: put every mask on a new line?

almiliMSFT: nit
```suggestion
                    & (pathIsUnderSharedOpaque ? ~FileAccessPolicy.AllowWrite : FileAccessPolicy.MaskNothing)); 
```

almiliMSFT: Would it be ok to unconditionally add `~FileAccessPolicy.AllowWrite` for all static dependencies, regardless of whether they are under a shared opaque directory or not.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1897,18 +1901,26 @@ private void AddStaticFileDependencyToFileAccessManifest(FileArtifact dependency
...
             var path = dependency.Path;
+
+            bool pathIsUnderSharedOpaque = TryGetContainingSharedOpaqueRoot(path, getOutmostRoot: true, out var sharedOpaqueRoot);
+
             m_fileAccessManifest.AddPath(
                 path,
                 values: FileAccessPolicy.AllowRead | FileAccessPolicy.AllowReadIfNonexistent,
-                mask: m_excludeReportAccessMask & ~FileAccessPolicy.AllowRealInputTimestamps); // Make sure we fake the input timestamp
+                // Make sure we fake the input timestamp
+                // The file dependency may be under the cone of a shared opaque, which will give write access
```
almiliMSFT: Doesn't the same happen when such a file is under the cone of an **exclusive** opaque directory?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs
smera: The problem of reading only a fraction of the sentence :)

---
In reply to: [284502181](https://github.com/microsoft/BuildXL/pull/316#discussion_r284502181) [](ancestors = 284502181)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
smera: Tried that first. The problem is that there can be inputs that are also untracked (e.g. a source is declared, but since it is overwritten, the output - which is the same path as the input - gets flagged as untracked). And in the way the manifest is constructed, one can only add masks but not remove them. So after adding ~Write here, there is no direct way to remove that flag in case that same path happens to be untracked.

---
In reply to: [284862416](https://github.com/microsoft/BuildXL/pull/316#discussion_r284862416) [](ancestors = 284862416)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
smera: Inputs cannot be part of exclusive opaques, those are blocked by construction

---
In reply to: [284863123](https://github.com/microsoft/BuildXL/pull/316#discussion_r284863123) [](ancestors = 284863123)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
almiliMSFT: thanks

PR #317, opened 2019-05-16
Add a way to exclude BuildXL.Explorer through env var

narasamdya:
BuildXL.Explorer takes a long time to build, and is currently in broken state.

Building it can take a long time in CB environment. This change introduces an env var that one can use to exclude building BuildXL.Explorer.

[AB#1532701](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1532701)

### DIFF
--- a/Public/Sdk/SelfHost/BuildXL/BuildXLSdk.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/BuildXLSdk.dsc
@@ -143,7 +143,7 @@ namespace Flags {
 
     /**
      * Temporary flag to exclude building BuildXL.Explorer.
-     * BuildXL.Explorer is broken but building it can take 1h in CB.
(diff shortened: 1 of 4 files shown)


Public/Sdk/SelfHost/BuildXL/BuildXLSdk.dsc
```diff
@@ -140,6 +140,13 @@ namespace Flags {
      */
     @@public
     export const genVSSolution = Environment.getFlag("[Sdk.BuildXL]GenerateVSSolution");
+
+    /**
+     * Temporary flag to exclude building BuildXL.Explorer.
+     * BuildXL.Explorer is broken but building it can take 1h in CB.
```
almiliMSFT:
```suggestion
     * BuildXL.Explorer is broken but building it can take a long time.
```

PR #319, opened 2019-05-16
[.NET Core] Remove more DISABLE_FEATURE constants

almiliMSFT:
- remove define constants:
  - DISABLE_FEATURE_HTTPEXCEPTION
  - DISABLE_FEATURE_SECURITY_ATTRIBUTES

[AB#1521781](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1521781)

### DIFF
--- a/Public/Src/FrontEnd/Core/FrontEndHostController.Nuget.cs
+++ b/Public/Src/FrontEnd/Core/FrontEndHostController.Nuget.cs
@@ -723,16 +723,7 @@ namespace BuildXL.FrontEnd.Core
                 {
                     return LogAndReturnDownloadFailure(e);
                 }
-#if !DISABLE_FEATURE_HTTPEXCEPTION
-                catch (HttpException e)
-                {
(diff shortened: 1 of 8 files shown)


Public/Src/FrontEnd/Core/FrontEndHostController.Nuget.cs
```diff
@@ -723,16 +723,7 @@ public async Task<Possible<string>> DownloadFile(LoggingContext loggingContext,
                 {
                     return LogAndReturnDownloadFailure(e);
                 }
-#if !DISABLE_FEATURE_HTTPEXCEPTION
-                catch (HttpException e)
```
Kr1sso: HttpException not needed?

almiliMSFT: Even full framework libraries now throw `HttpRequestException`

PR #322, opened 2019-05-16
Launch pips in VM

narasamdya:
This change allows us to launch pips in the VM. This change also includes more logging when `VmCommandProxy` is invoked. 

Some changes need to be added to CloudBuild's side:
- Subst `B:` in VM for `D:\dbs\el\bxlint`
- Net share `D:` in all involved host machines.


Public/Src/Utilities/Utilities/VmCommandProxy/LowPrivilegeAccountUtils.cs
```diff
@@ -12,7 +12,7 @@
 namespace BuildXL.Utilities.VmCommandProxy
 {
     /// <summary>
-    /// Class EngineSchedule.
+    /// Utilities for low privilege account.
```
norahuang: Thanks for fixing the comments

PR #323, opened 2019-05-16
Remove Sender information from openbond and protobuf abstractions

semihokur:
We pass sender information via grpc headers, so I do remove the explicit sender data from both openbond and protobuf abstractions.

### DIFF
--- a/Public/Src/Engine/Distribution.Grpc/Interfaces.proto
+++ b/Public/Src/Engine/Distribution.Grpc/Interfaces.proto
@@ -27,34 +27,20 @@ service Worker
 message RpcResponse {
 }
 
-message SenderInfo
-{
-    // The name of the sender of the message
-    string SenderName = 1;
-
-    // The id of the sender of the message
-    string SenderId = 2;
-
-    // The id of the build
-    string BuildId = 3;
-}
-
 // The response from an Attach event
 message AttachCompletionInfo
 {
-    SenderInfo Sender = 1;
-
     // Identifies the worker which makes the callback
-    uint32 WorkerId = 2;
+    uint32 WorkerId = 1;
 
     // The maximum number of simultaneous pip executions for the worker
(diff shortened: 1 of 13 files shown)


Public/Src/Engine/Distribution.Grpc/Interfaces.proto
```diff
@@ -103,49 +89,45 @@ message EventMessage
 // Information about completed pips and events logged on worker
 message WorkerNotificationArgs
 {
-    SenderInfo Sender = 1;
-
     // Identifies the worker which executed the pips
-    uint32 WorkerId = 2;
+    uint32 WorkerId = 1;
```
dannyvv: Since nobody depends on our protobuf yet, this is okay to do since it is an internal implementaiotn detail, but when this protocol is used for a running service that upgrades SxS this is considerd a bad practice and one can remove fields but should never 'renumber'.

Public/Src/Engine/Dll/Distribution/Grpc/GrpcWorkerServer.cs
```diff
@@ -56,7 +56,9 @@ public override Task<RpcResponse> Attach(BuildStartData message, ServerCallConte
         {
             var bondMessage = message.ToOpenBond();
 
-            m_workerService.AttachCore(bondMessage);
+            GrpcSettings.ParseHeader(context.RequestHeaders, out string sender, out string senderBuildId, out string traceId);
```
dannyvv: >out string senderBuildId [](start = 80, length = 24)

can use _ to indicate we are not using the fields

Public/Src/Engine/Distribution.Grpc/Interfaces.proto
semihokur: thanks for letting me know!

PR #325, opened 2019-05-17
Use regular locking in ConcurrentBoundedSortedCollection

almiliMSFT:
Previously, a `ReaderWriterLock` was used.  

On .NET Core, errors like "Attempt to release a lock that is not owned by the calling thread" were happening regularly in our CI validations.

This PR replaces this lock with a regular exclusive lock, since in our particular uses of `ConcurrentBoundedSortedCollection` we never have only concurrent reads, so it's questionable if we were getting any benefits from use a read-write lock.

[AB#1533036](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1533036)

### DIFF
--- a/Public/Src/Utilities/Collections/ConcurrentBoundedSortedCollection.cs
+++ b/Public/Src/Utilities/Collections/ConcurrentBoundedSortedCollection.cs
@@ -6,7 +6,6 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
 using System.Linq;
-using System.Threading;
 
 namespace BuildXL.Utilities.Collections
 {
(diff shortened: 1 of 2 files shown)


Public/Src/Utilities/Collections/ConcurrentBoundedSortedCollection.cs
```diff
@@ -57,62 +55,49 @@ public ConcurrentBoundedSortedCollection(int capacity, IComparer<TSort> comparer
...
             }
+
+            return false;
         }
 
         /// <inheritdoc />
         public IEnumerator<KeyValuePair<TSort, TValue>> GetEnumerator()
         {
-            return m_list.GetEnumerator();
+            lock (m_lock)
```
SergeyTeplyakov: I'm not sure this is actually helpful. Yep, we obtain enumerator under the lock, but we will use it without it causing the race condition.

almiliMSFT: But we clone the list first, so the enumerator will iterate over the clone

PR #326, opened 2019-05-17
Win Helium sandboxing: WCI disable and BindFlt exclusions

erikma:
Add the ability to disable WCI and pipe in BindFlt exceptions

Not used in this codebase but useful in AnyBuild prototype - on remote agent we use BuildXL as a library and Helium minus WCI sandboxing

### DIFF
--- a/Public/Src/Engine/Processes/Containers/ContainerConfiguration.cs
+++ b/Public/Src/Engine/Processes/Containers/ContainerConfiguration.cs
@@ -42,17 +42,35 @@ namespace BuildXL.Processes.Containers
         public IReadOnlyDictionary<AbsolutePath, IReadOnlyList<ExpandedAbsolutePath>> OriginalDirectories { get; }
 
         /// <summary>
+        /// On Windows, defines whether the WCI (Windows Container) filter driver is enabled. When false,
+        /// only BindFlt is used for output redirection.
+        /// </summary>
+        public bool IsWciFilterEnabled { get; }
+
+        /// <summary>
+        /// Paths that should not have the bindflt path transformation applied to them.
+        /// </summary>
+        public IReadOnlySet<string> BindFltExcludedPaths { get; }
+
+        /// <summary>
         /// No isolation
         /// </summary>
         public static ContainerConfiguration DisabledIsolation = new ContainerConfiguration(
                 pathTable: null,
                 redirectedDirectories: CollectionUtilities.EmptyDictionary<ExpandedAbsolutePath, IReadOnlyList<ExpandedAbsolutePath>>(),
-                originalDirectories: CollectionUtilities.EmptyDictionary<AbsolutePath, IReadOnlyList<ExpandedAbsolutePath>>());
+                originalDirectories: CollectionUtilities.EmptyDictionary<AbsolutePath, IReadOnlyList<ExpandedAbsolutePath>>(),
+                isWciFilterEnabled: false);
 
         /// <summary>
         /// Creates a configuration created for a specific process
         /// </summary>
-        public ContainerConfiguration(PathTable pathTable, IReadOnlyDictionary<ExpandedAbsolutePath, IReadOnlyList<ExpandedAbsolutePath>> redirectedDirectories, IReadOnlyDictionary<AbsolutePath, IReadOnlyList<ExpandedAbsolutePath>> originalDirectories)
+        public ContainerConfiguration(
+            PathTable pathTable,
+            IReadOnlyDictionary<ExpandedAbsolutePath, IReadOnlyList<ExpandedAbsolutePath>> redirectedDirectories,
+            IReadOnlyDictionary<AbsolutePath,
+            IReadOnlyList<ExpandedAbsolutePath>> originalDirectories,
+            bool isWciFilterEnabled = true,
+            IReadOnlySet<AbsolutePath> bindFltExcludedPaths = null)
         {
             Contract.Requires(redirectedDirectories.Count == 0 || pathTable != null);
             Contract.Requires(redirectedDirectories != null);
@@ -61,6 +79,10 @@ namespace BuildXL.Processes.Containers
             IsIsolationEnabled = redirectedDirectories.Count > 0;
             RedirectedDirectories = redirectedDirectories;
             OriginalDirectories = originalDirectories;
+            IsWciFilterEnabled = isWciFilterEnabled;
+            BindFltExcludedPaths = bindFltExcludedPaths != null ?
+                CollectionUtilities.ToReadOnlySet(bindFltExcludedPaths.Select(p => p.ToString(pathTable)))
+                : CollectionUtilities.EmptySet<string>();
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/Processes/Containers/ContainerConfiguration.cs
```diff
@@ -41,18 +41,36 @@ public sealed class ContainerConfiguration
...
+
         /// <summary>
         /// No isolation
         /// </summary>
         public static ContainerConfiguration DisabledIsolation = new ContainerConfiguration(
                 pathTable: null,
                 redirectedDirectories: CollectionUtilities.EmptyDictionary<ExpandedAbsolutePath, IReadOnlyList<ExpandedAbsolutePath>>(),
-                originalDirectories: CollectionUtilities.EmptyDictionary<AbsolutePath, IReadOnlyList<ExpandedAbsolutePath>>());
+                originalDirectories: CollectionUtilities.EmptyDictionary<AbsolutePath, IReadOnlyList<ExpandedAbsolutePath>>(),
+                isWciFilterEnabled: false);
```
dannyvv: Sorry, GitHub can't comment per word... This is for the default value of wci enabled: if we use default should it be true to match existing behavior?

Public/Src/Utilities/Native/Processes/Windows/ProcessUtilities.Win.Container.cs
```diff
@@ -217,8 +232,9 @@ private static void ConfigureWciFilter(IntPtr hJob, IReadOnlyCollection<Expanded
             
         }
 
-        private static void ConfigureBindFilter(IntPtr hJob, IReadOnlyCollection<ExpandedAbsolutePath> sourcePaths, string targetPath)
+        private static void ConfigureBindFilter(IntPtr hJob, IReadOnlyCollection<ExpandedAbsolutePath> sourcePaths, string targetPath, IEnumerable<string> exclusionPaths)
```
dannyvv: Just checking if the change in name is intentional...

Public/Src/Engine/Processes/Containers/ContainerConfiguration.cs
erikma: Total fail on my part. Fixed.

Public/Src/Utilities/Native/Processes/Windows/ProcessUtilities.Win.Container.cs
erikma: In this case yes, the method is BindFlt related

Public/Src/Utilities/Native/Processes/ProcessUtilities.cs
```diff
@@ -287,21 +287,30 @@ public static bool OSSupportsNestedJobs()
...
         /// - Bind filter is able to map a source folder into a target folder, so any process running in a container will get every access
         /// to the source path translated into an access to the target path
+        ///
         /// Working together, these filters can provide full virtualization: a process accessing a source path will be redirected under the hood
         /// to the target path, and the content that it will see there is fully controlled.
         /// </remarks>
         /// <param name="hJob">A pointer to a job object where a specific filter configuration will be associated with. 
         /// This is the result of calling CreateJobObject <see href="https://msdn.microsoft.com/en-us/library/windows/desktop/ms682409(v=vs.85).aspx"/></param>
         /// <param name="redirectedDirectories">The collection of source paths to be virtualize to destination paths</param>
+        /// <param name="isWciFilterEnabled">Enables WCI filter for inpuit virtualization</param>
```
smera: >inpuit [](start = 68, length = 6)

input

Public/Src/Utilities/Native/Processes/ProcessUtilities.cs
```diff
@@ -287,21 +287,30 @@ public static bool OSSupportsNestedJobs()
...
+        /// <param name="bindFltExclusions">Paths to not apply the bindflt path transformation to.</param>
         /// <param name="warnings">Any warnings that happened during the creation of the container. The container was created successfully regardless of these.</param>
         /// <exception cref="BuildXLException">If any unrecoverable error occurs when setting up the container</exception>
-        public static void AttachContainerToJobObject(IntPtr hJob, IReadOnlyDictionary<ExpandedAbsolutePath, IReadOnlyList<ExpandedAbsolutePath>> redirectedDirectories, out IEnumerable<string> warnings) 
-            => s_nativeMethods.AttachContainerToJobObject(hJob, redirectedDirectories, out warnings);
+        public static void AttachContainerToJobObject(
+            IntPtr hJob,
+            IReadOnlyDictionary<ExpandedAbsolutePath, IReadOnlyList<ExpandedAbsolutePath>> redirectedDirectories,
+            bool isWciFilterEnabled,
```
smera: >isWciFilterEnabled [](start = 17, length = 18)

I think a better name would be 'enableWciFilter'.
From what I can see, this flag is for instructing the API to configure WCI. The name makes it look as if we were querying something

Public/Src/Utilities/Native/Processes/ProcessUtilities.cs
```diff
@@ -287,21 +287,30 @@ public static bool OSSupportsNestedJobs()
...
         /// This is the result of calling CreateJobObject <see href="https://msdn.microsoft.com/en-us/library/windows/desktop/ms682409(v=vs.85).aspx"/></param>
         /// <param name="redirectedDirectories">The collection of source paths to be virtualize to destination paths</param>
+        /// <param name="isWciFilterEnabled">Enables WCI filter for inpuit virtualization</param>
+        /// <param name="bindFltExclusions">Paths to not apply the bindflt path transformation to.</param>
         /// <param name="warnings">Any warnings that happened during the creation of the container. The container was created successfully regardless of these.</param>
         /// <exception cref="BuildXLException">If any unrecoverable error occurs when setting up the container</exception>
-        public static void AttachContainerToJobObject(IntPtr hJob, IReadOnlyDictionary<ExpandedAbsolutePath, IReadOnlyList<ExpandedAbsolutePath>> redirectedDirectories, out IEnumerable<string> warnings) 
-            => s_nativeMethods.AttachContainerToJobObject(hJob, redirectedDirectories, out warnings);
+        public static void AttachContainerToJobObject(
```
smera: >AttachContainerToJobObject [](start = 27, length = 26)

The function looks a bit funky/asymmetric now (I mean, having the ability to disable one of the two filters, but not the other one)

Public/Src/Engine/Processes/Containers/ContainerConfiguration.cs
```diff
@@ -41,18 +41,36 @@ public sealed class ContainerConfiguration
...
+        /// <summary>
+        /// On Windows, defines whether the WCI (Windows Container) filter driver is enabled. When false,
+        /// only BindFlt is used for output redirection.
+        /// </summary>
+        public bool IsWciFilterEnabled { get; }
+
+        /// <summary>
+        /// Paths that should not have the bindflt path transformation applied to them.
+        /// </summary>
+        public IReadOnlySet<string> BindFltExcludedPaths { get; }
```
smera: >string [](start = 28, length = 6)

I'd prefer to see ExpandedAbsolutePath here, just for consistency

Public/Src/Utilities/Native/Processes/ProcessUtilities.cs
erikma: At this point I don't have a case for turning off BindFlt and keeping only WCI. Do you? Seems like overdesign if there's no matching scenario.

smera: No, feel free to ignore my comment. Was just pointing out that the signature looks slightly odd as it is, but no need to change.

PR #329, opened 2019-05-17
Avoid cache from crashing with OOM when enumerating

jbayardo:
As the cache grows large, our enumeration would run OOM due to the maximum object size limit. This change addresses this by splitting the enumeration into multiple lists.

### DIFF
--- a/Public/Src/Cache/ContentStore/InterfacesTest/FileSystem/MemoryFileSystem.cs
+++ b/Public/Src/Cache/ContentStore/InterfacesTest/FileSystem/MemoryFileSystem.cs
@@ -383,7 +383,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public virtual bool DirectoryExists(AbsolutePath path)
+        public bool DirectoryExists(AbsolutePath path)
         {
             lock (_drives)
             {
@@ -419,7 +419,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public virtual void CreateDirectory(AbsolutePath path)
+        public void CreateDirectory(AbsolutePath path)
         {
             lock (_drives)
             {
@@ -453,7 +453,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public virtual void DeleteDirectory(AbsolutePath path, DeleteOptions deleteOptions)
+        public void DeleteDirectory(AbsolutePath path, DeleteOptions deleteOptions)
         {
             lock (_drives)
             {
@@ -565,7 +565,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public virtual void MoveDirectory(AbsolutePath sourcePath, AbsolutePath destinationPath)
+        public void MoveDirectory(AbsolutePath sourcePath, AbsolutePath destinationPath)
         {
             lock (_drives)
             {
--- a/Public/Src/Cache/ContentStore/InterfacesTest/FileSystem/MemoryFileSystem.cs
+++ b/Public/Src/Cache/ContentStore/InterfacesTest/FileSystem/MemoryFileSystem.cs
@@ -162,7 +162,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public IEnumerable<FileInfo> EnumerateFiles(AbsolutePath path, EnumerateOptions options)
+        public virtual IEnumerable<FileInfo> EnumerateFiles(AbsolutePath path, EnumerateOptions options)
         {
             var infos = new List<FileInfo>();
             lock (_drives)
@@ -383,7 +383,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public bool DirectoryExists(AbsolutePath path)
+        public virtual bool DirectoryExists(AbsolutePath path)
         {
             lock (_drives)
             {
@@ -419,7 +419,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public void CreateDirectory(AbsolutePath path)
+        public virtual void CreateDirectory(AbsolutePath path)
         {
             lock (_drives)
             {
@@ -453,7 +453,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public void DeleteDirectory(AbsolutePath path, DeleteOptions deleteOptions)
+        public virtual void DeleteDirectory(AbsolutePath path, DeleteOptions deleteOptions)
         {
             lock (_drives)
             {
@@ -565,7 +565,7 @@ namespace BuildXL.Cache.ContentStore.InterfacesTest.FileSystem
         }
 
         /// <inheritdoc />
-        public void MoveDirectory(AbsolutePath sourcePath, AbsolutePath destinationPath)
+        public virtual void MoveDirectory(AbsolutePath sourcePath, AbsolutePath destinationPath)
         {
             lock (_drives)
             {
--- a/Public/Src/Cache/ContentStore/Library/Stores/ContentDirectorySnapshot.cs
+++ b/Public/Src/Cache/ContentStore/Library/Stores/ContentDirectorySnapshot.cs
@@ -43,7 +43,7 @@ namespace BuildXL.Cache.ContentStore.Stores
         private List<PayloadFromDisk<T>>[] InitializeSnapshot()
         {
             var snapshot = new List<PayloadFromDisk<T>>[byte.MaxValue];
-            for (var i = 0; i < _snapshot.Length; i++)
+            for (var i = 0; i < snapshot.Length; i++)
             {
                 snapshot[i] = new List<PayloadFromDisk<T>>();
             }
--- a/Public/Src/Cache/ContentStore/Library/Stores/ContentDirectorySnapshot.cs
+++ b/Public/Src/Cache/ContentStore/Library/Stores/ContentDirectorySnapshot.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
@@ -10,59 +9,28 @@ using BuildXL.Cache.ContentStore.Hashing;
 namespace BuildXL.Cache.ContentStore.Stores
 {
     /// <summary>
-    /// This class is specifically for usage within <see cref="ContentHashAddressableSnapshot{T}"/>. It is a reference type on purpose, in order to avoid issues
-    /// with the maximum object size limit.
-    /// </summary>
-    /// <typeparam name="T">Type tagged with a hash</typeparam>
-    public class PayloadFromDisk<T> : IComparable<PayloadFromDisk<T>>
-    {
-        /// <summary>
-        /// Hash for the <see cref="Payload"/>
-        /// </summary>
-        public readonly ContentHash Hash;
-
-        /// <summary>
-        /// Information for which <see cref="Hash"/> applies
-        /// </summary>
-        public readonly T Payload;
-
-        /// <nodoc />
-        public PayloadFromDisk(ContentHash hash, T payload)
-        {
-            Hash = hash;
-            Payload = payload;
-        }
-
-        /// <nodoc />
-        public int CompareTo(PayloadFromDisk<T> other)
-        {
-            return Hash.CompareTo(other.Hash);
-        }
-    }
-
-    /// <summary>
     /// This class represents an immutable snapshot of a collection of items at an undetermined point in time, used for enumerations of the file
     /// system. It is made specifically to avoid issues with enumerations of large amounts of files which go over the maximum object size
     /// restriction.
     /// </summary>
     /// <typeparam name="T">Type held inside the snapshot</typeparam>
-    public class ContentHashAddressableSnapshot<T> : IEnumerable<PayloadFromDisk<T>>
+    public class ContentDirectorySnapshot<T> : IEnumerable<PayloadFromDisk<T>>
     {
-        private List<PayloadFromDisk<T>>[] _snapshot;
+        private readonly List<PayloadFromDisk<T>>[] _snapshot;
 
         /// <nodoc />
-        public readonly long Count = 0;
+        public long Count { get; } = 0;
 
         /// <nodoc />
-        public ContentHashAddressableSnapshot()
+        public ContentDirectorySnapshot()
         {
-            InitializeSnapshot();
+            _snapshot = InitializeSnapshot();
         }
 
         /// <nodoc />
-        public ContentHashAddressableSnapshot(IEnumerable<PayloadFromDisk<T>> snapshot)
+        public ContentDirectorySnapshot(IEnumerable<PayloadFromDisk<T>> snapshot)
         {
-            InitializeSnapshot();
+            _snapshot = InitializeSnapshot();
 
             foreach (var payload in snapshot)
             {
@@ -71,15 +39,16 @@ namespace BuildXL.Cache.ContentStore.Stores
                 Count++;
             }
         }
-
-        /// <nodoc />
-        private void InitializeSnapshot()
+        
+        private List<PayloadFromDisk<T>>[] InitializeSnapshot()
         {
-            _snapshot = new List<PayloadFromDisk<T>>[256];
+            var snapshot = new List<PayloadFromDisk<T>>[byte.MaxValue];
             for (var i = 0; i < _snapshot.Length; i++)
             {
-                _snapshot[i] = new List<PayloadFromDisk<T>>();
+                snapshot[i] = new List<PayloadFromDisk<T>>();
             }
+
+            return snapshot;
         }
 
         /// <nodoc />
@@ -105,7 +74,7 @@ namespace BuildXL.Cache.ContentStore.Stores
             }
         }
 
-        /// <nodoc />
+        /// <inheritdoc />
         public IEnumerator<PayloadFromDisk<T>> GetEnumerator()
         {
             foreach (var bucket in _snapshot)
@@ -117,7 +86,7 @@ namespace BuildXL.Cache.ContentStore.Stores
             }
         }
 
-        /// <nodoc />
+        /// <inheritdoc />
         IEnumerator IEnumerable.GetEnumerator()
         {
             return GetEnumerator();
--- a/Public/Src/Cache/ContentStore/Library/Stores/ContentDirectorySnapshot.cs
+++ b/Public/Src/Cache/ContentStore/Library/Stores/ContentDirectorySnapshot.cs
@@ -4,6 +4,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using BuildXL.Cache.ContentStore.Hashing;
 
 namespace BuildXL.Cache.ContentStore.Stores
@@ -16,28 +17,19 @@ namespace BuildXL.Cache.ContentStore.Stores
     /// <typeparam name="T">Type held inside the snapshot</typeparam>
     public class ContentDirectorySnapshot<T> : IEnumerable<PayloadFromDisk<T>>
     {
-        private readonly List<PayloadFromDisk<T>>[] _snapshot;
+        private List<PayloadFromDisk<T>>[] _snapshot;
+        private BitArray _sorted;
 
         /// <nodoc />
-        public long Count { get; } = 0;
+        public long Count { get; private set; } = 0;
 
         /// <nodoc />
         public ContentDirectorySnapshot()
         {
             _snapshot = InitializeSnapshot();
-        }
 
-        /// <nodoc />
-        public ContentDirectorySnapshot(IEnumerable<PayloadFromDisk<T>> snapshot)
-        {
-            _snapshot = InitializeSnapshot();
-
-            foreach (var payload in snapshot)
-            {
-                var identifier = payload.Hash[0];
-                _snapshot[identifier].Add(payload);
-                Count++;
-            }
+            // Empty list is always sorted, so we initialize everything to true
+            _sorted = new BitArray(_snapshot.Length, true);
         }
         
         private List<PayloadFromDisk<T>>[] InitializeSnapshot()
@@ -52,16 +44,50 @@ namespace BuildXL.Cache.ContentStore.Stores
         }
 
         /// <nodoc />
+        public void Add(IEnumerable<PayloadFromDisk<T>> snapshot)
+        {
+            foreach (var payload in snapshot)
+            {
+                Add(payload);
+            }
+        }
+
+        /// <nodoc />
+        public void Add(PayloadFromDisk<T> payload)
+        {
+            // We are using the first byte of the hash to round-robin entries across multiple lists
+            byte identifier = payload.Hash[0];
+            _snapshot[identifier].Add(payload);
+            _sorted.Set(identifier, false);
+            Count++;
+        }
+
+        /// <nodoc />
         public List<PayloadFromDisk<T>> ListOrderedByHash()
         {
-            var result = this.ToList();
-            result.Sort();
+            var comparer = new ByHashPayloadFromDiskComparer<T>();
+            Parallel.For(0, _snapshot.Length, i =>
+            {
+                if (!_sorted.Get(i))
+                {
+                    _snapshot[i].Sort(comparer);
+                }
+            });
+
+            var result = new List<PayloadFromDisk<T>>();
+            foreach (var list in _snapshot)
+            {
+                // Each list has hashes whose most significant byte is exactly the bucket selector, so we can just append
+                result.AddRange(list);
+            }
+
             return result;
         }
 
         /// <summary>
         /// Generates a similar result to grouping the <see cref="PayloadFromDisk{T}"/> by their hashes. The order of the returned
-        /// groups may not be the same as in GroupBy.
+        /// groups may not be the same as in GroupBy. Moreover, the order across subsequent calls may not be preserved if the call
+        /// is interleaved with a call to <see cref="ListOrderedByHash"/>.
         /// </summary>
         public IEnumerable<IGrouping<ContentHash, PayloadFromDisk<T>>> GroupByHash()
         {
--- a/Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
+++ b/Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
@@ -85,7 +85,9 @@ namespace BuildXL.Cache.ContentStore.Stores
         /// <nodoc />
         public List<PayloadFromDisk<T>> ListOrderedByHash()
         {
-            return this.OrderBy(p => p.Hash).ToList();
+            var result = this.ToList();
+            result.Sort();
+            return result;
         }
 
         /// <summary>
--- a/Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
+++ b/Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
@@ -94,7 +94,10 @@ namespace BuildXL.Cache.ContentStore.Stores
             }
         }
 
-        /// <nodoc />
+        /// <summary>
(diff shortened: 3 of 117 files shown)


Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
```diff
@@ -16,6 +16,7 @@
 using BuildXL.Cache.ContentStore.Tracing;
 using BuildXL.Cache.ContentStore.Tracing.Internal;
 using BuildXL.Cache.ContentStore.Utils;
+using static BuildXL.Cache.ContentStore.Stores.FileSystemContentStoreInternal;
```
dcwuser: Please just qualify static method names with the type name. Makes code clearer.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
```diff
@@ -235,15 +234,15 @@ int findNextIndexToProcess(ContentHash lastProcessedHash)
...
+                processedBytes += contentHashes[currentHashIndex].Payload.Length;
 
                 var swTime = stopwatch.Elapsed;
                 if (swTime - progressTracker > _settings.SelfCheckProgressReportingInterval)
                 {
                     // It is possible to have multiple replicas with the same hash.
                     // We need to save the state only when *all* the replicas are processed.
                     // So we check if the next item has a different hash.
-                    if (currentHash != contentHashes[currentHashIndex + 1].hash)
+                    if (currentHash != contentHashes[currentHashIndex + 1].Hash)
```
dcwuser: What happens when currentHashIndex points to last element? (I know this + 1 was in code already, but I would like to be convinced of safety while we are looking at this code.)

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,130 @@
...
+using BuildXL.Cache.ContentStore.Hashing;
+
+namespace BuildXL.Cache.ContentStore.Stores
+{
+    /// <summary>
+    /// This class is specifically for usage within <see cref="ContentHashAddressableSnapshot{T}"/>. It is a reference type on purpose, in order to avoid issues
+    /// with the maximum object size limit.
+    /// </summary>
+    /// <typeparam name="T">Type tagged with a hash</typeparam>
+    public class PayloadFromDisk<T> : IComparable<PayloadFromDisk<T>>
```
dcwuser: Is it possible to make internal? (I know our code base is bad about this and tends to just make everything public, but I would like to change if possible.)

Public/Src/Cache/ContentStore/InterfacesTest/FileSystem/MemoryFileSystem.cs
```diff
@@ -162,7 +162,7 @@ protected virtual void Dispose(bool disposing)
         }
 
         /// <inheritdoc />
-        public IEnumerable<FileInfo> EnumerateFiles(AbsolutePath path, EnumerateOptions options)
+        public virtual IEnumerable<FileInfo> EnumerateFiles(AbsolutePath path, EnumerateOptions options)
```
dcwuser: >public [](start = 8, length = 6)

What is reason for making virtual?

Public/Src/Cache/ContentStore/Library/Stores/IContentDirectory.cs
```diff
@@ -10,6 +10,7 @@
 using BuildXL.Cache.ContentStore.Interfaces.Time;
 using BuildXL.Cache.ContentStore.Interfaces.Tracing;
 using BuildXL.Cache.ContentStore.UtilitiesCore;
+using static BuildXL.Cache.ContentStore.Stores.FileSystemContentStoreInternal;
```
dcwuser: Why?

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            _snapshot = new List<PayloadFromDisk<T>>[256];
+            for (var i = 0; i < _snapshot.Length; i++)
+            {
+                _snapshot[i] = new List<PayloadFromDisk<T>>();
+            }
+        }
+
+        /// <nodoc />
+        public IEnumerable<PayloadFromDisk<T>> EnumerateOrderedByHash()
```
dcwuser: >EnumerateOrderedByHash [](start = 47, length = 22)

Don't think I like this approach. The OrderBy call will create a backing list, and then the caller of this method creates another backing list. Just have a method that returns a list. Populate the list by the unordered enumeration. Sort the list in place. Return it.

dcwuser: This is probably my key point on this PR. Let's talk if you disagree.

---
In reply to: [285268197](https://github.com/microsoft/BuildXL/pull/329#discussion_r285268197) [](ancestors = 285268197)

Public/Src/Cache/ContentStore/InterfacesTest/FileSystem/MemoryFileSystem.cs
jbayardo: The mock filesystem inherits from `MemoryFileSystem` in order to avoid having to implement everything it does. The function is required to be virtual to be able to override it.

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
jbayardo: Making this internal implies that the Reconstruct interface would also become internal, and that will cascade to MemoryContentDirectory, which I will probably keep on cascading as it is public.

I can definitely do this if you want, but I think it will be a big enough amount of changes that it'd be better to do in a different PR, so as not to side-track this one.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
jbayardo: This was from a previous iteration, forgot to remove.

Public/Src/Cache/ContentStore/Library/Stores/IContentDirectory.cs
jbayardo: This was from a previous iteration, forgot to remove.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
jbayardo: I'm not sure why it was being done before, but it certainly seems like an error that did not trigger (probably something to do with the reporting intervals).

I have added a check to ensure we don't enter it at the last index, as the state will be saved immediately afterwards.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -1864,30 +1861,30 @@ private static RelativePath GetHashSubDirectory(ContentHash contentHash)
...
-            var contentHashes = new List<(ContentHash hash, FileInfo fileInfo)>();
+            var contentHashes = new List<PayloadFromDisk<FileInfo>>();
             if (_settings.UseNativeBlobEnumeration)
             {
-                EnumerateBlobPathsFromDisk(context, fileInfo => parseAndAccumulateContentHashes(fileInfo));
+                EnumerateBlobPathsFromDisk(context, fileInfo => ParseAndAccumulateContentHashes(fileInfo));
             }
             else
             {
                 foreach (var fileInfo in EnumerateBlobPathsFromDisk())
```
dcwuser: >EnumerateBlobPaths [](start = 41, length = 18)

Please add comments here (and elsewhere where we construct list from enumeration) describing that we were hit by large object limits, so we want to make the objects in the list reference types or small structures.

Public/Src/Cache/ContentStore/InterfacesTest/FileSystem/MemoryFileSystem.cs
dcwuser: By the way, your answers are showing up as separate comments rather than as threaded answers to my comments. I assume you are using the github web interface. Is it possible to get threaded comments in that interface?

---
In reply to: [285271628](https://github.com/microsoft/BuildXL/pull/329#discussion_r285271628) [](ancestors = 285271628)

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
dcwuser: Understood. No need to change, just wanted to make sure we aren't doing so unnecessarily.

---
In reply to: [285277556](https://github.com/microsoft/BuildXL/pull/329#discussion_r285277556) [](ancestors = 285277556)

Public/Src/Cache/ContentStore/Test/Stores/MemoryContentDirectoryTests.cs
```diff
@@ -15,6 +15,7 @@
 using ContentStoreTest.Test;
 using FluentAssertions;
 using Xunit;
+using static BuildXL.Cache.ContentStore.Stores.FileSystemContentStoreInternal;
```
dcwuser: Here's another static alias.

Public/Src/Cache/ContentStore/Library/Stores/IContentDirectory.cs
dcwuser: By the way, in addition to threading comments, when you resolve a comment, feel free to mark resolved.

---
In reply to: [285284696](https://github.com/microsoft/BuildXL/pull/329#discussion_r285284696) [](ancestors = 285284696)

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// This class is specifically for usage within <see cref="ContentHashAddressableSnapshot{T}"/>. It is a reference type on purpose, in order to avoid issues
+    /// with the maximum object size limit.
+    /// </summary>
+    /// <typeparam name="T">Type tagged with a hash</typeparam>
+    public class PayloadFromDisk<T> : IComparable<PayloadFromDisk<T>>
+    {
+        /// <summary>
+        /// Hash for the <see cref="Payload"/>
+        /// </summary>
+        public readonly ContentHash Hash;
```
SergeyTeplyakov: I think we generally prefer get-only properties over public fields and I would encourage to use them because they're easier to change in the future and the same in terms of coding.

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+        public readonly T Payload;
+
+        /// <nodoc />
+        public PayloadFromDisk(ContentHash hash, T payload)
+        {
+            Hash = hash;
+            Payload = payload;
+        }
+
+        /// <nodoc />
```
SergeyTeplyakov: very nit: <inheritdoc />

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+using BuildXL.Cache.ContentStore.Hashing;
+
+namespace BuildXL.Cache.ContentStore.Stores
+{
+    /// <summary>
+    /// This class is specifically for usage within <see cref="ContentHashAddressableSnapshot{T}"/>. It is a reference type on purpose, in order to avoid issues
+    /// with the maximum object size limit.
+    /// </summary>
+    /// <typeparam name="T">Type tagged with a hash</typeparam>
+    public class PayloadFromDisk<T> : IComparable<PayloadFromDisk<T>>
```
SergeyTeplyakov: BTW, I would slightly prefer to have an external comparer instead of making the payload comparable.

When a class is comparable, then one may think that it should be equatable as well. But this is quite a bit of unncessary work. So I would suggest to use a separate comparer that will compare this instance by Hash.

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    /// <summary>
+    /// This class represents an immutable snapshot of a collection of items at an undetermined point in time, used for enumerations of the file
+    /// system. It is made specifically to avoid issues with enumerations of large amounts of files which go over the maximum object size
+    /// restriction.
+    /// </summary>
+    /// <typeparam name="T">Type held inside the snapshot</typeparam>
+    public class ContentHashAddressableSnapshot<T> : IEnumerable<PayloadFromDisk<T>>
+    {
+        private List<PayloadFromDisk<T>>[] _snapshot;
```
SergeyTeplyakov: nit: readony?

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+                var identifier = payload.Hash[0];
+                _snapshot[identifier].Add(payload);
+                Count++;
+            }
+        }
+
+        /// <nodoc />
+        private void InitializeSnapshot()
+        {
+            _snapshot = new List<PayloadFromDisk<T>>[256];
```
SergeyTeplyakov: byte.MaxValue?

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+            foreach (var payload in snapshot)
+            {
+                var identifier = payload.Hash[0];
+                _snapshot[identifier].Add(payload);
+                Count++;
+            }
+        }
+
+        /// <nodoc />
+        private void InitializeSnapshot()
```
SergeyTeplyakov: nit:
List<PayloadFromDisk<T>>[] InitializeSnapshot() {
 // create and use the local variable.
}

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            foreach (var payload in snapshot)
+            {
+                var identifier = payload.Hash[0];
+                _snapshot[identifier].Add(payload);
+                Count++;
+            }
+        }
+
+        /// <nodoc />
```
SergeyTeplyakov: Don't have to have nodoc for private methods.

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                _snapshot[i] = new List<PayloadFromDisk<T>>();
+            }
+        }
+
+        /// <nodoc />
+        public List<PayloadFromDisk<T>> ListOrderedByHash()
+        {
+            var result = this.ToList();
+            result.Sort();
```
SergeyTeplyakov: This is fine, but another approach may be to keep the entries sorted from the beginning. What do you think about this approach?

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// <typeparam name="T">Type held inside the snapshot</typeparam>
+    public class ContentHashAddressableSnapshot<T> : IEnumerable<PayloadFromDisk<T>>
+    {
+        private List<PayloadFromDisk<T>>[] _snapshot;
+
+        /// <nodoc />
+        public readonly long Count = 0;
+
+        /// <nodoc />
+        public ContentHashAddressableSnapshot()
```
SergeyTeplyakov: This type is immutable. Does it make sense to have a separate constructor for creating an empty instance?

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// system. It is made specifically to avoid issues with enumerations of large amounts of files which go over the maximum object size
+    /// restriction.
+    /// </summary>
+    /// <typeparam name="T">Type held inside the snapshot</typeparam>
+    public class ContentHashAddressableSnapshot<T> : IEnumerable<PayloadFromDisk<T>>
+    {
+        private List<PayloadFromDisk<T>>[] _snapshot;
+
+        /// <nodoc />
+        public readonly long Count = 0;
```
SergeyTeplyakov: Could you please use public property instead?

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+            foreach (var bucket in _snapshot)
+            {
+                foreach (var group in bucket.GroupBy(p => p.Hash))
+                {
+                    yield return group;
+                }
+            }
+        }
+
+        /// <nodoc />
```
SergeyTeplyakov: nit: inheritdoc

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+            foreach (var bucket in _snapshot)
+            {
+                foreach (var payload in bucket)
+                {
+                    yield return payload;
+                }
+            }
+        }
+
+        /// <nodoc />
```
SergeyTeplyakov: nit: inheritdoc

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
SergeyTeplyakov: Coud you please move it to a separate file?

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    /// <summary>
+    /// This class represents an immutable snapshot of a collection of items at an undetermined point in time, used for enumerations of the file
+    /// system. It is made specifically to avoid issues with enumerations of large amounts of files which go over the maximum object size
+    /// restriction.
+    /// </summary>
+    /// <typeparam name="T">Type held inside the snapshot</typeparam>
+    public class ContentHashAddressableSnapshot<T> : IEnumerable<PayloadFromDisk<T>>
```
SergeyTeplyakov: I'm not sure I like the name. I know that this thing is more specific that ContentDirectorySnapshot<T>, but I think I better prefer more specific but easier to understand term.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
```diff
@@ -201,17 +200,15 @@ private void RemoveSelfCheckStateFromDisk(Context context)
...
                 // The loop was interrupted. Saving an incremental state.
-                var newStatus = selfCheckState.WithEpochAndPosition(_settings.SelfCheckEpoch, contentHashes[index].hash);
+                var newStatus = selfCheckState.WithEpochAndPosition(_settings.SelfCheckEpoch, contentHashes[index].Hash);
                 UpdateSelfCheckStateOnDisk(context, newStatus);
             }
 
             return Result.Success(new SelfCheckResult(invalidHashes: invalidEntries, totalProcessedFiles: processedEntries));
 
-            int findNextIndexToProcess(ContentHash lastProcessedHash)
+            int FindNextIndexToProcess(ContentHash lastProcessedHash)
```
SergeyTeplyakov: we use camelCase naming convention for local functions. So I think we should rename it back.

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
jbayardo: This is convenient when testing

---
In reply to: [285318012](https://github.com/microsoft/BuildXL/pull/329#discussion_r285318012) [](ancestors = 285318012)

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
jbayardo: That approach would incur unnecessary work in one of the use-cases as it stands today. I think it'd be better to keep it like this, and change it in the future if the majority of the workload ends up being this function.

---
In reply to: [285317930](https://github.com/microsoft/BuildXL/pull/329#discussion_r285317930) [](ancestors = 285317930)

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
jbayardo: This was done on purpose. It is more error-prone to duplicate code, and code won't compile if you attempt to compare, so it won't go silently.

---
In reply to: [285317541](https://github.com/microsoft/BuildXL/pull/329#discussion_r285317541) [](ancestors = 285317541)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -611,23 +611,22 @@ private void DeleteTempFolder()
...
             try
             {
-                var contentHashes = ReadContentHashesFromDisk(context);
-                _tracer.Debug(context, $"Enumerated {contentHashes.Count} entries by {stopwatch.ElapsedMilliseconds}ms.");
+                var contentHashes = ReadSnapshotFromDisk(context);
+                _tracer.Debug(context, $"Enumerated {contentHashes.Count} entries in {stopwatch.ElapsedMilliseconds}ms.");
 
-                var hashInfoPairs = new List<KeyValuePair<ContentHash, ContentFileInfo>>();
-                foreach (var grouping in contentHashes.GroupBy(hash => hash))
+                // We are using a list of classes instead of structs due to the maximum object size restriction
```
SergeyTeplyakov: This is very minor, but I think it make sense to move this grouping logic into a factory method inside ContentDirectorySnapshot? In this case it'll have more "behavior" and stop being just plain data, IMO.

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
SergeyTeplyakov: I'm fine with this approach, but the alternative is not more error-prone:

public class ByHashPayloadFromDiskComparer<T> : IComparer<PayloadFromDisk<T>> {
  public int Compare(PayloadFromDisk<T> left, PayloadFromDisk<T> right) => left.Hash.CompareTo(right.Hash);
}

---
In reply to: [285320584](https://github.com/microsoft/BuildXL/pull/329#discussion_r285320584) [](ancestors = 285320584,285317541)

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
SergeyTeplyakov: I see. An alternative would be still to order by each "bucket" individually and yield return the result.

I think in this case the O complexity is still the same, but actual time spend should be smaller.

And as a plus side we can sort all the "buckets" in parallel.

---
In reply to: [285320486](https://github.com/microsoft/BuildXL/pull/329#discussion_r285320486) [](ancestors = 285320486,285317930)

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
SergeyTeplyakov: I disagree with this approach: in tests you could have a helper that will pass an empty enumerable.

---
In reply to: [285320330](https://github.com/microsoft/BuildXL/pull/329#discussion_r285320330) [](ancestors = 285320330,285318012)

Public/Src/Cache/ContentStore/Library/Stores/ContentDirectorySnapshot.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        /// <nodoc />
+        public ContentDirectorySnapshot(IEnumerable<PayloadFromDisk<T>> snapshot)
+        {
+            _snapshot = InitializeSnapshot();
+
+            foreach (var payload in snapshot)
+            {
+                var identifier = payload.Hash[0];
```
dannyvv: >var [](start = 16, length = 3)

We should assert that this is a byte, if this is not a byte, then we should use an `unchecked` region to let it overlow into the byte.

dannyvv: I see now that all is fine, perhaps a comment stating taking the first byte of the hash to round-robin the entries accross multiple lists.

---
In reply to: [285707794](https://github.com/microsoft/BuildXL/pull/329#discussion_r285707794) [](ancestors = 285707794)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -611,23 +611,22 @@ private void DeleteTempFolder()
...
 
-                    hashInfoPairs.Add(
-                        new KeyValuePair<ContentHash, ContentFileInfo>(
-                            grouping.Key.hash,
-                            contentFileInfo));
+                    hashInfoPairs.Add(new PayloadFromDisk<ContentFileInfo>(grouping.Key, contentFileInfo));
                 }
 
-                return hashInfoPairs;
+                return new ContentDirectorySnapshot<ContentFileInfo>(hashInfoPairs);
```
dannyvv: >rectorySnapshot<ContentFileInfo>(hashInfoPairs); [](start = 36, length = 48)

Note: This is double-allocating the lists and dynamically growing the lists twice.. (i.e. the list in hashInfoPairs as well as the listss in the snapshot...
Since  you are already looping here... should the snapshot accept a single payloadfromdisk at a time?

Public/Src/Cache/ContentStore/Library/Stores/ContentDirectorySnapshot.cs
```diff
@@ -0,0 +1,95 @@
...
+            var result = this.ToList();
+            result.Sort();
+            return result;
+        }
+
+        /// <summary>
+        /// Generates a similar result to grouping the <see cref="PayloadFromDisk{T}"/> by their hashes. The order of the returned
+        /// groups may not be the same as in GroupBy.
+        /// </summary>
+        public IEnumerable<IGrouping<ContentHash, PayloadFromDisk<T>>> GroupByHash()
```
dcwuser: Not sure this will work. What if a hash appears in more than one bucket?

SergeyTeplyakov: I think this is fine, because we group all the hashes by it's first character in the first place.

---
In reply to: [285727199](https://github.com/microsoft/BuildXL/pull/329#discussion_r285727199) [](ancestors = 285727199)

dcwuser: Oh, right. I withdraw comment.

---
In reply to: [285727836](https://github.com/microsoft/BuildXL/pull/329#discussion_r285727836) [](ancestors = 285727836,285727199)

Public/Src/Cache/ContentStore/Library/Stores/ContentHashAddressableSnapshot.cs
dcwuser: These ideas are valid but are getting toward the side of adding WAY too much code. I would like to get the basic structure in place, then develop further if it works and that is justified. Please just sort each individually and append to list. (If you want to factor into a method that yields a sorted IEnumerable and a second method that uses it to produce a sorted list via appending, fine.) More than that is too much.

---
In reply to: [285690964](https://github.com/microsoft/BuildXL/pull/329#discussion_r285690964) [](ancestors = 285690964,285320486,285317930)

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
jbayardo: This is the only usage of the method, it seems a little bit overkill to make this change for now, as it would not be saving us from duplicate code.

---
In reply to: [285689134](https://github.com/microsoft/BuildXL/pull/329#discussion_r285689134) [](ancestors = 285689134)

PR #332, opened 2019-05-17
Add intermediate output path predictor

smera:
The copy task predictor is predicting as a source file a path that is actually an output. This is because the copy task is publishing the produced exe to the publish location, so even though the exe is indeed a source file for the task, it is not really a source file for the project. But this causes bxl to block writes on that path, since it believes it is an input.
This PR adds a predictor for the intermediate output path. This makes the current heuristic about inputs (which discards predicted inputs on output paths) kick in and avoid this problem.
Additionally, predicting the intermediate output path is a nice to have as well.

### DIFF
--- a/Public/External/BuildPrediction/BuildPrediction/StandardPredictors/IntermediateOutputPathIsOutputDir.cs
+++ b/Public/External/BuildPrediction/BuildPrediction/StandardPredictors/IntermediateOutputPathIsOutputDir.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using Microsoft.Build.Evaluation;
+using Microsoft.Build.Execution;
+
+namespace Microsoft.Build.Prediction.StandardPredictors
+{
+    /// <summary>
+    /// Scrapes the $(IntermediateOutputPath) if not found.
+    /// </summary>
+    internal class IntermediateOutputPathIsOutputDir : IProjectStaticPredictor
+    {
+        internal const string IntermediateOutputPathMacro = "IntermediateOutputPath";
+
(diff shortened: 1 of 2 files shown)


Public/External/BuildPrediction/BuildPrediction/StandardPredictors/IntermediateOutputPathIsOutputDir.cs
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using Microsoft.Build.Evaluation;
+using Microsoft.Build.Execution;
+
+namespace Microsoft.Build.Prediction.StandardPredictors
+{
+    /// <summary>
+    /// Scrapes the $(IntermediateOutputPath) if not found.
```
erikma: >if not found [](start = 46, length = 12)

if found

Public/External/BuildPrediction/BuildPredictionTests/StandardPredictors/IntermediateOutputPathIsOutputDirTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        static IntermediateOutputPathIsOutputDirTests()
+        {
+            MsBuildEnvironment.Setup(TestHelpers.GetAssemblyLocation());
+        }
+
+        [Fact]
+        public void IntermediateOutputPathFoundAsOutputDir()
+        {
+            const string IntermediateOutputPath = @"C:\repo\bin\x64";
```
erikma: >C:\repo\bin\x64 [](start = 52, length = 15)

nit: Consider 1 more case with relative path to cover both parts of the rooted path check in the predictor

smera: will do

PR #333, opened 2019-05-17
ClientConnectionManager becoming aware of channel state transitions

semihokur:
If the worker gets disconnected in an unexpected way, the master is timing out. These changes fix that.

### DIFF
--- a/Public/Src/Engine/Dll/Distribution/Grpc/ClientConnectionManager.cs
+++ b/Public/Src/Engine/Dll/Distribution/Grpc/ClientConnectionManager.cs
@@ -25,6 +25,8 @@ namespace BuildXL.Engine.Distribution.Grpc
         internal readonly Channel Channel;
         private readonly LoggingContext m_loggingContext;
         private readonly string m_buildId;
+        private readonly Task m_monitorConnectionTask;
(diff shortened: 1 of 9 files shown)


Public/Src/Engine/Dll/Distribution/Grpc/ClientConnectionManager.cs
```diff
@@ -43,11 +45,34 @@ public ClientConnectionManager(LoggingContext loggingContext, string ipAddress,
...
+        }
+
+        public async Task MonitorConnectionAsync()
+        {
+            await Task.Yield();
+
+            ChannelState state = ChannelState.Idle;
+            while (state != ChannelState.Shutdown)
+            {
+                await Channel.WaitForStateChangedAsync(state);
```
dannyvv: >WaitForStateChangedAsync [](start = 30, length = 24)

Do we trust this to not get into a hot loop?

semihokur: yes. tested in many builds

---
In reply to: [285705092](https://github.com/microsoft/BuildXL/pull/333#discussion_r285705092) [](ancestors = 285705092)

PR #338, opened 2019-05-18
Prevent Detours from modifying access parameters for untracked files

narasamdya:
BuildXL modifies access parameters on CreateFile-like functions even when the file passed to those functions is untracked. For example, we explicitly always add `FILE_SHARE_DELETE` or, on certain condition, downgrade desired access to `GENERIC_READ`. Such modifications are undesirable for unit tests, for example, tests that expect to have sharing violation. Good news is, those tests are typically run in a sandbox where the files are untracked. 

This change prevents Detours from modifying access parameters for untracked file.

[AB#1530341](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1530341)

### DIFF
--- a/Public/Src/Engine/UnitTests/Processes.Detours/PipExecutorDetoursTest.cs
+++ b/Public/Src/Engine/UnitTests/Processes.Detours/PipExecutorDetoursTest.cs
@@ -6163,6 +6163,46 @@ namespace Test.BuildXL.Processes.Detours
             return FileArtifact.CreateSourceFile(AbsolutePath.Create(pathTable, expandedPath));
         }
 
+        [Fact]
+        public async Task CallDeleteWithoutSharing()
+        {
+            var context = BuildXLContext.CreateInstanceForTesting();
+            var pathTable = context.PathTable;
+
+            using (var tempFiles = new TempFileStorage(canGetFileNames: true, rootPath: TemporaryDirectory))
+            {
+                var untrackedFile = tempFiles.GetFileName(pathTable, "untracked.txt");
+                WriteFile(pathTable, untrackedFile, "real");
+
+                var process = CreateDetourProcess(
+                    context,
+                    pathTable,
+                    tempFiles,
+                    argumentStr: "CallDeleteWithoutSharing",
+                    inputFiles: ReadOnlyArray<FileArtifact>.Empty,
+                    inputDirectories: ReadOnlyArray<DirectoryArtifact>.Empty,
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/UnitTests/Processes.Detours/PipExecutorDetoursTest.cs
```diff
@@ -6163,6 +6163,46 @@ private static FileArtifact WriteFile(PathTable pathTable, AbsolutePath filePath
             return FileArtifact.CreateSourceFile(AbsolutePath.Create(pathTable, expandedPath));
         }
 
+        [Fact]
+        public async Task CallDeleteWithoutSharing()
```
mpysson: >CallDeleteWithoutSharing [](start = 26, length = 24)

nit - could you add some comments or tweak the name to make it more clear what you're testing? UntrackedScopesDoNotGetFileShareDeleteAdded()?

Public/Src/Engine/UnitTests/Processes.Detours/PipExecutorDetoursTest.cs
```diff
@@ -6163,6 +6163,46 @@ private static FileArtifact WriteFile(PathTable pathTable, AbsolutePath filePath
...
+                    ignoreZwRenameFileInformation: false,
+                    monitorNtCreate: true,
+                    ignoreRepPoints: false,
+                    disableDetours: false,
+                    context: context,
+                    pip: process,
+                    errorString: out errorString);
+
+                VerifyNoFileAccesses(result);
+                VerifySharingViolation(context, result);
```
mpysson: >VerifySharingViolation(context, result); [](start = 16, length = 40)

Do we have test coverage for the other side of this? That when you are interacting with files that are tracked fileshare delete is added so other processes can delete?

narasamdya: I remove `FILE_SHARE_DELETE` altogether, so that we don't have any modification except what is explicitly requested, e.g., downgrading and force read sharing

Public/Src/Engine/UnitTests/Processes.Detours/PipExecutorDetoursTest.cs
narasamdya: Added unit test to cover the tracked path.

PR #340, opened 2019-05-18
Add MsBuildResolver and CMakeResolver examples

jbayardo:


Examples/MsBuildHelloWorld/config.bc
```diff
@@ -0,0 +1,9 @@
+config({
+    resolvers: [
+        {
+            kind: "MsBuild",
+            root: d`.`,
```
dannyvv: side-note: Should we rename this to projectRoot like cmake?

smera: We could... don't have a strong opinion about this :)

---
In reply to: [285703454](https://github.com/microsoft/BuildXL/pull/340#discussion_r285703454) [](ancestors = 285703454)

cdmihai: Inconsistency raises the WTF/minute metric :)

---
In reply to: [285705018](https://github.com/microsoft/BuildXL/pull/340#discussion_r285705018) [](ancestors = 285705018,285703454)

dannyvv:
Not needed in this PR, but we should file a workitem to have these examples build as part of our unittests.

PR #344, opened 2019-05-20
Change BuildXLLkgVersion.cmd to use win-x64 instead of net472

almiliMSFT:


Shared/Scripts/BuildXLLkgVersionPublic.cmd
```diff
@@ -1,4 +1,5 @@
-﻿REM @echo off
+REM @echo off
+REM TODO: update LKG_NAME to win-x64 once we start publishing that nuget to MyGet
```
almiliMSFT: This has too wait until we start publishing Microsoft.BuildXL.win-x64 to MyGet.

The pipelines have been updated, just need to wait until we start publishing

PR #346, opened 2019-05-20
Avoid NRE when GRPC layer is not initialized properly

SergeyTeplyakov:
If a content directory fails to initialize then the grpc part remains uninitialized causing NRE during shutdown.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
+++ b/Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
@@ -450,7 +450,10 @@ namespace BuildXL.Cache.ContentStore.Service
 
             _portDisposer?.Dispose();
 
-            await _grpcServer.KillAsync();
+            if (_grpcServer != null)
+            {
+                await _grpcServer.KillAsync();
+            }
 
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
```diff
@@ -450,7 +450,10 @@ protected override async Task<BoolResult> ShutdownCoreAsync(OperationContext con
 
             _portDisposer?.Dispose();
 
-            await _grpcServer.KillAsync();
+            if (_grpcServer != null)
+            {
+                await _grpcServer.KillAsync();
```
dannyvv: does null coalescing not work with async?

SergeyTeplyakov: await _grpcServer?.KillAsync() will throw NRE if the field is null because this is equivalent to:

await (_grpcServer != nul ? _grpcServer.KillAsync() : null);

Meaning the code will be transformed into 'await null'.

---
In reply to: [285773484](https://github.com/microsoft/BuildXL/pull/346#discussion_r285773484) [](ancestors = 285773484)

PR #347, opened 2019-05-20
Avoiding contract violation when EnsureInitialized is called before S…

SergeyTeplyakov:
…tartupAsync is done

`QuotaKeeper` may start purging at startup and complete this procedure before the LLS is initialized.

This PR adds an "uber" initialization task that the code awaits on regardless of if `StartupAsync` is already called or not.

Fix for AP#1532692

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
@@ -27,7 +27,6 @@ using BuildXL.Cache.ContentStore.UtilitiesCore;
 using BuildXL.Cache.ContentStore.Utils;
 using BuildXL.Native.IO;
 using BuildXL.Utilities.Collections;
-using BuildXL.Utilities.Tasks;
 using BuildXL.Utilities.Tracing;
 using DateTimeUtilities = BuildXL.Cache.ContentStore.Utils.DateTimeUtilities;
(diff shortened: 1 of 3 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -111,6 +112,12 @@ public sealed class LocalLocationStore : StartupShutdownBase
         /// </summary>
         private Task<BoolResult> _postInitializationTask;
 
+        /// <summary>
+        /// This field represents a task that is finishes when the initialization is fully done.
```
kewilli: >is finishes [](start = 46, length = 11)

Nit: Either `finishes` or `is finished`

ljcollins25:

            
        public override Task<BoolResult> StartupAsync(Context context)
        {
            var startupTask = base.StartupAsync(context);
            _startupCompletion.LinkToTask(startupTask);
            return startupTask;
        }


// Ensure startup was called then wait for it to complete successfully (or error)
            Contract.Assert(StartupStarted);
            _startupCompletion.Task.GetAwaiter().GetResult().ThrowIfFailure();

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs:1 in 1c9a6a6. [](commit_id = 1c9a6a608d798e91e822c5e3b2ca8eea8bc74683, deletion_comment = False)

PR #350, opened 2019-05-21
Turn off auto-including response files and node reuse for the MSBuildResolver

smera:


erikma:
What prompted this change? (Should the response be a comment in PipConstructor as to why we're doing it?)

smera:
> What prompted this change? (Should the response be a comment in PipConstructor as to why we're doing it?)

This is a relatively new option in MSBuild that is meant for running just 'msbuild' on a given directory and a response file will be searched (looking up the dir structure, I think) and used. This is never the case for bxl, where 'defaults' are decided using a separate mechanism.

We could expose this as an opt-in option, but I didn't see the case for it so far

PR #355, opened 2019-05-22
[.NET Core] EventWaitHandle.TryOpenExisting with Synchronize rights only

almiliMSFT:
.NET Core does not have a method for opening an existing `EventWaitHandle` with access rights explicitly specified, i.e., `EventWaitHandle` doesn't have a static method like
```csharp
bool TryOpenExisting(string name, EventWaitHandleRights rights, out EventWaitHandle result)
```
Both .NET Core and .NET Framework do have 
```csharp
bool TryOpenExisting(string name, out EventWaitHandle result)
```
which implicitly requests the `EventWaitHandleRights.Synchronize | EventWaitHandleRights.Modify` rights (see [EventWaitHandle.OpenExistingWorker](https://github.com/dotnet/corefx/blob/b86783ef9525f22b0576278939264897464f9bbd/src/Common/src/CoreLib/System/Threading/EventWaitHandle.Windows.cs#L14)).

Currently, when we create an `EventWaitHandle`, we only allow `EventWaitHandleRights.Synchronize`.  Consequently, if we try to open that same event handle from a .NET Core process, we get an access denied error:
```
System.UnauthorizedAccessException: Access to the path 'Global\ContentServerReadyEvent' is denied.
```

This PR:
  - implements a workaround where a lower-level `OpenEventW` (from "kernel32.dll") is p-invoked to achieve the existing behavior of opening an existing `EventWaitHandle` with `Synchronize` rights only.

[AB#1541060](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1541060)

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Utils/IpcUtilities.cs
+++ b/Public/Src/Cache/ContentStore/Library/Utils/IpcUtilities.cs
@@ -137,13 +137,17 @@ namespace BuildXL.Cache.ContentStore.Utils
         [DllImport("kernel32", EntryPoint = "OpenEventW", SetLastError = true, CharSet = CharSet.Unicode)]
         internal static extern SafeWaitHandle OpenEvent(uint desiredAccess, bool inheritHandle, string name);
 
-        private static ConstructorInfo EventWaitHandleConstructor =
-            typeof(EventWaitHandle).GetConstructor(BindingFlags.NonPublic | BindingFlags.Instance, null, new[] { typeof(SafeWaitHandle) }, null);
+        private static ConstructorInfo EventWaitHandleConstructor = typeof(EventWaitHandle).GetConstructor(
+            BindingFlags.NonPublic | BindingFlags.Instance,
+            null,
+            new[] { typeof(SafeWaitHandle) },
+            null);
 
         private static bool EventWaitHandleTryOpenExisting(string name, out EventWaitHandle handle)
         {
             // In .NET Core, only EventWaitHandle.TryOpenExisting(name, out handle) is supported,
-            // which requires both Synchronize and Modify rights on the handle (see: https://github.com/dotnet/corefx/blob/b86783ef9525f22b0576278939264897464f9bbd/src/Common/src/CoreLib/System/Threading/EventWaitHandle.Windows.cs#L14)
+            // which requires both Synchronize and Modify rights on the handle
+            // (see: https://github.com/dotnet/corefx/blob/b86783ef9525f22b0576278939264897464f9bbd/src/Common/src/CoreLib/System/Threading/EventWaitHandle.Windows.cs#L14)
             //
             // When we create this handle, we grant only Synchronize, which is why here we have
             // to work around this issue by directly p-invoking OpenEvent from kernel32.dll.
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Library/Utils/IpcUtilities.cs
```diff
@@ -183,10 +200,13 @@ private static class EventWaitHandleAccessRules
...
             {
-                var currentUser = UserUtilities.CurrentUserName();
+                // In CoreCLR, EventWaitHandle.TryOpenExisting() does not support setting access rights explicitly.
+                // Instead, a handle is always opened with EventWaitHandleRights.Synchronize | EventWaitHandleRights.Modify
+                // (see: https://github.com/dotnet/corefx/blob/b86783ef9525f22b0576278939264897464f9bbd/src/Common/src/CoreLib/System/Threading/EventWaitHandle.Windows.cs#L14)
+                // Because of that, here we have to request both Synchronize and Modify rights.
                 return new EventWaitHandleAccessRule(
                     new SecurityIdentifier(WellKnownSidType.WorldSid, null),
-                    EventWaitHandleRights.Synchronize,
+                    EventWaitHandleRights.Synchronize | EventWaitHandleRights.Modify,
```
dannyvv: >ventWaitHandleRights.Modif [](start = 57, length = 26)

I know this makes it 'work' but it does open up a security vulnerability... If BuildXL server process runs, now any process of any user on the machine has the ability to unblock the waithandle and cause issues for the user building.

dannyvv:
@mpysson is added to the review. #Closed

Public/Src/Cache/ContentStore/Library/Utils/IpcUtilities.cs
```diff
@@ -183,10 +200,13 @@ private static class EventWaitHandleAccessRules
         {
             public static EventWaitHandleAccessRule PublicSynchronizeAccessRule()
             {
-                var currentUser = UserUtilities.CurrentUserName();
+                // In CoreCLR, EventWaitHandle.TryOpenExisting() does not support setting access rights explicitly.
+                // Instead, a handle is always opened with EventWaitHandleRights.Synchronize | EventWaitHandleRights.Modify
```
dannyvv: MIke: can you weigh in since you did the initial security work of the pipes and event handles.

dannyvv: I know this makes it 'work' but it does open up a security vulnerability... If BuildXL server process runs, now any process of any user on the machine has the ability to unblock the waithandle and cause issues for the user building.

almiliMSFT: True, any process can call `Set()` on it with this change.  We can revert this change and just go with the hacky `EventWaitHandleTryOpenExisting` (that works too).

almiliMSFT: Reverted

almiliMSFT: reverted

PR #356, opened 2019-05-22
Switch to Microsoft.DotNet.XUnitConsoleRunner

Kr1sso:
With xUnit version 2.4.1 being the last release before 3.0, the xUnit team does not accept changes / PRs anymore - especially because the concurrency issues are in unsupported code to begin with. The Dotnet Arcade repo publishes their own XUnitConsoleRunner with all the fixes we need included. This PR switches the console runner assembly for .NETCoreApp 3.0 runs.


dannyvv:
@huihu: Do you know if QTest will pick up this better version of XUnit? Else we might still have issues when run under qtest

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -867,9 +867,9 @@ private static bool ShouldRemoveEmptyDirectories(IConfiguration configuration, s
 
             if (OperatingSystemHelper.IsUnixOS)
             {
-                // Don't scrub the .NET Core lock file when running the CoreCLR on Unix even if its parent directory is specified as scrubabble. 
-                // Some build tools use the '/tmp' folder as temporary file location (e.g. xcodebuild, clang, etc.) for dumping state and reports. 
-                // Unfortunately scrubbing the dotnet state files can lead to a missbehaving CoreCLR in subsequent or parallel runs where several 
+                // Don't scrub the .NET Core lock file when running the CoreCLR on Unix even if its parent directory is specified as scrubbable.
+                // Some build tools use the '/tmp' folder as temporary file location (e.g. xcodebuild, clang, etc.) for dumping state and reports.
+                // Unfortunately scrubbing the dotnet state files can lead to a miss-behaving CoreCLR in subsequent or parallel runs where several
```
almiliMSFT:
```suggestion
                // Unfortunately scrubbing the dotnet state files can lead to a misbehaving CoreCLR in subsequent or parallel runs where several
```

norahuang:
> @Huihu: Do you know if QTest will pick up this better version of XUnit? Else we might still have issues when run under qtest

I don't think so. Qtest use the xunit runner. For selfhost build, it use the standard xunit.runner.visualstudio one, with version 2.4.1. Office has their own xunit runner adaptor. So we do not need to care their case.

PR #358, opened 2019-05-22
PAT based VSTS auth for macOS only

Kr1sso:
Only do PAT based authentication when targeting macOS BuildXL


Public/Src/Cache/VerticalStore/BuildCacheAdapter/BuildCacheUtils.cs
```diff
@@ -50,7 +50,7 @@ internal static class BuildCacheUtils
 
             VssCredentialsFactory credentialsFactory;
 
-#if !FEATURE_CORECLR
+#if !PLATFORM_OSX
```
almiliMSFT: 👍

PR #360, opened 2019-05-22
USN record header's size does not match valid USN_RECORD_V3

norahuang:
Fixes [AB#1531444](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1531444)


Public/Src/Utilities/Native/IO/Windows/FileSystem.Win.cs
```diff
@@ -1349,7 +1349,7 @@ public unsafe NtStatus FlushPageCacheToFilesystem(SafeFileHandle handle)
                         Contract.Assume(
                             currentRecordHeader->RecordLength >= NativeUsnRecordV2.MinimumSize &&
                             currentRecordHeader->RecordLength <= NativeUsnRecordV2.MaximumSize,
-                            "Size in record header does not correspond to a valid USN_RECORD_V2.");
+                            "Size in record header does not correspond to a valid USN_RECORD_V2. Header record length:" + currentRecordHeader->RecordLength);
```
narasamdya: Convert  `Contract.Assume(x, message)` to
```
if (!x) { Contract.Assert(false, message); }
```

With the above change, you are doing string concat in a method called in hot loop.

norahuang: I don't see the difference between Contract.Assume(x, message) and if (!x) { Contract.Assert(false, message); } here. The Concat for the message still need to be happen in the loop. 
Can you please explain about what is your concern here?

dannyvv: In the first case, the message is always computed taking up cpu cycles and because it are strings also gives extra allocations.
In the second case, the full final message is only computed when it about to log it.

Frequently this doesn't mater on low-volume methods, but the filesystem is such low level that it is very perf sensitive.

There is an Info message in one of the analyzers. RAXXXX reported on each compile. This should probably be upgraded to a warning in debug and error in release.

norahuang: Thank you very much, Danny. Will change it.

PR #361, opened 2019-05-22
Report the existence of outputs on workers

semihokur:
Fixes [AB#1530983](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1530983)

The existence of all types of outputs was not reported on workers. Also, the existence of dynamic directory outputs was not reported on master. 

An example: If a pip probes one of the parent directories of an output (static or dynamic), it becomes ExistingDirectoryProbe after execution. If we do cache lookup for that pip in the next build, that ObservedInput would be AbsentPathProbe, causing a cachemiss.


semihokur:
The unit test does not seem possible because I could not simulate 'Directory Probe' with 'Operation' class. We do have Probe there; but even if the pip calls 'Directory.Exists' or asks for directory attributes, Detours does not report that access.

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -281,6 +281,10 @@ public bool ReportWorkerPipInputContent(LoggingContext loggingContext, FileArtif
             Contract.Assert(IsDistributedWorker);
 
             SetFileArtifactContentHashResult result = SetFileArtifactContentHash(artifact, info, PipOutputOrigin.NotMaterialized);
+
+            // Notify the host with content that was reported
+            m_host.ReportContent(artifact, info, PipOutputOrigin.NotMaterialized);
```
dannyvv: Shouldn't this be ReportDynamicCOntent?

mpysson: It's used down on line 590

dannyvv: OOPS :) Sorry missed that change :)

---
In reply to: [286714023](https://github.com/microsoft/BuildXL/pull/361#discussion_r286714023) [](ancestors = 286714023)

mpysson:
> @ljcollins25 and @olkononenko The unit test does not seem possible because I could not simulate 'Directory Probe' with 'Operation' class. We do have Probe there; but even if the pip calls 'Directory.Exists' or asks for directory attributes, Detours does not report that access.

You may need to pinvoke in the test process. It looks like the BCL Directory.Exists() ends up calling win32 FindFirstFile() which detours probably doesn't classify the same way. Please add a UT if all possible before merging

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5410,6 +5410,20 @@ void IFileContentManagerHost.ReportContent(FileArtifact artifact, in FileMateria
...
+        {
+            Contract.Requires(artifact.IsValid);
+            Contract.Requires(artifact.IsOutputFile);
+
+            if (origin != PipOutputOrigin.NotMaterialized)
+            {
+                State.FileSystemView.ReportRealFileSystemExistence(artifact.Path, PathExistence.ExistsAsFile);
+            }
+
+            State.FileSystemView.ReportOutputFileSystemExistence(artifact.Path, PathExistence.ExistsAsFile);
```
olkononenko: >PathExistence.ExistsAsFile [](start = 80, length = 26)

I'm not sure about "exists as a file" here. We can step into this method with origin == 'PipOrigin.NotMaterialized', and we'll mark the file as existent.

narasamdya:
It's concerning if Detours doesn't report probing. If you call Directory.Exists, most likely .NET will call GetFileAttributes. For that method, Detours will report the access as Probe. SandboxedProcess report will translates it as FileProbe, and then observed input processor determines that it's existing directory probe.

If possible, can you place a break point [http://index/?leftProject=BuildXL.Processes&leftSymbol=i9cbsgcisops&file=SandboxedProcessPipExecutor.cs&line=2551](http://index/?leftProject=BuildXL.Processes&leftSymbol=i9cbsgcisops&file=SandboxedProcessPipExecutor.cs&line=2551) 
then see what its returned value consists of. We can debug it together if you want. Thanks.

Public/Src/Engine/Scheduler/Scheduler.cs
semihokur: removed the method completely

---
In reply to: [286725574](https://github.com/microsoft/BuildXL/pull/361#discussion_r286725574) [](ancestors = 286725574)

PR #367, opened 2019-05-23
Trace PinBulk and PlaceFileBulk in ContentSessionBase

JuanCarlosGI:
Previously PerformOperationAsync only supported operations which returned a ResultBase.

PerformNonResultOperationAsync was created, which accepts any return type, with an additional method resultBaseFatory which gives the option to translate the result into a ResultBase for telemetry purposes.

PinBulk and PlaceFileBulk now call PerformNonResultOperationAsync. This should allow for better telemetry and logging.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
+++ b/Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
@@ -161,7 +161,7 @@ namespace BuildXL.Cache.ContentStore.Sessions
             Counter retryCounter);
 
         /// <inheritdoc />
-        public async Task<IEnumerable<Task<Indexed<PlaceFileResult>>>> PlaceFileAsync(
+        public Task<IEnumerable<Task<Indexed<PlaceFileResult>>>> PlaceFileAsync(
             Context context,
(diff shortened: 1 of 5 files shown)


Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
```diff
@@ -159,15 +170,15 @@ protected ContentSessionBase(string name)
...
-                {
-                    return await PlaceFileCoreAsync(cancellableContext, hashesWithPaths, accessMode, replacementMode, realizationMode, urgencyHint, _counters[ContentSessionBaseCounters.PlaceFileBulkRetries]);
-                }
-            }
+            return WithOperationContext(
+                context,
+                token,
+                operationContext => operationContext.PerformNonResultOperationAsync(
+                    Tracer,
+                    () => PlaceFileCoreAsync(operationContext, hashesWithPaths, accessMode, replacementMode, realizationMode, urgencyHint, _counters[ContentSessionBaseCounters.PlaceFileBulkRetries]),
```
dannyvv: >PlaceFileCoreAsync [](start = 26, length = 18)

Side-Note: Some concern about debuggability and callstack bloat with all the levels of lambda indirections, but that applies to the whole cache code..

PR #368, opened 2019-05-23
Fix IndexOutOfRange error when content snapshot takes hash that starts with FF

SergeyTeplyakov:
Fixing of by one error.


Public/Src/Cache/ContentStore/Test/Stores/ContentDirectorySnapshotTests.cs
```diff
@@ -1,19 +1,30 @@
...
-using Xunit.Abstractions;
 using FluentAssertions;
 
 namespace BuildXL.Cache.ContentStore.Test.Stores
 {
     // TODO: fix bug 1541363
-    class ContentDirectorySnapshotTests
+    public class ContentDirectorySnapshotTests
     {
+
```
kewilli: Nit: Remove empty line

PR #375, opened 2019-05-24
Don't use shared cache by default in selfhost builds

almiliMSFT:
The default is to build with win-x64, in which case shared cache doesn't work anyway

### DIFF
--- a/Shared/Scripts/bxl.ps1
+++ b/Shared/Scripts/bxl.ps1
@@ -240,7 +240,7 @@ if ($DeployStandaloneTest) {
 }
 
 if ($Vs) {
-    $AdditionalBuildXLArguments += "/p:[Sdk.BuildXL]GenerateVSSolution=true /p:[Sdk.BuildXL]ExcludeBuildXLExplorer=1 /q:DebugNet472 /vs";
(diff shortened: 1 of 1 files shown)


dannyvv:
Do we have a workitem to make it work and undo this?

almiliMSFT:
> Do we have a workitem to make it work and undo this?

Added a comment with a work item reference

Shared/Scripts/bxl.ps1
```diff
@@ -239,7 +240,7 @@ if ($DeployStandaloneTest) {
 }
 
 if ($Vs) {
-    $AdditionalBuildXLArguments += "/p:[Sdk.BuildXL]GenerateVSSolution=true /vs";
+    $AdditionalBuildXLArguments += "/p:[Sdk.BuildXL]GenerateVSSolution=true /p:[Sdk.BuildXL]ExcludeBuildXLExplorer=1 /q:DebugNet472 /vs";
```
mpysson: not sure what `"/p:[Sdk.BuildXL]GenerateVSSolution=true` does

almiliMSFT: excludes projects that are not buildable in VS

Issue #376, opened 2019-05-25
Two scripts in Shared\Scripts contain unicode BOM trash.


jwdonahue:
I just cloned the latest and ran bxl.cmd:

> bxl.cmd
Preparing to build.
'∩╗┐REM' is not recognized as an internal or external command,

Batch files should be stored as UTF-8, no BOM.

jwdonahue:
I took a few moments to search for the scripts with BOM's:

BuildXLLkgVersion.cmd:1:∩╗┐REM @echo off
BuildXLLkgVersionPublic.cmd:1:∩╗┐REM @echo off

I'll issue a PR shortly.

dannyvv:
Thanks for taking care of the pr as well!

PR #377, opened 2019-05-25
Fix Microsoft/BuildXL#376 Shared\Scripts contains unicode BOM trash.

jwdonahue:
Removed the BOM's from BuildXLLkgVersion.cmd and BuildXLLkgVerionPublic.cmd files.  Saved as UTF-8 no BOM.


dannyvv:
Thanks for your contribution Joseph!

PR #379, opened 2019-05-28
Fix crash when calling worker with a disposed connection

semihokur:


Public/Src/Engine/Dll/Distribution/Grpc/ClientConnectionManager.cs
```diff
@@ -88,7 +99,7 @@ public async Task CloseAsync()
 
             if (waitForConnection)
             {
-                bool connectionSucceeded = await TryConnectChannelAsync(operation, watch);
+                bool connectionSucceeded = await TryConnectChannelAsync(GrpcSettings.InactiveTimeout, operation, watch);
```
dannyvv: >InactiveTimeout [](start = 85, length = 15)

Just checking if InactiveTimeout is intentional, since the method is call I'd expect the calltimeout here.

semihokur: it is intentional. this is called for the first initiation with the worker; so the timeout is larger.

PR #382, opened 2019-05-28
Relocate Temp directory when running in VM

narasamdya:
Recall that, for VM execution, we `net share` our enlistment drive. Then, in VM we `net use` that drive from the VM with the same drive letter. Suppose that the drive letter is `D:`. Accessing `D:\f` from VM means accessing `D:\f` in the host.

Some operations that pips perform (particularly, unit test pips) need to be done locally on VM's disk, for example, creating symlinks, junctions, setting ACLs, etc. 

Let's say that the drive provided by the VM for sandboxed processes is `T:`. Suppose that the pip specified `D:\out\object\Pip123\t_0` as its temp directory. Here, we simply create a directory symlink (junctions won't work) from `D:\out\object\Pip123\t_0` to `T:\tmp\Pip123\0`. The directory `T:\tmp\Pip123\0` may not exist yet; it's `SandboxedProcessExecutor`'s responsibility for creating that target directory. Thus, any access in VM to `D:\out\object\Pip123\t_0\f` actually accessing file `f` locally at `T:\tmp\Pip123\0\f`. 

Further, some of our tests create junctions, but creating junctions with `D:\...` won't work because junctions are evaluated locally. Fortunately, our tests query `%TEMP%` to get the temporary folder. Thus, we need to set `%TEMP%` to the redirected path.

This change set also includes updates to `VmCommandProxy` abstraction to adapt to the newly implemented API in `VmCommandProxy`.

[AB#1518291](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1518291)


Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -671,6 +664,20 @@ public async Task<SandboxedProcessPipExecutionResult> RunAsync(CancellationToken
             }
         }
 
+        private bool ShouldSandboxedProcessExecuteExternal
+            => // Execution mode is external
+               m_sandboxConfig.AdminRequiredProcessExecutionMode.ExecuteExternal()
```
ljcollins25: >m_sandboxConfig.AdminRequiredProcessExecutionMode.ExecuteExternal() [](start = 15, length = 67)

Should it error in the case where process admin requires external execution but it cannot be launched externally for other reasons?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2098,13 +2121,82 @@ private bool CleanTempDirectory(AbsolutePath tempDirectoryPath)
...
+                {
+                    // Path exists as a file or a symlink (directory/file symlink).
+                    FileUtilities.DeleteFile(path, waitUntilDeletionFinished: true, tempDirectoryCleaner: m_tempDirectoryCleaner);
+                }
+
+                if (FileUtilities.DirectoryExistsNoFollow(path))
+                {
+                    // Path exists as a real directory: wipe out that directory.
+                    FileUtilities.DeleteDirectoryContents(path, deleteRootDirectory: true, tempDirectoryCleaner: m_tempDirectoryCleaner);
+                }
```
ljcollins25: Why not call PreparePathForDirectory?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
narasamdya: The pip is annotated with `privilegeLevel: Admin` field to denote that the pip needs to run as admin. The sandboxed configuration controls where to execute the pip. If, for example, it's set to external/VM, but then BuildXL is run on Unix or container is enabled, then it would be annoying if we failed the pip. I can log a warning instead of error in this case.

---
In reply to: [288712689](https://github.com/microsoft/BuildXL/pull/382#discussion_r288712689) [](ancestors = 288712689)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
narasamdya: Because I need to wipe out the existence of the path. In `PreparePathForDirectory`, if the directory exists, then we just wipe out its contents, but not the directory itself.

---
In reply to: [288714193](https://github.com/microsoft/BuildXL/pull/382#discussion_r288714193) [](ancestors = 288714193)

PR #384, opened 2019-05-28
Added csv option for PipExecutionPerformanceAnalyzer

semihokur:
It is a WDG-ask

### DIFF
--- a/Public/Src/Tools/Execution.Analyzer/Analyzers.Partner/PipExecutionPerformanceAnalyzer.cs
+++ b/Public/Src/Tools/Execution.Analyzer/Analyzers.Partner/PipExecutionPerformanceAnalyzer.cs
@@ -20,12 +20,18 @@ namespace BuildXL.Execution.Analyzer
         public Analyzer InitializePipExecutionPerformanceAnalyzer()
         {
             string outputFilePath = null;
+            bool isCsvFormat = false;
             foreach (var opt in AnalyzerOptions)
             {
                 if (opt.Name.Equals("outputFile", StringComparison.OrdinalIgnoreCase) ||
                     opt.Name.Equals("o", StringComparison.OrdinalIgnoreCase))
                 {
                     outputFilePath = ParseSingletonPathOption(opt, outputFilePath);
+                }
+                else if (opt.Name.Equals("csv", StringComparison.OrdinalIgnoreCase) ||
+                    opt.Name.Equals("c", StringComparison.OrdinalIgnoreCase))
+                {
+                    isCsvFormat = true;
                 }
                 else
                 {
@@ -38,7 +44,7 @@ namespace BuildXL.Execution.Analyzer
                 throw Error("outputFile parameter is required");
             }
 
-            return new PipExecutionPerformanceAnalyzer(GetAnalysisInput(), outputFilePath);
+            return new PipExecutionPerformanceAnalyzer(GetAnalysisInput(), outputFilePath, isCsvFormat);
         }
 
         private static void WritePipExecutionPerformanceAnalyzerHelp(HelpWriter writer)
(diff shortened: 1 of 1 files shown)


Public/Src/Tools/Execution.Analyzer/Analyzers.Partner/PipExecutionPerformanceAnalyzer.cs
```diff
@@ -56,23 +63,67 @@ internal sealed class PipExecutionPerformanceAnalyzer : Analyzer
...
             m_writer = new StreamWriter(outputFilePath);
+            m_isCsvFormat = isCsvFormat;
         }
 
-        /// <inheritdoc />
         public override int Analyze()
+        {
+            if (m_isCsvFormat)
+            {
+                ToCsv();
```
olkononenko: nit: maybe rename the methods (this and ToJson) to reflect the kind of data we are printing?

Public/Src/Tools/Execution.Analyzer/Analyzers.Partner/PipExecutionPerformanceAnalyzer.cs
```diff
@@ -20,13 +20,19 @@ internal partial class Args
...
             {
                 if (opt.Name.Equals("outputFile", StringComparison.OrdinalIgnoreCase) ||
                     opt.Name.Equals("o", StringComparison.OrdinalIgnoreCase))
                 {
                     outputFilePath = ParseSingletonPathOption(opt, outputFilePath);
                 }
+                else if (opt.Name.Equals("csv", StringComparison.OrdinalIgnoreCase) ||
+                    opt.Name.Equals("c", StringComparison.OrdinalIgnoreCase))
+                {
+                    isCsvFormat = true;
```
olkononenko: nit: "csv" argument not only changes the format, but also the data included in the output file, so I'd use some kind of a  report name (i.e., describing the data) instead here. (rename both the arg and the variable)

Public/Src/Tools/Execution.Analyzer/Analyzers.Partner/PipExecutionPerformanceAnalyzer.cs
```diff
@@ -20,13 +20,19 @@ internal partial class Args
...
             string outputFilePath = null;
+            bool isCsvFormat = false;
             foreach (var opt in AnalyzerOptions)
             {
                 if (opt.Name.Equals("outputFile", StringComparison.OrdinalIgnoreCase) ||
                     opt.Name.Equals("o", StringComparison.OrdinalIgnoreCase))
                 {
                     outputFilePath = ParseSingletonPathOption(opt, outputFilePath);
                 }
+                else if (opt.Name.Equals("csv", StringComparison.OrdinalIgnoreCase) ||
```
dannyvv: >if [](start = 21, length = 2)

Nit: Why not do the same as some analyzers and look for the extension of the output file i.a. `a.csv` would be a csv file...

Public/Src/Tools/Execution.Analyzer/Analyzers.Partner/PipExecutionPerformanceAnalyzer.cs
```diff
@@ -56,23 +63,67 @@ internal sealed class PipExecutionPerformanceAnalyzer : Analyzer
...
+        {
+            m_writer.WriteLine("Description,Step,Wall,User,Kernel,User/Wall,ReadByes,WriteBytes,OtherBytes,Processes,Memory");
+            foreach (var processIdAndExecutionPerformance in m_processPerformance.OrderByDescending(pip => pip.Value.ProcessExecutionTime.TotalMilliseconds))
+            {
+                var process = CachedGraph.PipGraph.GetPipFromPipId(processIdAndExecutionPerformance.Key) as Process;
+                Contract.Assert(process != null);
+
+                var performance = processIdAndExecutionPerformance.Value;
+
+                m_writer.Write(process.GetShortDescription(CachedGraph.Context) + ",");
```
dannyvv: >GetShortDescription [](start = 39, length = 19)

Description (especially in office) can have a comma, so we should encode comma's in here. either by putting quotes around it per CSV format, or simplye doing a string replace with another character...

dannyvv: >"," [](start = 82, length = 3)

Since you are using individual writes, you could avoid the string concatenation and the allocations for ath you could write this as `m_writer.Write(expr); m_writer.Write(',');` (or perhaps with a helper method `WriteColum(string expr, bool isLast = false)`

PR #385, opened 2019-05-29
Temp cleaner disposal should wait until clean-paths thread joins

narasamdya:
Temp cleaner disposal should wait until clean path thread joins.

The bug occurs due to the use of timeout in joining clean-paths thread. If joining timeouts, then the clean-paths thread still executes and tries to take an element from the blocking collection. However, the collection may get disposed by the thread disposing the temp cleaner. Taking an element from a disposed blocking collection throws an exception.

[AB#1546091](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1546091)

### DIFF
--- a/Public/Src/Engine/Scheduler/TempCleaner.cs
+++ b/Public/Src/Engine/Scheduler/TempCleaner.cs
@@ -123,11 +123,8 @@ namespace BuildXL.Scheduler
         {
             if (!m_disposed)
             {
-                m_disposed = true;
-
                 m_cancellationSource.Cancel();
-                m_pathsToDelete.CompleteAdding();
-                m_cleanerThread.Join();
+                WaitPendingTasksForCompletion();
 
                 Tracing.Logger.Log.PipTempCleanerSummary(
                     Events.StaticContext,
@@ -140,6 +137,8 @@ namespace BuildXL.Scheduler
 
                 m_pathsToDelete.Dispose();
                 m_cancellationSource.Dispose();
+
+                m_disposed = true;
             }
         }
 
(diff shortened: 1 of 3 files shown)


Public/Src/Engine/Scheduler/TempCleaner.cs
```diff
@@ -126,17 +126,17 @@ public void Dispose()
                 m_disposed = true;
 
                 m_cancellationSource.Cancel();
-                bool terminated = m_cleanerThread.Join(TimeSpan.FromMinutes(1));
+                m_pathsToDelete.CompleteAdding();
+                m_cleanerThread.Join();
```
mpysson: Alternately, call WaitPendingTasksForCompletion() above line 126

dannyvv: >rThread.Join(); [](start = 24, length = 15)

I thought the design was that we should not hold up the build any longer than a minute to clean temp directories after the build completes. Cleaning will be picked up by the next build in the background or when immediately needed.

narasamdya: You're right, this seems to be a change in the design. I'm still wanting to make the build ends quickly without waiting for the temp cleaner to finish. That's the reason I put the timeout in `Thread.Join`. However, that doesn't play nicely with the blocking collection. 

Currently, our hope for quick termination relies on the cancellation of directory/file deletion. This should be OK for our selfhost. But apparently, that is not sufficient for lab/cloud build. I think we need to understand better why cancellation is not so great there.

PR #388, opened 2019-05-30
Correctly compute priority by separating critical path priority from a given spec priority

erickulcyk:
When priority was first added, the computation was done incorrectly.  There are two pieces to priority, the one specified in the spec file and the one computed by critical path.  To compute the new values for these, the outgoing pips priority needs to be separated into the specified priority and the critical path priority.  Each of these values in then maxxed to find the largest priority that the pip should have.

### DIFF
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -84,7 +84,7 @@ namespace BuildXL.Scheduler
         /// The max priority assigned to pips in the initial critical path
         /// prioritization
         /// </summary>
-        private const int MaxInitialPipPriority = int.MaxValue - 1;
+        private const int MaxInitialPipPriority = (1 << 24) - 1;
 
         /// <summary>
         /// The priority of IPC pips when entering the ChooseWorker queue. This is greater than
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5013,14 +5013,18 @@ private void PrioritizeAndSchedule(LoggingContext loggingContext, IEnumerable<No
...
                             if (!graph.IsSinkNode(node))
                             {
                                 foreach (var edge in graph.GetOutgoingEdges(node))
                                 {
-                                    var otherCriticalPath = GetPipRuntimeInfo(edge.OtherNode).Priority;
-                                    criticalPath = Math.Max(criticalPath, otherCriticalPath);
+                                    var otherPriority = GetPipRuntimeInfo(edge.OtherNode).Priority;
+
+                                    // Priority consists of given priority in the specs (bits 24-31, and the critical path priority (bits 0-23)
+                                    criticalPath = Math.Max(criticalPath, otherPriority & ((1 << 24) - 1));
```
dannyvv: >otherPriority [](start = 74, length = 13)

could you add these two lines within `unchecked` region? We have a bug where we don't pass the right flag today in release builds.

narasamdya: >1 << 24) - 1) [](start = 92, length = 13)

`MaxInitialPriority`

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5013,14 +5013,18 @@ private void PrioritizeAndSchedule(LoggingContext loggingContext, IEnumerable<No
...
+                                    priorityBase = Math.Max(priorityBase, otherPriority >> 24);
```
narasamdya: >24 [](start = 91, length = 2)

Can this number `24` be represented as a constant? So that changing one will entails changing the others.

PR #390, opened 2019-05-30
Fix the printout of bxl command line arguments

almiliMSFT:
When printed out, the extension of the entry assembly is adjusted as follows:
  - if the extension is .dll and we are running on Windows --> the extension is changed to .exe
  - if the extension is .dll and we are running on non-Windows --> the extension is dropped
  - else --> no adjustment needed

[AB#1547983](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1547983)

### DIFF
--- a/Public/Src/Utilities/Utilities/AssemblyHelper.cs
+++ b/Public/Src/Utilities/Utilities/AssemblyHelper.cs
@@ -98,7 +98,7 @@ namespace BuildXL.Utilities
             if (entryAssemblyLocation.EndsWith(".dll"))
             {
                 return OperatingSystemHelper.IsUnixOS
-                    ? Path.GetFileNameWithoutExtension(entryAssemblyLocation)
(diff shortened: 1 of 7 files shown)


Public/Src/Utilities/Utilities/AssemblyHelper.cs
```diff
@@ -79,5 +80,44 @@ public static string GetLocation(this Assembly assembly, bool computeAssemblyLoc
...
+            }
+            else
+            {
+                return entryAssemblyLocation;
+            }
+        }
+
+        /// <summary>
+        /// Returns the same as <see cref="Environment.GetCommandLineArgs"/> except that the first
+        /// element (which is executable name) has its extension adjusted by calling
```
almiliMSFT:
```suggestion
        /// element (which is the name of the executable) is adjusted by calling
```

PR #391, opened 2019-05-30
Cache GRPC channels

kewilli:
Currently, we create a new GRPC channel for every GRPC copy. We can be much smarter about this and cache the constructed channels.


Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
     /// </summary>
     public sealed class GrpcCopyClient : IShutdown<BoolResult>
     {
+        // Link to DistributedContentStoreSettings.MaxConcurrentCopyOperations
+        private const int _maxClientCount = 512;
+
+        private static readonly object lockObject = new object();
+        internal static readonly ConcurrentDictionary<(string, int, bool), GrpcCopyClient> _clientDict = new ConcurrentDictionary<(string, int, bool), GrpcCopyClient>(Environment.ProcessorCount, _maxClientCount);
```
dannyvv: >_clientDict [](start = 91, length = 11)

Just checking:
Since you use Tuple<,,> you can't control the casing of the string value. Which seems to be the host. Is there any chance there could be a variation in the host (i.e. IP vs DNS, IPv4 vs IPv4 or even casing differences in case of DNS name)?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+                                {
+                                    // Cannot await within a lock
+                                    client._channel.ShutdownAsync().Start();
+                                }
+                            }
+                        }
+                    }
+                }
+
+                await Task.Delay(1000 * 60 * 30, ct);
```
dannyvv: >30 [](start = 45, length = 2)

`TimeSpan.FromMinutes(30)` ?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+        private async static Task BackgroundCleanupAsync(CancellationTokenSource cts)
+        {
+            _backgroundCleaningTaskTokenSource = cts;
+            var ct = _backgroundCleaningTaskTokenSource.Token;
+
+            while (!ct.IsCancellationRequested)
+            {
+                if (!_clientDict.IsEmpty)
+                {
+                    var oneHourAgo = DateTime.UtcNow - TimeSpan.FromHours(1);
```
dannyvv: >FromHours [](start = 64, length = 9)

This is border line close to the timeout of 30 minutes. Don't we run into a chance here where some times flip between just before the 1/2 hour timeout and some just after? Wouldn't it be better to have something relatively deterministic in the nr of cycles of the cleanup? i.e. 55 minutes or a 14 minute wait in the cleanup?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+                    {
+                        var client = kvp.Value;
+                        if (client._uses == 0 && client._lastUseTime < oneHourAgo)
+                        {
+                            lock (lockObject)
+                            {
+                                if (_clientDict.TryRemove((client._host, client._grpcPort, client._useCompression), out GrpcCopyClient removedClient))
+                                {
+                                    // Cannot await within a lock
+                                    client._channel.ShutdownAsync().Start();
```
dannyvv: >ShutdownAsync [](start = 52, length = 13)

Does the channel guarantee that the existing message is completed on it before it kills it?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+            }
+        }
+
+        internal static void RestartBackgroundCleanup()
+        {
+            lock (lockObject)
+            {
+                if (_backgroundCleaningTask != null)
+                {
+                    _backgroundCleaningTaskTokenSource.Cancel();
```
dannyvv: >_backgroundCleaningTaskTokenSource [](start = 20, length = 34)

Since you use `Task.Run(` below technically if the RestartBackgroundcleanup gets called twice fast it could be that the tcs is still null when this line is hit.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -67,9 +129,27 @@ private GrpcCopyClient(string host, int grpcPort, bool useCompression)
...
-            // TODO: Add caching of GrpcCopyClient objects
-            // TODO: Add case where _clientPool has exceeded some maximum count
-            return new GrpcCopyClient(host, grpcPort, useCompression);
+            if (_clientDict.TryGetValue((host, grpcPort, useCompression), out GrpcCopyClient existingClient))
+            {
+                Interlocked.Increment(ref existingClient._uses);
+                existingClient._lastUseTime = DateTime.UtcNow;
+                return existingClient;
+            }
+            else if (_clientDict.Count > _maxClientCount)
```
dannyvv: >ntDict.Count > _maxClie [](start = 26, length = 23)

Curious why you didn't do this check inside the GetOrAdd lambda, now there is a race where many threads can get past this check and blow the nr of Clients beyand the maxClient so now it is kind of like a soft limmit.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
dannyvv: ah, turns out this is not really restart, just star. given the name I assumed you had a check that on each creation of a client you doublechecked you were running the collector...

---
In reply to: [289193315](https://github.com/microsoft/BuildXL/pull/391#discussion_r289193315) [](ancestors = 289193315)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
SergeyTeplyakov: I think we need to extract this into ChannelCache or something. It looks very complicated right now.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
         private bool _useCompression;
+        internal DateTime _lastUseTime;
+        internal int _uses;
+
+        static GrpcCopyClient()
+        {
+            RestartBackgroundCleanup();
+        }
+
+        private async static Task BackgroundCleanupAsync(CancellationTokenSource cts)
```
SergeyTeplyakov: nit: static async

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+        {
+            _backgroundCleaningTaskTokenSource = cts;
```
SergeyTeplyakov: This is a very unusual pattern when the operation actually changes the static state...

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
SergeyTeplyakov: I think we should create cache instance and pass it here.

I really don't like to have a bunch of static state here...

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
     /// </summary>
     public sealed class GrpcCopyClient : IShutdown<BoolResult>
     {
+        // Link to DistributedContentStoreSettings.MaxConcurrentCopyOperations
+        private const int _maxClientCount = 512;
```
SergeyTeplyakov: Are you going to do this?

I think we should do this up front. (and, again, inside ConnectionCache or something like that).

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+        internal static readonly ConcurrentDictionary<(string, int, bool), GrpcCopyClient> _clientDict = new ConcurrentDictionary<(string, int, bool), GrpcCopyClient>(Environment.ProcessorCount, _maxClientCount);
+        private static Task _backgroundCleaningTask;
+        private static CancellationTokenSource _backgroundCleaningTaskTokenSource;
+
         private readonly Channel _channel;
         private readonly ContentServer.ContentServerClient _client;
         private readonly string _host;
         private readonly int _grpcPort;
         private bool _useCompression;
+        internal DateTime _lastUseTime;
```
SergeyTeplyakov: I would say that this should be stored externally from the client.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+                    }
+                }
+
+                await Task.Delay(1000 * 60 * 30, ct);
+            }
+        }
+
+        internal static void RestartBackgroundCleanup()
+        {
+            lock (lockObject)
```
SergeyTeplyakov: _restartBackgroundCleanupLock?

SergeyTeplyakov: This is very, very dangerous.

Static constructors are special and locks there could cause an app to deadlock.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
         private readonly ContentServer.ContentServerClient _client;
         private readonly string _host;
         private readonly int _grpcPort;
         private bool _useCompression;
+        internal DateTime _lastUseTime;
+        internal int _uses;
+
+        static GrpcCopyClient()
+        {
+            RestartBackgroundCleanup();
```
SergeyTeplyakov: I think the overall stuff would be way more simple with an explicit ConnectionCache class that will do stuff in background after StartAsync method on it is called.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
SergeyTeplyakov: 1. This should be configured
2. If its not configured it is easier to read the code by using TimeStamp.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
SergeyTeplyakov: _clientCache?

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+                            }
+                        }
+                    }
+                }
+
+                await Task.Delay(1000 * 60 * 30, ct);
+            }
+        }
+
+        internal static void RestartBackgroundCleanup()
```
SergeyTeplyakov: >RestartBackgroundCleanup [](start = 29, length = 24)

I can't understand why the solution is so complicated? I don't think we can actually "restart" the process... Right? I see that this method is called only once here.

I assume that this method is called in test.

And this is another reason why extracting the logic into a separate class will help to make this code so much easer to read.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
SergeyTeplyakov: I think this should be a type, not a tuple with no names...

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
SergeyTeplyakov: This is very strange that we call Start on task.

THere is only one case when you can do that: is when a task is created using constructor. But this is a very rare case.

So we should remove 'Start' or add a giant comment explaining that this is fine here.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
SergeyTeplyakov: Shoudl be configured.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+        }
+
+        private async static Task BackgroundCleanupAsync(CancellationTokenSource cts)
+        {
+            _backgroundCleaningTaskTokenSource = cts;
+            var ct = _backgroundCleaningTaskTokenSource.Token;
+
+            while (!ct.IsCancellationRequested)
+            {
+                if (!_clientDict.IsEmpty)
```
SergeyTeplyakov: I don't think it make sense to check IsEmpty explicitely...

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -67,9 +129,27 @@ private GrpcCopyClient(string host, int grpcPort, bool useCompression)
         /// <param name="useCompression">Whether or not GZip is enabled for copies.</param>
         public static GrpcCopyClient Create(string host, int grpcPort, bool useCompression = false)
         {
-            // TODO: Add caching of GrpcCopyClient objects
-            // TODO: Add case where _clientPool has exceeded some maximum count
-            return new GrpcCopyClient(host, grpcPort, useCompression);
+            if (_clientDict.TryGetValue((host, grpcPort, useCompression), out GrpcCopyClient existingClient))
+            {
+                Interlocked.Increment(ref existingClient._uses);
```
SergeyTeplyakov: Here is another reason why we need a separate type: we should have counters here to track the number of times we reused the clients and the number of times we created new ones.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -24,28 +26,85 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
...
+                    var oneHourAgo = DateTime.UtcNow - TimeSpan.FromHours(1);
+
+                    foreach (var kvp in _clientDict)
+                    {
+                        var client = kvp.Value;
+                        if (client._uses == 0 && client._lastUseTime < oneHourAgo)
+                        {
+                            lock (lockObject)
+                            {
+                                if (_clientDict.TryRemove((client._host, client._grpcPort, client._useCompression), out GrpcCopyClient removedClient))
```
SergeyTeplyakov: We need trace the shutdown.

And we need to do this in try/catch.

I assume that Shutdown may fail due to some infrastructure issues and we don't want for the entire loop to stop.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -67,9 +129,27 @@ private GrpcCopyClient(string host, int grpcPort, bool useCompression)
...
-            return new GrpcCopyClient(host, grpcPort, useCompression);
+            if (_clientDict.TryGetValue((host, grpcPort, useCompression), out GrpcCopyClient existingClient))
+            {
+                Interlocked.Increment(ref existingClient._uses);
+                existingClient._lastUseTime = DateTime.UtcNow;
+                return existingClient;
+            }
+            else if (_clientDict.Count > _maxClientCount)
+            {
+                throw new CacheException($"Attempting to create {nameof(GrpcCopyClient)} to increase cached count above maximum allowed ({_maxClientCount})");
```
SergeyTeplyakov: I think in this case we should clean-up first and only when all the clients are used we should fail.

Today do to an unbnormal spike in client creation you can get here and fail, even though you have a bunch of instances that are no longer used.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: Great point! This should be casing agnostic, and I don't think we should worry about having multiple hosts which evaluate to the same thing. All GrpcCopyClient constructors use the same method to construct the host.

---
In reply to: [289191783](https://github.com/microsoft/BuildXL/pull/391#discussion_r289191783) [](ancestors = 289191783)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: Agreed. Moving to a separate class.

---
In reply to: [289207822](https://github.com/microsoft/BuildXL/pull/391#discussion_r289207822) [](ancestors = 289207822)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: Moving to non-static constructor in separate class

---
In reply to: [289207259](https://github.com/microsoft/BuildXL/pull/391#discussion_r289207259) [](ancestors = 289207259)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: It does not guarantee anything. The assumption here is that the client will only have 0 uses if it is no longer in use.

---
In reply to: [289192667](https://github.com/microsoft/BuildXL/pull/391#discussion_r289192667) [](ancestors = 289192667)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: In the cache's dictionary? That seems more complex than just keeping it tied to the client itself.

---
In reply to: [289207076](https://github.com/microsoft/BuildXL/pull/391#discussion_r289207076) [](ancestors = 289207076)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: Instead, we can collect the tasks and wait on the entire set.

---
In reply to: [289208202](https://github.com/microsoft/BuildXL/pull/391#discussion_r289208202) [](ancestors = 289208202)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: `GetOrAdd`'s factory argument doesn't affect whether the factory's result will be added. We could guard with a lock, but it's okay if this is only a soft limit.

---
In reply to: [289194267](https://github.com/microsoft/BuildXL/pull/391#discussion_r289194267) [](ancestors = 289194267)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
kewilli: I think this is fine for the time-being. If we end up with too many unused clients that last too long, then we can fix for the short-term via config options (reduce wait time between cleanups and reduce maximum age) and revisit this.

---
In reply to: [289209028](https://github.com/microsoft/BuildXL/pull/391#discussion_r289209028) [](ancestors = 289209028)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClientCache.cs
```diff
@@ -0,0 +1,146 @@
...
+
+namespace BuildXL.Cache.ContentStore.Service.Grpc
+{
+    /// <summary>
+    /// Cache for <see cref="GrpcCopyClient"/>.
+    /// </summary>
+    public sealed class GrpcCopyClientCache
+    {
+        // Link to DistributedContentStoreSettings.MaxConcurrentCopyOperations
+        internal readonly ConcurrentDictionary<GrpcCopyClientKey, GrpcCopyClient> _clientDict;
```
SergeyTeplyakov: Minor, but I would prefer to keep it internal add an internal method that will do what you want this field for.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClientCache.cs
```diff
@@ -0,0 +1,146 @@
...
+#pragma warning disable ERP022 // Unobserved exception in generic exception handler
+                catch (Exception)
+                {
+                    // If Task.WhenAll throws in an await, it unwraps the AggregateException and only
+                    // throws the first inner exception. We want to see all failed shutdowns.
+                    _context.Error($"Shutdown of unused GRPC clients failed after removal from client cache. {allTasks.Exception}");
+                }
+#pragma warning restore ERP022 // Unobserved exception in generic exception handler
+
+                await Task.Delay(TimeSpan.FromMinutes(_waitBetweenCleanupInMinutes), ct);
```
SergeyTeplyakov: I think during cancellation we can get UnobservedTaskException, because this Delay will fail with TaskCancelled, and I can't see that _backgroundCleaningTask exception is handled...

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClientCache.cs
```diff
@@ -0,0 +1,146 @@
...
+        public GrpcCopyClient Create(string host, int grpcPort, bool useCompression = false)
+        {
+            var key = new GrpcCopyClientKey(host, grpcPort, useCompression);
+            if (_clientDict.TryGetValue(key, out GrpcCopyClient existingClient))
+            {
+                Interlocked.Increment(ref existingClient._uses);
+                existingClient._lastUseTime = DateTime.UtcNow;
+                return existingClient;
+            }
+            else if (_clientDict.Count > _maxClientCount)
```
SergeyTeplyakov: I think we should try to evict not used connections first and only after that to fail.

Public/Src/Cache/ContentStore/DistributedTest/Stores/GrpcCopyContentTests.cs
```diff
@@ -32,12 +32,67 @@ public class GrpcCopyContentTests : TestBase
...
+        [Fact]
+        public void ValidateBackgroundCleanup()
+        {
+            var key = new GrpcCopyClientKey(LocalHost, 11, true);
+            using (var client = _clientCache.Create(key.Host, key.GrpcPort, key.UseCompression))
+            { 
+                client._lastUseTime = DateTime.UtcNow - TimeSpan.FromHours(2);
+            }
+
+            _clientCache.StartBackgroundCleanup();
```
SergeyTeplyakov: I'm not sure we need it, right? We start it during startup anyways.

Public/Src/Cache/ContentStore/DistributedTest/Stores/GrpcCopyContentTests.cs
```diff
@@ -32,12 +32,67 @@ public class GrpcCopyContentTests : TestBase
...
+            var endTime = DateTime.UtcNow + TimeSpan.FromMinutes(1);
+            while (DateTime.UtcNow < endTime)
+            {
+                if (!_clientCache._clientDict.TryGetValue(key, out GrpcCopyClient foundClient))
+                {
+                    return;
+                }
+
+                Task.Delay(1000).GetAwaiter().GetResult();
```
SergeyTeplyakov: We should switch to IClock and not to wait for a minute for this.

The current approach is more brittle and will take way more time than needed to run the tests.

Public/Src/Cache/ContentStore/DistributedTest/Stores/GrpcCopyContentTests.cs
```diff
@@ -32,12 +32,67 @@ public class GrpcCopyContentTests : TestBase
...
+        public void IssueSameClientManyTimes()
+        {
+            GrpcCopyClient sameClient = null;
+            for (int i = 0; i < 1000; i++)
+            {
+                using (var client = _clientCache.Create(LocalHost, 42, false))
+                {
+                    if (i == 0)
+                    {
+                        sameClient = client;
```
SergeyTeplyakov: this is fine, but I would suggest to create another client explicitely outside of the loop.

Public/Src/Cache/ContentStore/DistributedTest/Stores/GrpcCopyContentTests.cs
```diff
@@ -32,12 +32,67 @@ public class GrpcCopyContentTests : TestBase
...
+            using (var client = _clientCache.Create(key.Host, key.GrpcPort, key.UseCompression))
+            { 
+                client._lastUseTime = DateTime.UtcNow - TimeSpan.FromHours(2);
+            }
+
+            _clientCache.StartBackgroundCleanup();
+            var endTime = DateTime.UtcNow + TimeSpan.FromMinutes(1);
+            while (DateTime.UtcNow < endTime)
+            {
+                if (!_clientCache._clientDict.TryGetValue(key, out GrpcCopyClient foundClient))
```
SergeyTeplyakov: I'm not sure that this test worth making state internal...

Public/Src/Cache/DistributedCache.Host/Configuration/DistributedContentSettings.cs
```diff
@@ -178,6 +180,24 @@ private DistributedContentSettings(IDictionary<string, string> connectionSecretN
...
+        /// Maximum cached age for GRPC clients.
+        /// </summary>
+        [DataMember]
+        public int MaxGrpcClientAgeMinutes { get; set; } = 55;
+
+        /// <summary>
+        /// Time between GRPC cache cleanups.
+        /// </summary>
+        [DataMember]
+        public int GrpcClientCleanupDelayMinutes { get; set; } = 17;
```
SergeyTeplyakov: Why 17:)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClientKey.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            return obj is GrpcCopyClientKey key && Equals(key);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            return BuildXL.Utilities.HashCodeHelper.Combine(Host.GetHashCode(), GrpcPort, UseCompression ? 1 : 0);
```
SergeyTeplyakov: nit: return (Host, GrpcPort, UseCompression).GetHashCode();

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClientKey.cs
```diff
@@ -0,0 +1,40 @@
...
+    internal struct GrpcCopyClientKey : IEquatable<GrpcCopyClientKey>
+    {
+        public GrpcCopyClientKey(string host, int grpcPort, bool useCompression)
+        {
+            Host = host;
+            GrpcPort = grpcPort;
+            UseCompression = useCompression;
+        }
+
+        public string Host;
```
SergeyTeplyakov: Could you please move them to the top of the file and make them get-only properties.

And please make the struct readonly.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClientCache.cs
```diff
@@ -0,0 +1,146 @@
...
+            _context = context;
+            _maximumAgeInMinutes = maxClientAgeMinutes;
+            _waitBetweenCleanupInMinutes = waitBetweenCleanupMinutes;
+            _maxClientCount = maxClientCount;
+
+            _clientDict = new ConcurrentDictionary<GrpcCopyClientKey, GrpcCopyClient>(Environment.ProcessorCount, _maxClientCount);
+            StartBackgroundCleanup();
+        }
+
+        internal void StartBackgroundCleanup()
```
SergeyTeplyakov: I'm not sure, but the fact that this method internal and exposed made a lot of stuff way more complicated.

First of all, I can't see why we need to expose this method at all...

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClientCache.cs
```diff
@@ -0,0 +1,146 @@
...
+using BuildXL.Cache.ContentStore.Exceptions;
+using BuildXL.Cache.ContentStore.Interfaces.Results;
+using BuildXL.Cache.ContentStore.Interfaces.Tracing;
+
+namespace BuildXL.Cache.ContentStore.Service.Grpc
+{
+    /// <summary>
+    /// Cache for <see cref="GrpcCopyClient"/>.
+    /// </summary>
+    public sealed class GrpcCopyClientCache
```
SergeyTeplyakov: I think cache should be disposable.

This thing should stop background processing when Dispose method is called and it should dispose all the GRPCClients.

Public/Src/Cache/ContentStore/DistributedTest/Stores/GrpcCopyContentTests.cs
kewilli: The intent here is to force cleanup to happen now. We don't want to wait. I'll add a comment.

---
In reply to: [290518317](https://github.com/microsoft/BuildXL/pull/391#discussion_r290518317) [](ancestors = 290518317)

Public/Src/Cache/ContentStore/DistributedTest/Stores/GrpcCopyContentTests.cs
kewilli: Are you suggesting that this test be removed? Or are you suggesting that we create a new client every N milliseconds until it is the original client?

---
In reply to: [290518941](https://github.com/microsoft/BuildXL/pull/391#discussion_r290518941) [](ancestors = 290518941)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClientCache.cs
kewilli: For testing, really. I didn't want to write a test which relies on cleanup without having a way to initiate cleanup on demand. I don't want to keep this cleanup feature without having tests to validate it.

---
In reply to: [290519565](https://github.com/microsoft/BuildXL/pull/391#discussion_r290519565) [](ancestors = 290519565)

Public/Src/Cache/DistributedCache.Host/Configuration/DistributedContentSettings.cs
kewilli: Random starting point that is not a factor of the maximum client age. This should change in the future.

---
In reply to: [290519019](https://github.com/microsoft/BuildXL/pull/391#discussion_r290519019) [](ancestors = 290519019)

Public/Src/Cache/ContentStore/DistributedTest/Stores/GrpcCopyContentTests.cs
kewilli: Because I think that IClock is messy. Would switching to a significantly shorter delay like 100ms be better?

---
In reply to: [290518437](https://github.com/microsoft/BuildXL/pull/391#discussion_r290518437) [](ancestors = 290518437)

kewilli:
AB#1547942

PR #392, opened 2019-05-31
Early worker release for distributed builds

semihokur:


Public/Src/Engine/Dll/Engine.cs
```diff
@@ -1100,6 +1100,13 @@ public static void PopulateLoggingAndLayoutConfiguration(CommandLineConfiguratio
                 mutableConfig.Logging.StoreFingerprints = true;
             }
 
+            // EarlyWorkerRelease is only enabled for Office ProductBuild lab and OSG lab builds.
+            if (mutableConfig.Logging.Environment != ExecutionEnvironment.OfficeProductBuildLab &&
```
ljcollins25: Also disable if mutableConfig.Distribution.ReplicateOutputsToWorkers == true

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2036,6 +2050,45 @@ public void UpdateStatus(bool overwriteable = true, int expectedCallbackFrequenc
...
+
+        /// <summary>
+        /// Decide whether we can release a remote worker. This method is executed every 2 seconds depending on the frequency of LogStatus timer.
+        /// </summary>
+        private void PerformEarlyReleaseWorker(long numProcessPipsPending, long numProcessPipsRunning)
+        {
+            long numProcessPipsWaiting = numProcessPipsPending - numProcessPipsRunning;
+
+            // If the available remote workers perform at half capacity in future, how many process pips we can concurrently execute:
+            int totalSlots = (int)Math.Ceiling(Workers.Where(a => a.IsRemote && a.IsAvailable).Sum(a => a.TotalProcessSlots * 0.5) + LocalWorker.TotalProcessSlots);
```
ljcollins25: >int totalSlots = (int)Math.Ceiling(Workers.Where(a => a.IsRemote && a.IsAvailable).Sum(a => a.TotalProcessSlots * 0.5) + LocalWorker.TotalProcessSlots); [](start = 12, length = 152)

I think the logic should be slightly different.

var workerToReleaseCandidate = Workers.Where(a => a.IsRemote && a.IsAvailable).OrderBy(a => a.AcquiredProcessSlots).FirstOrDefault();
if (workerToReleaseCandidate != null) return;

// NOTE: Multiplier is configurable.
int totalSlots = m_scheduleConfiguration.EarlyWorkerReleaseCapacityMultiplier * (int)Math.Ceiling(Workers.Where(a => a.IsRemote && a.IsAvailable).Sum(a => a.TotalProcessSlots) + LocalWorker.TotalProcessSlots);

// Release worker if numProcessPipsWaiting can be satisfied by remaining workers
if (numProcessPipsWaiting > 0 && (numProcessPipsWaiting < totalSlots - workerToReleaseCandidate.TotalProcessSlots))
{
     workerToReleaseCandidate.EarlyReleaseAsync();
}

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -381,58 +381,112 @@ public override async void Start()
...
+            var drainStopwatch = new StopwatchVar();
 
-                // Only wait a short amount of time for exit (15 seconds) if worker is not successfully attached.
-                if (m_attachCompletion.Task.Status != TaskStatus.RanToCompletion || !await m_attachCompletion.Task)
+            using (drainStopwatch.Start())
+            {
+                if (AcquiredSlots != 0)
                 {
-                    exitCancellation.CancelAfter(TimeSpan.FromSeconds(15));
+                    await DrainCompletion.Task;
```
ljcollins25: >await DrainCompletion.Task; [](start = 20, length = 27)

Why not always await this?

Public/Src/App/Bxl/Args.cs
```diff
@@ -306,6 +306,9 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                             "diagnostic",
                             "diag",
                             opt => loggingConfiguration.Diagnostic |= CommandLineUtilities.ParseEnumOption<DiagnosticLevels>(opt)),
+                        OptionHandlerFactory.CreateBoolOption(
```
ljcollins25: >CreateBoolOption [](start = 45, length = 16)

Add option of release capacity multiplier. See comment in scheduler.

Public/Src/Engine/Scheduler/Tracing/ExecutionLog.Events.cs
```diff
@@ -1205,6 +1205,12 @@ public struct StatusEventData : IExecutionLogEventData<StatusEventData>
         /// </summary>
         public int UnresponsivenessFactor;
 
+        /// <nodoc />
+        public long ProcessPipsPending;
+
+        /// <nodoc />
+        public long ProcessPipsRunning;
```
mpysson: Please doc exactly what these mean. Does Running mean the running state? an external child process actually executing? etc

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -381,58 +381,112 @@ public override async void Start()
...
+            using (disconnectStopwatch.Start())
+            {
+                await DisconnectAsync();
+            }
+
+            Scheduler.Tracing.Logger.Log.WorkerReleasedEarly(
+                m_appLoggingContext,
+                Name,
+                (long)drainStopwatch.TotalElapsed.TotalMilliseconds,
+                (long)disconnectStopwatch.TotalElapsed.TotalMilliseconds);
```
mpysson: Can you add counters for these so they go into the stats file and telemetry? I'm thinking the total drain duration, disconnect duration, and number of workers released early. Please don't just add send the WorkerReleasedEarly event to telemetry too since that will be complicated to join against due to multiple events per session and we're low on our unique telemetry event quota.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -1994,6 +2000,12 @@ public void UpdateStatus(bool overwriteable = true, int expectedCallbackFrequenc
                     writeFileDone: writeFileStats.DoneCount,
                     writeFileNotDone: writeFileStats.Total - writeFileStats.DoneCount - writeFileStats.IgnoredCount);
 
+                // Number of process pips that are not completed yet.
+                long numProcessPipsPending = m_processStateCountersSnapshot[PipState.Waiting] + m_processStateCountersSnapshot[PipState.Ready] + m_processStateCountersSnapshot[PipState.Running];
+
+                // PipState.Running does not mean that the pip is actually running. The pip might be waiting for a slot.
+                long numProcessPipsRunning = Workers.Sum(a => a.AcquiredSlotsForProcessPips);
```
mpysson: >Running [](start = 35, length = 7)

If PipState.Running doesn't mean "running" should we be reusing the ambiguous term running here?

I guess this feedback is bigger than the change in this PR but these states have become super confusing to the point where you have to hunt down the implementation that sets the value to know what it actually means. It'd be great to give some thought as to how to clean this up the future

Public/Src/Engine/Scheduler/Scheduler.cs
mpysson: @semihokur - How does this algorithm compare to the one the QB uses?

semihokur: I have no idea what algorithm they uses.  This is what we discussed in the perf meeting though. This seems a good start and we can do several iterations for that logic. 

---
In reply to: [289506464](https://github.com/microsoft/BuildXL/pull/392#discussion_r289506464) [](ancestors = 289506464)

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
semihokur: ADded the reason as a comment

---
In reply to: [289427814](https://github.com/microsoft/BuildXL/pull/392#discussion_r289427814) [](ancestors = 289427814)

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
semihokur: Add the followings:

RemoteWorker_EarlyReleaseSavingDurationMs
RemoteWorker_EarlyReleaseDrainDurationMs

---
In reply to: [289496810](https://github.com/microsoft/BuildXL/pull/392#discussion_r289496810) [](ancestors = 289496810)

PR #395, opened 2019-05-31
[Unix] Temporarily fix file permissions on cache materialization

Kr1sso:
To unblock our distributed BuildXL build scenarios, this PR adds temporary cache config flags that allow for file permission override when run on non-windows systems. This means that once a file is retrieved from cache it will be marked with the file permission mask `0777` after placing. This will be removed once we have proper file permission handling implemented.


Public/Src/Cache/ContentStore/UtilitiesCore/FileSystemConstants.cs
```diff
@@ -12,6 +13,26 @@ namespace BuildXL.Cache.ContentStore.Interfaces.FileSystem
...
+
+        /// <nodoc />
+        public static void OverrideFileAccessMode(bool changePermissions, string path)
+        {
+#if PLATFORM_OSX
+            if (changePermissions)
+            {
+                // Force 0777 on the file at 'path' - this is a temporary hack when placing files as our cache layer
+                // currently does not track Unix file access flags when putting / placing files
+                LibC.chmod(path, 0x0001 | 0x0002 | 0x0004 | 0x0040 | 0x0080 | 0x0100 | 0x0008 | 0x0010 | 0x0020);
```
almiliMSFT: This looks clearer to me:
```suggestion
                LibC.chmod(path, Convert.ToInt32("777", 8));
```

Public/Src/Cache/ContentStore/UtilitiesCore/FileSystemConstants.cs
```diff
@@ -12,6 +13,26 @@ namespace BuildXL.Cache.ContentStore.Interfaces.FileSystem
...
     {
+        private static class LibC
+        {
+            /// <nodoc />
+            [DllImport("libc", SetLastError = true)]
+            public static extern int chmod(string path, int mode);
+        }
+
+        /// <nodoc />
+        public static void OverrideFileAccessMode(bool changePermissions, string path)
```
mpysson: >OverrideFileAccessMode [](start = 27, length = 22)

This utility doesn't seem to fit with the purpose of the FileSystemContants class. Is there a better home for it?

Public/Src/Cache/ContentStore/UtilitiesCore/FileSystemConstants.cs
Kr1sso: 👍

Public/Src/Cache/ContentStore/UtilitiesCore/FileSystemConstants.cs
Kr1sso: I've added a `UnixHelpers` class, which will make it easier to delete once we have proper support implemented in cache.

PR #398, opened 2019-06-02
Flight consistent temp directory creation behind a flag

narasamdya:
Changes that made temp directory creation consistent break customers. Here, we put it (hopefully temporarily) behind a flag

[AB#1549282](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1549282)

### DIFF
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -366,6 +366,9 @@ namespace BuildXL
                         OptionHandlerFactory.CreateBoolOption(
                             "enableGrpc",
                             sign => distributionConfiguration.IsGrpcEnabled = sign),
+                        OptionHandlerFactory.CreateBoolOption(
+                            "enableIncrementalFrontEnd",
+                            sign => frontEndConfiguration.EnableIncrementalFrontEnd = sign),
                         OptionHandlerFactory.CreateBoolOptionWithValue(
                             "enableLazyOutputs",
                             (opt, sign) => HandleLazyOutputMaterializationOption(opt, sign, schedulingConfiguration)),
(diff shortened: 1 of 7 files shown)


Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -630,6 +630,28 @@ private void DoCreateDir()
         {
             string directoryPath = FileOrDirectoryToString(Path);
 
+            bool failedIfExists = false;
```
Kr1sso: Nit: `failIfExists`?

Public/Src/App/Bxl/Args.cs
```diff
@@ -366,18 +366,18 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
...
+                            "enableIncrementalFrontEnd",
+                            sign => frontEndConfiguration.EnableIncrementalFrontEnd = sign),
                         OptionHandlerFactory.CreateBoolOptionWithValue(
                             "enableLazyOutputs",
                             (opt, sign) => HandleLazyOutputMaterializationOption(opt, sign, schedulingConfiguration)),
                         OptionHandlerFactory.CreateOption(
                             "engineCacheDirectory",
                             opt => layoutConfiguration.EngineCacheDirectory = CommandLineUtilities.ParsePathOption(opt, pathTable)),
                         OptionHandlerFactory.CreateBoolOption(
-                            "enableIncrementalFrontEnd",
```
mpysson: Nit - in the future I'd suggest not making cleanup type of changes in something we're going to have to cherry pick to prevent merge conflicts

mpysson:
> Changes that made temp directory creation consistent break customers. Here, we put it (hopefully temporarily) behind a flag

The flag isn't going to be very temporary since Office isn't going to backport this change to their LTSB branches.

narasamdya:
Speaking of this, perhaps we need to allocate some window every 3-6 months to introduce all breaking changes. One part of it is to enumerate all argument options, and deprecate unused options and features. 

I don't know the deal with LTSBs, but I would expect that once breaking changes are introduced, they have to lived with old versions of BXL.

---
In reply to: [498342783](https://github.com/microsoft/BuildXL/pull/398#issuecomment-498342783) [](ancestors = 498342783)

Public/Src/App/Bxl/Args.cs
narasamdya: Noted.

---
In reply to: [289939300](https://github.com/microsoft/BuildXL/pull/398#discussion_r289939300) [](ancestors = 289939300)

PR #401, opened 2019-06-03
Retry process if exits with Azure Watson 0xDEAD exit code

narasamdya:
Process can nondeterministically  exit with 0xDEAD exit code in CB. This is the exit code returned by Azure Watson dump after catching the process crash. The root cause of the crash is unknown, but the primary suspect is the way Detours handle NtClose.  One needs to have a deep investigation on detoured NtClose.

This change is an attempt to improve reliability by retrying the process.

### DIFF
--- a/Public/Src/Engine/Dll/Engine.cs
+++ b/Public/Src/Engine/Dll/Engine.cs
@@ -1066,7 +1066,7 @@ namespace BuildXL.Engine
                 }
 
                 mutableConfig.Logging.StoreFingerprints = initialCommandLineConfiguration.Logging.StoreFingerprints ?? false;
-                mutableConfig.Sandbox.RetryOnDeadExitCode = true;
+                mutableConfig.Sandbox.RetryOnAzureWatsonExitCode = true;
             }
             else
             {
--- a/Public/Src/Engine/Dll/Engine.cs
+++ b/Public/Src/Engine/Dll/Engine.cs
@@ -1066,6 +1066,7 @@ namespace BuildXL.Engine
                 }
 
                 mutableConfig.Logging.StoreFingerprints = initialCommandLineConfiguration.Logging.StoreFingerprints ?? false;
+                mutableConfig.Sandbox.RetryOnDeadExitCode = true;
             }
             else
             {
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -38,6 +38,8 @@ namespace BuildXL.Processes
         /// </summary>
         public const int MaxConsoleLength = 2048; // around 30 lines
 
+        private const uint DeadExitCode = 0xDEAD;
+
         private static readonly string s_appDataLocalMicrosoftClrPrefix =
             Path.Combine(SpecialFolderUtilities.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "Microsoft", "CLR");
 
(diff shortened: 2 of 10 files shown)


Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -38,6 +38,8 @@ public sealed class SandboxedProcessPipExecutor : ISandboxedProcessFileStorage
         /// </summary>
         public const int MaxConsoleLength = 2048; // around 30 lines
 
+        private const uint DeadExitCode = 0xDEAD;
```
Kr1sso: Maybe a comment that this is specific to cloudbuild only?

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -1066,6 +1066,7 @@ public static void PopulateLoggingAndLayoutConfiguration(CommandLineConfiguratio
                 }
 
                 mutableConfig.Logging.StoreFingerprints = initialCommandLineConfiguration.Logging.StoreFingerprints ?? false;
+                mutableConfig.Sandbox.RetryOnDeadExitCode = true;
```
dannyvv: >Dead [](start = 45, length = 4)

Maybe rename this to `AzureWatsonExitCode` Now it reads there is a generic exit code when a process crashes.

dannyvv: Note comment applies to all usage of 'dead' (other than the const that azure defined)

---
In reply to: [289928447](https://github.com/microsoft/BuildXL/pull/401#discussion_r289928447) [](ancestors = 289928447)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
mpysson: I like Danny's suggestion to name this AzureWatsonExitCode. When AzureWatson isn't active, this isn't applicable

Public/Src/Utilities/Configuration/ISandboxConfiguration.cs
```diff
@@ -220,5 +220,15 @@ public interface ISandboxConfiguration
...
+
+        /// <summary>
+        /// Retries process whose exit code or its children's exit code is 0xDEAD.
+        /// </summary>
+        /// <remarks>
+        /// Process can nondeterministically exit with 0xDEAD exit code in CB. This is the exit code
+        /// returned by Azure Watson dump after catching the process crash. The root cause
+        /// of the crash is unknown, but the primary suspect is the way Detours handle NtClose.
+        /// </remarks>
+        bool RetryOnDeadExitCode { get; }
```
mpysson: What's the scope of this? Is it just going to be for QB? Just for the specific pip that's hitting the detours break?

mpysson: Traditionally we've treated a compiler or other tool crash as any other build failure. Having a distinct exit code for crashes in AzureWatson enabled environments is an interesting angle. WDG has asked about having the ability to treat tool crashes different in reporting. No action from the BuildXL, just sharing some thoughts

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -1210,6 +1210,15 @@ internal static string PipWriteFileFailedMessage(string pipDescription, string p
             Message = Events.PipPrefix + "The number of messages sent by detoured processes did not match the number received by the {MainExecutableName} process. Refer to the {ShortProductName} log for more information.")]
         public abstract void LogMismatchedDetoursErrorCount(LoggingContext context, long pipSemiStableHash, string pipDescription);
 
+        [GeneratedEvent(
+            (int)EventId.PipExitedWithDeadExitCode,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Error,
+            Keywords = (int)Events.Keywords.UserMessage,
+            EventTask = (int)Events.Tasks.PipExecutor,
+            Message = Events.PipPrefix + "Pip exited with 0xDEAD exit code. Refer to the {ShortProductName} log for more information.")]
```
mpysson: So is the pip going to be retried? Could you make the message more explicit about what's going to happen? If you do retry, shouldn't this be a warning instead of an error?

Public/Src/Utilities/Configuration/ISandboxConfiguration.cs
narasamdya: For QuickBuild case, there's a queue and there are known targets in that queue that fail 95% of time with this exit code. I try to isolate the problem, and it seems to root at our detoured `NtClose`. 

Office also experiences this issue. But so far, from the Azure Watson's dumps, they cannot make definite conclusion what may have cause it. For Office, it's less often, and its non-deterministic nature makes it difficult to perform investigation.

---
In reply to: [289950639](https://github.com/microsoft/BuildXL/pull/401#discussion_r289950639) [](ancestors = 289950639)

narasamdya: " to treat tool crashes different in reporting" -- how different in reporting do they want? Or, what kind of reporting do they want?

---
In reply to: [289961049](https://github.com/microsoft/BuildXL/pull/401#discussion_r289961049) [](ancestors = 289961049)

Public/Src/Engine/Scheduler/Tracing/Log.cs
narasamdya: This log is only used after we perform several retries. Currently, with that exit code, we produces an error. So, this will also gives an error.

PR #403, opened 2019-06-03
Improve ContentLocationDatabase performance

jbayardo:
This PR includes changes to the ContentLocationDatabase so as to avoid communication with RocksDB as much as possible, and to make that communication faster. Changes are:

- RocksDB configuration changes for faster point queries.
- A in-memory cache between the ContentLocationDatabase and the database that helps keep as many operations as possible in-memory

There are some further improvements that could be achieved:
- Point lookups (RocksDB GET) are still the main performance bottleneck, so further work on improving its speed would be great.
- ContentLocationEntry is immutable, so we keep allocating instances that are short-lived anyways and that generates GC pressure.
- The tracing from the NagleOperationsTracer is the main driver of allocations, and should be considered for improvement.


Public/Src/Cache/ContentStore/Distributed/NuCache/InMemory/MemoryContentLocationEventStore.cs
```diff
@@ -138,6 +138,16 @@ public void Send(ContentLocationEventData eventData)
                     OnEvent(eventData);
                 }
             }
+
+            /// <summary>
+            /// The above function, without the lock. Used because of perf benchmarks getting contention on
```
norahuang: It is better to clearly indicate which function, instead of saying "the above function".

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -9,20 +9,151 @@
...
+using BuildXL.Utilities.ParallelAlgorithms;
 using BuildXL.Utilities.Tasks;
+using BuildXL.Utilities.Threading;
 using BuildXL.Utilities.Tracing;
 using static BuildXL.Cache.ContentStore.Distributed.Tracing.TracingStructuredExtensions;
 using AbsolutePath = BuildXL.Cache.ContentStore.Interfaces.FileSystem.AbsolutePath;
 
 namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 {
+    internal class FlushableCache
```
SergeyTeplyakov: Could you please move it to a separate file?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -119,6 +128,36 @@ public void ConfigureGarbageCollection(bool shouldDoGc)
...
+            if (_configuration.CacheEnabled)
+            {
+                // This clear is actually safe, as no operations should happen concurrently with this function.
+                _inMemoryCache.UnsafeClear();
+
+                Interlocked.Exchange(ref _cacheUpdatesSinceLastFlush, 0);
+
+                lock (_cacheFlushTimerLock)
+                {
+                    IsInMemoryCacheEnabled = isDatabaseWritable;
```
ljcollins25: >IsInMemoryCacheEnabled [](start = 20, length = 22)

Should this be set to false if _configuration.CacheEnabled == false

jbayardo: This branch can only be taken if _configuration.CacheEnabled is true in the first place. IsInMemoryCacheEnabled is initialized to false by default

---
In reply to: [291336968](https://github.com/microsoft/BuildXL/pull/403#discussion_r291336968) [](ancestors = 291336968)

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
```diff
@@ -0,0 +1,147 @@
...
+            using (_exchangeLock.AcquireReadLock())
+            {
+                // The entry could be a tombstone, so we need to make sure the user knows content has actually been
+                // deleted, which is why we check for null.
+                if (_cache.TryGetValue(hash, out entry))
+                {
+                    _database.Counters[ContentLocationDatabaseCounters.TotalNumberOfCacheHit].Increment();
+                    return entry != null;
+                }
+                else if (_flushingCache.TryGetValue(hash, out entry))
```
ljcollins25: >_flushingCache.TryGetValue(hash, out entry) [](start = 25, length = 43)

Only need to check flushingCache while flushing. Add boolean _isFlushing and set/unset it before and after flushing.

jbayardo: No longer applies due to latest changes, as discussed in person

---
In reply to: [291623570](https://github.com/microsoft/BuildXL/pull/403#discussion_r291623570) [](ancestors = 291623570)

Public/Src/Utilities/KeyValueStore/IKeyValueStore.cs
```diff
@@ -82,5 +82,21 @@ public interface IKeyValueStore<in TKey, TValue>
...
+        /// The key of each entry.
+        /// </param>
+        /// <param name="values">
+        /// The value of each entry. Must be the same size as the keys. If a value is null, then a Delete will be 
+        /// performed; otherwise, a Put.
+        /// </param>
+        /// <param name="columnFamilyName">
+        /// The column family to use.
+        /// </param>
+        void ApplyBatch(IEnumerable<TKey> keys, IEnumerable<TValue> values, string columnFamilyName = null);
```
ljcollins25: >IEnumerable<TKey> keys, IEnumerable<TValue> values [](start = 24, length = 50)

Just use IEnumerable<KeyValuePair<TKey, TValue>>?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
```diff
@@ -7,6 +7,31 @@
 
 namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 {
+    /// <summary>
+    /// Configuration type for <see cref="FlushableCache"/>
+    /// </summary>
+    public class FlushableCacheConfiguration
```
ljcollins25: >FlushableCacheConfiguration [](start = 17, length = 27)

Can we just inline this into the normal ContentLocationDatabaseConfiguration?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -355,19 +456,96 @@ public BoolResult RestoreCheckpoint(OperationContext context, AbsolutePath check
...
+                return BoolResult.SuccessTask;
+            }
+
+            return context.PerformOperationAsync(
+                Tracer,
+                async () =>
+                {
+                    try
+                    {
+                        await _inMemoryCache.FlushAsync(context);
```
ljcollins25: >FlushAsync [](start = 45, length = 10)

Its worth noting that this could execute synchronously. Maybe we should do an await Task.Yield() above.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -106,10 +108,17 @@ public static ContentLocationDatabase Create(IClock clock, ContentLocationDataba
             }
         }
 
+        /// <todoc />
```
dcwuser: Don't leave <todoc /> on public methods.

SergeyTeplyakov: This is a placeholder, kind of WIP. So yep, we should comment this.

---
In reply to: [292137263](https://github.com/microsoft/BuildXL/pull/403#discussion_r292137263) [](ancestors = 292137263)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -131,6 +170,18 @@ protected override Task<BoolResult> StartupCoreAsync(OperationContext context)
                     Timeout.InfiniteTimeSpan);
             }
 
+            if (_configuration.CacheEnabled && _configuration.CacheFlushingMaximumInterval != Timeout.InfiniteTimeSpan)
+            {
+                _inMemoryCacheFlushTimer = new Timer(
+                    _ => {
+                        Counters[ContentLocationDatabaseCounters.NumberOfCacheFlushesTriggeredByTimer].Increment();
+                        ForceCacheFlush(context);
```
ljcollins25: >ForceCacheFlush [](start = 24, length = 15)

Maybe just call ForceCacheFlushAsync().FireAndForget() to avoid blocking two threads.

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
```diff
@@ -0,0 +1,165 @@
...
+                {
+                    var actionBlock = new ActionBlockSlim<KeyValuePair<ShortHash, ContentLocationEntry>>(_configuration.FlushDegreeOfParallelism, kv =>
+                    {
+                        // Do not lock on GetLock here, as it will cause a deadlock with
+                        // SetMachineExistenceAndUpdateDatabase. It is correct not do take any locks as well, because
+                        // no Store can happen while flush is running.
+                        _database.Persist(context, kv.Key, kv.Value);
+                    });
+
+                    foreach (var kv in _flushingCache)
```
ljcollins25: >_flushingCache [](start = 39, length = 14)

In theory you could call GetPages() in to create batches rather than using a single item database.Persist()

dcwuser:
Just wanted to say, very happy to see: (1) making member and type access modifiers as restricted as posssible (e.g. internal or protected instead of public), (2) explicitly declared types instead of var (making code review easier), (3) plenty of comments.

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
```diff
@@ -0,0 +1,165 @@
...
+            // This lock is required to ensure no flushes happen concurrently. We may loose updates if that happens.
+            // AcquireAsync is used so as to avoid multiple concurrent tasks just waiting; this way we return the
+            // task to the thread pool in between.
+            using (await _flushTurnstile.AcquireAsync())
+            {
+                PerformFlush(context);
+            }
+        }
+
+        private void PerformFlush(OperationContext context)
```
ljcollins25: >PerformFlush [](start = 21, length = 12)

I'm tempted to say this should just be inlined into FlushAsync(). It wasn't immediately clear that this is always called inside the flushTurnstile lock

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// </summary>
+    internal class FlushableCache
+    {
+        private readonly FlushableCacheConfiguration _configuration;
+        private readonly ContentLocationDatabase _database;
+
+        private ConcurrentBigMap<ShortHash, ContentLocationEntry> _cache = new ConcurrentBigMap<ShortHash, ContentLocationEntry>();
+        private ConcurrentBigMap<ShortHash, ContentLocationEntry> _flushingCache = new ConcurrentBigMap<ShortHash, ContentLocationEntry>();
+
+        private readonly SemaphoreSlim _flushTurnstile = new SemaphoreSlim(1);
```
ljcollins25: >_flushTurnstile [](start = 39, length = 15)

_flushMutex would be clearer to me

Public/Src/Utilities/KeyValueStore/IKeyValueStore.cs
jbayardo: Tried this at the beginning; but its not possible: the type is contravariant on TKey

---
In reply to: [292134200](https://github.com/microsoft/BuildXL/pull/403#discussion_r292134200) [](ancestors = 292134200)

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
jbayardo: Refactored it out just to reduce nesting. I have added a comment.

---
In reply to: [292140919](https://github.com/microsoft/BuildXL/pull/403#discussion_r292140919) [](ancestors = 292140919)

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
jbayardo: Indeed, I'd rather not do this change in this PR for now, but in a follow-up if we end up seeing that single-threaded flushing of WriteBatches is indeed problematic

---
In reply to: [292139868](https://github.com/microsoft/BuildXL/pull/403#discussion_r292139868) [](ancestors = 292139868)

Public/Src/Utilities/KeyValueStore/IKeyValueStore.cs
ljcollins25: Does a tuple work?

---
In reply to: [292149253](https://github.com/microsoft/BuildXL/pull/403#discussion_r292149253) [](ancestors = 292149253,292134200)

jbayardo: No, it has the same issue

PR #404, opened 2019-06-03
Restore FILE_SHARE_DELETE in Detoured CreateFile/NtCreateFile

narasamdya:
We need to add FILE_SHARE_DELETE to dwShareMode to leverage NTFS hardlinks to avoid copying cache content, i.e., we need to be able to delete one of many links to a file. Unfortunately, share-mode is aggregated only per file rather than per-link, so in order to keep unused links delete-able, we should ensure in-use links are delete-able as well. However, adding FILE_SHARE_DELETE may be unexpected, for example, some unit tests may test for sharing violation. Thus, we only add FILE_SHARE_DELETE if the file is untracked.

[AB#1530341](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1530341)

### DIFF
--- a/Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
+++ b/Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
@@ -2089,7 +2089,7 @@ HANDLE WINAPI Detoured_CreateFileW(
     // content, i.e., we need to be able to delete one of many links to a file. Unfortunately, share-mode is aggregated only per file
     // rather than per-link, so in order to keep unused links delete-able, we should ensure in-use links are delete-able as well.
     // However, adding FILE_SHARE_DELETE may be unexpected, for example, some unit tests may test for sharing violation. Thus,
-    // we only add FILE_SHARE_DELETE if the file is untracked.
+    // we only add FILE_SHARE_DELETE if the file is tracked.
     
     // We also add FILE_SHARE_READ when it is safe to do so, since some tools accidentally ask for exclusive access on their inputs.
 
@@ -5157,7 +5157,7 @@ NTSTATUS NTAPI Detoured_ZwCreateFile(
     // content, i.e., we need to be able to delete one of many links to a file. Unfortunately, share-mode is aggregated only per file
     // rather than per-link, so in order to keep unused links delete-able, we should ensure in-use links are delete-able as well.
     // However, adding FILE_SHARE_DELETE may be unexpected, for example, some unit tests may test for sharing violation. Thus,
-    // we only add FILE_SHARE_DELETE if the file is untracked.
+    // we only add FILE_SHARE_DELETE if the file is tracked.
 
     // We also add FILE_SHARE_READ when it is safe to do so, since some tools accidentally ask for exclusive access on their inputs.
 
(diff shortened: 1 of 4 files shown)


Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -2085,10 +2085,11 @@ HANDLE WINAPI Detoured_CreateFileW(
...
     // It is safe to go ahead and perform the real CreateFile() call, and then to reason about the results after the fact.
 
-    // Note that we can add FILE_SHARE_DELETE to dwShareMode. I.e., in order to leverage NTFS hardlinks to avoid copying cache
-    // content, we need to be able to delete one of many links to a file. Unfortunately, share-mode is aggregated only per file
+    // Note that we need to add FILE_SHARE_DELETE to dwShareMode to leverage NTFS hardlinks to avoid copying cache
+    // content, i.e., we need to be able to delete one of many links to a file. Unfortunately, share-mode is aggregated only per file
     // rather than per-link, so in order to keep unused links delete-able, we should ensure in-use links are delete-able as well.
-    // However, adding FILE_SHARE_DELETE may be unexpected, for example, some unit tests may test for sharing violation.
+    // However, adding FILE_SHARE_DELETE may be unexpected, for example, some unit tests may test for sharing violation. Thus,
+    // we only add FILE_SHARE_DELETE if the file is untracked.
```
mpysson: >we only add FILE_SHARE_DELETE if the file is untracked. [](start = 7, length = 55)

Is this backwards? We only add it if the file is tracked

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -5735,7 +5739,7 @@ NTSTATUS NTAPI Detoured_ZwOpenFile(
     {
         DWORD readSharingIfNeeded = policyResult.ShouldForceReadSharing(accessCheck) ? FILE_SHARE_READ : 0UL;
         desiredAccess = !forceReadOnlyForRequestedRWAccess ? desiredAccess : (desiredAccess & FILE_GENERIC_READ);
-        sharedAccess = sharedAccess | readSharingIfNeeded;
+        sharedAccess = sharedAccess | readSharingIfNeeded | FILE_SHARE_DELETE;
     }
```
mpysson: Duplicated 4 times :(

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -2085,10 +2085,11 @@ HANDLE WINAPI Detoured_CreateFileW(
     // read request which may or may not have been approved (due to special exceptions for directories and non-existent files).
     // It is safe to go ahead and perform the real CreateFile() call, and then to reason about the results after the fact.
 
-    // Note that we can add FILE_SHARE_DELETE to dwShareMode. I.e., in order to leverage NTFS hardlinks to avoid copying cache
-    // content, we need to be able to delete one of many links to a file. Unfortunately, share-mode is aggregated only per file
+    // Note that we need to add FILE_SHARE_DELETE to dwShareMode to leverage NTFS hardlinks to avoid copying cache
```
mpysson: >    // N [](start = 0, length = 8)

There's a slight chance that we may still run into problems here. The QTest sandbox is created with CopyFile pips, which use hardlinks by default. So conceivably a test could have a handle open to a file in its sandbox while BuildXL tries to delete another hardlink to the same file that's used elsewhere in the build. In all likelihood I don't think we'll run into that, but it'll be a pain to debug if we do.

I guess we could only allow this when files are copied rather than hardlinked. That could be config based or you could check the link count of files. Doing the latter may have perf implications.

Again, I think we'll probably be fine with this. Just wanted to bring this up in case you think it's worth addressing now.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
narasamdya: I'll refactor it later.

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
narasamdya: Thanks for the scenario. I think we should have a unit test for this.

PR #405, opened 2019-06-03
Add some debugging info to Contracts in BitSet

olkononenko:
It appears that GetIndexByPath(path) might return values bigger than PathTable.Count. Adding debugging info to the contracts to confirm that this is the case.

[AB#1532584](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1532584)

### DIFF
--- a/Public/Src/Utilities/Collections/BitSet.cs
+++ b/Public/Src/Utilities/Collections/BitSet.cs
@@ -100,7 +100,10 @@ namespace BuildXL.Utilities.Collections
         /// </summary>
         public void Add(int index)
         {
-            Contract.Requires(index >= 0 && index < Length);
+            if (index < 0 || index >= Length)
+            {
+                Contract.Requires(false, $"index={index} must be within [0, {Length}) range.");
+            }            
 
             ulong newEntry = GetEntry(index) | (1UL << (index % 64));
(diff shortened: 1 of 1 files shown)


Public/Src/Utilities/Collections/BitSet.cs
```diff
@@ -100,7 +100,10 @@ private bool TryCompareExchangeEntry(int index, ulong value, ulong comparand)
         /// </summary>
         public void Add(int index)
         {
-            Contract.Requires(index >= 0 && index < Length);
+            if (index < 0 || index >= Length)
+            {
+                Contract.Requires(false, $"index={index} must be within [0, {Length}) range.");
```
almiliMSFT:
```suggestion
                Contract.Assert(false, $"index={index} must be within [0, {Length}) range.");
```

mpysson: Did you mean a ] instead of ) ?
> $"index={index} must be within [0, {Length}) range.");

$"index={index} must be within [0, {Length}**]** range.");

olkononenko: No, the supported range is [0, Length), i.e., Length is not included.

PR #409, opened 2019-06-04
Increase process timeout period for sandboxed test cases

Kr1sso:
Some of our tests are timing out and failing our CI pipelines with runtimes slightly above 10 minutes. This PR pumps the timeout as a measure to not fail tests on slight deviations.

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcess.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcess.cs
@@ -81,8 +81,8 @@ namespace BuildXL.Processes
                 s_binaryPaths = new BinaryPaths(); // this can take a while; performs I/O
             }
 
-            // If unspecified make the injection timeout 10 mins. Also, make it no less than 10 mins.
-            m_timeoutMins = info.Timeout.HasValue ? ((uint)info.Timeout.Value.TotalMinutes) : 10;
+            // If unspecified make the injection timeout 15 mins. Also, make it no less than 10 mins.
+            m_timeoutMins = info.Timeout.HasValue ? ((uint)info.Timeout.Value.TotalMinutes) : 15;
             if (m_timeoutMins < 10)
             {
                 m_timeoutMins = 10;
@@ -595,7 +595,7 @@ namespace BuildXL.Processes
                     }
                 }
             }
-            
+
 
             return survivingChildProcesses;
         }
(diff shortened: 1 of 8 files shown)


Public/Src/Engine/Processes/SandboxedProcess.cs
```diff
@@ -81,11 +81,11 @@ internal SandboxedProcess(SandboxedProcessInfo info)
                 s_binaryPaths = new BinaryPaths(); // this can take a while; performs I/O
             }
 
-            // If unspecified make the injection timeout 10 mins. Also, make it no less than 10 mins.
-            m_timeoutMins = info.Timeout.HasValue ? ((uint)info.Timeout.Value.TotalMinutes) : 10;
-            if (m_timeoutMins < 10)
+            // If unspecified make the injection timeout 15 mins. Also, make it no less than 15 mins.
+            m_timeoutMins = info.Timeout.HasValue ? ((uint)info.Timeout.Value.TotalMinutes) : 15;
+            if (m_timeoutMins < 15)
```
almiliMSFT: can you use `DefaultProcessTimeout` from `UnSandboxedProcess.cs`? (or the other way around)

Public/Src/Utilities/Configuration/Mutable/SandboxConfiguration.cs
```diff
@@ -16,7 +16,7 @@ public SandboxConfiguration()
             m_unsafeSandboxConfig = new UnsafeSandboxConfiguration();
 
             FailUnexpectedFileAccesses = true;
-            DefaultTimeout = 10 * 60 * 1000;
+            DefaultTimeout = 15 * 60 * 1000;
```
almiliMSFT: This is going to change the default timeout for all pips in all builds, right?  If so, I think this default value is also mentioned in the help text, so it should be updated there as well

Public/Src/Utilities/Utilities/VmCommandProxy/VmInitializer.cs
```diff
@@ -25,7 +25,7 @@ public class VmInitializer
         /// <remarks>
         /// According to CB team, VM initialization roughly takes 2-3 minutes.
         /// </remarks>
-        private const int InitVmTimeoutInMinute = 10;
+        private const int InitVmTimeoutInMinute = 15;
```
almiliMSFT: Is this related?  It looks to me that it is not

Kr1sso: Good catch, search replace issue.

Public/Src/Utilities/Configuration/Mutable/SandboxConfiguration.cs
Kr1sso: Help text is just a generic message explaining the value semantics.

PR #410, opened 2019-06-04
Fix explorer by initialing global hashing algo

dannyvv:


Public/Src/Explorer/Server/Program.cs
```diff
@@ -16,6 +16,10 @@ public static void Main(string[] args)
             {
                 Debugger.Launch();
             }
+
+            // Globally this is needed to initialize the hash code
+            BuildXL.Storage.ContentHashingUtilities.SetDefaultHashType();
```
olkononenko: Is it safe to always use the 'default' hash algo?

dannyvv: All codebases use this function. It should match what BuildXL is using so if that implementation changes, it should still match :)

PR #414, opened 2019-06-05
Don't log error when processes retry since retry could succeed

mpysson:
We recently added some logging in SandboxedProcessPipExecutor for the exit code on retries. This logged an error which breaks assertions about no errors being logged on success since the pip sometimes succeeds on retry.

Fixes [AB#1550935](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1550935)

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -1218,12 +1218,6 @@ namespace BuildXL.Processes
 
                         if (m_pip.RetryExitCodes.Contains(result.ExitCode) && m_remainingUserRetryCount > 0)
                         {
-                            // Retry if user specifies that the exit code can be retried.
-                            if (await TrySaveAndLogStandardOutputAsync(result) && await TrySaveAndLogStandardErrorAsync(result))
-                            {
-                                await TryLogErrorAsync(result, exitedWithSuccessExitCode);
-                            }
-
                             return SandboxedProcessPipExecutionResult.RetryProcessDueToUserSpecifiedExitCode(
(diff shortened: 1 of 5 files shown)


mpysson:
@narasamdya - We need this fix for our week's release candidate. I've already pulled it back into that branch so I can run the patched version against our validation suite overnight. Because of this I'd like to commit this change roughly as-is unless there are big problems and then address feedback in a subsequent PR.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1218,12 +1218,6 @@ private void LogCommandLineTooLong(SandboxedProcessInfo info)
 
                         if (m_pip.RetryExitCodes.Contains(result.ExitCode) && m_remainingUserRetryCount > 0)
                         {
-                            // Retry if user specifies that the exit code can be retried.
-                            if (await TrySaveAndLogStandardOutputAsync(result) && await TrySaveAndLogStandardErrorAsync(result))
-                            {
-                                await TryLogErrorAsync(result, exitedWithSuccessExitCode);
```
dannyvv: >TryLogErrorAsync [](start = 38, length = 16)

Should we log the error details as an info here? So we have telemetry and data on which processes are failing periodically and can investigate?

mpysson: Yes. I also want to address my todo in the test framework and extend this validation to all integration tests. I'll tweak that in a follow up PR.

dannyvv: sounds good

PR #415, opened 2019-06-05
Allow orchestrator re-run with agent re-use

Kr1sso:
This PR allows consecutive orchestrated runs on the same VSTS build definition, re-using agent data and allowing for continue agent info advertising through the AzDev APIs.

### DIFF
--- a/Public/Src/Tools/Orchestrator/Vsts/Api.cs
+++ b/Public/Src/Tools/Orchestrator/Vsts/Api.cs
@@ -172,15 +172,10 @@ namespace BuildXL.Orchestrator.Vsts
 
             if (record != null)
             {
-                // Try to remove all values for the default keys first, this makes sure updating for the same agent works and has no effect if variables are empty
-                record.Variables.Remove(Constants.MachineType);
-                record.Variables.Remove(Constants.MachineHostName);
-                record.Variables.Remove(Constants.MachineIpV4Address);
-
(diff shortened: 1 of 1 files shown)


Public/Src/Tools/Orchestrator/Vsts/Api.cs
```diff
@@ -172,6 +172,12 @@ public async Task SetMachineReadyToBuild(string hostName, string ipV4Address, bo
 
             if (record != null)
             {
+                // Try to remove all values for the default keys first, this makes sure updating for the same agent works and has no effect if variables are empty
+                record.Variables.Remove(Constants.MachineType);
+                record.Variables.Remove(Constants.MachineHostName);
+                record.Variables.Remove(Constants.MachineIpV4Address);
+
+                // Add agent info for the build orchestration
                 record.Variables.Add(Constants.MachineType, (isMaster ? AgentType.Master : AgentType.Worker).ToString());
```
dannyvv: >cord.Variabl [](start = 18, length = 12)

Since this is an IDictionary you could also update this to use indexers:

```
record.Variables[Constants.MachineHostName] = hostName;
```

which overrides if already defined, no need to do two operations...

Kr1sso: Good point, changed!

PR #422, opened 2019-06-06
Refactor TestProcess to not use PathTable to speed up tests

mpysson:
The SchedulerIntegrationTests call TestProcess.exe hundreds of times. Tweaking it to not use the PathTable shaves off about 100ms of JIT per invocation which really adds up in our test suite. The end result is a bit inelegant but I think it's worth it for the faster test execution time.

### DIFF
--- a/Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
+++ b/Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
@@ -240,7 +240,7 @@ namespace Test.BuildXL.Executables.TestProcess
         public Type OpType { get; private set; }
 
         /// <summary>
-        /// Path to file. NOTE: This will not be accessable within TestProcess.exe
+        /// Path to file. NOTE: This will not be accessible within TestProcess.exe
         /// </summary>
         public FileOrDirectoryArtifact Path { get; private set; }
 
--- a/Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
+++ b/Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
@@ -239,30 +239,34 @@ namespace Test.BuildXL.Executables.TestProcess
         /// </summary>
         public Type OpType { get; private set; }
 
+#if TestProcess
+        /// <summary>
+        /// Should only be used in the context of TestProcess.exe
+        /// </summary>
+        private string PathAsString { get; set; }
+#endif
+
         /// <summary>
         /// Path to file. NOTE: This will not be accessible within TestProcess.exe
         /// </summary>
         public FileOrDirectoryArtifact Path { get; private set; }
 
         /// <summary>
-        /// Should only be used in the context of TestProcess.exe
-        /// </summary>
-        private string PathAsString { get; set; }
-
-        /// <summary>
         /// Content to write to file
         /// </summary>
         public string Content { get; private set; }
+
+#if TestProcess
+        /// <summary>
+        /// Should only be used in the context of TestProcess.exe
+        /// </summary>
+        private string LinkPathAsString { get; set; }
+#endif
 
         /// <summary>
         /// Path of new symlink or hardlink to create. NOTE: This will not be accessible within TestProcess.exe
         /// </summary>
         public FileOrDirectoryArtifact LinkPath { get; private set; }
-
-        /// <summary>
-        /// Should only be used in the context of TestProcess.exe
-        /// </summary>
-        private string LinkPathAsString { get; set; }
 
         /// <summary>
         /// Flag for correcting file or directory symlink
@@ -298,6 +302,7 @@ namespace Test.BuildXL.Executables.TestProcess
             AdditionalArgs = additionalArgs;
         }
 
+#if TestProcess
         private static Operation FromCommandLine(Type type, string path = null, string content = null, string linkPath = null, SymbolicLinkFlag? symLinkFlag = null, string additionalArgs = null)
         {
             Contract.Requires(content == null || !content.Contains(Environment.NewLine));
@@ -312,6 +317,7 @@ namespace Test.BuildXL.Executables.TestProcess
 
             return result;
         }
+#endif
 
         /// <summary>
         /// Executes the associated filesystem operation
@@ -1041,9 +1047,9 @@ namespace Test.BuildXL.Executables.TestProcess
             }
 
             Type opType;
-            string pathAsString = null;
+            string pathAsString = "{Invalid}";
             SymbolicLinkFlag symLinkFlag;
-            string linkPathAsString = null;
+            string linkPathAsString = "{Invalid}";
 
             if (!Enum.TryParse<Type>(opArgs[0], out opType))
             {
--- a/Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
+++ b/Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
@@ -324,6 +324,17 @@ namespace Test.BuildXL.Executables.TestProcess
         /// </summary>
         public void Run()
         {
+            // Make sure the string version of the paths are set in case the operation is run outside of the context of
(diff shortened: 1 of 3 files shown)


Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -853,10 +876,10 @@ private void DoEnumerateDir()
 
         private void DoCreateSymlink()
         {
-            string target = AdditionalArgs ?? FileOrDirectoryToString(Path);
+            string target = AdditionalArgs ?? PathAsString;
 
             // delete whatever might exist at the link location
-            var linkPath = FileOrDirectoryToString(LinkPath);
+            var linkPath = LinkPathAsString;
             FileUtilities.DeleteFile(linkPath);
```
Kr1sso: This asserts that linkPath is non null/empty and should be conditionalized to only delete if a file exists at `linkPath`.

Kr1sso:
```suggestion
            if (FileUtilities.Exists(linkPath))
            {
                FileUtilities.DeleteFile(linkPath);
            }
```

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -240,20 +240,30 @@ public enum SymbolicLinkFlag : uint
         public Type OpType { get; private set; }
 
         /// <summary>
-        /// Path to file
+        /// Path to file. NOTE: This will not be accessable within TestProcess.exe
```
dannyvv: >TestProcess [](start = 67, length = 11)

Should this then simply be conditionally compiled bteween the two?

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -1004,7 +1026,11 @@ private void DoSucceedOnRetry()
         /// Converts command line arg input to process ops
         /// Input format: "OpType?Path?Content?LinkPath?SymLinkFlag"
         /// </summary>
-        public static Operation CreateFromCommandLine(PathTable pathTable, string commandLineArg)
+        /// <remarks>
+        /// This and the execution of Operations within the context of TestProcess.exe should not utilize the PathTable
+        /// since this adds an appreciable amount of JIT overhead to our test execution
```
dannyvv: >of JIT overhead [](start = 50, length = 15)

Since we are now mostly on CoreCLR and we can ilgen the assembly. Would this still be the case? Should we just disable this test on full framework?

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
almiliMSFT:
```suggestion
        /// Path to file. NOTE: This will not be accessible within TestProcess.exe
```

almiliMSFT:
```suggestion
        /// Path to file. NOTE: This will not be accessible within TestProcess.exe
```

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -240,20 +240,30 @@ public enum SymbolicLinkFlag : uint
...
+        private string PathAsString { get; set; }
+
         /// <summary>
         /// Content to write to file
         /// </summary>
         public string Content { get; private set; }
 
         /// <summary>
-        /// Path of new symlink or hardlink to create
+        /// Path of new symlink or hardlink to create. NOTE: This will not be accessable within TestProcess.exe
```
almiliMSFT:
```suggestion
        /// Path of new symlink or hardlink to create. NOTE: This will not be accessible within TestProcess.exe
```

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -240,20 +240,30 @@ public enum SymbolicLinkFlag : uint
...
         /// </summary>
         public FileOrDirectoryArtifact LinkPath { get; private set; }
 
+        /// <summary>
+        /// Should only be used in the context of TestProcess.exe
+        /// </summary>
+        private string LinkPathAsString { get; set; }
```
olkononenko: >set [](start = 47, length = 3)

'private set' ? I don't think this field should be changeable outside this class

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -288,6 +298,21 @@ private Operation(Type type, FileOrDirectoryArtifact? path = null, string conten
...
+        {
+            Contract.Requires(content == null || !content.Contains(Environment.NewLine));
+
+            var result = new Operation(type: type,
+                content: content,
+                symLinkFlag: symLinkFlag,
+                doNotInfer: false,
+                additionalArgs: additionalArgs);
+            result.PathAsString = path;
+            result.LinkPathAsString = linkPath;
```
olkononenko: nit: should we create another ctor that takes strings?

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
mpysson: FileUtilities.DeleteFile() doesn't fail if the file isn't there, so making that change yields no behavioral difference.

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
mpysson: I was thinking on the same lines, but as long as we have most of our customers on full framework i want to keep running the full framework tests. I didn't look into perf implications of this change on NetCore.

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
mpysson: good idea. done.

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
mpysson: The property itself is private. You cannot specify access modifiers on getters/setters when that's the case

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
Kr1sso: The assert failing in the unit tests is due to link path being empty/null I thought. `Exists` should return false on invalid path and thus we don't fail the assert in `DeleteFile`. I thought that was macOS unit test failure I saw.

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
mpysson: I preferred a static helper method over a constructor here in order to give it a name to make it clear in what context it should be used. A constructor would have made it appear more general purpose.

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
mpysson: Oh, I missed your point. Previously the FileOrDirectoryToString(LinkPath) would have been returning a string with "{Invalid}" if the LinkPath weren't specified. So it would have been calling FileUtilities.DeleteFile() with a path that didn't reference a real file. Now it's calling it with an empty string. I admittedly hadn't seen the test failure you were talking about since the pipeline is otherwise broken

mpysson: Switched it to default to {Invalid} for the filename to preserve the prior implementation's behavior

PR #426, opened 2019-06-06
Update Nuget hash versions to force specs to be generated

dannyvv:
We sometimes see unreliable nuget spec generation, bumping version to force regeneration


dannyvv:
Issue with nuget feed lost a bond package, abandoning...

PR #427, opened 2019-06-06
Deal with relative paths in PATH

norahuang:
Transform relative paths in PATH env var to absolute paths to match behavior of Environment.getPathValues();


Public/Src/FrontEnd/UnitTests/Script.Interpretation/InterpreterTests.cs
```diff
@@ -488,18 +490,21 @@ namespace M
...
-                    return path.Replace("/", "\\").Replace("\\;", ";").Replace("p`", string.Empty).Replace("`", string.Empty);
+                    return path.Replace("/", "\\").Replace("\\;", separator).Replace("p`", string.Empty).Replace("`", string.Empty);
                 };
 
             Func<string, string> toUpperInvariant = s => string.IsNullOrWhiteSpace(s) ? s : s.ToUpperInvariant();
 
             var userName = Environment.GetEnvironmentVariable("USERNAME");
             var sysRoot = Environment.GetEnvironmentVariable("SystemRoot");
-            var paths = Environment.GetEnvironmentVariable("Path");
+            var paths = Environment.GetEnvironmentVariable("Path");
```
dannyvv: >           [](start = 67, length = 10)

nit: extra spaces

Public/Src/FrontEnd/UnitTests/Script.Interpretation/InterpreterTests.cs
```diff
@@ -488,18 +490,21 @@ namespace M
...
                 };
 
             Func<string, string> toUpperInvariant = s => string.IsNullOrWhiteSpace(s) ? s : s.ToUpperInvariant();
 
             var userName = Environment.GetEnvironmentVariable("USERNAME");
             var sysRoot = Environment.GetEnvironmentVariable("SystemRoot");
-            var paths = Environment.GetEnvironmentVariable("Path");
+            var paths = Environment.GetEnvironmentVariable("Path");          
+            var pathValues = paths.Split(new[] { separator }, StringSplitOptions.RemoveEmptyEntries);
+            var absolutePaths = pathValues.Aggregate((currentAccumPaths, nextPath) => currentAccumPaths + separator + Path.GetFullPath(nextPath));
```
dannyvv: >+ [](start = 104, length = 1)

Fine for unittest, but in product code we should use a string builder since this is sum(1..n) allocations

dannyvv:
Curious why this fails, this test should not depend on the current host since it runs under BuildXL which is limmiting the environment varaibesl being passed in...
Is this broken perhaps by QTest?

norahuang:
> Curious why this fails, this test should not depend on the current host since it runs under BuildXL which is limmiting the environment varaibesl being passed in...
> Is this broken perhaps by QTest?

The latest Qtest version set the path with "B:\\out\\frontend\\Nuget\\pkgs\\CB.QTest.19.5.29.221321\\tools\\V150\\Common7\\IDE\\CommonExtensions\\Microsoft\\IntelliTrace\\..\\..\\..\\..\\..\\Team Tools\\Dynamic Code Coverage Tools\\amd64". This relative path makes the case constantly fail when enable code coverage feature. I wanted to ask them change the env vars value. But it seems inappropriate to request this innocent change,

dannyvv:
> > Curious why this fails, this test should not depend on the current host since it runs under BuildXL which is limmiting the environment varaibesl being passed in...
> > Is this broken perhaps by QTest?
> 
> The latest Qtest version set the path with "B:\out\frontend\Nuget\pkgs\CB.QTest.19.5.29.221321\tools\V150\Common7\IDE\CommonExtensions\Microsoft\IntelliTrace\..\..\..\..\..\Team Tools\Dynamic Code Coverage Tools\amd64". This relative path makes the case constantly fail when enable code coverage feature. I wanted to ask them change the env vars value. But it seems inappropriate to request this innocent change,

Ah! So qtest 'changes' the PATH environment variable???? Sneaky....

PR #428, opened 2019-06-07
Make sure that VS solution generator skips over netcoreapp processes

almiliMSFT:


Public/Src/IDE/Generator/IdeGenerator.cs
```diff
@@ -206,7 +206,7 @@ private void ProcessRawReferences(IReadOnlyList<MsbuildFile> msbuildFiles)
                             item.SetMetadata("Name", referenceName);
                             msbuildFile.ProjectReferences.Add(referencedMsbuildFile);
                         }
-                        else if (msbuildFile is CsprojFile)
+                        else if (msbuildFile is CsprojFile) // && !reference.ToString(m_context.PathTable).Contains("NETCore.App"))
```
almiliMSFT:
```suggestion
                        else if (msbuildFile is CsprojFile)
```

Public/Src/IDE/Generator/CsprojFile.cs
```diff
@@ -41,6 +41,20 @@ internal sealed class CsprojFile : MsbuildFile
 
         internal override void VisitProcess(Process process, ProcessType pipCategory)
         {
+            var qualifier = Context.QualifierTable.GetQualifier(process.Provenance.QualifierId);
+
+            // only consider processes targeting Windows
+            if (qualifier.TryGetValue(Context.StringTable, "targetRuntime", out var targetRuntime) && targetRuntime != "win-x64")
```
almiliMSFT:
```suggestion
            if (qualifier.TryGetValue(Context.StringTable, "targetRuntime", out var targetRuntime) && 
                targetRuntime != "win-x64")
```

Public/Src/IDE/Generator/CsprojFile.cs
```diff
@@ -41,6 +41,20 @@ internal sealed class CsprojFile : MsbuildFile
...
+            var qualifier = Context.QualifierTable.GetQualifier(process.Provenance.QualifierId);
+
+            // only consider processes targeting Windows
+            if (qualifier.TryGetValue(Context.StringTable, "targetRuntime", out var targetRuntime) && targetRuntime != "win-x64")
+            {
+                return;
+            }
+            
+            // HACK: skip over processes targeting netcoreapp framework
+            if (qualifier.TryGetValue(Context.StringTable, "targetFramework", out var targetFramework) && targetFramework.Contains("netcoreapp"))
```
almiliMSFT:
```suggestion
            if (qualifier.TryGetValue(Context.StringTable, "targetFramework", out var targetFramework)
                && targetFramework.Contains("netcoreapp"))
```

dannyvv:
> I'm wondering if we should just spend the effort and make it play nice with .NET Core, so the generated solution fully works as expected?

Agree, would love for us to schedule time to emit an Sdk-Style project and handle multipel qualifeirs

PR #431, opened 2019-06-07
Fix PutStream for non-seekable streams

jbayardo:
With a non-seekable stream, we would return a success, but no content would actually be inserted. This fixes the issue.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Sessions/ServiceClientContentSession.cs
+++ b/Public/Src/Cache/ContentStore/Library/Sessions/ServiceClientContentSession.cs
@@ -38,65 +38,47 @@ namespace BuildXL.Cache.ContentStore.Sessions
         }
 
         /// <inheritdoc />
-        protected override Task<PutResult> PutStreamCoreAsync(
+        protected override async Task<PutResult> PutStreamCoreAsync(
             OperationContext operationContext, HashType hashType, Stream stream, UrgencyHint urgencyHint, Counter retryCounter)
         {
             // We need a seekable stream, that can give its length. If the input stream is seekable, we can use it directly.
             // Otherwise, we need to create a temp file for this purpose.
-            Stream putStream;
-            long position;
-            Stream disposableStream;
-
-            if (stream.CanSeek)
-            {
-                putStream = stream;
-                position = stream.Position;
-                disposableStream = null;
-            }
-            else
+            var putStream = stream;
+            Stream disposableStream = null;
+            if (!stream.CanSeek)
             {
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Library/Sessions/ServiceClientContentSession.cs
```diff
@@ -38,65 +38,47 @@ public class ServiceClientContentSession : ReadOnlyServiceClientContentSession,
...
+            var putStream = stream;
+            Stream disposableStream = null;
+            if (!stream.CanSeek)
             {
                 putStream = TempFileStreamFactory.Create(operationContext, stream);
-                position = 0;
                 disposableStream = putStream;
             }
 
             using (disposableStream)
```
SergeyTeplyakov: disposableStream is still not used... Is it ok?

SergeyTeplyakov: Ignore this. It is fine. putStream is used. So we're good.

---
In reply to: [292134755](https://github.com/microsoft/BuildXL/pull/431#discussion_r292134755) [](ancestors = 292134755)

Public/Src/Cache/ContentStore/Test/Sessions/ServiceClientContentSessionTests.cs
```diff
@@ -23,6 +23,8 @@
 using ContentStoreTest.Test;
 using FluentAssertions;
 using Xunit;
+using System.IO;
```
kewilli: Nit: Please sort your usings :)

Public/Src/Cache/ContentStore/Library/Sessions/ServiceClientContentSession.cs
```diff
@@ -38,65 +38,47 @@ public class ServiceClientContentSession : ReadOnlyServiceClientContentSession,
         }
 
         /// <inheritdoc />
-        protected override Task<PutResult> PutStreamCoreAsync(
+        protected override async Task<PutResult> PutStreamCoreAsync(
```
dcwuser: What was reason for async/await-ifying this? As a rule, if all pieces of code are fast, it's better to not use async/await, because async/await incurs the overhead of allocating a state-carrying object and waiting for the scheduler to schedule you.

jbayardo: We need to dispose of the stream in the case we need to create a temporary file. The alternative is to use ContinueWith and return a Task as we were doing before. The await syntax seems much more explicit about intent.

PR #432, opened 2019-06-07
Always compile our bits with 'checked : true'

olkononenko:
I've run validation builds in BVT queues - there was no overflow. 

[AB#1534271](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1534271)

### DIFF
--- a/Public/Src/Engine/Dll/HistoricTableSizes.cs
+++ b/Public/Src/Engine/Dll/HistoricTableSizes.cs
@@ -102,7 +102,7 @@ namespace BuildXL.Engine
         /// <summary>
         /// Sum of all table sizes in bytes.
         /// </summary>
-        public int TotalSizeInBytes() => AllTableStats.Sum(s => s.SizeInBytes);
+        public long TotalSizeInBytes() => AllTableStats.Sum(s => s.SizeInBytes);
 
         /// <nodoc/>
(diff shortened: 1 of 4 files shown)


Public/Src/Engine/Dll/HistoricTableSizes.cs
```diff
@@ -199,7 +199,7 @@ public override bool Equals(object obj)
         /// <inheritdoc/>
         public override int GetHashCode()
         {
-            return HashCodeHelper.Combine(Count, SizeInBytes);
+            return HashCodeHelper.Combine(Count, (int)(SizeInBytes & 0xFFFFFFFF), (int) (SizeInBytes >> 32));
```
olkononenko: this is dirty hacking - i'd appreciate any input here

dannyvv: Another alternavie would be using bitconverter to create a bytearray to be add to hascode helper. But I like this pattern better way more effecient.

Allthough perhaps this belongs inside the hasCodehelper itself.


---
In reply to: [291774166](https://github.com/microsoft/BuildXL/pull/432#discussion_r291774166) [](ancestors = 291774166)

dannyvv:
Note: The title has a Typo, I corrected it in the PR, but GitHub doesn't pick that up when merging so you'll have to manually do it again on merge....

PR #434, opened 2019-06-08
Reset NagleQueue timer when the event processing is resumed.

SergeyTeplyakov:
[AB#1552616](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1552616)

This issue can cause events to get stuck in nagle queue for a long time causing CTMIS.

### DIFF
--- a/Public/Src/Cache/ContentStore/Test/Utils/NagleQueueTests.cs
+++ b/Public/Src/Cache/ContentStore/Test/Utils/NagleQueueTests.cs
@@ -13,6 +13,31 @@ namespace ContentStoreTest.Utils
     public class NagleQueueTests
     {
         [Fact]
+        public void ResumeShouldTriggerBatchOnTime()
+        {
+            bool processBatchIsCalled = false; ;
+            var queue = NagleQueue<int>.Create(
+                processBatch: data =>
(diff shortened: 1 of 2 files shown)


dcwuser:
So cool to see a mystery resolved by a simple code fix.

Public/Src/Cache/ContentStore/Test/Utils/NagleQueueTests.cs
```diff
@@ -12,6 +12,57 @@ namespace ContentStoreTest.Utils
...
+                                  processBatchIsCalled++;
+                                  return Task.FromResult(42);
+                              },
+                maxDegreeOfParallelism: 1,
+                interval: TimeSpan.FromMilliseconds(10),
+                batchSize: 10);
+
+            queue.Enqueue(42);
+            
+            Thread.Sleep(100);
```
kewilli: These `Sleep` calls will lead to flakiness, right? These tests tend to run at 100% CPU, so there's a chance the queue might not complete its operation before the assertion is called. Does it make sense to decrease the queue's interval even further to 1ms?

PR #437, opened 2019-06-10
Extract ResourcePool from GrpcCopyClientCache

kewilli:


Public/Src/Cache/ContentStore/App/CopyFile.cs
```diff
@@ -64,12 +65,6 @@ internal sealed partial class Application
                     {
                         _logger.Debug($"Copy of {hashString} to {finalPath} was successful");
                     }
-
-                    var shutdownResult = rpcClient.ShutdownAsync(context).Result;
```
SergeyTeplyakov: Just curious, does it mean that the code without this change, the old code was broken because we were shutting down pulled clients?

SergeyTeplyakov: If this is the case, then I would suggest adding an assertion in clientCache.CreateAsync to make sure the instance is opened and not shut down.

---
In reply to: [292213563](https://github.com/microsoft/BuildXL/pull/437#discussion_r292213563) [](ancestors = 292213563)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -339,9 +281,9 @@ private async Task<T> RunClientActionAndThrowIfFailedAsync<T>(Context context, F
         /// <inheritdoc />
         public void Dispose()
         {
-            lock (this)
+            if (!ShutdownStarted || !ShutdownCompleted)
             {
-                _uses--;
+                throw new CacheException($"{nameof(GrpcCopyClient)} must be shutdown before disposing.");
```
SergeyTeplyakov: I'm not sure, but I think it should be fine to dispose the insance if ShutdownAsync fails.

But I think this code will fail.

Should it be:

`if (ShutdownStarted && !ShutdownCompleted) `

PR #438, opened 2019-06-10
[MacSandbox] track process name upon execve

almiliMSFT:


Public/Src/Sandbox/MacOs/Sandbox/Src/Listeners.cpp
```diff
@@ -197,8 +201,10 @@ int Listeners::mpo_cred_label_update_execve(kauth_cred_t old_cred,
                                             size_t macpolicyattrlen,
                                             int *disjointp)
 {
-    // Track vfork(), make sure not to re-add the tracked process as this handler also gets called on execve()
+    // this 'mpo_cred_label_update_execve' handler can be called both upon 'vfork' and upon 'exec',
+    // which is why which we have to handle both 'fork' and 'exec' here. 
     mpo_cred_label_associate_fork(old_cred, p);
+    handle_exec(vp);
```
Kr1sso: We had this in the first iteration and decided on removing it, do you remember the reason?

almiliMSFT: When we first introduced this handler, it was still simply delegating its work to `mpo_cred_label_associate_fork`, but, `mpo_cred_label_associate_fork` was doing `proc_name(proc->p_pid)` to obtain the name.  Calling `proc_name` on fork is not very useful, because at that time it returns the name of the parent process, and not the name of the vnode that is about to be executed.

PR #439, opened 2019-06-10
Translate host UNC drive path to allow journal tests in VM

narasamdya:
This is one of three changes required for running journal tests in VM. In this change, we translate host UNC drive path that is often access by tools like dotnet.exe. 

This change also include introduction of some special environment variables that the unit tests will use to work properly when run in VMs.

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -2048,7 +2048,7 @@ namespace BuildXL.Processes
         {
             if (ShouldSandboxedProcessExecuteInVm)
             {
-                environmentVariables = environmentVariables.Override(new[] { new KeyValuePair<string, string>(VmSpecialEnvironmentVariable.IsInVm, "1")});
+                environmentVariables = environmentVariables.Override(new[] { new KeyValuePair<string, string>(VmSpecialEnvironmentVariables.IsInVm, "1")});
             }
         }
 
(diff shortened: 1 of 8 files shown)


Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2020,6 +2044,14 @@ private bool PrepareWorkingDirectory()
             return true;
         }
 
+        private void PrepareEnvironmentVariables(ref IBuildParameters environmentVariables)
+        {
+            if (ShouldSandboxedProcessExecuteInVm)
+            {
+                environmentVariables = environmentVariables.Override(new[] { new KeyValuePair<string, string>(VmSpecialEnvironmentVariable.IsInVm, "1")});
```
norahuang: Would "VmSpecialEnvironmentVariables" be more appropriate?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -861,6 +863,11 @@ private bool ShouldSandboxedProcessExecuteExternal
 
             info.RedirectedTempFolders = m_tempFolderRedirectionForVm.Select(kvp => (kvp.Key.ToString(m_pathTable), kvp.Value.ToString(m_pathTable))).ToArray();
 
+            if (m_sandboxConfig.AdminRequiredProcessExecutionMode.ExecuteExternalVm())
```
norahuang: Why don't use ShouldSandboxedProcessExecuteInVm here? Does "ShouldSandboxedProcessExecuteExternal" matter when translate Unc drive?

narasamdya: This method is called when `ShouldSandboxedProcessExecuteExternal` is true. Then, we only need to translate host UNC drive if in fact we're about to run the pip in VM. If not, then the pip will only run on the host but not as a child process of BuildXL. Because the pip still runs on the host, there's no need to translate host UNC drive.

---
In reply to: [292239677](https://github.com/microsoft/BuildXL/pull/439#discussion_r292239677) [](ancestors = 292239677)

PR #445, opened 2019-06-11
Increase weight for frequently timeout tests

norahuang:


Public/Src/FrontEnd/UnitTests/Script.Ast/Test.BuildXL.FrontEnd.Script.dsc
```diff
@@ -19,6 +19,7 @@ namespace Script {
         sources: globR(d`.`, "*.cs"),
         runTestArgs: {
             parallelGroups: categoriesToRunInParallel,
+            weight: 2,
```
narasamdya: >weight: 2, [](start = 12, length = 10)

Please add a comment in the spec why you increase the weight.

norahuang:
I messed up this PR by push changes of another branch to here. Since the change of this one is simple, I will abandon it and create a new one.

PR #446, opened 2019-06-11
Order pip fingerprints when adding to json

erickulcyk:
Adding order independent items to the fingerprinter should actually sort them before adding them to the json text.  To do this, a comparer is given for FileArtifacts, FileArtifactsWithAttributes, and DirectoryArtifacts.  The actual pip fingerprinter ignores this comparer as it XORs the values together, but the json fingerprinters sorts the items before adding them.


Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -250,16 +254,16 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process
             AddFileOutput(fingerprinter, "StandardError", process.StandardError);
             AddFileOutput(fingerprinter, "StandardOutput", process.StandardOutput);
 
-            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f));
-            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d));
+            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f), m_expandedPathFileArtifactComparer);
+            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d), m_expandedPathDirectoryArtifactComparer);
```
olkononenko: >m_expandedPathDirectoryArtifactComparer [](start = 207, length = 39)

for directories, we only add their paths to the FP, so I don't think we need a dedicated class, i.e., I think something like this
m_directoryArtifactComparer = Comparer<DirectoryArtifact>.Create((d1, d2) => m_pathTable.ExpandedPathComparer.Compare(d1.Path, d2.Path));
should be enough

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -250,16 +254,16 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process
...
+            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f), m_expandedPathFileArtifactComparer);
+            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d), m_expandedPathDirectoryArtifactComparer);
 
-            fingerprinter.AddOrderIndependentCollection<FileArtifactWithAttributes, ReadOnlyArray<FileArtifactWithAttributes>>("Outputs", process.FileOutputs, (fp, f) => AddFileOutput(fp, f));
-            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryOutputs", process.DirectoryOutputs, (h, p) => h.Add(p.Path));
+            fingerprinter.AddOrderIndependentCollection<FileArtifactWithAttributes, ReadOnlyArray<FileArtifactWithAttributes>>("Outputs", process.FileOutputs, (fp, f) => AddFileOutput(fp, f), m_expandedPathFileArtifactWithAttributesComparer);
```
olkononenko: >m_expandedPathFileArtifactWithAttributesComparer [](start = 192, length = 48)

there should be no FileArtifacts that have the same path but different PathExistence. I'd go with Compare.Create(...) that would only look at paths.

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -100,6 +102,8 @@ public string FingerprintSalt
             m_pipDataLookup = pipDataLookup ?? new PipDataLookup(file => PipData.Invalid);
             PathExpander = pathExpander ?? PathExpander.Default;
             m_expandedPathFileArtifactComparer = new ExpandedPathFileArtifactComparer(m_pathTable.ExpandedPathComparer, pathOnly: false);
+            m_expandedPathDirectoryArtifactComparer = new ExpandedPathDirectoryArtifactComparer(m_pathTable.ExpandedPathComparer, pathOnly: false);
```
olkononenko: >false [](start = 140, length = 5)

'true' - see my comment in AddWeakFP method

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -100,6 +102,8 @@ public string FingerprintSalt
             m_pipDataLookup = pipDataLookup ?? new PipDataLookup(file => PipData.Invalid);
             PathExpander = pathExpander ?? PathExpander.Default;
             m_expandedPathFileArtifactComparer = new ExpandedPathFileArtifactComparer(m_pathTable.ExpandedPathComparer, pathOnly: false);
+            m_expandedPathDirectoryArtifactComparer = new ExpandedPathDirectoryArtifactComparer(m_pathTable.ExpandedPathComparer, pathOnly: false);
+            m_expandedPathFileArtifactWithAttributesComparer = new ExpandedPathFileArtifactWithAttributesComparer(m_pathTable.ExpandedPathComparer, pathOnly: false);
```
olkononenko: >false [](start = 158, length = 5)

'true' - ?

PR #447, opened 2019-06-11
Remove unused copyfile logs

kewilli:
Removing:

* Attempt #0: Copying VSO0:5190E226C8606C2CE7CD1BB9BF03DED3C740155D5FD60999BFF8A4A38483871F00 with 1 locations
  * Copies are already groups by attempt, and location count is given by Redis' GetBulk call

* DistributedContentCopier.RemoteCopyFile start. (VSO0:5190E226C8606C2CE7CD1BB9BF03DED3C740155D5FD60999BFF8A4A38483871F00 size=230511 from=[...
  * While removing this stops us from tracking incomplete copies, this hasn't been an issue. I am confident in our copy error logging to catch all related issues.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
@@ -133,6 +133,9 @@ namespace BuildXL.Cache.ContentStore.Distributed.Stores
                 var badContentLocations = new HashSet<MachineLocation>();
                 var missingContentLocations = new HashSet<MachineLocation>();
                 int attemptCount = 0;
+
+                Tracer.Debug(operationContext, $"Copying {hashInfo.ContentHash} with {hashInfo.Locations.Count} locations");
+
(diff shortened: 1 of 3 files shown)


Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -135,7 +135,6 @@ protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
                 int attemptCount = 0;
                 while (attemptCount < _retryIntervals.Count && (putResult == null || !putResult))
                 {
-                    Tracer.Debug(operationContext, $"Attempt #{attemptCount}: Copying {hashInfo.ContentHash} with {hashInfo.Locations.Count} locations");
```
dcwuser: Please leave this, but outside while loop. Yes, you can infer this number from Redis logs, but this component should still log enough information that we can see what it is doing even if sources of input data change or change their logging.

PR #448, opened 2019-06-11
Mark BuildXL executables 64-bit only

dannyvv:
Mark BuildXL executables 64-bit only rather than AnyCpu since all our native code which all our exe's depend on are 64-bit only..

### DIFF
--- a/Public/Sdk/SelfHost/BuildXL/BuildXLSdk.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/BuildXLSdk.dsc
@@ -47,8 +47,6 @@ const brandingDefines = [
 
 @@public
 export interface Arguments extends Managed.Arguments {
-    cscArgs?: Csc.Arguments;
-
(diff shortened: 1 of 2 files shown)


Public/Sdk/SelfHost/BuildXL/BuildXLSdk.dsc
```diff
@@ -286,6 +284,7 @@ export function executable(args: Arguments): Managed.Assembly {
         ],
         tools: {
             csc: {
+                platform: <"x64">"x64",
```
Kr1sso: Is this some copy paste mistake?

narasamdya: A potential bug in type checking forces Danny to do the casting.

narasamdya:
I am going to push this change since my work on VM execution depends on this change.

PR #452, opened 2019-06-12
Changes in tests needed to run them in VM

narasamdya:
These changes are needed to run tests in VM. 

One notable change is the way we check the ability to scan the journal. Normally, we check the root path of the executing assembly to know if journal scanning can be performed. However, in the VM scenario the executing assembly is on the host, and one cannot check journal capability of the host from VM. 

Another less safe change is skipping file tracking if the root path is in a journal incapable volume. Our unit tests (particularly the integration tests) track the test executable (the one that execute Operations.cs). But that executable is located on the host, and tracking the host USN from VM does not work. Unfortunately, once tracking failed, the file change tracker switches into disabled state, and this failed all incremental scheduling tests.

[AB#1552081](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1552081)


Public/Sdk/Public/Managed/Testing/XUnit/xunit.dsc
```diff
@@ -65,6 +65,14 @@ export function runConsoleTest(args: TestRunArguments): Result {
 
     const result = Transformer.execute(execArguments);
 
+    if (Environment.hasVariable("BUILDXL_IS_IN_CLOUDBUILD")) {
+        const qualifierRelative = r`${qualifier.configuration}/${qualifier.targetFramework}/${qualifier.targetRuntime}`;
+        const parallelRelative = args.parallelBucketIndex !== undefined ? `${args.parallelBucketIndex}` : `0`;
+        const xunitLogDir = d`${Context.getMount("LogsDirectory").path}/XUnit/${Context.getLastActiveUseModuleName()}/${Context.getLastActiveUseName()}/${qualifierRelative}/${parallelRelative}`;
+
+        const copies = result.getOutputFiles().map(f => Transformer.copyFile(f, p`${xunitLogDir}/${f.name}`));
+    }
```
narasamdya: I need this to understand why our tests failed.

PR #453, opened 2019-06-12
Improve CaSaaS use of Secret Store

jbayardo:
Allows us to replace the secrets being used by both EventHub and Azure Blob without needing to restart CASaaS. 

For Azure Blob, this is officially supported, and hence we make use of the feature.

EventHub supports secrets but they handle it inside their own code, and don't seem to have a way to plug them in externally.  Hence, we stop traffic, regenerate the object, and restart traffic again. The changing of secrets is a very infrequent operation, so this shouldn't affect operation significantly.


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStoreConfiguration.cs
```diff
@@ -203,32 +205,62 @@ public DistributedCentralStoreConfiguration(AbsolutePath cacheRoot)
         public int MaxSimultaneousCopies { get; set; } = 10;
     }
 
+    /// <summary>
+    /// Provides Azure Storage authentication options for <see cref="BlobCentralStorage"/>
+    /// </summary>
+    public class AzureStorageCredentials
+    {
+        /// <nodoc />
+        public string ConnectionString { get; private set; } = null;
```
SergeyTeplyakov: Lets make it immutable without private setter, and you don't have to set it to null.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStoreConfiguration.cs
```diff
@@ -203,32 +205,62 @@ public DistributedCentralStoreConfiguration(AbsolutePath cacheRoot)
...
+
+        public AzureStorageCredentials(string connectionString)
+        {
+            Contract.Requires(connectionString != null);
+            ConnectionString = connectionString;
+        }
+
+        public AzureStorageCredentials(StorageCredentials storageCredentials)
+        {
+            Contract.Requires(storageCredentials != null);
```
SergeyTeplyakov: Add this here: Contract.Requires(credentials.StorageCredentials.IsSAS);

jbayardo: Are we sure that's what we want? not requiring this would allow us to also change passwords while still running, for example.

jbayardo:
Please ignore this PR; I was testing something and so I closed it. The PR for this work item is @ https://github.com/microsoft/BuildXL/pull/459

PR #456, opened 2019-06-12
Log correct replica count and effectiveLastAccessTime in eviction

JuanCarlosGI:
Fix bug where replica count is always 1

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
@@ -1089,7 +1089,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                             Counters[ContentLocationStoreCounters.EffectiveLastAccessTimeLookupMiss].Increment();
                         }
 
-                        effectiveLastAccessTimes.Add(new ContentHashWithLastAccessTimeAndReplicaCount(contentHash.Hash, contentHash.LastAccessTime,  replicaCount, effectiveLastAccessTime: effectiveLastAccessTime ?? contentHash.LastAccessTime));
+                        effectiveLastAccessTimes.Add(new ContentHashWithLastAccessTimeAndReplicaCount(contentHash.Hash, contentHash.LastAccessTime, replicaCount, effectiveLastAccessTime: effectiveLastAccessTime ?? contentHash.LastAccessTime));
                     }
 
                     return Result.Success<IReadOnlyList<ContentHashWithLastAccessTimeAndReplicaCount>>(effectiveLastAccessTimes);
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
@@ -1089,7 +1089,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                             Counters[ContentLocationStoreCounters.EffectiveLastAccessTimeLookupMiss].Increment();
                         }
 
-                        effectiveLastAccessTimes.Add(new ContentHashWithLastAccessTimeAndReplicaCount(contentHash.Hash, effectiveLastAccessTime ?? contentHash.LastAccessTime, replicaCount, originalLastAccessTime: contentHash.LastAccessTime));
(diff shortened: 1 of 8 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1055,11 +1055,11 @@ public async Task<BoolResult> TrimBulkAsync(OperationContext context, IReadOnlyL
                 () =>
                 {
                     var effectiveLastAccessTimes = new List<ContentHashWithLastAccessTimeAndReplicaCount>();
-                    int replicaCount = 1;
 
                     foreach (var contentHash in contentHashes)
                     {
                         DateTime? effectiveLastAccessTime = null;
+                        int replicaCount = 1;
```
SergeyTeplyakov: nit: effectiveReplicaCount? to follow the same naming scheme to effectiveLastAccessTime?

Public/Src/Cache/ContentStore/Distributed/NuCache/TransitioningContentLocationStore.cs
```diff
@@ -254,7 +254,8 @@ public Task<BoolResult> UnregisterAsync(Context context, IReadOnlyList<ContentHa
             // Priority queue orders by least first. So we compare by last access time to get the least last access time (i.e. oldest) first.
             var priorityQueue = new PriorityQueue<ContentHashWithLastAccessTimeAndReplicaCount>(
                 pageSize * 2,
-                Comparer<ContentHashWithLastAccessTimeAndReplicaCount>.Create((c1, c2) => c1.LastAccessTime.CompareTo(c2.LastAccessTime)));
+                // Assume that EffectiveLastAccessTime will always have a value.
+                Comparer<ContentHashWithLastAccessTimeAndReplicaCount>.Create((c1, c2) => c1.EffectiveLastAccessTime.Value.CompareTo(c2.EffectiveLastAccessTime.Value)));
```
SergeyTeplyakov: I'm not 100% sure that the assumption is true for all the cases.

Does it make sense to be null-safe here?

JuanCarlosGI: In the function called below (GetEffectiveLastAccessTimes) we make sure that it is never null. I would rather something break if the assumption is false because it would mean something is wrong with the code

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
SergeyTeplyakov: Just curious, does it make sense to use nullable int and actually use null if we don't know for sure the number of replicas?

Public/Src/Cache/ContentStore/Distributed/NuCache/TransitioningContentLocationStore.cs
SergeyTeplyakov: Understood. Thanks.

---
In reply to: [293474376](https://github.com/microsoft/BuildXL/pull/456#discussion_r293474376) [](ancestors = 293474376)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
JuanCarlosGI: Decided not to make nullable since it would add no value and we would have to add null checks in more places.

dcwuser:
>         public TimeSpan Age { get; }

I think it would be better to compute this when called rather than compute and store when instantiated.

---
Refers to: Public/Src/Cache/ContentStore/Library/Tracing/EvictResult.cs:92 in 3bfe711. [](commit_id = 3bfe7110e4df709e5a4b7d7ebd697afde5b24fdf, deletion_comment = False)

Public/Src/Cache/ContentStore/Library/Tracing/EvictResult.cs
```diff
@@ -80,7 +95,7 @@ public EvictResult(Exception exception, string message = null)
         public override string ToString()
         {
             return Succeeded
-                ? $"Success Size={EvictedSize} Files={EvictedFiles} Pinned={PinnedSize} LastAccessTime={LastAccessTime} ReplicaCount={ReplicaCount} Age={Age}"
+                ? $"Success Size={EvictedSize} Files={EvictedFiles} Pinned={PinnedSize} EffectiveLastAccessTime={EffectiveLastAccessTime} OriginalLastAccessTime={LastAccessTime} ReplicaCount={ReplicaCount} Age={Age}"
```
dcwuser: Can we also print EffectiveAge?

PR #457, opened 2019-06-12
Changes needed to run tests in VM

narasamdya:
These changes are needed to run tests in VM.

One notable change is the way we check the ability to scan the journal. Normally, we check the root path of the executing assembly to know if journal scanning can be performed. However, in the VM scenario the executing assembly is on the host, and one cannot check journal capability of the host from VM.

Another less safe change is skipping file tracking if the root path is in a journal incapable volume. Our unit tests (particularly the integration tests) track the test executable (the one that execute Operations.cs). But that executable is located on the host, and tracking the host USN from VM does not work. Unfortunately, once tracking failed, the file change tracker switches into disabled state, and this failed all incremental scheduling tests.

[AB#1552081](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1552081)


Public/Sdk/Public/Managed/Testing/XUnit/xunit.dsc
```diff
@@ -65,6 +65,14 @@ export function runConsoleTest(args: TestRunArguments): Result {
 
     const result = Transformer.execute(execArguments);
 
+    if (Environment.hasVariable("BUILDXL_IS_IN_CLOUDBUILD")) {
+        const qualifierRelative = r`${qualifier.configuration}/${qualifier.targetFramework}/${qualifier.targetRuntime}`;
+        const parallelRelative = args.parallelBucketIndex !== undefined ? `${args.parallelBucketIndex}` : `0`;
+        const xunitLogDir = d`${Context.getMount("LogsDirectory").path}/XUnit/${Context.getLastActiveUseModuleName()}/${Context.getLastActiveUseName()}/${qualifierRelative}/${parallelRelative}`;
+
+        const copies = result.getOutputFiles().map(f => Transformer.copyFile(f, p`${xunitLogDir}/${f.name}`));
```
norahuang: copies are not used later on. Do you need to save it?

narasamdya: I don't need it; potentially a left over from my experiments.

---
In reply to: [293204603](https://github.com/microsoft/BuildXL/pull/457#discussion_r293204603) [](ancestors = 293204603)

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -1471,19 +1471,21 @@ private JournalState GetJournalStateWithVolumeMap(VolumeMap volumeMap, LoggingCo
             Contract.Requires(loggingContext != null);
 
             // Under some configurations, we access the change journal (i.e., actually scan the journal).
-            var journal = Configuration.Engine.ScanChangeJournal
-                ? JournalAccessorGetter.TryGetJournalAccessor(
-                    loggingContext,
+            if (Configuration.Engine.ScanChangeJournal)
```
norahuang: if (OperatingSystemHelper.IsUnixOS || !Configuration.Engine.ScanChangeJournal) and then "if (Configuration.Engine.ScanChangeJournal)" and the last "return JournalState.DisabledJournal;" can be removed. It is easier to read in this logic.

narasamdya: I have a different opinion. The Unix exclusion can be thought of as a shortcut for Unix. Thus, the rest of the code is only applied to Win.

---
In reply to: [293577631](https://github.com/microsoft/BuildXL/pull/457#discussion_r293577631) [](ancestors = 293577631)

Public/Src/Utilities/Storage/ChangeTracking/FileChangeTracker.cs
```diff
@@ -730,10 +730,53 @@ public Possible<PathExistence> TryProbeAndTrackPath(string path, bool? isReadOnl
...
+            if (!m_volumeMap.SkipTrackingJournalIncapableVolume)
+            {
+                // Immediately disable tracking.
+                DoDisable(path, trackingFailure);
+            }
+            else
             {
-                Logger.Log.DisableChangeTracker(m_loggingContext, path, trackingFailure.DescribeIncludingInnerFailures());
+                // Check if the path is in the volume that can be skipped.
+                // This is expensive, and should only be used during testing.
```
norahuang: Is that mean this block of code will be removed in the future?

narasamdya: No it isn't. But the SkipTrackingJournalIncapableVolume should only be set true for testing.

---
In reply to: [293586421](https://github.com/microsoft/BuildXL/pull/457#discussion_r293586421) [](ancestors = 293586421)

Public/Sdk/Public/Managed/Testing/XUnit/xunit.dsc
```diff
@@ -65,6 +65,14 @@ export function runConsoleTest(args: TestRunArguments): Result {
 
     const result = Transformer.execute(execArguments);
 
+    if (Environment.hasVariable("BUILDXL_IS_IN_CLOUDBUILD")) {
```
almiliMSFT:
```suggestion
    if (Environment.hasVariable(BuildXL.Utilities.BuildParameters.IsInCloudBuildVariableName)) {
```

almiliMSFT: The `"BUILDXL_IS_IN_CLOUDBUILD"` string is also used in `qtestFramework.dsc` and `Tool.QTestRunner.dsc`, so maybe it would be worth moving it to `BuildXLSdk.dsc` and adding something like
```typescript
@@public export const isInCloudBuild = Environment.hasVariable("BUILDXL_IS_IN_CLOUDBUILD");
```

mpysson: I know it's convenient but I'm not a big fan of our "is in cloudbuild" type of flags. I'd rather us tie this sort of thing to the capabilities needed, not the system and our current understanding of its capabilities are. Maybe in 10 years we'll be be more advanced than the cloud and be running in spacebuild.

At the very least please comment why the test needs to behave differently and what assumptions it's making about the environment here

narasamdya: I was thinking of the same thing. However, we cannot pull BuildXLSdk.dsc here because it means that the SDK that we publish to customers will depend on BuildXLSdk.dsc.

Either
- have a baked Environment.isInCloudBuild(), or
- add in process pip "logOutput: " ... similar to "output", but it goes to log folder.
For sure, both should not be in this PR.

---
In reply to: [293865962](https://github.com/microsoft/BuildXL/pull/457#discussion_r293865962) [](ancestors = 293865962)

Public/Src/Utilities/UnitTests/TestUtilities.XUnit/JournalUtils.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// Gets an instance of <see cref="IChangeJournalAccessor"/> for tests.
+        /// </summary>
+        /// <param name="volumeMap">Volume map.</param>
+        /// <returns>An instance of <see cref="IChangeJournalAccessor"/>.</returns>
+        public static Possible<IChangeJournalAccessor> TryGetJournalAccessorForTest(VolumeMap volumeMap)
+        {
+            string path = null;
+            bool usingTempFile = false;
+
+            if (!HasRelocatedTempInVm)
```
mpysson: >HasRelocatedTempInVm [](start = 17, length = 20)

Does this need to be conditional? Why not just always get a JournalAccessor based on the temp directory? On the test VM it will be on a machine local drive and it will be under the object directory in a non-VM run

Public/Sdk/Public/Managed/Testing/XUnit/xunit.dsc
mpysson: I think it would be a nice behavior to copy the xunit logs to the build session's log directory regardless of whether it's running in CB or not. Do you see problems with that?

Then there's how to implement it. Since we now have a log mount it could be exposed that way and have build logic authors do it in DScript. Or we could have an annotation on the pips that some files are outputs and thus should be handled special. What do you think?

Public/Src/Utilities/UnitTests/TestUtilities.XUnit/JournalUtils.cs
narasamdya: I was trying to preserve the old behavior before switching over.

---
In reply to: [294012874](https://github.com/microsoft/BuildXL/pull/457#discussion_r294012874) [](ancestors = 294012874)

Public/Sdk/Public/Managed/Testing/XUnit/xunit.dsc
narasamdya: I think it's aligned to what I replied to Aleks, and what I mentioned in the user story. I want to have `Context.getNewOutputLogDirectory(...)`, which is similar to `Context.getNewOutputDirectory(...)`, but rooted at the log directory, instead of at the object directory.

---
In reply to: [294013891](https://github.com/microsoft/BuildXL/pull/457#discussion_r294013891) [](ancestors = 294013891)

PR #459, opened 2019-06-12
Improve CaSaaS use of Secret Store

jbayardo:
Allows us to replace the secrets being used by both EventHub and Azure Blob without needing to restart CASaaS.

For Azure Blob, this is officially supported, and hence we make use of the feature.

EventHub supports secrets but they handle it inside their own code, and don't seem to have a way to plug them in externally. Hence, we stop traffic, regenerate the object, and restart traffic again. The changing of secrets is a very infrequent operation, so this shouldn't affect operation significantly.


Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
```diff
@@ -42,22 +42,29 @@ public class BlobCentralStorage : CentralStorage
         /// <nodoc />
         public BlobCentralStorage(BlobCentralStoreConfiguration configuration)
         {
-            Contract.Requires(configuration.ConnectionStrings.Count != 0);
+            Contract.Requires(configuration.Credentials.Count != 0);
```
dannyvv: >Credentials [](start = 44, length = 11)

I realize you are not chaing things here, but I'm checking that we somwhere have a friendly user message that there must be at least one credential int he configuration, since this is a hard crash which will likely not result in something actionalbe for end user...

Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
```diff
@@ -42,22 +42,29 @@ public class BlobCentralStorage : CentralStorage
...
-                (connectionString, index) =>
+            _containers = _configuration.Credentials.Select(
+                (credentials, index) =>
                 {
-                    var storage = CloudStorageAccount.Parse(connectionString);
-                    var blobClient = storage.CreateCloudBlobClient();
-                    return (blobClient.GetContainerReference(configuration.ContainerName), shardId: index);
-                }).ToArray();
+                    CloudStorageAccount storageAccount = null;
+                    if (credentials.ConnectionString != null)
```
dannyvv: >null [](start = 56, length = 4)

`string.isNullOrEmpty`?

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -112,30 +112,51 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
...
 
+        private readonly object _eventHubLock = new object();
+
+        public void UpdateCredentials(OperationContext context, EventHubsConnectionStringBuilder connectionStringBuilder)
+        {
+            lock (_eventHubLock)
+            {
+                DoSuspendProcessing(context).ThrowIfFailure();
+                RegenerateEventHubClientFromCredentials(connectionStringBuilder);
+                DoStartProcessing(context, GetLastProcessedSequencePoint()).ThrowIfFailure();
```
dannyvv: >DoStartProcessing [](start = 16, length = 17)

this method seems to acquire the log itself already...

dannyvv: line 431

---
In reply to: [293123691](https://github.com/microsoft/BuildXL/pull/459#discussion_r293123691) [](ancestors = 293123691)

jbayardo: It is done on purpose, since each of these methods acquire the lock separately, and it is possible that this may interleave with a different call to suspend, taking the reentrant lock guarantees that no other call will interleave

---
In reply to: [293123691](https://github.com/microsoft/BuildXL/pull/459#discussion_r293123691) [](ancestors = 293123691)

Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
jbayardo: Added requires where they should be and added a message to them. Is that good enough, or should I use a different mechanism?

---
In reply to: [293122690](https://github.com/microsoft/BuildXL/pull/459#discussion_r293122690) [](ancestors = 293122690)

SergeyTeplyakov:
We need to test this change in CB to make sure that nuget change won't break cloudbuild.

Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
```diff
@@ -42,22 +42,29 @@ public class BlobCentralStorage : CentralStorage
         /// <nodoc />
         public BlobCentralStorage(BlobCentralStoreConfiguration configuration)
         {
-            Contract.Requires(configuration.ConnectionStrings.Count != 0);
-
             _configuration = configuration;
 
-            _containers = _configuration.ConnectionStrings.Select(
-                (connectionString, index) =>
+            _containers = _configuration.Credentials.Select(
```
SergeyTeplyakov: I would create a factory method (exgternal or in Credentials) that will create CloubBlobContainer.

In this case the logic in this class will be essentially the same as it was before.

And I think it make sense to move this logic away from this type because its not essential to it. (and it'll make Credentials type easier to use).

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStoreConfiguration.cs
```diff
@@ -203,32 +205,87 @@ public DistributedCentralStoreConfiguration(AbsolutePath cacheRoot)
         public int MaxSimultaneousCopies { get; set; } = 10;
     }
 
+    /// <summary>
+    /// Provides Azure Storage authentication options for <see cref="BlobCentralStorage"/>
+    /// </summary>
+    public class AzureBlobStorageCredentials
+    {
+        /// <nodoc />
+        public readonly string ConnectionString;
```
SergeyTeplyakov: I meant to use get-only property, not readonly fields.

Readonly fields are fine, but we usually prefer using get-only properties in case we'll need to change the implementation without changing the clients of the code.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStoreConfiguration.cs
```diff
@@ -203,32 +205,87 @@ public DistributedCentralStoreConfiguration(AbsolutePath cacheRoot)
         public int MaxSimultaneousCopies { get; set; } = 10;
     }
 
+    /// <summary>
+    /// Provides Azure Storage authentication options for <see cref="BlobCentralStorage"/>
+    /// </summary>
+    public class AzureBlobStorageCredentials
```
SergeyTeplyakov: As I suggested, it is actually possible that this class would have just one public method: CreateCloudBlobContainer and two constructors.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStoreConfiguration.cs
```diff
@@ -203,32 +205,87 @@ public DistributedCentralStoreConfiguration(AbsolutePath cacheRoot)
...
+            StorageCredentials = storageCredentials;
+            AccountName = accountName;
+            EndpointSuffix = endpointSuffix;
+        }
+    }
+
     /// <summary>
     /// Configuration of a central store backed by azure blob storage.
     /// </summary>
     public class BlobCentralStoreConfiguration : CentralStoreConfiguration
```
SergeyTeplyakov: I actually was wrong, and the entire code that deals with this class is in this codebase not in Cloudbuild.

You can change DistributedContentStoreFactory type from BuildXL.Cache.Host.SErvice to use your new api.

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -112,30 +112,52 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
...
-            _partitionSender = _eventHubClient.CreatePartitionSender(PartitionId);
+            RegenerateEventHubClientFromCredentials(new EventHubsConnectionStringBuilder(_configuration.EventHubConnectionString)
+            {
+                EntityPath = _configuration.EventHubName,
+            });
 
             return BoolResult.Success;
         }
 
+        private readonly object _eventHubLock = new object();
```
SergeyTeplyakov: Please move this to the rest of the fields.

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -112,30 +112,52 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
...
+
+        /// <nodoc />
+        public void UpdateCredentials(OperationContext context, EventHubsConnectionStringBuilder connectionStringBuilder)
+        {
+            lock (_eventHubLock)
+            {
+                DoSuspendProcessing(context).ThrowIfFailure();
```
SergeyTeplyakov: Please trace the fact that credentials are being updated.

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -112,30 +112,52 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
...
+                RegenerateEventHubClientFromCredentials(connectionStringBuilder);
+                DoStartProcessing(context, GetLastProcessedSequencePoint()).ThrowIfFailure();
+            }
+        }
+
+        private void RegenerateEventHubClientFromCredentials(EventHubsConnectionStringBuilder connectionStringBuilder)
```
SergeyTeplyakov: I would say the name should be: CreateEventHubClient.

We usually creating instances, not "regenerating" them, IMO.

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -112,30 +112,52 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
...
+                EntityPath = _configuration.EventHubName,
+            });
 
             return BoolResult.Success;
         }
 
+        private readonly object _eventHubLock = new object();
+
+        /// <nodoc />
+        public void UpdateCredentials(OperationContext context, EventHubsConnectionStringBuilder connectionStringBuilder)
```
SergeyTeplyakov: I think we need to put the ability to update credentials into the configuration because the code that knows about the update (recides in BuildXL.Cache.Host.Service project) won't be able to access EventHubConentLocationEventStore instance.

PR #462, opened 2019-06-13
Add a way to specify passthrough environment variables for the MSBuild resolver

smera:
- Expose a new primitive type 'unit' in DScript
- Enhance the MSBuild resolver environment to take either string values for env variable values, or 'unit' representing a passthrough

E.g.:

```
environment: Map.empty<string, (PassthroughEnvironmentVariable | string)>()
                      .add("Path", Environment.getStringValue("PATH"))
                      .add("USERPROFILE", Unit.unit())

```

### DIFF
--- a/Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
+++ b/Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
@@ -91,6 +91,9 @@ interface DownloadSettings {
     hash?: string,
 }
 
+/** We represent a passthrough environment variable with the value unit */ 
+type PassthroughEnvironmentVariable = Unit;
+
 /**
  * Resolver for MSBuild project-level build execution, utilizing the MsBuild static graph API to
  * find MSBuild files and convert them to a pip graph
@@ -160,8 +163,11 @@ interface MsBuildResolver extends ResolverBase, UntrackingSettings {
      * Note: if this field is not specified any change in an environment variable will potentially cause
      * cache misses for all pips. This is because there is no way to know which variables were actually used during the build.
      * Therefore, it is recommended to specify the environment explicitly.
+     * The value can be either a string or a PassthroughEnvironmentVariable, the latter representing that the associated variable will be exposed
+     * but its value won't be considered part of the build inputs for tracking purposes. This means that any change in the value of the 
+     * variable won't cause a rebuild.
      */
-    environment?: Map<string, string>;
+    environment?: Map<string, (PassthroughEnvironmentVariable | string)>;
 
     /**
      * Global properties to use for all projects.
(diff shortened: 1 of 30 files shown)


Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
```diff
@@ -160,8 +163,11 @@ interface MsBuildResolver extends ResolverBase, UntrackingSettings {
      * Note: if this field is not specified any change in an environment variable will potentially cause
      * cache misses for all pips. This is because there is no way to know which variables were actually used during the build.
      * Therefore, it is recommended to specify the environment explicitly.
+     * The value can be either a string or a PassthroughEnvironmentVariable, the latter representing that the associated variable will be exposed
+     * but its value won't be considered part of the build inputs for tracking purposes. This means that any change in the value of the 
+     * variable won't cause a rebuild.
      */
-    environment?: Map<string, string>;
+    environment?: Map<string, (PassthroughEnvironmentVariable | string)>;
```
dannyvv: >Map [](start = 18, length = 3)

Are you sure you wnat ot use a map, and not just a list of key-value pairs? That would be a lot nicer to author since we currently lack a nice literal construction for maps like C# has...

dannyvv: or altenatively a keyed object... (forgot syntax...)

---
In reply to: [293480516](https://github.com/microsoft/BuildXL/pull/462#discussion_r293480516) [](ancestors = 293480516)

Public/Src/FrontEnd/Script/Util/ConfigurationConverter.cs
```diff
@@ -809,6 +856,12 @@ private object ConvertAny(object valueToConvert, Type resultType, object overrid
                 return ConvertToRelativePath(overridingOrMergeIntoValue ?? valueToConvert, context);
             }
 
+            // target: UnionType
```
dannyvv: >UnionType [](start = 23, length = 9)

Union or Unit?

dannyvv: Ahh this is for the discriminating union you are adding... please ignore :)

---
In reply to: [293481811](https://github.com/microsoft/BuildXL/pull/462#discussion_r293481811) [](ancestors = 293481811)

Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
smera: Agree with the lack of nice literal construction. On the other hand, a map fits more naturally since it never makes sense to specify a variable twice. And we can always consider adding literal-specific stuff for maps. We already diverged from TS :).
So unless you feel strong about this one, I'd prefer to leave it as is.

smera:
azp /run  [Optional]Primary BuildXL PR Validation(Schedule Builds Only)

PR #463, opened 2019-06-13
DeterminismProbe+ disables Incremental Scheduling

olkononenko:
[AB#1553973](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1553973)

### DIFF
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -1255,7 +1255,7 @@ namespace BuildXL
 
                 if (ideConfiguration.IsEnabled)
                 {
-                    // Disable incrementalScheduling if the /vs is passed. Ide generator needs to catch all scheduled nodes and should not ignore the skipped ones due to the incremental scheduling
+                    // Disable incrementalScheduling if the /vs is passed. IDE generator needs to catch all scheduled nodes and should not ignore the skipped ones due to the incremental scheduling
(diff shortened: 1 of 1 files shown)


dannyvv:
>                     // Disable incrementalScheduling if the /vs is passed. Ide generator needs to catch all scheduled nodes and should not ignore the skipped ones due to the incremental scheduling

Perhaps these 3 should be folded in a single check?
Or is the logic more related to the indivudla flag than the incrementalscehduling flag...

---
Refers to: Public/Src/App/Bxl/Args.cs:1258 in 617f45d. [](commit_id = 617f45df5343a3d594ae56ceb61c38336452fa3d, deletion_comment = False)

olkononenko:
>                     // Disable incrementalScheduling if the /vs is passed. Ide generator needs to catch all scheduled nodes and should not ignore the skipped ones due to the incremental scheduling

this logic is for individual flags... technically there might be different sets of options that we need to tweak for each of these flags.

---
In reply to: [501795313](https://github.com/microsoft/BuildXL/pull/463#issuecomment-501795313) [](ancestors = 501795313)

---
Refers to: Public/Src/App/Bxl/Args.cs:1258 in 617f45d. [](commit_id = 617f45df5343a3d594ae56ceb61c38336452fa3d, deletion_comment = False)

PR #467, opened 2019-06-13
Add testAssembly in the description for QTest pips

norahuang:
[AB#1552021](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1552021)

### DIFF
--- a/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
@@ -188,7 +188,7 @@ export function runQTest(args: QTestArguments): Result {
     let result = Transformer.execute({
         tool: args.qTestTool ? args.qTestTool : qTestTool,
         tags: tags,
-        description: args.description,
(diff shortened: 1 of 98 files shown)


Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -188,7 +188,7 @@ export function runQTest(args: QTestArguments): Result {
     let result = Transformer.execute({
         tool: args.qTestTool ? args.qTestTool : qTestTool,
         tags: tags,
-        description: args.description,
+        description: `${args.description} || Running QTest for ${args.testAssembly}`,
```
dannyvv: Using format string is not needed....
The format quote should be after ||

norahuang: I actually want both of them to be in the description. But since args.description can be none I will put it in the back so it will look nicely.

norahuang:
Qtest needs more than this. So abandon this PR.

PR #472, opened 2019-06-14
Read next file chunk while waiting for GRPC response

kewilli:
GRPC's streaming protocol works as follows:
* Server sends request
* Client receives request, sends ACK
* Server waits to send next request until ACK is received

We can take advantage of this waiting time. Currently, we do nothing but wait. With this change, we read the next chunk of the file we are copying while waiting for the client's ACK.

In local tests with 0ms RTT, this showed a 20% copy time reduction in large files with small buffers. The returns were negligible for large buffers (where we spend more time copying than waiting) and small files (where we only send one request). In a lab with non-zero RTT, I would expect to see gains in these areas.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
+++ b/Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
@@ -293,8 +293,6 @@ namespace BuildXL.Cache.ContentStore.Service.Grpc
             Debug.Assert(!(responseStream is null));
 
             int chunkSize = 0;
-            async Task<int> ReadNextChunk() { chunkSize = await reader.ReadAsync(buffer, 0, buffer.Length, ct); return chunkSize; }
-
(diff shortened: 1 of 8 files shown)


Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -292,23 +292,29 @@ private async Task<(long Chunks, long Bytes)> StreamContentAsync(Stream reader,
             Debug.Assert(!(reader is null));
             Debug.Assert(!(responseStream is null));
 
+            int chunkSize = 0;
+            async Task<int> ReadNextChunk() { chunkSize = await reader.ReadAsync(buffer, 0, buffer.Length, ct); return chunkSize; }
```
SergeyTeplyakov: nit: could you please move it to the end of the method and switch to 'readNextChunk'? (this is the naming convention we use for local methods).

PR #474, opened 2019-06-14
Add ContentLocationDabase's in-memory cache configuration options

jbayardo:
Adds ContentLocationDatabase's in-memory cache configuration options to CloudBuild's layer so we can actually deploy


Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -87,6 +87,36 @@ public sealed class DistributedContentStoreFactory
                     dbConfig.LocalDatabaseGarbageCollectionInterval = TimeSpan.FromMinutes(_distributedSettings.ContentLocationDatabaseGcIntervalMinutes.Value);
                 }
 
+                if (_distributedSettings.ContentLocationDatabaseCacheEnabled.HasValue)
```
SergeyTeplyakov: Please use ApplyIfNotNull helper here.

PR #475, opened 2019-06-15
Add better support for opaque directories in DependencyAnalyzer

shivanshu3:
Add file members of output directories of pips to the output files list in DependencyAnalyzer.


dannyvv:
I have manually trigged the builds because github didn't kick them off automatically after I added you to the contributors.

dannyvv:
It seems github is not properly kicking off builds, trying again...

PR #476, opened 2019-06-15
Reduce DScript AST conversion parallelism

mpysson:
@smera tells me that this extra layer of parallelism was a vestige of the past that was pinned in non-DScript V2 due to lack of testing for the new default. The AST conversion is already parallelized at another layer in V2 and legacy modes.

Based on testing in CloudBuild (thanks @norahuang) it's apparent that this setting is what's causing the occasional degenerate graph construction times for some very large builds. In local smaller scale builds there is definitely less variance without the extra layer of parallelism. In aggregate the duration is about the same. I think it makes sense to make this change for sake of dramatically improving the degenerate cases of the large CB builds.

### DIFF
--- a/Public/Src/FrontEnd/Script/RuntimeModel/AstConversionConfiguration.cs
+++ b/Public/Src/FrontEnd/Script/RuntimeModel/AstConversionConfiguration.cs
@@ -23,8 +23,7 @@ namespace BuildXL.FrontEnd.Script.RuntimeModel.AstBridge
             PolicyRules = policyRules ?? CollectionUtilities.EmptyArray<string>();
             UnsafeOptions = UnsafeConversionConfiguration.GetConfigurationFromEnvironmentVariables();
 
-            // In DScript V2 AST converter itself does not convert nodes in parallel.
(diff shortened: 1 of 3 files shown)


Public/Src/FrontEnd/Script/RuntimeModel/AstConversionConfiguration.cs
```diff
@@ -23,8 +23,7 @@ public sealed class AstConversionConfiguration
             PolicyRules = policyRules ?? CollectionUtilities.EmptyArray<string>();
             UnsafeOptions = UnsafeConversionConfiguration.GetConfigurationFromEnvironmentVariables();
 
-            // In DScript V2 AST converter itself does not convert nodes in parallel.
-            DegreeOfParalellism = useLegacyOfficeLogic ? degreeOfParallelism : 1;
```
almiliMSFT: The `degreeOfParallelism` parameter is unused now, so it should be removed.

almiliMSFT:
Suggestion for renaming this PR: `Reduce DScript AST conversion parallelism`

PR #477, opened 2019-06-17
Enable additional DScript interpreter test for macOS

Kr1sso:
Enable SuperLongBinaryExpressionShouldNotLeadToStackoverflow test for macOS now that we know the culprit of stack overflows. On macOS all threads besides the main thread have "only" 512 Kilobytes of stack size (see https://developer.apple.com/library/archive/qa/qa1419/_index.html). `COMPlus_DefaultStackSize` helps here.

### DIFF
--- a/Public/Src/Deployment/Tests.Osx/test_runner.sh
+++ b/Public/Src/Deployment/Tests.Osx/test_runner.sh
@@ -14,6 +14,9 @@ chmod +x "$bxlSh"
 find "$MY_DIR/TestProj/tests" -name "*.TestProcess" -exec chmod +x {} \;
 find "$MY_DIR/TestProj/tests" -name "*CoreDump*" -exec chmod +x {} \;
 
+# Allow for up to 4mb of thread stack size
+export COMPlus_DefaultStackSize=400000
+
 # run the build
 # Temporarily disable this until we have the new macOS based CI
 # /kextMeasureProcessCpuTimes
--- a/Public/Src/Deployment/Tests.Osx/test_runner.sh
+++ b/Public/Src/Deployment/Tests.Osx/test_runner.sh
@@ -14,9 +14,6 @@ chmod +x "$bxlSh"
 find "$MY_DIR/TestProj/tests" -name "*.TestProcess" -exec chmod +x {} \;
 find "$MY_DIR/TestProj/tests" -name "*CoreDump*" -exec chmod +x {} \;
 
-# Allow for up to 4mb of thread stack size
-export COMPlus_DefaultStackSize=400000
-
 # run the build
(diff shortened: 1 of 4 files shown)


Public/Src/FrontEnd/UnitTests/Script.Interpretation/InterpretBinaryExpressions.cs
```diff
@@ -116,7 +116,7 @@ public void PathAndStringAdd()
             Assert.True(((string)result).EndsWith("/myfile.txt`"));
         }
 
-        [FactIfSupported(requiresWindowsBasedOperatingSystem: true)]
+        [Fact]
```
almiliMSFT: Right now, this unit test will test that our `test_runner.sh` script is correct (instead of testing BuildXL).  So for example, if someone runs this test from a debugger or something, the test will fail. 

I don't know how to fix this.  Setting `COMPlus_DefaultStackSize` from the `bxl` process is probably already too late.

In any case, add a comment here saying that this env variable needs to be set to something like 400000 for this test to pass

Public/Src/Deployment/Tests.Osx/test_runner.sh
```diff
@@ -14,6 +14,9 @@ chmod +x "$bxlSh"
 find "$MY_DIR/TestProj/tests" -name "*.TestProcess" -exec chmod +x {} \;
 find "$MY_DIR/TestProj/tests" -name "*CoreDump*" -exec chmod +x {} \;
 
+# Allow for up to 4mb of thread stack size
+export COMPlus_DefaultStackSize=400000
```
almiliMSFT: move this to [`Public/Src/Sandbox/MacOS/scripts/bxl.sh`](https://github.com/microsoft/BuildXL/blob/master/Public/Src/Sandbox/MacOs/scripts/bxl.sh)

dannyvv: Should this also be added when we launch the server process (in case a customer misses it?)

---
In reply to: [294366144](https://github.com/microsoft/BuildXL/pull/477#discussion_r294366144) [](ancestors = 294366144)

Public/Src/FrontEnd/UnitTests/Script.Interpretation/InterpretBinaryExpressions.cs
Kr1sso: As you've correctly pointed out this has to happen before the CoreCLR initialization - we could do some hacky bxl restart logic but I guess its not worth it. The error happens once DScript gets to a certain complexity level and the type checker recurses to deep, hitting the default stack size limits. I'll add a comment, so we don't forget.

PR #479, opened 2019-06-17
Allow users to specify the source from which a qtest target is built.

anirudhsanthiar:
This is essential for the correct operation of qtest's flaky test detection and reporting.

### DIFF
--- a/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
@@ -336,7 +336,7 @@ export interface QTestArguments extends Transformer.RunnerArguments {
     /** Specifies the environment variables to forward to qtest */
     qTestEnvironmentVariables?: Transformer.EnvironmentVariable[];
     /** Specify the path relative to enlistment root of the sources from which the test target is built */
-    testSourceDir?: string;
+    testSourceDir?: RelativePath;
 }
 /**
  * Test results from a vstest.console.exe run
--- a/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
@@ -172,7 +172,7 @@ export function runQTest(args: QTestArguments): Result {
         Cmd.option("--qTestAdditionalOptions ", args.qTestAdditionalOptions, args.qTestAdditionalOptions ? true : false),
         Cmd.option("--qTestContextInfo ", qTestContextInfoPath),
         Cmd.option("--qTestBuildType ", args.qTestBuildType || "unset"),
-        Cmd.option("--testSourceDir", args.testSourceDir)
+        Cmd.option("--testSourceDir ", args.testSourceDir)
     ];          
 
     let unsafeOptions = {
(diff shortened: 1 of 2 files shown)


Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -334,6 +335,8 @@ export interface QTestArguments extends Transformer.RunnerArguments {
     qTestBuildType?: string;
     /** Specifies the environment variables to forward to qtest */
     qTestEnvironmentVariables?: Transformer.EnvironmentVariable[];
+    /** Specify the path relative to enlistment root of the sources from which the test target is built */
+    testSourceDir?: string;
```
narasamdya: >string [](start = 20, length = 6)

If this is a path, why do you make it as a string? We have Path in our DScript type.

Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -171,7 +171,8 @@ export function runQTest(args: QTestArguments): Result {
         Cmd.flag("--qTestIgnoreQTestSkip", args.qTestIgnoreQTestSkip),
         Cmd.option("--qTestAdditionalOptions ", args.qTestAdditionalOptions, args.qTestAdditionalOptions ? true : false),
         Cmd.option("--qTestContextInfo ", qTestContextInfoPath),
-        Cmd.option("--qTestBuildType ", args.qTestBuildType || "unset")
+        Cmd.option("--qTestBuildType ", args.qTestBuildType || "unset"),
+        Cmd.option("--testSourceDir", args.testSourceDir)
```
narasamdya: >args.testSourceDir [](start = 38, length = 18)

This should be path, instead of string. But then, since this is just a directory, you can do `Artifact.none(args.testSourceDir)`.

Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
anirudhsanthiar: This is meant to be a relative path to a source folder. Is Path the right type to use here? What are the invariants guaranteed by the Path type? Will passing a relative path from enlistment root violate any of those invariants?

Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
anirudhsanthiar: Artifact.none does not accept RelativePath

Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
dannyvv: @anirudhsanthiar Perhaps then make the type RelativePath...

callers can specify these at the callsite with.
```
r`my/folder`
```

Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
dannyvv: If relative path no need for artifact...

PR #485, opened 2019-06-18
QTest: Add Recycle Bin to untracked directories

plade:
https://dev.azure.com/mseng/1ES/_workitems/edit/1552522

Files from Recycle bin cause failure due to disallowed file access.
Added Recycle bin from the drive of the sandbox to untracked directories.

### DIFF
--- a/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
@@ -182,6 +182,8 @@ export function runQTest(args: QTestArguments): Result {
         untrackedScopes: [
             d`d:/data`,
             d`d:/app`,
+            // Untracking Recyclebin here to primarily unblock user scenarios that
+            // deal with soft-delete and restoration of files from recycle bin.
             d`${sandboxDir.pathRoot}/$Recycle.Bin`,
             ...addIf(Environment.hasVariable("QAUTHMATERIALROOT"), Environment.getDirectoryValue("QAUTHMATERIALROOT")),
         ]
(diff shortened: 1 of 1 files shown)


Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -181,6 +181,7 @@ export function runQTest(args: QTestArguments): Result {
         untrackedScopes: [
             d`d:/data`,
             d`d:/app`,
+            d`${sandboxDir.pathRoot}/$Recycle.Bin`,
```
dannyvv: Is this because QTest performs operations on the recycle bin?

If it is customer (test code or code under test) then that test should actually do the work.

plade: The customer test case in question perfoms a Win32 Delete operation in one of the QTests. DFA flagging tracks the recycle bin as of today, which results in the test being flagged.

dannyvv: So then technically the customer should declare the dependency...  the Qtest ask can't aggregate all possible file pattern of customers. Since recycle bin is sort of potentially generic for tests I'm okay, but technically this belongs in the customers' set of dependencies/untracked space...

mpysson: The QTest SDK has a user specifiable hook for adding additional dependencies but there isn't one for untracking directories. I think it's ok to special case this since we really don't want/expect additional per-test untracking. The majority of accesses are in the QTest sandbox which is already untracked. In short, I think the change is fine

dannyvv: K, sounds good

PR #488, opened 2019-06-19
Add translator for GlobalUntrackScope

norahuang:
Translate the paths in GlobalUntrackedScope.


Public/Src/App/Bxl/BuildXLApp.cs
```diff
@@ -896,6 +896,8 @@ private AppResult RunWithLoggingScope(Func<PerformanceMeasurement, AppResult> ru
                         translatedLogDirectory = pathTranslator != null ? pathTranslator.Translate(logDirectory) : logDirectory;
                     }
 
+                    TranslateGlobalUntrackedScope();
```
dannyvv: >TranslateGlobalUntrackedScope [](start = 20, length = 29)

Would it make more sense to do this in: `PopulateAndValidateConfiguration` since that is were we do all our configuration mutations for the engine..

Public/Src/Utilities/Configuration/ISandboxConfiguration.cs
```diff
@@ -256,6 +256,6 @@ public interface ISandboxConfiguration
         /// Moreover, this global configuration from cammand line will bypass cache,
         /// which means pips and graph will be cached ignoring paths specified in this configure
         /// </remarks>
-        IReadOnlyList<AbsolutePath> GlobalUnsafeUntrackedScopes { get; }
+        List<AbsolutePath> GlobalUnsafeUntrackedScopes { get; set; }
```
dannyvv: The configuration interfaces should remain immutable..

The implementation is mutable...

Public/Src/App/Bxl/BuildXLApp.cs
dannyvv: The configuration in taht function is properly mutable already so you can leave the interface readonly

---
In reply to: [295508556](https://github.com/microsoft/BuildXL/pull/488#discussion_r295508556) [](ancestors = 295508556)

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -1113,6 +1113,9 @@ public static void PopulateLoggingAndLayoutConfiguration(CommandLineConfiguratio
                 mutableConfig.Schedule.EarlyWorkerRelease = false;
             }
 
+            PathTranslator.CreateIfEnabled(mutableConfig.Logging.SubstSource, mutableConfig.Logging.SubstTarget, pathTable, out var translator);
+            mutableConfig.Sandbox.GlobalUnsafeUntrackedScopes = mutableConfig.Sandbox.GlobalUnsafeUntrackedScopes.Select(path => AbsolutePath.Create(pathTable, translator.Translate(path.ToString(pathTable)))).ToList();
```
narasamdya: We don't want to perform translation here. The paths given in the `GlobalUnsafeUntrackedScopes` should contain those specified in the command line argument, and unmodified.

The translated version should be given at the sandboxed process layer. At that layer, we have a directory translator that is embedded in the file access manifest. You can use that translator to prepare the paths.

norahuang:
> I think we should make it a requirement that the caller of bxl make the command line, environment variables, and spec files consistent with respect to subst.

I was hesitate to change the existing environment variables from CB. 
Now I changed the environment variable __CREDENTIAL_PROVIDER_LOG_DIR to point the directory junction in session root instead of the target one in enlistment dir.
I have make the change in CB with the CB dependencies PR.
I will abandon this one then.

mpysson:
> > I think we should make it a requirement that the caller of bxl make the command line, environment variables, and spec files consistent with respect to subst.
> 
> I was hesitate to change the existing environment variables from CB.
> Now I changed the environment variable __CREDENTIAL_PROVIDER_LOG_DIR to point the directory junction in session root instead of the target one in enlistment dir.
> I have make the change in CB with the CB dependencies PR.
> I will abandon this one then.

Discussed in person. I retract my concern and agree that given the other constraints it makes sense to put this into bxl for now. Thanks for drilling into the various facets of the full picture for me.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1606,9 +1606,11 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
             }
 
             // Untrack the globally untracked paths specified in the configuration 
+            var reverseTranslator = m_fileAccessManifest.DirectoryTranslator.GetReverseTranslator();
```
narasamdya: >DirectoryTranslator [](start = 57, length = 19)

You need to check if this always exists. It could be null if I'm not mistaken.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1606,9 +1606,11 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
             }
 
             // Untrack the globally untracked paths specified in the configuration 
+            var reverseTranslator = m_fileAccessManifest.DirectoryTranslator.GetReverseTranslator();
             foreach (var path in m_sandboxConfig.GlobalUnsafeUntrackedScopes)
             {
-                m_fileAccessManifest.AddScope(path, mask: m_excludeReportAccessMask, values: FileAccessPolicy.AllowAll | FileAccessPolicy.AllowRealInputTimestamps);
+                var translatedPath = AbsolutePath.Create(m_pathTable, reverseTranslator.Translate(path.ToString(m_pathTable)));
+                m_fileAccessManifest.AddScope(translatedPath, mask: m_excludeReportAccessMask, values: FileAccessPolicy.AllowAll | FileAccessPolicy.AllowRealInputTimestamps);
```
narasamdya: >translatedPath [](start = 46, length = 14)

Why don't you add both original and translated paths as untracked?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1606,8 +1606,23 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
...
+                if (reverseTranslator != null)
+                {
+                    var pathString = path.ToString(m_pathTable);
+                    var translatedPathString = reverseTranslator?.Translate(pathString);
+                    var translatedPath = AbsolutePath.Create(m_pathTable, translatedPathString);
+                  
+                    if (pathString != translatedPathString)
+                    {
+                        m_fileAccessManifest.AddScope(translatedPath, mask: m_excludeReportAccessMask, values: FileAccessPolicy.AllowAll | FileAccessPolicy.AllowRealInputTimestamps);
+                        Tracing.Logger.Log.TranslatePathInGlobalUnsafeUntrackedScopes(loggingContext, m_pip.SemiStableHash, m_pip.GetDescription(m_context), pathString, translatedPathString);
```
almiliMSFT: why is only this path logged and not the one 3 lines below?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1606,8 +1606,23 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
...
 
             // Untrack the globally untracked paths specified in the configuration 
+            var reverseTranslator = m_fileAccessManifest.DirectoryTranslator?.GetReverseTranslator();
             foreach (var path in m_sandboxConfig.GlobalUnsafeUntrackedScopes)
             {
+                // translate the path and untrack the translated one
+                if (reverseTranslator != null)
+                {
+                    var pathString = path.ToString(m_pathTable);
+                    var translatedPathString = reverseTranslator?.Translate(pathString);
```
dannyvv: >? [](start = 64, length = 1)

`?` not needed...  check is above already

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1606,8 +1606,23 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
...
+                    var translatedPath = AbsolutePath.Create(m_pathTable, translatedPathString);
+                  
+                    if (pathString != translatedPathString)
```
dannyvv: You might have to do case insenstive string comparisson here..
At least 'ordinal' since now it is culture specific.

almiliMSFT: Even better: compare `AbsolutePaths` (which should take care of any platform-specific requirements for whether to use case sensitive or case insensitive)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: I only log the translation, don't want to spam the log, since this will happen in every pip.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1606,9 +1606,23 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
...
+            {               
+                foreach (var path in m_sandboxConfig.GlobalUnsafeUntrackedScopes)
+                {
+                    var pathString = path.ToString(m_pathTable);
+                    var translatedPathString = m_fileAccessManifest.DirectoryTranslator.Translate(pathString);
+                    var translatedPath = AbsolutePath.Create(m_pathTable, translatedPathString);
+
+                    if (!path.Equals(translatedPath))
+                    {
+                        //untrack the translated path
```
narasamdya: >u [](start = 26, length = 1)

nit: add space

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1606,9 +1606,23 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
             }
 
             // Untrack the globally untracked paths specified in the configuration 
-            foreach (var path in m_sandboxConfig.GlobalUnsafeUntrackedScopes)
-            {
-                m_fileAccessManifest.AddScope(path, mask: m_excludeReportAccessMask, values: FileAccessPolicy.AllowAll | FileAccessPolicy.AllowRealInputTimestamps);
+            if (m_fileAccessManifest.DirectoryTranslator != null)
```
narasamdya: >(m_fileAccessManifest.DirectoryTranslator [](start = 15, length = 41)

Does this mean that global untracked scopes won't work if there's no directory translator.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1606,9 +1606,23 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
             }
 
             // Untrack the globally untracked paths specified in the configuration 
-            foreach (var path in m_sandboxConfig.GlobalUnsafeUntrackedScopes)
-            {
-                m_fileAccessManifest.AddScope(path, mask: m_excludeReportAccessMask, values: FileAccessPolicy.AllowAll | FileAccessPolicy.AllowRealInputTimestamps);
+            if (m_fileAccessManifest.DirectoryTranslator != null) 
+            {               
+                foreach (var path in m_sandboxConfig.GlobalUnsafeUntrackedScopes)
+                {
```
narasamdya: >{ [](start = 16, length = 1)

Why don't we have in this block the following?
```
{
    m_fileAccessManifest.AddScope(path, mask: m_excludeReportAccessMask, values: FileAccessPolicy.AllowAll | FileAccessPolicy.AllowRealInputTimestamps);
    if (m_fileAccessManifest.DirectoryTranslator != null) { // cheap check
        var pathString = path.ToString(m_pathTable);
        var translatedPathString = m_fileAccessManifest.DirectoryTranslator.Translate(pathString);
        var translatedPath = AbsolutePath.Create(m_pathTable, translatedPathString);
        m_fileAccessManifest.AddScope(translatedPath, mask: m_excludeReportAccessMask, values: FileAccessPolicy.AllowAll | FileAccessPolicy.AllowRealInputTimestamps);
    }
}
```

narasamdya:

Any changes in this file?

---
Refers to: Public/Src/Utilities/Configuration/Mutable/SandboxConfiguration.cs:1 in 0cbad19. [](commit_id = 0cbad1989891360a32f58df91bac134e3ae6c55c, deletion_comment = False)

narasamdya:
Can you provide unit tests?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: oh, thanks, I made the logic wrong. Will change it

norahuang:
> 
> Any changes in this file?
> 
> Refers to: Public/Src/Utilities/Configuration/Mutable/SandboxConfiguration.cs:1 in 0cbad19. [](commit_id = 0cbad19, deletion_comment = False)

Not meaningful changes, but trimming the extra spaces.

norahuang:
> Can you provide unit tests?

Added

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1605,9 +1605,24 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
...
             foreach (var path in m_sandboxConfig.GlobalUnsafeUntrackedScopes)
             {
+                // translate the path and untrack the translated one                
+                if (m_fileAccessManifest.DirectoryTranslator != null)
+                {
+                    var pathString = path.ToString(m_pathTable);
+                    var translatedPathString = m_fileAccessManifest.DirectoryTranslator.Translate(pathString);
+                    var translatedPath = AbsolutePath.Create(m_pathTable, translatedPathString);
+
+                    if (!path.Equals(translatedPath))
```
narasamdya: >path.Equals(translatedPath)) [](start = 25, length = 28)

Do we care about case sensitiveness?

Public/Src/Engine/UnitTests/Processes.Detours/SandboxedProcessPipExecutorTest.cs
```diff
@@ -1876,6 +1876,88 @@ public async Task ProcessUsingUntrackedEnvironmentVariable()
             }
         }
 
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public async Task ProcessGlobalUntrackedScope(bool translate)
```
narasamdya: >ProcessGlobalUntrackedScope [](start = 26, length = 27)

Writing Detours test only exercises Windows case. Why don't you write an integration test, so that both Windows and Mac are covered?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: According to Aleks' comment, the AbsolutePath comparation will take care of platform specific case sensitiveness.

almiliMSFT: I believe that's true, but it's worth double checking

Public/Src/Engine/UnitTests/Processes.Detours/SandboxedProcessPipExecutorTest.cs
norahuang: moved to integration test.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: > path.Equals(translatedPath)) [](start = 25, length = 28)
> 
> Do we care about case sensitiveness?

I checked d: equals to D:

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/UnsafeGlobalUntrackedScopesTests.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace IntegrationTest.BuildXL.Scheduler
+{
+    public class UnsafeGlobalUntrackedScopesTests : SchedulerIntegrationTestBase
+    {
+        public UnsafeGlobalUntrackedScopesTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Theory]
+        //[InlineData(true)]
```
almiliMSFT: intentional?

norahuang: no, thanks for catching.

PR #491, opened 2019-06-20
Refinements for error tracking in telemetry

mpysson:
- Classify distributed build network failures with higher precedence than all other infrastructural errors
- Plumb first error message head & tail into completion events to aid debugging internal errors


Public/Src/Utilities/Utilities/Tracing/TrackingEventListener.cs
```diff
@@ -6,6 +6,7 @@
 using System.Globalization;
 using System.Threading;
 using BuildXL.Utilities.Collections;
+using System;
```
olkononenko: > System; [](start = 5, length = 8)

nit: sort using's?

Public/Src/Utilities/Utilities/Tracing/TrackingEventListener.cs
```diff
@@ -41,13 +38,17 @@ public sealed class TrackingEventListener : BaseEventListener
...
         /// The event source to listen to.
         /// </param>
+        /// <param name="baseTime">
+        /// The starting time that events are compared against for displaying relative time for messages.
+        /// </param>
         /// <param name="warningMapper">
         /// An optional delegate that is used to map warnings into errors or to suppress warnings.
         /// </param>
-        public TrackingEventListener(Events eventSource, WarningMapper warningMapper = null)
+        public TrackingEventListener(Events eventSource, DateTime baseTime = default(DateTime), WarningMapper warningMapper = null)
```
olkononenko: >DateTime baseTime = default(DateTime) [](start = 57, length = 37)

nit: use nullable here?

Public/Src/Utilities/Utilities/Tracing/ErrorCategoryDetails.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// Details about errors logged for a particular error category
+    /// </summary>
+    public class ErrorCagetoryDetails
+    {
+        private int m_numErrors;
+
+        /// <summary>
+        /// The name of the first error of this category logged
+        /// </summary>
+        public string FirstErrorName { get; set; }
```
olkononenko: > set;  [](start = 43, length = 6)

I think we can remove setters on these properties (everything should be set via RegisterError method, right?)

Public/Src/Engine/Dll/Distribution/WorkerService.cs
```diff
@@ -903,7 +903,7 @@ private sealed class StatusReporter : BaseEventListener
                 private Snapshot m_pipStateSnapshot;
 
                 public StatusReporter(LoggingContext loggingContext, WorkerServicePipStateManager stateManager)
-                    : base(Events.Log, warningMapper: null, eventMask: new EventMask(enabledEvents: new int[] { (int)EventId.PipStatus }, disabledEvents: null))
+                    : base(Events.Log, baseTime:DateTime.MinValue, warningMapper: null, eventMask: new EventMask(enabledEvents: new int[] { (int)EventId.PipStatus }, disabledEvents: null))
```
olkononenko: >MinValue [](start = 57, length = 8)

why are we using MinValue here?

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -319,7 +319,7 @@ internal static string PipWriteFileFailedMessage(string pipDescription, string p
             Keywords = (int)Events.Keywords.UserMessage,
             EventTask = (ushort)Events.Tasks.Scheduler,
             Message = "[{pipDescription}] Failed to materialize pip dependencies content from cache: {errorMessage}")]
-        internal abstract void PipMaterializeDependenciesFromCacheFailure(LoggingContext loggingContext, string pipDescription, string errorMessage);
+        public abstract void PipMaterializeDependenciesFromCacheFailure(LoggingContext loggingContext, string pipDescription, string errorMessage);
```
olkononenko: >public [](start = 8, length = 6)

looks like it's only needed for the test, right? can we achive the same effect by adding 'internalsVisibleTo : [ name of test namespace/dsc ]' ?

olkononenko:
>         /// CAUTION!!

is this comment up-to-date? if yes, should we update the 'upcoming release' section on the wiki so we won't forget to include it in the release email?

---
Refers to: Public/Src/App/Bxl/Tracing/Log.cs:89 in e78ab04. [](commit_id = e78ab04bf647886e8f77d72006109377861d517d, deletion_comment = False)

Public/Src/Utilities/Utilities/Tracing/ErrorCategoryDetails.cs
mpysson: Still need setters to modify them in that method, I made them private though

---
In reply to: [296053601](https://github.com/microsoft/BuildXL/pull/491#discussion_r296053601) [](ancestors = 296053601)

mpysson:
>         /// CAUTION!!

technically additions are ok. Only changes to existing fields are breaks. Updated comment

---
In reply to: [504240796](https://github.com/microsoft/BuildXL/pull/491#issuecomment-504240796) [](ancestors = 504240796)

---
Refers to: Public/Src/App/Bxl/Tracing/Log.cs:89 in e78ab04. [](commit_id = e78ab04bf647886e8f77d72006109377861d517d, deletion_comment = False)

PR #494, opened 2019-06-20
Fix non-internal sandbox build change detection

Kr1sso:
Fix source file dependencies for sandbox build phase to correctly build code changes when the sandbox is not running during the bxl execution (non-internal builds). Previously, source file changes would not trigger rebuild.

### DIFF
--- a/Public/Src/Sandbox/MacOs/projects.dsc
+++ b/Public/Src/Sandbox/MacOs/projects.dsc
@@ -21,13 +21,23 @@ namespace Sandbox {
         derivedDataOutDir: StaticDirectory
     }
 
-    const sandboxSealDir = Transformer.sealSourceDirectory(
-        d`${Context.getMount("Sandbox").path}`,
-        Transformer.SealSourceDirectoryOption.allDirectories);
+    function sourceFileDependencies()
+    {
+        const sandboxDir = d`${Context.getMount("Sandbox").path}`;
+        const thirdPartyDir = d`../../../../third_party`;
 
-    const thirdPartySealDir = Transformer.sealSourceDirectory(
-        d`../../../../third_party`,
-        Transformer.SealSourceDirectoryOption.allDirectories);
+        const sourceSealOptions =  Transformer.SealSourceDirectoryOption.allDirectories;
+        const globPattern = "*";
+
+        return BuildXLSdk.Flags.isMicrosoftInternal ?
+            [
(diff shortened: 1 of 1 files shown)


Public/Src/Sandbox/MacOs/projects.dsc
```diff
@@ -21,13 +21,23 @@ namespace Sandbox {
...
     }
 
-    const sandboxSealDir = Transformer.sealSourceDirectory(
-        d`${Context.getMount("Sandbox").path}`,
-        Transformer.SealSourceDirectoryOption.allDirectories);
-
-    const thirdPartySealDir = Transformer.sealSourceDirectory(
-        d`../../../../third_party`,
-        Transformer.SealSourceDirectoryOption.allDirectories);
+    function sourceFileDependencies()
```
almiliMSFT: If this function is called multiple times, it will seal those directories multiple times (once per invocation).

That's why I had those as variables before.

```suggestion
    const sourceFileDependencies = (() =>
```

Public/Src/Sandbox/MacOs/projects.dsc
```diff
@@ -21,13 +21,23 @@ namespace Sandbox {
...
+
+        return BuildXLSdk.Flags.isMicrosoftInternal ?
+            [
+                Transformer.sealSourceDirectory(sandboxDir, sourceSealOptions),
+                Transformer.sealSourceDirectory(thirdPartyDir, sourceSealOptions)
+            ] : [
+                ...Transformer.sealDirectory(sandboxDir, globR(sandboxDir, globPattern)).contents,
+                ...Transformer.sealDirectory(thirdPartyDir, globR(thirdPartyDir, globPattern)).contents
+            ];
+    }
```
almiliMSFT:
```suggestion
    })();
```

Public/Src/Sandbox/MacOs/projects.dsc
```diff
@@ -45,10 +55,10 @@ namespace Sandbox {
             ],
             dependencies: [
                 ...(args.dependencies || []),
-                sandboxSealDir,
-                thirdPartySealDir
+                ...sourceFileDependencies()
```
almiliMSFT:
```suggestion
                ...sourceFileDependencies
```

Public/Src/Sandbox/MacOs/projects.dsc
Kr1sso: Good catch!

PR #498, opened 2019-06-20
Don't delete contents of preserved opaque directories on cache replay

mpysson:
This updates the opaque directory cache replay logic to not nuke the directory before replay if the directory is preserved and running in preserve outputs mode. The goal here is to enable gradle pips to be incremental again

### DIFF
--- a/Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
+++ b/Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
@@ -1579,9 +1579,12 @@ namespace BuildXL.Scheduler.Artifacts
 
                     if (sealDirectoryKind == SealDirectoryKind.Opaque)
                     {
-                        // Dynamic directories must be deleted before materializing files
-                        // We don't want this to happen for shared dynamic ones
-                        AddDirectoryDeletion(state, artifact.DirectoryArtifact);
+                        if (Configuration.Sandbox.UnsafeSandboxConfiguration.PreserveOutputs != PreserveOutputsMode.Enabled || !PipArtifacts.IsPreservedOutputByPip(state.PipInfo.UnderlyingPip, directory.Path, Context.PathTable))
+                        {
+                            // Dynamic directories must be deleted before materializing files
+                            // We don't want this to happen for shared dynamic ones
+                            AddDirectoryDeletion(state, artifact.DirectoryArtifact);
+                        }
 
                         // For dynamic directories we need to specify the value of
                         // allow read only since the host will not know about the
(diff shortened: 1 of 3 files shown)


Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/OpaqueDirectoryTests.cs
```diff
@@ -70,6 +70,46 @@ public void OpaqueDirectoryConsumptionCachingBehavior()
...
+            RunScheduler().AssertCacheHit(pip.Process.PipId);
+
+            // If running with Incremental scheduling the stray file will still exist, otherwise it will not.
+            if (Configuration.Schedule.IncrementalScheduling || Configuration.Schedule.GraphAgnosticIncrementalScheduling)
+            {
+                XAssert.IsTrue(File.Exists(externallyProducedOutput), "Expected {0} to exist when using incremental scheduling", externallyProducedOutput);
+            }
+            else
+            {
+                XAssert.IsFalse(File.Exists(externallyProducedOutput), "Did not expect {0} to exist when using incremental scheduling", externallyProducedOutput);
```
narasamdya: >when using  [](start = 100, length = 11)

"when not using incremental scheduling"

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -1579,9 +1579,12 @@ private void PopulateArtifactsToMaterialize(PipArtifactsState state, bool? allow
 
                     if (sealDirectoryKind == SealDirectoryKind.Opaque)
                     {
-                        // Dynamic directories must be deleted before materializing files
-                        // We don't want this to happen for shared dynamic ones
-                        AddDirectoryDeletion(state, artifact.DirectoryArtifact);
+                        if (Configuration.Sandbox.UnsafeSandboxConfiguration.PreserveOutputs != PreserveOutputsMode.Enabled || !PipArtifacts.IsPreservedOutputByPip(state.PipInfo.UnderlyingPip, directory.Path, Context.PathTable))
```
narasamdya: >if [](start = 24, length = 2)

I'm leaning towards replying the empty directory inside an opaque directory. But given we need to deliver this to customer, I guess this change is OK.

dannyvv: nit: rhs of `||` on new line for readability..

semihokur: i cannot update Mike's branch. As the rest of the changes looks good, I will merge it.

PR #499, opened 2019-06-20
Increase default number of retries for IPC connections

almiliMSFT:


Public/Src/Utilities/Ipc/Common/Utils.cs
```diff
@@ -83,7 +83,7 @@ internal static async Task<Possible<TConn>> ConnectAsync<TConn>(int maxRetry, Ti
                 if (numAttempts > 0)
                 {
                     // TODO: consider implementing exponential backoff instead
```
mpysson: You added a linear backoff. Is that enough to warrant removing the todo?

almiliMSFT: Probably.  I'll remove it.

PR #502, opened 2019-06-21
Nuget full framework dependency spec generator changes

Kr1sso:
The Nuget spec file explicitly lists group dependencies for specific target frameworks. The logic so far has been to only include those dependencies (and their transitive closure) if the target framework of the qualifier evaluated matches the one specified in the nuspec files group dependency section. This is especially important for .NETCore, where continuous releases include more assemblies in the runtime folder and would collide when using all specified dependencies without looking at the target framework.

The right solution is explicitly specifying for what target framework a dependency and its transitive closure should be included using e.g. `withQualifier()`. We keep this for .NETStandard/Core going forward. For full framework builds (which we are gradually deprecating) and for the fact that package maintainers don't keep their Nuget dependencies in check to explicitly make statements about what dependencies are needed for which target framework, we now always include all dependencies at their latest version.

### DIFF
--- a/Public/Src/FrontEnd/Nuget/NugetSpecGenerator.cs
+++ b/Public/Src/FrontEnd/Nuget/NugetSpecGenerator.cs
@@ -124,6 +124,14 @@ namespace BuildXL.FrontEnd.Nuget
             var valid = analyzedPackage.TargetFrameworks.Exists(moniker => m_nugetFrameworkMonikers.FullFrameworkVersionHistory.Contains(moniker) || m_nugetFrameworkMonikers.NetCoreVersionHistory.Contains(moniker));
             Contract.Assert(valid, "Target framework monikers must exsist and be registered with internal target framework version helpers.");
 
+            var fullFrameworkDeps = m_nugetFrameworkMonikers.FullFrameworkVersionHistory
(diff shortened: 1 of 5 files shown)


Public/Src/FrontEnd/Nuget/NugetSpecGenerator.cs
```diff
@@ -142,27 +155,26 @@ private List<ICaseClause> CreateSwitchCasesForTargetFrameworks(NugetAnalyzedPack
...
-                        {
-                            runtime.Add(CreateSimpleBinary(assembly));
-                        }
+                        runtime.AddRange(libAssemblies.Select(l => CreateSimpleBinary(l)));
                     }
 
-                    // Dependency items
-                    if (analyzedPackage.DependenciesPerFramework.TryGetValue(monikers.First(), out IReadOnlyList<INugetPackage> dependencySpecificFrameworks))
+                    // For full framework dependencies we unconditionally include all the distinct dependencies from the nuspec file,
+                    // .NETStandard dependencies are only included if the monikor and the parsed target framework match!
```
almiliMSFT:
```suggestion
                    // .NETStandard dependencies are only included if the moniker and the parsed target framework match!
```

PR #506, opened 2019-06-21
Reduce tracing

SergeyTeplyakov:
This PR reduces the number of messages sent to Kusto by doing the following things:

1. Stop tracing "starts" in `FileSystemContentStoreInternal` (should cut about 10% of overall messages because FSCS is responsible for roughly 30% of all the messages.
2. More efficient traces for `RegisterLocalLocationAsync`. Originally RRLA messages are responsible for 20-25% of all the messages and now there is only 1 message instead of 5.
3. Using short hashes as much as possible. This will reduce the size of the messages and will make investigations slightly easier.
4. Cut redundant messages for copying files.

PR #508, opened 2019-06-21
Remove noisy GRPC logging

kewilli:


kewilli:
Abandoning in favor of #506.

PR #510, opened 2019-06-22
Fix allocation issues in ConcurrentBigMap

dannyvv:
ConcurrentBigmap was using KeyValuePair in collections causing lots of unneeded boxing when adding
items to the maps. This change uses a proper struct that can safely and performently be used as entries
in hash based collections.

### DIFF
--- a/Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
+++ b/Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
@@ -1,0 +1,58 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace BuildXL.Utilities.Collections
+{
+    /// <summary>
+    /// This is a key value pair for the concurrent big map.
+    /// </summary>
+    /// <remarks>
+    /// We can't use the built-in KeyValuePair because it does not implement GetHashCode and Equals.
+    /// Using KeyValue would result in a lot more memory traffic when using the ConcurrentBigMap due to
+    /// allocations made by boxing key/values when they are added to some internal members of ConcurrentBigMap
+    /// </remarks>
+    public readonly struct ConcurrentBigMapEntry<TKey, TValue> : IEquatable<ConcurrentBigMapEntry<TKey, TValue>>
+    {
+        /// <nodoc />
+        public ConcurrentBigMapEntry(TKey key, TValue value)
+        {
+            Key = key;
+            Value = value;
+        }
+
+        /// <nodoc />
+        public TKey Key { get; }
+
+        /// <nodoc />
+        public TValue Value { get; }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            return StructUtilities.Equals(this, obj);
+        }
+
+        /// <inheritdoc />
+        public bool Equals(ConcurrentBigMapEntry<TKey, TValue> other)
+        {
+            return EqualityComparer<TKey>.Default.Equals(Key, other.Key) &&
+                EqualityComparer<TValue>.Default.Equals(Value, other.Value);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            return (Key, Value).GetHashCode();
+        }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            // Use the original KeyValuePair pair tostring for convenience.
+            return new KeyValuePair<TKey, TValue>(Key, Value).ToString();
+        }
+    }
+}
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
@@ -478,7 +478,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         internal abstract void Persist(OperationContext context, ShortHash hash, ContentLocationEntry entry);
 
         /// <nodoc />
-        internal virtual void PersistBatch(OperationContext context, IEnumerable<KeyValuePair<ShortHash, ContentLocationEntry>> pairs)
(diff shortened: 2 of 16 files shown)


Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace BuildXL.Utilities.Collections
+{
+    /// <summary>
+    /// This is a key value pair for the concurrent big map.
+    /// </summary>
+    /// <remarks>
+    /// We can't use the built-in KeyValuePair because it does not implement GetHashCode and Equals
+    /// </remarks>
+    public struct ConcurrentBigMapEntry<TKey, TValue> : IEquatable<ConcurrentBigMapEntry<TKey, TValue>>
```
almiliMSFT: Should/can this be declared as `readonly` struct?

Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace BuildXL.Utilities.Collections
+{
+    /// <summary>
+    /// This is a key value pair for the concurrent big map.
+    /// </summary>
+    /// <remarks>
+    /// We can't use the built-in KeyValuePair because it does not implement GetHashCode and Equals
```
almiliMSFT: If we can't use `KeyValuePair`, how did we use it until now?

Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <nodoc />
+        public TKey Key { get; }
+
+        /// <nodoc />
+        public TValue Value { get; }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            return obj is KeyValuePair<TKey, TValue> pair &&
```
almiliMSFT:
```suggestion
            return obj is KeyValuePair<TKey, TValue> pair && Equals(pair);
```

Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
almiliMSFT: > This change uses a proper struct that can safely and performently be used as entries in hash based collections.

How is this struct more proper or more safe or more performant than `KeyValuePair`?  Is it all because this one implements Equals/HashCode?

dannyvv: yes, I was thinking about that this morning as well.

Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
dannyvv: With a lot of extra allocations... I had a test case in a prototype where adding lots of entries were showing up that boxing was allocating hundreds of megs of memory since KeyValuePair<T,U> does is not equatable or implement  gethashcode and Equals therefore was boxing inside of the dictionaries. See: https://devblogs.microsoft.com/premier-developer/performance-implications-of-default-struct-equality-in-c/

Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
dannyvv: Actually this should have been `StructUtilities.Equals` ...

Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
dannyvv: Correct, and it is not about safety, it is about perf due to nr of allocations.

dannyvv: updated comment to reflect

almiliMSFT: Maybe add that comment to the remarks as well.

Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
almiliMSFT: thanks for the clarification

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -56,6 +56,7 @@
 using BuildXL.Processes.Containers;
 using static BuildXL.Scheduler.FileMonitoringViolationAnalyzer;
 using BuildXL.Utilities.VmCommandProxy;
+using BuildXL.Utilities.Configuration.Mutable;
 #if FEATURE_MICROSOFT_DIAGNOSTICS_TRACING
```
kewilli: Nit: sort usings

Public/Src/Utilities/Collections/ConcurrentBigMapEntry.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace BuildXL.Utilities.Collections
+{
+    /// <summary>
+    /// This is a key value pair for the concurrent big map.
+    /// </summary>
+    /// <remarks>
+    /// We can't use the built-in KeyValuePair because it does not implement GetHashCode and Equals.
+    /// Using KeyValue would result in a lot more memory traffic when using the ConcurrentBigMap due to
+    /// allocations made by boxing key/values when they are added to:
```
SergeyTeplyakov: I think you haven't finished the sentence. 'added to:'

ljcollins25:
As far as I know, Equals/GetHashCode should never be called on KeyValuePair.

SergeyTeplyakov:
I draw the wrong conclusion based on the traces I've seen on Friday.

Indeed we have 10% of memory traffic from GenericEqualityComparer.Equals and GenericEqualityComparer.GetHashCode but the boxed type is FileArtifact now KeyValuePair<FileArtifact, OtherStuff>.

So the issue is there but it is not caused by KeyValuePair.

---
In reply to: [505123953](https://github.com/microsoft/BuildXL/pull/510#issuecomment-505123953) [](ancestors = 505123953)

PR #514, opened 2019-06-24
Use server GC on ContentStoreApp.exe

kewilli:


Public/Src/Cache/ContentStore/App/App.config
```diff
@@ -7,6 +7,7 @@
     </connectionManagement>
   </system.net>
   <runtime>
+    <gcServer enabled="true" />
```
dannyvv: >erver enable [](start = 8, length = 12)

Note that NetCore has it's own format for doing this... You'll have to update the runtime.json for this.

almiliMSFT: If this is being built through `BuildXLSdk.dsc`, runtimeconfig.json will be generated by the SDK and it will already contain all the [config options](https://github.com/microsoft/BuildXL/blob/37124d26b1c8b5905c792fbe7d256c26c424a5c4/Public/Sdk/Public/Managed/runtimeConfigFiles.dsc#L207-L212) needed to enable server GC

almiliMSFT: Note that in Bxl's App.config we additionally set
```xml
    <Thread_UseAllCpuGroups enabled="true"/>
    <GCCpuGroup enabled="true"/>
```
Not sure what's the significance of those two parameters though.

PR #515, opened 2019-06-25
Speed up loggen by adding a 'lite' version

dannyvv:
When we use the 'lite' version this does two things:
a) Reduces the C# files that loggenerator has to parse to only files under the `Tracing` folder
b) Reduces the C# references to a fixed small set just needed by just log generation. This will get most loggens out of the critical path of our selfhost.


Public/Sdk/SelfHost/BuildXL/BuildXLSdk.dsc
```diff
@@ -61,6 +61,9 @@ export interface Arguments extends Managed.Arguments {
 
     /** Whether to run LogGen. */
     generateLogs?: boolean;
+    
+    /** Wheter we can use the fast lite version of loggen */
```
smera: Whether

PR #520, opened 2019-06-25
Set buffer size on GRPC client

kewilli:


Public/Src/Cache/ContentStore/Grpc/CopyConstants.cs
```diff
@@ -5,8 +5,6 @@ namespace BuildXL.Cache.ContentStore.Grpc
 {
     public static class CopyConstants
     {
-        public static readonly int DefaultBufferSize = 8192;
-
-        public static readonly int CompressionSize = DefaultBufferSize * 8;
```
dcwuser: I understand this isn't needed for client, but if you delete this, how is server deciding when to compress?

PR #522, opened 2019-06-25
Configure GZip barrier

kewilli:


Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -46,14 +47,16 @@ public class GrpcContentServer
             Capabilities serviceCapabilities,
             ISessionHandler<IContentSession> sessionHandler,
             Dictionary<string, IContentStore> storesByName,
-            int? bufferSize = null)
+            int? bufferSize = null,
+            int? gzipSizeBarrier = null)
```
ljcollins25: >gzipSizeBarrier [](start = 17, length = 15)

I think this could just take a LocalServerConfiguration.

PR #523, opened 2019-06-26
[ContentStoreApp] Gracefully handle SIGTERM

almiliMSFT:


Public/Src/Cache/ContentStore/App/Program.cs
```diff
@@ -12,7 +15,33 @@ public static class Program
...
+                // must cancel this event because otherwise CoreCLR will proceed to kill this process
+                eventArgs.Cancel = true;
+            };
+
+            // handle graceful termination (i.e., SIGTERM) by requesting cancellation from the app
+            AppDomain.CurrentDomain.ProcessExit += (sender, eventArgs) =>
+            {
+                cancellationTokenSource.Cancel();
+                // we cannot cancel this event, so to prevent CoreCLR from killing this process before
+                // we are done cancelling and cleaning up we wait here until our app task completes.
```
almiliMSFT:
```suggestion
                // we are done finishing up we wait here until our app task completes.
```

Public/Src/Cache/ContentStore/App/Program.cs
```diff
@@ -12,7 +15,33 @@ public static class Program
     {
         private static int Main(string[] args)
         {
-            using (var app = new Application())
+            var cancellationTokenSource = new CancellationTokenSource();
+            var runAppTask = Task.Run(() => RunApp(args, cancellationTokenSource.Token));
+
+            // handle Ctrl+C (i.e., SIGINT) by requesting cancellation from the app
+            Console.CancelKeyPress += (sender, eventArgs) =>
```
dannyvv: Should we move this into a shared helper?

We do have BuildXL.Utilities.ToolSupport which would be a natural place.

The primary executabels we have are:
* bxl (assuming handled)
* cachetool
* bxlanalyzer
* bxlscriptanalyzer
* bxp-server

almiliMSFT: ContentStoreApp doesn't currently reference the ToolSupport assembly.  I don't want to add this dependency just because of this.

bxl doesn't use the ToolSupport's `ToolProgram` class either.  It already handles Ctrl+C but not SIGTERM.  

bxl-server doesn't derive from `ToolProgram` either.

Other apps don't seem to need to handle cancellation.

almiliMSFT: I extracted this into a new utility class and updated bxl to use it to handle termination requests.

I tried to add `ToolSupport` as a dependency of `ContentStoreApp` but qualifiers didn't agree (ContentStoreApp needs to support net461) so I left that as is.

almiliMSFT: I tried handling SIGTERM for bxl but it didn't work.  I think what happens is, EventListener stops working after SIGTERM was detected, and so when we start cancelling pips we log error messages (as we should) but those do not get picked up by trackingEventListeners so later on our assertion fails:
```
Assertion failed: The build has failed but the logging infrastructure has not encountered an error. 
```
So in the end, I reverted everything to where it was

PR #525, opened 2019-06-26
Make sure all accesses under shared opaques get reported

smera:
Not all accesses corresponding to inputs falling under the cone of a shared opaque were reported. Undeclared accesses and sealed directory accesses were skipped. This PR fixes that problem, but considering this is a breaking change, the fix takes place under an unsafe flag that is on by default.
The plan is to fix affected customers first and then swap the default (or even remove the flag)

### DIFF
--- a/Public/Src/Engine/Dll/Tracing/Log.cs
+++ b/Public/Src/Engine/Dll/Tracing/Log.cs
@@ -1089,8 +1089,8 @@ namespace BuildXL.Engine.Tracing
             (ushort)EventId.ConfigUnsafeIgnoreUndeclaredAccessesUnderSharedOpaques,
             EventGenerators = EventGenerators.LocalOnly,
             EventLevel = Level.Warning,
-            Keywords = (int)Events.Keywords.UserMessage,
-            EventTask = (int)Events.Tasks.Engine,
+            Keywords = (int)Keywords.UserMessage,
+            EventTask = (int)Tasks.Engine,
             Message = "/unsafe_IgnoreUndeclaredAccessesUnderSharedOpaques enabled: Undeclared accesses under shared opaques will not be reported. This is an unsafe configuration since it removes all guarantees of build correctness.")]
         public abstract void ConfigUnsafeIgnoreUndeclaredAccessesUnderSharedOpaques(LoggingContext context);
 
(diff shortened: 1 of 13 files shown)


smera:
> Just making sure I understand correctly, with this we continue running as is, only setting the flag to false surfaces the DFA errors?

Yes. The bug will still be there unless you run bxl with /unsafe_IgnoreUndeclaredAccessesUnderSharedOpaques- . So people can try to fix the new DFAs  by passing this flag.

Kr1sso:
> > Just making sure I understand correctly, with this we continue running as is, only setting the flag to false surfaces the DFA errors?
> 
> Yes. The bug will still be there unless you run bxl with /unsafe_IgnoreUndeclaredAccessesUnderSharedOpaques- . So people can try to fix the new DFAs by passing this flag.

Thanks, makes perfect sense.

Public/Src/Engine/Dll/Tracing/Log.cs
```diff
@@ -1085,6 +1085,15 @@ public abstract partial class Logger
             Message = "/unsafe_UnexpectedFileAccessesAreErrors disabled: Invalid file accesses by tools are configured to trigger warnings instead of errors. This is an unsafe (diagnostic only) configuration since it removes all guarantees of build correctness.")]
         public abstract void ConfigUnsafeUnexpectedFileAccessesAsWarnings(LoggingContext context);
 
+        [GeneratedEvent(
+            (ushort)EventId.ConfigUnsafeIgnoreUndeclaredAccessesUnderSharedOpaques,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Warning,
+            Keywords = (int)Keywords.UserMessage,
+            EventTask = (int)Tasks.Engine,
+            Message = "/unsafe_IgnoreUndeclaredAccessesUnderSharedOpaques enabled: Undeclared accesses under shared opaques will not be reported. This is an unsafe configuration since it removes all guarantees of build correctness.")]
```
almiliMSFT: "Undeclared accesses under shared opaques will not be reported" sounds very counter-intuitive to me, because the main idea behind opaque directories is exactly to not have to declare any accesses.

smera: In terms of outputs, yes. In terms of inputs you are still only allowed to read what you declare. I could rephrase it as 'undeclared reads' if that sounds better.

almiliMSFT: maybe it does sound a bit better, but I'm not insisting

PR #531, opened 2019-06-27
Substitute process shim NRE fix and logging

erikma:
Fix issue with null lpCommandLine (use lpApplicationName instead), and add debug tracing

### DIFF
--- a/Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
+++ b/Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
@@ -1794,6 +1794,7 @@ BOOL WINAPI Detoured_CreateProcessW(
 {
     bool injectedShim = false;
     BOOL ret = MaybeInjectSubstituteProcessShim(
+        lpApplicationName,
         lpCommandLine,
         lpProcessAttributes,
         lpThreadAttributes,
@@ -1806,6 +1807,7 @@ BOOL WINAPI Detoured_CreateProcessW(
         injectedShim);
     if (injectedShim)
     {
+        Dbg(L"Injected shim for lpCommandLine='%s', returning 0x%08X from CreateProcessW", lpCommandLine, ret);
         return ret;
     }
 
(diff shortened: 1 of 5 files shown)


Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1806,6 +1807,7 @@ BOOL WINAPI Detoured_CreateProcessW(
         injectedShim);
     if (injectedShim)
     {
+        Dbg(L"Injected shim for lpCommandLine='%s', returning 0x%08X from CreateProcessW", lpCommandLine, ret);
```
dmgonch: >lpCommandLine [](start = 32, length = 13)

can lpCommandLine be NULL?

Public/Src/Sandbox/Windows/DetoursServices/SubstituteProcessExecution.cpp
```diff
@@ -251,12 +252,15 @@ BOOL WINAPI MaybeInjectSubstituteProcessShim(
...
+        // when lpCommandLine is null we just use lpApplicationName as the command line to parse.
+        // When lpCommandLine is not null, it contains the command, possibly with quotes containing spaces,
+        // as the first whitespace-delimited token; we can ignore lpApplicationName in this case.
+        Dbg(L"Shim: Finding command and args from lpApplicationName='%s', lpCommandLine='%s'", lpApplicationName, lpCommandLine);
+        LPCWSTR cmdLine = lpCommandLine == nullptr ? lpApplicationName : lpCommandLine;
         wstring command;
         wstring commandArgs;
-        FindApplicationNameFromCommandLine(lpCommandLine, command, commandArgs);
+        FindApplicationNameFromCommandLine(cmdLine, command, commandArgs);
+        Dbg(L"Shim: Found command='%s', args='%s' from lpApplicationName='%s', lpCommandLine='%s'", command.c_str(), commandArgs.c_str(), lpCommandLine);
```
dmgonch: >lpCommandLine [](start = 138, length = 13)

can ne NULL - will it crash then?

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
erikma: Added additional check, having both lpApp and lpCmd null should not be possible but better to be defensive.

PR #533, opened 2019-06-28
Allow absolute path and frontend to accept paths beyond root drive

sahilmgandhi:
[AB#1557991](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1557991) 

Allow the absolute paths for windows/unix to go past the root (ie C:/aaa/bbb/../../../../ should resolve to C:/). This is particularly useful for any Domino auto-generated paths during the build phase. 

Furthermore, if the frontend passes in any paths that go beyond the root, we no longer throw a path invalid error, and merely return the root directory.

### DIFF
--- a/Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/LiteralConverter.cs
+++ b/Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/LiteralConverter.cs
@@ -227,13 +227,16 @@ namespace BuildXL.FrontEnd.Script.RuntimeModel.AstBridge
         {
             // parent folder count is a number of ../ expression in the original path literal
             var folder = currentFolder;
+            // If we go beyond the root directory, stay at the root directory. Ie C:/AAA/BBB/../../../../.. should be C:/
+            var tempFolder = currentFolder;
             for (int i = 0; i < parentFolderCount; i++)
             {
-                folder = folder.GetParent(Context.PathTable);
-                if (!folder.IsValid)
+                tempFolder = tempFolder.GetParent(Context.PathTable);
+                if (!tempFolder.IsValid)
                 {
                     break;
                 }
+                folder = tempFolder;
             }
 
             if (!folder.IsValid)
--- a/Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/LiteralConverter.cs
+++ b/Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/LiteralConverter.cs
@@ -226,22 +226,23 @@ namespace BuildXL.FrontEnd.Script.RuntimeModel.AstBridge
         private AbsolutePath ComputeAbsolutePath(AbsolutePath currentFolder, RelativePath relativePath, int parentFolderCount)
         {
             // parent folder count is a number of ../ expression in the original path literal
+            if (!currentFolder.IsValid)
+            {
+                return AbsolutePath.Invalid;
+            }
+
             var folder = currentFolder;
-            // If we go beyond the root directory, stay at the root directory. Ie C:/AAA/BBB/../../../../.. should be C:/
-            var tempFolder = currentFolder;
+
             for (int i = 0; i < parentFolderCount; i++)
             {
-                tempFolder = tempFolder.GetParent(Context.PathTable);
+                var tempFolder = folder.GetParent(Context.PathTable);
+                // If we go beyond the root directory (tempFolder will be invalid), stay at the root directory. 
+                // Ie. C:/AAA/BBB/../../../../.. should be C:/
                 if (!tempFolder.IsValid)
                 {
                     break;
                 }
                 folder = tempFolder;
-            }
-
-            if (!folder.IsValid)
-            {
-                return AbsolutePath.Invalid;
             }
 
             return folder.Combine(Context.PathTable, relativePath);
--- a/Public/Src/Utilities/UnitTests/Utilities/AbsolutePathTestsUnix.cs
+++ b/Public/Src/Utilities/UnitTests/Utilities/AbsolutePathTestsUnix.cs
@@ -41,6 +41,11 @@ namespace Test.BuildXL.Utilities
             XAssert.AreEqual(@"/", p.ToString(pt));
             XAssert.IsTrue(AbsolutePath.TryCreate(pt, @"/usr/etc/aaa/../../../../../../", out p));
             XAssert.AreEqual(@"/", p.ToString(pt));
+            XAssert.IsTrue(AbsolutePath.TryCreate(pt, @"/usr/a/../b.txt", out p));
+            XAssert.AreEqual(@"/usr/b.txt", p.ToString(pt));
+            XAssert.IsTrue(AbsolutePath.TryCreate(pt, @"/usr/a/../../../../b.txt", out p));
+            XAssert.AreEqual(@"/b.txt", p.ToString(pt));
+
             XAssert.IsFalse(AbsolutePath.TryCreate(pt, @"..", out p));
             XAssert.IsFalse(AbsolutePath.TryCreate(pt, @".", out p));
             XAssert.IsFalse(AbsolutePath.TryCreate(pt, @"\", out p));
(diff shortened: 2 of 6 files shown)


Public/Src/Utilities/UnitTests/Utilities/AbsolutePathTestsUnix.cs
```diff
@@ -34,8 +34,13 @@ public void TryCreate()
...
 
-            XAssert.IsFalse(AbsolutePath.TryCreate(pt, @"/..", out p));
-            XAssert.IsFalse(AbsolutePath.TryCreate(pt, @"/../..", out p));
+
+            XAssert.IsTrue(AbsolutePath.TryCreate(pt, @"/..", out p));
+            XAssert.AreEqual(@"/", p.ToString(pt));
+            XAssert.IsTrue(AbsolutePath.TryCreate(pt, @"/../..", out p));
+            XAssert.AreEqual(@"/", p.ToString(pt));
+            XAssert.IsTrue(AbsolutePath.TryCreate(pt, @"/usr/etc/aaa/../../../../../../", out p));
+            XAssert.AreEqual(@"/", p.ToString(pt));
```
dannyvv: Do we really want this behavior? 
I agree on the absolute path case since that matches the win32 api's... But I'm not 100% sure we want this for relative paths, it seems that going out of a relative path should remain an error...

dannyvv: And I should have looked better :D This is unix paths, not relative paths...


---
In reply to: [298701845](https://github.com/microsoft/BuildXL/pull/533#discussion_r298701845) [](ancestors = 298701845)

sahilmgandhi: Indeed, so if the path is passed into the RelativePath's tryCreate directly, it will error out. But if it is coming through an AbsolutePath's tryCreate call, then we allow it to go through since we know that there is a root directory to fall back to.

Public/Src/Utilities/Utilities/RelativePath.cs
```diff
@@ -96,7 +96,7 @@ public static bool TryCreate<T>(StringTable table, T relativePath, out RelativeP
         /// Try to create a RelativePath from a string.
         /// </summary>
         /// <returns>Return the parser result indicating success, or what was wrong with the parsing.</returns>
-        public static ParseResult TryCreate<T>(StringTable table, T relativePath, out RelativePath result, out int characterWithError)
+        public static ParseResult TryCreate<T>(StringTable table, T relativePath, out RelativePath result, out int characterWithError, bool isAbsolute = false)
```
mpysson: > bool isAbsolute = false [](start = 134, length = 24)

The purpose of this isn't obvious when looking at the function signature. Could you please add xml parameter comments for this and the other parameters of this function?

Example:
        /// <param name="pathFormat">Optional override for path format.</param>

Public/Src/Utilities/UnitTests/Utilities/RelativePathTests.cs
```diff
@@ -40,10 +40,10 @@ public void TryCreate()
             XAssert.IsTrue(RelativePath.TryCreate(st, @"AAA\CCC", out p));
             XAssert.AreEqual(@"AAA\CCC", p.ToString(st));
 
+            XAssert.IsFalse(RelativePath.TryCreate(st, @"..", out p));
```
mpysson: >XAssert.IsFalse(RelativePath.TryCreate(st, @"..", out p)); [](start = 12, length = 58)

Is there a change here? Looks like it was just reordered. Please revert if that was just an unintentional change due to debugging

Public/Src/Utilities/Utilities/RelativePath.cs
```diff
@@ -169,14 +169,17 @@ public static ParseResult TryCreate<T>(StringTable table, T relativePath, out Re
                                 || (relativePath[index + 2] == '/'))
                             {
                                 // component is a sole .. so try to go up
-                                if (components.Count == 0)
+                                if (components.Count == 0 && !isAbsolute)
```
mpysson: >isAbsolute [](start = 62, length = 10)

I'm thinking it would be good to rename the isAbsolute parameter to something that more closely describes the behavioral difference of the validation provided by TryCreate. What do you think about "failDotDotOutOfScrope"?

Public/Src/Utilities/UnitTests/Utilities/AbsolutePathTestsUnix.cs
dannyvv: Thanks, sorry for not reading carefully enough :)

Public/Src/Utilities/UnitTests/Utilities/RelativePathTests.cs
sahilmgandhi: Ahh yea looks like I re-ordered it to do a test, and then didn't move it back in the end. Ill change it back

Public/Src/Utilities/Utilities/RelativePath.cs
sahilmgandhi: Sure I will do go ahead and do that.

Public/Src/Utilities/Utilities/RelativePath.cs
sahilmgandhi: I think fromAbsolutePath is more accurate here, since if the components.Count == 0 && the function was not called form AbsolutePath then we should return the error.

olkononenko:
>         public void TryCreate()

Could you add a couple of tests that would cover ".." followed by path elements scenario (both Windows and unix). I'd say we need these two: C:\a\..\b.txt and C:\a\..\..\b.txt


---
Refers to: Public/Src/Utilities/UnitTests/Utilities/AbsolutePathTests.cs:31 in 171fbfc. [](commit_id = 171fbfc6e7d1a156321664c895d945ead02b995e, deletion_comment = False)

Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/LiteralConverter.cs
```diff
@@ -227,13 +227,16 @@ private AbsolutePath ComputeAbsolutePath(AbsolutePath currentFolder, RelativePat
         {
             // parent folder count is a number of ../ expression in the original path literal
             var folder = currentFolder;
+            // If we go beyond the root directory, stay at the root directory. Ie C:/AAA/BBB/../../../../.. should be C:/
+            var tempFolder = currentFolder;
```
olkononenko: > var tempFolder = currentFolder; [](start = 11, length = 32)

I'd move this inside the loop

olkononenko:
>             }

looks like, with your change , you no longer need this check --- (1) `AbsolutePath.Invalid` will never be assigned to `folder`, (2) if `currentFolder` has an invalid value, we'd blow up earlier at `GetParent` call.

---
Refers to: Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/LiteralConverter.cs:245 in 171fbfc. [](commit_id = 171fbfc6e7d1a156321664c895d945ead02b995e, deletion_comment = False)

olkononenko:
>             }

it also might be worth moving the check to the top of the method, i.e., return AP.Invalid if `currentFolder` is invalid.

---
In reply to: [506882183](https://github.com/microsoft/BuildXL/pull/533#issuecomment-506882183) [](ancestors = 506882183)

---
Refers to: Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/LiteralConverter.cs:245 in 171fbfc. [](commit_id = 171fbfc6e7d1a156321664c895d945ead02b995e, deletion_comment = False)

Public/Src/Utilities/Utilities/RelativePath.cs
sahilmgandhi: Changing it to "fromAbsolutePath" in next commit

Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/LiteralConverter.cs
sahilmgandhi: Done

---
In reply to: [298751888](https://github.com/microsoft/BuildXL/pull/533#discussion_r298751888) [](ancestors = 298751888)

Public/Src/Utilities/Utilities/RelativePath.cs
mpysson: Discussed a tweak offline

PR #534, opened 2019-06-28
Fix flaky test ConsumerProducersCanOnlyReadFromProducersInSharedOpaqueWithSameRoot

smera:
Order pips to avoid a race


Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs
```diff
@@ -310,7 +310,7 @@ public void ConsumerProducersCanOnlyReadFromProducersInSharedOpaqueWithSameRoot(
             var pipB = CreateAndScheduleSharedOpaqueProducer(
                 sharedOpaqueDir,
                 fileToProduceStatically: FileArtifact.Invalid,
-                sourceFileToRead: CreateSourceFile(),
+                sourceFileToRead: dummyOutputA,
```
dannyvv: >dummyOutputA [](start = 34, length = 12)

Thanks!

PR #536, opened 2019-06-28
Report violations under shared opaques for writes in existing files

smera:
Shared opaques define a cone where writes are always allowed. The only exception are known dependencies, where those specific paths are blocked. However, this doesn't prevent writes on files that were not created by the writing pip and exist before the pip started. 
This PR addresses that problem by identifying the first write to a file that happens in the context of a pip, and determining whether the file being written existed before the first write happened. 
This new type of violations are only produced under a specific policy flag - for now set only for shared opaques and when undeclared source read mode is on (effectively, only for MSBuild and Ninja). We can consider extending this policy to all shared opaques in general.
For now this is a Windows-specific feature. We can think about adding this for Mac as well.

### DIFF
--- a/Public/Src/Engine/Processes/FileAccessPolicy.cs
+++ b/Public/Src/Engine/Processes/FileAccessPolicy.cs
@@ -85,12 +85,11 @@ namespace BuildXL.Processes
         /// Override writes allowed by policy based on file existence checks. 
         /// </summary>
         /// <remarks>
-        /// Override writes allowed by policy based on file existence checks. 
-        /// Used mainly in the context of shared opaques, where the whole cone under the opaque root is write-allowed by policy (except known inputs).
+        /// Used entirely in the context of shared opaques, where the whole cone under the opaque root is write-allowed by policy (except known inputs).
         /// This policy makes sure that writes on undeclared inputs that fall under the write-allowed cone are flagged as DFAs.
         /// The way to dermine undeclared inputs is based on file existence: if a pip tries to write into a file - allowed by policy - but
-        /// that was not created by the pip (i.e. the file was there before the first write), then it is a write on an undeclared input
-        /// Observe detours never blocks in this case, denying the access is surfaced as a DFA after the write happened.
+        /// that was not created by the pip (i.e. the file was there before the first write attempted by this pip), then it is a write on an undeclared input
+        /// Observe that sandboxing never blocks in this case, denying the access is surfaced as a DFA after the write happened.
         /// </remarks>
         OverrideAllowWriteForExistingFiles = 0x400,
 
--- a/Public/Src/Engine/Processes/FileAccessPolicy.cs
+++ b/Public/Src/Engine/Processes/FileAccessPolicy.cs
@@ -82,6 +82,19 @@ namespace BuildXL.Processes
         AllowRealInputTimestamps = 0x200,
 
         /// <summary>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// </summary>
+        /// <remarks>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// Used mainly in the context of shared opaques, where the whole cone under the opaque root is write-allowed by policy (except known inputs).
+        /// This policy makes sure that writes on undeclared inputs that fall under the write-allowed cone are flagged as DFAs.
+        /// The way to dermine undeclared inputs is based on file existence: if a pip tries to write into a file - allowed by policy - but
+        /// that was not created by the pip (i.e. the file was there before the first write), then it is a write on an undeclared input
+        /// Observe detours never blocks in this case, denying the access is surfaced as a DFA after the write happened.
+        /// </remarks>
+        OverrideAllowWriteForExistingFiles = 0x400,
+
+        /// <summary>
         /// If set, then we will report attempts to access files under this scope, whether they exist or not (combination of <see cref="ReportAccessIfExistent"/>
         /// and <see cref="ReportAccessIfNonexistent"/>).
         /// </summary>
--- a/Public/Src/Engine/Processes/ReportedFileOperation.cs
+++ b/Public/Src/Engine/Processes/ReportedFileOperation.cs
@@ -173,6 +173,13 @@ namespace BuildXL.Processes
         ChangedReadWriteToReadAccess,
 
         /// <summary>
+        /// This is a quazy operation. We issue this only when <see cref="FileAccessPolicy.OverrideAllowWriteForExistingFiles"/> is set, representing
+        /// that an allow for write check was performed for a given path for the first time (in the scope of a process, another process in the same process 
+        /// tree may also report this for the same path)
+        /// </summary>
+        FirstAllowWriteCheckInProcess,
+
+        /// <summary>
         /// Access of reparse point target.
         /// </summary>
         ReparsePointTarget,
--- a/Public/Src/Engine/Processes/ReportedFileOperation.cs
+++ b/Public/Src/Engine/Processes/ReportedFileOperation.cs
@@ -166,14 +166,14 @@ namespace BuildXL.Processes
         ZwOpenFile,
 
         /// <summary>
-        /// This is a quazy operation. We issue this
+        /// This is a quasi operation. We issue this
         /// report when Detours is changing file open
         /// request with Read/Write access to Read access only.
         /// </summary>
         ChangedReadWriteToReadAccess,
 
         /// <summary>
-        /// This is a quazy operation. We issue this only when <see cref="FileAccessPolicy.OverrideAllowWriteForExistingFiles"/> is set, representing
+        /// This is a quasi operation. The sandbox issues this only when <see cref="FileAccessPolicy.OverrideAllowWriteForExistingFiles"/> is set, representing
         /// that an allow for write check was performed for a given path for the first time (in the scope of a process, another process in the same process 
         /// tree may also report this for the same path)
         /// </summary>
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -1798,7 +1798,13 @@ namespace BuildXL.Processes
                             FileAccessPolicy.AllowAll | // Symlink creation is allowed under opaques.
                             FileAccessPolicy.AllowRealInputTimestamps |
                             // For shared opaques, we need to know the (write) accesses that occurred, since we determine file ownership based on that.
-                            (directory.IsSharedOpaque? FileAccessPolicy.ReportAccess : FileAccessPolicy.Deny);
+                            (directory.IsSharedOpaque? FileAccessPolicy.ReportAccess : FileAccessPolicy.Deny) |
+                            // For shared opaques and if allowed undeclared source reads is enabled, make sure that any file used as an undeclared input under the 
+                            // shared opaque gets deny write access. Observe that with exclusive opaques they are wiped out before the pip runs, so it is moot to check for inputs
+                            // TODO: considering configuring this policy for all shared opaques, and not only when AllowedUndeclaredSourceReads is set. The case of a write on an undeclared
+                            // input is more likely to happen when undeclared sources are allowed, but also possible otherwide. For now, this is just a conservative way to try this feature
+                            // out for a subset of our scenarios.
+                            (m_pip.AllowUndeclaredSourceReads && directory.IsSharedOpaque ? FileAccessPolicy.OverrideAllowWriteForExistingFiles : FileAccessPolicy.Deny);
 
                         // For exclusive opaques, we don't need reporting back and the content is discovered by enumerating the disk
                         var mask = directory.IsSharedOpaque ? FileAccessPolicy.MaskNothing : m_excludeReportAccessMask;
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -1802,7 +1802,7 @@ namespace BuildXL.Processes
                             // For shared opaques and if allowed undeclared source reads is enabled, make sure that any file used as an undeclared input under the 
                             // shared opaque gets deny write access. Observe that with exclusive opaques they are wiped out before the pip runs, so it is moot to check for inputs
                             // TODO: considering configuring this policy for all shared opaques, and not only when AllowedUndeclaredSourceReads is set. The case of a write on an undeclared
-                            // input is more likely to happen when undeclared sources are allowed, but also possible otherwide. For now, this is just a conservative way to try this feature
+                            // input is more likely to happen when undeclared sources are allowed, but also possible otherwise. For now, this is just a conservative way to try this feature
                             // out for a subset of our scenarios.
                             (m_pip.AllowUndeclaredSourceReads && directory.IsSharedOpaque ? FileAccessPolicy.OverrideAllowWriteForExistingFiles : FileAccessPolicy.Deny);
 
--- a/Public/Src/Engine/Processes/SandboxedProcessReports.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessReports.cs
@@ -658,7 +658,7 @@ namespace BuildXL.Processes
                 // is reported for a given path, and ignore subsequent reports.
                 // Races are ignored: a race means two child processes are racing to create or delete the same file
                 // - something that is not a good build behavior anyway - and the outcome will be that we will 
-                // non -deterministically deny the access
+                // non-deterministically deny the access
                 if (path != null && !m_overrideAllowedWritePaths.ContainsKey(path))
                 {
                     // We should override write allowed accesses for this path if the status of the special operation was 'denied'
@@ -668,7 +668,7 @@ namespace BuildXL.Processes
                 return true;
             }
 
-            FileAccessStatusMethod method;
+            FileAccessStatusMethod method = FileAccessStatusMethod.PolicyBased;
 
             // If we are processing an allowed write, but this should be overridden based on file existence, 
             // we change the status here
@@ -680,10 +680,6 @@ namespace BuildXL.Processes
             {
                 status = FileAccessStatus.Denied;
                 method = FileAccessStatusMethod.FileExistenceBased;
-            }
-            else
-            {
-                method = FileAccessStatusMethod.PolicyBased;
             }
 
             var reportedAccess =
--- a/Public/Src/Engine/Processes/SandboxedProcessReports.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessReports.cs
@@ -35,6 +35,7 @@ namespace BuildXL.Processes
         private readonly ConcurrentDictionary<uint, ReportedProcess> m_processesExits = new ConcurrentDictionary<uint, ReportedProcess>();
 
         private readonly Dictionary<string, string> m_pathCache = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, bool> m_overrideAllowedWritePaths = new Dictionary<string, bool>();
         private readonly IDetoursEventListener m_detoursEventListener;
 
         public readonly List<ReportedProcess> Processes = new List<ReportedProcess>();
@@ -647,6 +648,44 @@ namespace BuildXL.Processes
                 }
             }
 
+            if (operation == ReportedFileOperation.FirstAllowWriteCheckInProcess)
+            {
+                // This operation represents that a given path was checked for write access for the first time
+                // within the scope of a process. The status of the operation represents whether that access should
+                // have been allowed/denied, based on the existence of the file.
+                // However, we need to determine whether to deny the access based on the first time the path was
+                // checked for writes across the whole process tree. This means checking the first time this operation 
+                // is reported for a given path, and ignore subsequent reports.
+                // Races are ignored: a race means two child processes are racing to create or delete the same file
+                // - something that is not a good build behavior anyway - and the outcome will be that we will 
+                // non -deterministically deny the access
+                if (path != null && !m_overrideAllowedWritePaths.ContainsKey(path))
+                {
+                    // We should override write allowed accesses for this path if the status of the special operation was 'denied'
+                    m_overrideAllowedWritePaths[path] = (status == FileAccessStatus.Denied);
+                }
+
+                return true;
+            }
+
+            FileAccessStatusMethod method;
+
+            // If we are processing an allowed write, but this should be overridden based on file existence, 
+            // we change the status here
+            if (path != null &&
+                (requestedAccess & RequestedAccess.Write) != 0 &&
+                status == FileAccessStatus.Allowed &&
+                m_overrideAllowedWritePaths.TryGetValue(path, out bool shouldOverrideAllowedAccess) &&
+                shouldOverrideAllowedAccess)
+            {
+                status = FileAccessStatus.Denied;
+                method = FileAccessStatusMethod.FileExistenceBased;
+            }
+            else
+            {
+                method = FileAccessStatusMethod.PolicyBased;
+            }
+
             var reportedAccess =
                 new ReportedFileAccess(
                     operation,
@@ -662,7 +701,8 @@ namespace BuildXL.Processes
                     flagsAndAttributes,
                     manifestPath,
                     path,
-                    enumeratePattern);
+                    enumeratePattern,
+                    method);
 
             HandleReportedAccess(reportedAccess);
             return true;
@@ -848,6 +888,7 @@ namespace BuildXL.Processes
                     { "CreateSymbolicLink_Source", ReportedFileOperation.CreateSymbolicLinkSource },
                     { "ReparsePointTarget", ReportedFileOperation.ReparsePointTarget },
                     { "ChangedReadWriteToReadAccess", ReportedFileOperation.ChangedReadWriteToReadAccess },
+                    { "FirstAllowWriteCheckInProcess", ReportedFileOperation.FirstAllowWriteCheckInProcess },
                     { "MoveFileWithProgress_Source", ReportedFileOperation.MoveFileWithProgressSource },
                     { "MoveFileWithProgress_Dest", ReportedFileOperation.MoveFileWithProgressDest },
                     { "MultipleOperations", ReportedFileOperation.MultipleOperations },
--- a/Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/AllowedUndeclaredReadsTests.cs
+++ b/Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/AllowedUndeclaredReadsTests.cs
@@ -1,25 +1,22 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Collections;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Diagnostics.ContractsLight;
 using System.IO;
 using System.Linq;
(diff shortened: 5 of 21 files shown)


Public/Src/Sandbox/Windows/DetoursServices/DetoursHelpers.cpp
```diff
@@ -1025,3 +1025,11 @@ bool EnumerateDirectory(
 
     return true;
 }
+
+bool ExistsAsFile(const wchar_t* path)
+{
+    DWORD dwAttrib = Real_GetFileAttributesW(path);
```
narasamdya: >Real_ [](start = 21, length = 5)

The convention is we should not call Real_* or Detoured_* . We just call the function.

Public/Src/Engine/Processes/FileAccessPolicy.cs
```diff
@@ -81,6 +81,19 @@ public enum FileAccessPolicy : ushort
         /// </remarks>
         AllowRealInputTimestamps = 0x200,
 
+        /// <summary>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// </summary>
+        /// <remarks>
+        /// Override writes allowed by policy based on file existence checks.
```
almiliMSFT: No need to copy this line from `<summary>`
```suggestion
```

Public/Src/Engine/Processes/FileAccessPolicy.cs
```diff
@@ -81,6 +81,19 @@ public enum FileAccessPolicy : ushort
         /// </remarks>
         AllowRealInputTimestamps = 0x200,
 
+        /// <summary>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// </summary>
+        /// <remarks>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// Used mainly in the context of shared opaques, where the whole cone under the opaque root is write-allowed by policy (except known inputs).
```
almiliMSFT: Is it not used `entirely` int eh context of shared opaques?
```suggestion
        /// Used entirely in the context of shared opaques, where the whole cone under the opaque root is write-allowed by policy (except known inputs).
```

Public/Src/Engine/Processes/FileAccessPolicy.cs
```diff
@@ -81,6 +81,19 @@ public enum FileAccessPolicy : ushort
...
+        /// <summary>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// </summary>
+        /// <remarks>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// Used mainly in the context of shared opaques, where the whole cone under the opaque root is write-allowed by policy (except known inputs).
+        /// This policy makes sure that writes on undeclared inputs that fall under the write-allowed cone are flagged as DFAs.
+        /// The way to dermine undeclared inputs is based on file existence: if a pip tries to write into a file - allowed by policy - but
+        /// that was not created by the pip (i.e. the file was there before the first write), then it is a write on an undeclared input
+        /// Observe detours never blocks in this case, denying the access is surfaced as a DFA after the write happened.
```
almiliMSFT:
```suggestion
        /// Observe that sandboxing never blocks in this case, denying the access is surfaced as a DFA after the write happened.
```

Public/Src/Engine/Processes/ReportedFileOperation.cs
```diff
@@ -172,6 +172,13 @@ public enum ReportedFileOperation : byte
         /// </summary>
         ChangedReadWriteToReadAccess,
 
+        /// <summary>
+        /// This is a quazy operation. We issue this only when <see cref="FileAccessPolicy.OverrideAllowWriteForExistingFiles"/> is set, representing
```
almiliMSFT: What's "quazy"?

Public/Src/Sandbox/Windows/DetoursServices/DetoursHelpers.cpp
smera: Will change

---
In reply to: [298778259](https://github.com/microsoft/BuildXL/pull/536#discussion_r298778259) [](ancestors = 298778259)

Public/Src/Engine/Processes/ReportedFileOperation.cs
almiliMSFT: > We issue this
Is the sandbox expected to send this report, or is it the BuildXL engine that figures this out based the "standard" reports from the sandbox?  Please clarify this in the comment

Public/Src/Engine/Processes/FileAccessPolicy.cs
smera: Yeah, guess I was thinking about how this can evolve in the future. WIll change.

---
In reply to: [299145360](https://github.com/microsoft/BuildXL/pull/536#discussion_r299145360) [](ancestors = 299145360)

Public/Src/Engine/Processes/ReportedFileOperation.cs
smera: 'quasi'. Copy paste from the value above

---
In reply to: [299159736](https://github.com/microsoft/BuildXL/pull/536#discussion_r299159736) [](ancestors = 299159736)

smera: Sandbox. Will change

---
In reply to: [299162179](https://github.com/microsoft/BuildXL/pull/536#discussion_r299162179) [](ancestors = 299162179)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/AllowedUndeclaredReadsTests.cs
```diff
@@ -383,6 +383,66 @@ public void ReadViolationsAreDetectedEvenWhenRunningFromCache()
             AssertErrorEventLogged(LogEventId.DependencyViolationWriteInUndeclaredSourceRead);
         }
 
+        /// <summary>
+        /// TODO: blocking writes on existing undeclared inputs is not implemented on Mac yet
+        /// </summary>
+        [FactIfSupported(requiresWindowsBasedOperatingSystem: true)]
```
almiliMSFT: Can you summarize why this functionality needs to be implemented inside the sandbox instead of as part of the engine?

Couldn't you instead enumerate the content of all shared opaque directories (after they've been scrubbed of shared opaque outputs), remember those files as source files, and then inside the engine issue a DFA for every undeclared write to any of those source files?  You could even add those source files explicitly to the FileAccessManifest and disallow writes there.

Public/Src/Sandbox/Windows/DetoursServices/PolicyResult.cpp
```diff
@@ -102,3 +103,51 @@ void PolicyResult::ReportIndeterminatePolicyAndSetLastError(FileOperationContext
...
         -1);
 }
+
+bool PolicyResult::AllowWrite() const {
+
+    bool isWriteAllowedByPolicy = (m_policy & FileAccessPolicy_AllowWrite) != 0;
+
+    // Send a special message to managed code if the policy to override allowed writes based on file existence is set
+    // and the write is allowed by policy (for the latter, if the write is denied, there is nothing to override)
+    if (OverrideAllowWriteForExistingFiles() && isWriteAllowedByPolicy) {
```
almiliMSFT:
```suggestion
    if (isWriteAllowedByPolicy  && OverrideAllowWriteForExistingFiles()) {
```

Public/Src/Sandbox/Windows/DetoursServices/PolicyResult.cpp
```diff
@@ -102,3 +103,51 @@ void PolicyResult::ReportIndeterminatePolicyAndSetLastError(FileOperationContext
...
+                L"FirstAllowWriteCheckInProcess",
+                GENERIC_WRITE,
+                FILE_SHARE_WRITE,
+                OPEN_ALWAYS,
+                FILE_ATTRIBUTE_NORMAL,
+                this->GetCanonicalizedPath().GetPathString());
+
+            ReportFileAccess(
+                operationContext,
+                fileExists? FileAccessStatus_Denied : FileAccessStatus_Allowed,
```
almiliMSFT:
```suggestion
                fileExists ? FileAccessStatus_Denied : FileAccessStatus_Allowed,
```

Public/Src/Engine/Processes/FileAccessPolicy.cs
```diff
@@ -81,6 +81,19 @@ public enum FileAccessPolicy : ushort
...
 
+        /// <summary>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// </summary>
+        /// <remarks>
+        /// Override writes allowed by policy based on file existence checks. 
+        /// Used mainly in the context of shared opaques, where the whole cone under the opaque root is write-allowed by policy (except known inputs).
+        /// This policy makes sure that writes on undeclared inputs that fall under the write-allowed cone are flagged as DFAs.
+        /// The way to dermine undeclared inputs is based on file existence: if a pip tries to write into a file - allowed by policy - but
+        /// that was not created by the pip (i.e. the file was there before the first write), then it is a write on an undeclared input
```
narasamdya: >by the pip [](start = 33, length = 10)

What's the behavior when the file is created by another pip?

narasamdya: >the first write [](start = 76, length = 15)

Is this the first write by the same pip? Or could it be the first write by another pip?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1798,7 +1798,13 @@ private void TryPrepareFileAccessMonitoringForPip(Process pip, HashSet<AbsoluteP
...
                             FileAccessPolicy.AllowRealInputTimestamps |
                             // For shared opaques, we need to know the (write) accesses that occurred, since we determine file ownership based on that.
-                            (directory.IsSharedOpaque? FileAccessPolicy.ReportAccess : FileAccessPolicy.Deny);
+                            (directory.IsSharedOpaque? FileAccessPolicy.ReportAccess : FileAccessPolicy.Deny) |
+                            // For shared opaques and if allowed undeclared source reads is enabled, make sure that any file used as an undeclared input under the 
+                            // shared opaque gets deny write access. Observe that with exclusive opaques they are wiped out before the pip runs, so it is moot to check for inputs
+                            // TODO: considering configuring this policy for all shared opaques, and not only when AllowedUndeclaredSourceReads is set. The case of a write on an undeclared
+                            // input is more likely to happen when undeclared sources are allowed, but also possible otherwide. For now, this is just a conservative way to try this feature
```
narasamdya: >otherwide [](start = 117, length = 9)

otherwise

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1798,7 +1798,13 @@ private void TryPrepareFileAccessMonitoringForPip(Process pip, HashSet<AbsoluteP
                             FileAccessPolicy.AllowAll | // Symlink creation is allowed under opaques.
                             FileAccessPolicy.AllowRealInputTimestamps |
                             // For shared opaques, we need to know the (write) accesses that occurred, since we determine file ownership based on that.
-                            (directory.IsSharedOpaque? FileAccessPolicy.ReportAccess : FileAccessPolicy.Deny);
+                            (directory.IsSharedOpaque? FileAccessPolicy.ReportAccess : FileAccessPolicy.Deny) |
+                            // For shared opaques and if allowed undeclared source reads is enabled, make sure that any file used as an undeclared input under the 
+                            // shared opaque gets deny write access. Observe that with exclusive opaques they are wiped out before the pip runs, so it is moot to check for inputs
+                            // TODO: considering configuring this policy for all shared opaques, and not only when AllowedUndeclaredSourceReads is set. The case of a write on an undeclared
```
narasamdya: >The case of a write [](start = 152, length = 19)

Can you elaborate more (in this comment) why the case of write on undeclared input is more likely to happen when undeclared sources are allowed?

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -647,6 +648,44 @@ private bool FileAccessReportLineReceived<T>(ref T data, FileAccessReportProvide
...
+            {
+                // This operation represents that a given path was checked for write access for the first time
+                // within the scope of a process. The status of the operation represents whether that access should
+                // have been allowed/denied, based on the existence of the file.
+                // However, we need to determine whether to deny the access based on the first time the path was
+                // checked for writes across the whole process tree. This means checking the first time this operation 
+                // is reported for a given path, and ignore subsequent reports.
+                // Races are ignored: a race means two child processes are racing to create or delete the same file
+                // - something that is not a good build behavior anyway - and the outcome will be that we will 
+                // non -deterministically deny the access
```
narasamdya: >non -deterministicall [](start = 19, length = 21)

non-deterministically

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/AllowedUndeclaredReadsTests.cs
smera: Enumerating arbitrary cones before each pip runs is not an option perf-wise. Shared opaques can be arbitrarily big. E.g. for the MSBuild case, shared opaques are usually the entire repo, including a mix of sources and outputs.

---
In reply to: [299174940](https://github.com/microsoft/BuildXL/pull/536#discussion_r299174940) [](ancestors = 299174940)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
smera: Because the mode implies that many legit source files are not declared, so the chances of writing into an existing file that is not caught by the regular manifest is higher.

There is another secondary reason as well, which I didn't think it was worth writing down: by doing it in this way we are technically enabling this option for MSBuild/Ninja only. After giving this some mileage, we could consider extending this for all shared opaques.

---
In reply to: [299183232](https://github.com/microsoft/BuildXL/pull/536#discussion_r299183232) [](ancestors = 299183232)

Public/Src/Engine/Processes/FileAccessPolicy.cs
smera: Before the first write by this pip. I.e. the pip didn't create the file.

---
In reply to: [299179727](https://github.com/microsoft/BuildXL/pull/536#discussion_r299179727) [](ancestors = 299179727)

smera: Same. Under this policy a DFA is generated whenever a pip tries to write on a file that it didn't create itself. That's the shared opaque invariant (all content has rewrite count 1)

---
In reply to: [299179577](https://github.com/microsoft/BuildXL/pull/536#discussion_r299179577) [](ancestors = 299179577)

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -647,6 +648,44 @@ private bool FileAccessReportLineReceived<T>(ref T data, FileAccessReportProvide
...
+            }
+
+            FileAccessStatusMethod method;
+
+            // If we are processing an allowed write, but this should be overridden based on file existence, 
+            // we change the status here
+            if (path != null &&
+                (requestedAccess & RequestedAccess.Write) != 0 &&
+                status == FileAccessStatus.Allowed &&
+                m_overrideAllowedWritePaths.TryGetValue(path, out bool shouldOverrideAllowedAccess) &&
```
narasamdya: >m_overrideAllowedWritePaths [](start = 16, length = 27)

If `TryGetValue` doesn't check for size first, then I suggest that you check `m_overrideAllowedWritePaths.Count != 0`.

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -647,6 +648,44 @@ private bool FileAccessReportLineReceived<T>(ref T data, FileAccessReportProvide
...
+                if (path != null && !m_overrideAllowedWritePaths.ContainsKey(path))
+                {
+                    // We should override write allowed accesses for this path if the status of the special operation was 'denied'
+                    m_overrideAllowedWritePaths[path] = (status == FileAccessStatus.Denied);
+                }
+
+                return true;
+            }
+
+            FileAccessStatusMethod method;
```
narasamdya: >FileAccessStatusMethod method; [](start = 12, length = 30)

`FileAccessStatusMethod method = FileAccessStatusMethod.PolicyBased;`

Thus you can reduce the unnecessary `else { ... }` branch.

Public/Src/Engine/Processes/SandboxedProcessReports.cs
smera: It's a regular dictionary, I'd expect that to cost the same...

---
In reply to: [299188507](https://github.com/microsoft/BuildXL/pull/536#discussion_r299188507) [](ancestors = 299188507)

Public/Src/Sandbox/Windows/DetoursServices/DataTypes.h
```diff
@@ -128,6 +128,13 @@ enum FileAccessPolicy
     // this flag is specified
     FileAccessPolicy_AllowRealInputTimestamps = 0x200,
 
+    // Override writes allowed by policy based on file existence checks. 
+    // Used mainly in the context of shared opaques, where the whole cone under the opaque root is write-allowed by policy (except known inputs).
+    // This policy makes sure that writes on undeclared inputs that fall under the write-allowed cone are flagged as DFAs.
+    // The way to dermine undeclared inputs is based on file existence: if a pip tries to write into a file - allowed by policy - but
```
narasamdya: >dermine [](start = 18, length = 7)

typo: determine

Public/Src/Sandbox/Windows/DetoursServices/DetoursHelpers.h
```diff
@@ -122,6 +122,8 @@ bool EnumerateDirectory(
     bool treatReparsePointAsFile,
     _Inout_ std::vector<std::pair<std::wstring, DWORD>>& filesAndDirectories);
 
+bool ExistsAsFile(const wchar_t* path);
```
narasamdya: >const wchar_t* [](start = 18, length = 14)

`__in PCWSTR path`

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -787,6 +797,11 @@ private void DoWriteFile(string file, string content)
         {
             try
             {
+                if (AdditionalArgs == AllUppercasePath)
+                {
+                    file = file.ToUpperInvariant();
+                }
```
narasamdya: For Windows, it's a bit weird that we have a test that exercises casing because if the file entry (could be a directory name) is already there, then even though we write with all upper case, the directory name is still the one when it was first created.

smera: This is making sure that tracking done on detours side for paths were write allowed are checked is actually case insensitive. E.g. the pip first creates 'FOO.txt', and then writes again in 'foo.txt'. If the tracking was case sensitive we will think that the second write is a first write on an existing file, and therefore deny it.

---
In reply to: [299192998](https://github.com/microsoft/BuildXL/pull/536#discussion_r299192998) [](ancestors = 299192998)

Public/Src/Sandbox/Windows/DetoursServices/FilesCheckedForAccess.cpp
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "FilesCheckedForAccess.h"
+#include "string.h"
+
+FilesCheckedForAccess FilesCheckedForAccess::Create()
+{
+    return FilesCheckedForAccess();
+}
```
narasamdya: Do we need `Create()` ?

smera: Good catch, we don't. Leftover from another impl

---
In reply to: [299195215](https://github.com/microsoft/BuildXL/pull/536#discussion_r299195215) [](ancestors = 299195215)

Public/Src/Engine/Processes/FileAccessPolicy.cs
narasamdya: Can you complete the sentence in the comment "the file was there before the first write by this pip" ?

---
In reply to: [299186944](https://github.com/microsoft/BuildXL/pull/536#discussion_r299186944) [](ancestors = 299186944,299179727)

narasamdya: I like the words "a file that it didn't create itself"

---
In reply to: [299187361](https://github.com/microsoft/BuildXL/pull/536#discussion_r299187361) [](ancestors = 299187361,299179577)

smera: Sure

---
In reply to: [299195913](https://github.com/microsoft/BuildXL/pull/536#discussion_r299195913) [](ancestors = 299195913,299186944,299179727)

smera: Is my own dumb attempt at the Russell's paradox :)

---
In reply to: [299196126](https://github.com/microsoft/BuildXL/pull/536#discussion_r299196126) [](ancestors = 299196126,299187361,299179577)

Public/Src/Sandbox/Windows/DetoursServices/PolicyResult.cpp
```diff
@@ -102,3 +103,51 @@ void PolicyResult::ReportIndeterminatePolicyAndSetLastError(FileOperationContext
...
+    // and the write is allowed by policy (for the latter, if the write is denied, there is nothing to override)
+    if (OverrideAllowWriteForExistingFiles() && isWriteAllowedByPolicy) {
+        
+        // Let's check if this path was already checked for allow writes in this process. Observe this structure lifespan is the same 
+        // as the current process so other child processes won't share it. 
+        // But for the current process it will avoid probing the file system over and over for the same path.
+        FilesCheckedForAccess* filesCheckedForWriteAccess = GetGlobalFilesCheckedForAccesses();
+        
+        if (filesCheckedForWriteAccess->TryRegisterPath(m_canonicalizedPath)) {
+            DWORD error = GetLastError();
```
narasamdya: >DWORD error = GetLastError(); [](start = 12, length = 29)

Why do you need to get the last error here? I think it's just simply ERROR_SUCCESS

Public/Src/Sandbox/Windows/DetoursServices/FilesCheckedForAccess.cpp
```diff
@@ -0,0 +1,47 @@
...
+    std::pair<std::unordered_set<std::wstring>::iterator, bool> result;
+    
+    EnterCriticalSection(&m_lock);
+    result = m_pathSet.insert(path.GetPathString());
+    LeaveCriticalSection(&m_lock);
+
+    return result.second;
+}
+
+bool FilesCheckedForAccess::IsRegistered(const CanonicalizedPath& path) {
```
narasamdya: Where is this used?

Public/Src/Sandbox/Windows/DetoursServices/PolicyResult.cpp
```diff
@@ -102,3 +103,51 @@ void PolicyResult::ReportIndeterminatePolicyAndSetLastError(FileOperationContext
...
+            bool fileExists = ExistsAsFile(m_canonicalizedPath.GetPathString());
+
+            AccessCheckResult accessCheck = AccessCheckResult(RequestedAccess::Read, ResultAction::Allow, ReportLevel::Ignore);
+            FileOperationContext operationContext(
+                L"FirstAllowWriteCheckInProcess",
+                GENERIC_WRITE,
+                FILE_SHARE_WRITE,
+                OPEN_ALWAYS,
+                FILE_ATTRIBUTE_NORMAL,
+                this->GetCanonicalizedPath().GetPathString());
```
narasamdya: I think this should be
```
FileOperationContext::CreateForRead(L""FirstAllowWriteCheckInProcess", this->GetCanonicalizedPath().GetPathString());
```

Public/Src/Sandbox/Windows/DetoursServices/PolicyResult.cpp
smera: What I'm after is so last error is preserved. The previous implementation of AllowWrite didn't do any OS calls, so I was not sure if any caller was assuming that AllowWrite didn't change last error. So just making sure ExistsAsFile is not changing that value.

---
In reply to: [299198238](https://github.com/microsoft/BuildXL/pull/536#discussion_r299198238) [](ancestors = 299198238)

Public/Src/Sandbox/Windows/DetoursServices/FilesCheckedForAccess.cpp
smera: Nowhere as of now. Just added it for completeness.

---
In reply to: [299198574](https://github.com/microsoft/BuildXL/pull/536#discussion_r299198574) [](ancestors = 299198574)

Public/Src/Sandbox/Windows/DetoursServices/PolicyResult.cpp
smera: Cool, didn't know we had that. Will change.

---
In reply to: [299200131](https://github.com/microsoft/BuildXL/pull/536#discussion_r299200131) [](ancestors = 299200131)

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
narasamdya: Can this kind of test be done in SandboxedProcessPipExecutorTest? 

---
In reply to: [299194088](https://github.com/microsoft/BuildXL/pull/536#discussion_r299194088) [](ancestors = 299194088,299192998)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/AllowedUndeclaredReadsTests.cs
almiliMSFT: I meant: enumerate them once after scrubbing (for scrubbing you have to enumerate them anyway). 

Enumerating once before the build instead of before every pip should be enough, because, I believe, we already detect the case when 2 different pips end up writing the same undeclared file (via their respective shared opaque directories)

smera: Medium-term we want to get rid of scrubbing for shared opaques (or at least minimize it to some unavoidable set of cases). E.g. when Helium is present, scrubbing can be removed altogether. So I wouldn't like to pile up on that feature. 
I'm also concerned about generating a 'preventive' detours manifest for every pip that is probably going to be many orders of magnitude bigger than what we have today. But still my main reason for not going that route is the first argument.

---
In reply to: [299208300](https://github.com/microsoft/BuildXL/pull/536#discussion_r299208300) [](ancestors = 299208300)

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
smera: I think that's possible, but a little bit harder to implement. I'm currently using Operation, which is pretty handy, and the corresponding test (WritingToExistentFileProducedBySamePipIsAllowed) complements some aspects of other side-by-side tests, that require the integration test machinery. So unless you are strong about it, I'd prefer to leave it as is.

---
In reply to: [299202987](https://github.com/microsoft/BuildXL/pull/536#discussion_r299202987) [](ancestors = 299202987,299194088,299192998)

PR #539, opened 2019-07-01
Fix update of last processed sequence point in event store.

ljcollins25:


Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -269,15 +260,15 @@ async Task<SendToActionBlockResult> sendToActionBlockAsync()
                 foreach (var messageGroup in messages.GroupBy(GetProcessingIndex))
                 {
                     var eventProcessingBlock = _eventProcessingBlocks[messageGroup.Key];
-                    var input = ProcessEventsInput.Create(context, messageGroup);
+                    var input = new ProcessEventsInput(state, messageGroup);
                     bool success = await eventProcessingBlock.SendAsync(input);
                     if (!success)
                     {
+                        // NOTE: This case should not actually occur.
+                        // Complete the operation in case we couldn't send to the action block to prevent pending event queue from getting backlogged.
```
dcwuser: Assert? Is it better that we die and notice this than that we continue without noticing?

ljcollins25: I think this might happen during Shutdown actually. Under that assumption, crashing is not ideal.

PR #543, opened 2019-07-01
Content Repudiation in FileSystemContentStoreInternal

kewilli:


Public/Src/Cache/ContentStore/Interfaces/Stores/IContentStore.cs
```diff
@@ -28,6 +29,11 @@ public interface IContentStore : IStartupShutdown
         ///     Gets a current stats snapshot.
         /// </summary>
         Task<GetStatsResult> GetStatsAsync(Context context);
+
+        /// <summary>
+        ///     Remove given content from all sessions.
+        /// </summary>
+        Task<DeleteResult> DeleteAsync(Context context, ContentHash contentHash);
     }
```
dcwuser: I had assumed this would go on Session rather that Store, but one could argue that it should count as a session-less operation and should therefore go here. Are you making that argument or did it end up here by accident?

Public/Src/Cache/ContentStore/Interfaces/Results/DeleteResult.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="DeleteResult"/> class.
+        /// </summary>
+        public DeleteResult(ContentHash contentHash, long evictedSize, long pinnedSize, long replicaCount)
+        {
+            ContentHash = contentHash;
+            EvictedSize = evictedSize;
+            PinnedSize = pinnedSize;
+            ReplicaCount = replicaCount;
+        }
```
dcwuser: I believe ReplicaCount is a fully internal concept. Not sure we want to surface it to deleters, or if there is a reason we should.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -2074,6 +2074,13 @@ public Task<EvictResult> EvictAsync(Context context, ContentHashWithLastAccessTi
             return EvictCoreAsync(context, contentHashInfo, force: false, onlyUnlinked, evicted);
         }
 
+        /// <inheritdoc />
+        public async Task<DeleteResult> DeleteAsync(Context context, ContentHash contentHash)
+        {
```
dcwuser: Should this be a method on IContentStoreInternal?

Public/Src/Cache/ContentStore/Library/Stores/ReadOnlyEmptyContentStore.cs
```diff
@@ -37,6 +37,9 @@ public class ReadOnlyEmptyContentStore : StartupShutdownBase, IContentStore
 
         /// <inheritdoc />
         public Task<GetStatsResult> GetStatsAsync(Context context) => Task.FromResult(new GetStatsResult(_tracer.GetCounters()));
+
+        /// <inheritdoc />
+        Task<DeleteResult> IContentStore.DeleteAsync(Context context, ContentHash contentHash) => throw new System.NotImplementedException();
     }
```
dcwuser: Should actually be straightforward to implement on this: always false.

Public/Src/Cache/ContentStore/Library/Tracing/EvictResult.cs
```diff
@@ -91,6 +92,23 @@ public EvictResult(ContentHashWithLastAccessTimeAndReplicaCount contentHash, lon
         /// </summary>
         public DateTime CreationTime { get; }
 
+        /// <summary>
+        ///     Convert to the <see cref="DeleteResult"/> class.
+        /// </summary>
+        public DeleteResult ToDeleteResult(ContentHash contentHash)
+        {
```
dcwuser: I would argue that this conversion method more naturally belongs on DeleteResult, because DeleteResult depends on EvictResult, not vice-versa.

Public/Src/Cache/ContentStore/Test/Stores/TestFailingContentStore.cs
```diff
@@ -64,5 +65,11 @@ public Task<GetStatsResult> GetStatsAsync(Context context)
         {
             return Task.FromResult(new GetStatsResult(FailureMessage));
         }
+
+        /// <inheritdoc />
+        public Task<DeleteResult> DeleteAsync(Context context, ContentHash contentHash)
+        {
```
dcwuser: This is another one that is straightforward to implement: always error.

Public/Src/Cache/DistributedCache.Host/Service/Internal/MultiplexedContentStore.cs
```diff
@@ -337,5 +337,11 @@ public async Task<FileExistenceResult> CheckFileExistsAsync(Context context, Con
 
             return fileExistenceResult ?? new FileExistenceResult(FileExistenceResult.ResultCode.Error, $"Could not find a content store which implements {nameof(IStreamStore)} in {nameof(MultiplexedContentStore)}.");
         }
+
+        /// <inheritdoc />
+        public Task<DeleteResult> DeleteAsync(Context context, ContentHash contentHash)
```
dcwuser: Another straightforward implementation: call Delete on each content store.

Public/Src/Cache/ContentStore/Interfaces/Stores/IContentStore.cs
kewilli: Yes, I am making this argument. If our intent is to remove content from a _machine_, then our most base implementation should not be for a session.

Public/Src/Cache/ContentStore/Library/Tracing/EvictResult.cs
kewilli: I would agree, but `DeleteResult` must be defined in Interfaces with `IContentStore`, while `EvictResult` currently lives in Library. It seems dangerous to move `EvictResult` from Library to Interfaces, but I can give it a try if needed.

Another option would be to move this converter into its own extension method in Library, but I think that's even more messy than defining it in `EvictResult`. Thoughts?

---
In reply to: [299291008](https://github.com/microsoft/BuildXL/pull/543#discussion_r299291008) [](ancestors = 299291008)

PR #547, opened 2019-07-02
Update enconding of script files to be UTF-8 instead of UTF-8-BOM

edgarrs:


Shared/Scripts/BuildXLLkgVersion.cmd
```diff
@@ -1,4 +1,3 @@
-﻿REM @echo off
```
almiliMSFT: I think the script in our release pipeline that automatically updates these LKG version files is the one that adds BOM.  Did you update those?

edgarrs: Good catch, I have not

dannyvv: Edgar has updated the CI scripts, so it should be good from now on.

PR #548, opened 2019-07-03
Force server redeployment on pipe exceptions

Kr1sso:
Poison server deployment on broken reader pipe exceptions and to force a re-deployment to eventually fix binary issues.

### DIFF
--- a/Public/Src/App/Bxl/AppServer.cs
+++ b/Public/Src/App/Bxl/AppServer.cs
@@ -981,7 +981,7 @@ namespace BuildXL
                                         // Make sure to poison the server deployment if an exception is caught, this forces the server bits
                                         // to be re-deployed and helps clean up a broken deployment state.
                                         ServerDeployment.PoisonServerDeployment(serverDeploymentDirectory);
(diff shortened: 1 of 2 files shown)


Public/Src/App/Bxl/AppServer.cs
```diff
@@ -976,14 +977,22 @@ internal Connection(NamedPipeClientStream clientStream, IConsole console)
                                             }
                                         }
                                     },
-                                    ex => { throw new BuildXLException("Error while reading data from server process. Inner exception reason: " + ex?.Message, ex); });
+                                    ex => {
+                                        // Make sure to poison the server deployment if an exception is caught, this forces the server bits
+                                        // to be re-deployed and helps clean up a broken deployment state.
+                                        ServerDeployment.PoisonServerDeployment(serverDeploymentDirectory);
+                                        throw new BuildXLException("Error while reading data from server process. Inner exception reason: " + ex?.Message, ex);
```
mpysson:
```suggestion
                                        throw new BuildXLException("Error while reading data from server process. Server will be redeployed in subsequent invocation. Inner exception reason: " + ex?.Message, ex);
```

PR #549, opened 2019-07-04
Check source file consistency after materialization to correctly handling hashing with directory symlinks.

ljcollins25:


Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -1541,18 +1553,22 @@ public static bool CreateSymlinkEagerly(LoggingContext loggingContext, IConfigur
...
-                return ArtifactMaterializationResult.DeleteFilesRequiredAbsentFailed;
-            }
-
             // Place Files:
             if (!await PlaceFilesAsync(operationContext, pipInfo, state))
             {
                 return ArtifactMaterializationResult.PlaceFileFailed;
             }
 
+            if (IsDistributedWorker)
```
narasamdya: >IsDistributedWorker [](start = 16, length = 19)

Should we also say `!SourceFileMaterializationEnabled` ?

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -1531,8 +1531,20 @@ public static bool CreateSymlinkEagerly(LoggingContext loggingContext, IConfigur
...
+                return ArtifactMaterializationResult.DeleteFilesRequiredAbsentFailed;
+            }
+
+            if (IsDistributedWorker && SourceFileMaterializationEnabled)
+            {
+                // NOTE: In the typical case (SourceFileMaterializationEnabled=false), we only verify source files after
+                // placement to ensure directory symlinks which can make outputs appear as source files work. This relies
+                // on the logic that pips which depend on the source file version of a file under a directory symlink also
+                // have a dependency on the corresponding file at its real location where it appears as an output and therefore
+                // will be materialized during PlaceFilesAsync.
```
narasamdya: This need a better comment. Perhaps provide an example to make the scenario easily understandable.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/DirectorySymlinkTests.cs
```diff
@@ -643,6 +643,44 @@ public void EnumerateDirectoryViaDirectorySymlinkShouldBeObservedAsDirectoryEnum
             RunScheduler().AssertCacheMiss(processWithOutputs.Process.PipId);
         }
 
+        /// <summary>
+        /// Creates a shared opaque directory producer & consumer and verifies their usage and caching behavior
+        /// </summary>
+        [FactIfSupported(requiresSymlinkPermission: true)]
+        public void PipDependsOnFileAndSourceFileAlternatePathUnderDirectorySymlink()
```
narasamdya: >PipDependsOnFileAndSourceFileAlternatePathUnderDirectorySymlink [](start = 20, length = 63)

I'm not so sure how this test relates to your change or to the scenario we're dealing with. It doesn't seem to introduce any directory symlink. Also, the change only takes effect if the process has `IsDistributedWorker`, no?

PR #550, opened 2019-07-04
Increase copy delay for "bad" sources

kewilli:
In some cases, multiple CASaaS instances can overwhelm a single FsServer with too many requests for content (>2000 active requests). In this case, the FsServer client will catch an FsServerException and return a CopyFileResult with a result of SourcePathError. If all locations have this issue (which should only be common in the case where a single machine has new content which is needed by all non-master machines), then we can avoid an immediate retry and give the source machine a chance to work through its backlog. This also adds some randomness to the delay in an attempt to spread out the requests.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
@@ -129,6 +129,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.Stores
 
             try
             {
+                Random r = new Random();
                 PutResult putResult = null;
                 var badContentLocations = new HashSet<MachineLocation>();
                 var missingContentLocations = new HashSet<MachineLocation>();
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -171,7 +171,16 @@ protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
                     }
 
                     TimeSpan waitDelay = _retryIntervals[attemptCount];
-                    Tracer.Warning(operationContext, $"{AttemptTracePrefix(attemptCount)} All replicas {hashInfo.Locations.Count} failed. Retrying for hash {hashInfo.ContentHash}...");
+                    bool extendedForBadLocations = false;
+
+                    if (badContentLocations.Count == hashInfo.Locations.Count)
+                    {
+                        // Double delay because all locations are bad. This allows the machines more time to recover, especially if they are overwhelmed by requests.
+                        waitDelay += TimeSpan.FromTicks((long)(waitDelay.Ticks * ThreadSafeRandom.Generator.NextDouble()));
```
jbayardo: In the first few iterations, this will actually wait for -strictly- less than we would have originally, which would only make the problem worse. I'm not sure how this interacts with the other behaviors of this object, but I'm thinking `attemptCount = 2;` should work? so as to enter the exponential decay behavior directly.

kewilli: Wouldn't this change increase `waitDelay` by to up-to double the current delay? Note the `+=` operator.

I'm worried that increasing `attemptCount` would skip useful retries.

kewilli: After discussing offline, I agree that reducing the delay in this case is acceptable. The new delay's randomized distribution is `[0.5 * delay, delay)`.

Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -171,7 +171,16 @@ protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
                     }
 
                     TimeSpan waitDelay = _retryIntervals[attemptCount];
-                    Tracer.Warning(operationContext, $"{AttemptTracePrefix(attemptCount)} All replicas {hashInfo.Locations.Count} failed. Retrying for hash {hashInfo.ContentHash}...");
+                    bool extendedForBadLocations = false;
+
+                    if (badContentLocations.Count == hashInfo.Locations.Count)
+                    {
+                        // Double delay because all locations are bad. This allows the machines more time to recover, especially if they are overwhelmed by requests.
```
jbayardo: Comment is out of date

PR #551, opened 2019-07-04
Fix errors getting into custom logs

olkononenko:
This PR changes the behavior of /customlog flag. Now only the event specifically listed will be included in the custom log (previously such logs also included all Error level events).

This PR does not change dev.log, i.e., it stil includes all error level events.

[AB#1558752](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1558752)

### DIFF
--- a/Public/Src/Utilities/Configuration/ILoggingConfiguration.cs
+++ b/Public/Src/Utilities/Configuration/ILoggingConfiguration.cs
@@ -118,7 +118,7 @@ namespace BuildXL.Utilities.Configuration
         /// Creates a custom log file for a specific set of event IDs. Event list should be comma separated integers excluding the DX prefix.
         /// </summary>
         [NotNull]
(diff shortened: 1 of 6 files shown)


Public/Src/Utilities/Configuration/ILoggingConfiguration.cs
```diff
@@ -118,7 +118,7 @@ public interface ILoggingConfiguration : IWarningHandling
         /// Creates a custom log file for a specific set of event IDs. Event list should be comma separated integers excluding the DX prefix.
```
mpysson: >Event list should be comma separated integers excluding the DX prefix. [](start = 71, length = 70)

This doc looks like its stale from CustomLog being a string. Could you update and include what EventLevel should be set to & what it means if unspecified?

PR #552, opened 2019-07-05
Outputs order handling

norahuang:
This is a proposed solution to fix the reordering issue of the dependencies and outputs.

The outputs are sorted in fingerprint computing. The original order in the spec doesn't matter. So when you change only the dependencies' or/and outputs' orders you will still get cache hit.
However, the dependencies and outputs keeps their original order in the metadata. So when the cache hit pips try to retrieve the output/dependencies contents from metadata by the new spec ordering, they will be messed up and fail the build.

In this PR I updated the PipCacheDescriptor.bond, adding the absolutepath for the StaticOutputHashes. So when we retrieve the metadata, the corresponding outputartifact is going to be associated with the absolutepath instead of index.

[AB#1537359](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1537359)

### DIFF
--- a/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
+++ b/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
@@ -254,11 +254,11 @@ namespace BuildXL.Scheduler.Fingerprints
             AddFileOutput(fingerprinter, "StandardError", process.StandardError);
             AddFileOutput(fingerprinter, "StandardOutput", process.StandardOutput);
 
-            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f), m_expandedPathFileArtifactComparer);            
-            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d), m_directoryComparer);
+            fingerprinter.AddCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f));            
+            fingerprinter.AddCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d));
                         
-            fingerprinter.AddOrderIndependentCollection<FileArtifactWithAttributes, ReadOnlyArray<FileArtifactWithAttributes>>("Outputs", process.FileOutputs, (fp, f) => AddFileOutput(fp, f), m_expandedPathFileArtifactWithAttributesComparer);
-            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryOutputs", process.DirectoryOutputs, (h, p) => h.Add(p.Path), m_directoryComparer);
+            fingerprinter.AddCollection<FileArtifactWithAttributes, ReadOnlyArray<FileArtifactWithAttributes>>("Outputs", process.FileOutputs, (fp, f) => AddFileOutput(fp, f));
+            fingerprinter.AddCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryOutputs", process.DirectoryOutputs, (h, p) => h.Add(p.Path));
 
             fingerprinter.AddOrderIndependentCollection<AbsolutePath, ReadOnlyArray<AbsolutePath>>("UntrackedPaths", process.UntrackedPaths, (h, p) => h.Add(p), m_pathTable.ExpandedPathComparer);
             fingerprinter.AddOrderIndependentCollection<AbsolutePath, ReadOnlyArray<AbsolutePath>>("UntrackedScopes", process.UntrackedScopes, (h, p) => h.Add(p), m_pathTable.ExpandedPathComparer);
--- a/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
+++ b/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
@@ -254,11 +254,11 @@ namespace BuildXL.Scheduler.Fingerprints
             AddFileOutput(fingerprinter, "StandardError", process.StandardError);
             AddFileOutput(fingerprinter, "StandardOutput", process.StandardOutput);
 
-            fingerprinter.AddCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f));            
-            fingerprinter.AddCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d));
-                        
-            fingerprinter.AddCollection<FileArtifactWithAttributes, ReadOnlyArray<FileArtifactWithAttributes>>("Outputs", process.FileOutputs, (fp, f) => AddFileOutput(fp, f));
-            fingerprinter.AddCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryOutputs", process.DirectoryOutputs, (h, p) => h.Add(p.Path));
+            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f), m_expandedPathFileArtifactComparer);            
+            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d), m_directoryComparer);
+
+            fingerprinter.AddOrderIndependentCollection<FileArtifactWithAttributes, ReadOnlyArray<FileArtifactWithAttributes>>("Outputs", process.FileOutputs, (fp, f) => AddFileOutput(fp, f), m_expandedPathFileArtifactWithAttributesComparer);
+            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryOutputs", process.DirectoryOutputs, (h, p) => h.Add(p.Path), m_directoryComparer);
 
             fingerprinter.AddOrderIndependentCollection<AbsolutePath, ReadOnlyArray<AbsolutePath>>("UntrackedPaths", process.UntrackedPaths, (h, p) => h.Add(p), m_pathTable.ExpandedPathComparer);
             fingerprinter.AddOrderIndependentCollection<AbsolutePath, ReadOnlyArray<AbsolutePath>>("UntrackedScopes", process.UntrackedScopes, (h, p) => h.Add(p), m_pathTable.ExpandedPathComparer);
--- a/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
+++ b/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
@@ -259,7 +259,7 @@ namespace BuildXL.Scheduler.Fingerprints
 
             fingerprinter.AddOrderIndependentCollection<FileArtifactWithAttributes, ReadOnlyArray<FileArtifactWithAttributes>>("Outputs", process.FileOutputs, (fp, f) => AddFileOutput(fp, f), m_expandedPathFileArtifactWithAttributesComparer);
             fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryOutputs", process.DirectoryOutputs, (h, p) => h.Add(p.Path), m_directoryComparer);
-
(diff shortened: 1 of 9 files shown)


mpysson:
> 2\. only has performance decrease impact when outputs/dependencies get reordered in the specs, which is rare.

I'm not sure how true this is with Office and WDG having auto-generated spec files

mpysson:
> I have two options to fix this problem:

mpysson:
> 2\. add sorting when storing and retrieving the metadata. This solution require clean up of the old caches, otherwise it will cause potentially all build with cache hit fail. Since with the new sorting in retrieve cache, the output/dependencies will be messed up.

Indeed it looks like PipExecutor.TryCreatePipCacheDescriptorFromMetadata() is assuming the current pip's Outputs order is the same as the order in the PipCacheDescriptorV2Metadata retrieved from cache. Do you think we could get away with just sorting when PipCacheDescriptorV2Metadata is stored? That keeps the cache hit path faster without adding sorting there. We would have to bump the fingerprint version to do a one-time eviction since we cannot presume metadata created with prior versions is sorted. But one-time evictions aren't a huge deal.

norahuang:
> 2. add sorting when storing and retrieving the metadata. This solution require clean up of the old caches, otherwise it will cause potentially all build with cache hit fail. Since with the new sorting in retrieve cache, the output/dependencies will be messed up.
> 
> Indeed it looks like PipExecutor.TryCreatePipCacheDescriptorFromMetadata() is assuming the current pip's Outputs order is the same as the order in the PipCacheDescriptorV2Metadata retrieved from cache. Do you think we could get away with just sorting when PipCacheDescriptorV2Metadata is stored? That keeps the cache hit path faster without adding sorting there. We would have to bump the fingerprint version to do a one-time eviction since we cannot presume metadata created with prior versions is sorted. But one-time evictions aren't a huge deal.

We will need to sort both PipCacheDescriptorV2Metadata and pip.outputs to make them align.

mpysson:
> > 1. add sorting when storing and retrieving the metadata. This solution require clean up of the old caches, otherwise it will cause potentially all build with cache hit fail. Since with the new sorting in retrieve cache, the output/dependencies will be messed up.
> > 
> > Indeed it looks like PipExecutor.TryCreatePipCacheDescriptorFromMetadata() is assuming the current pip's Outputs order is the same as the order in the PipCacheDescriptorV2Metadata retrieved from cache. Do you think we could get away with just sorting when PipCacheDescriptorV2Metadata is stored? That keeps the cache hit path faster without adding sorting there. We would have to bump the fingerprint version to do a one-time eviction since we cannot presume metadata created with prior versions is sorted. But one-time evictions aren't a huge deal.
> 
> We will need to sort both PipCacheDescriptorV2Metadata and pip.outputs to make them align.

You're right. Maybe option 1 is the way to go then. The fact that we only very infrequently fail this validation means we aren't really taking advantage of getting cache hits on reordering the definition of outputs.

Another approach would be to save the relative file paths in the metadata instead of just the file path if we wanted to uphold hits on reordering without having to do sorting.

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -254,11 +254,11 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process
             AddFileOutput(fingerprinter, "StandardError", process.StandardError);
             AddFileOutput(fingerprinter, "StandardOutput", process.StandardOutput);
 
-            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f), m_expandedPathFileArtifactComparer);            
-            fingerprinter.AddOrderIndependentCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d), m_directoryComparer);
+            fingerprinter.AddCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f));            
+            fingerprinter.AddCollection<DirectoryArtifact, ReadOnlyArray<DirectoryArtifact>>("DirectoryDependencies", process.DirectoryDependencies, (fp, d) => AddDirectoryDependency(fp, d));
```
mpysson: Do we actually have an order dependency on this?

It would be good to add some tests for these reordering scenarios at the SchedulerIntegrationTest level since the PipFingerprinter tests don't cover cache replay

norahuang: I will add an integration test

narasamdya:
This change reverts the change that we made some time ago where we need order-independent addition because we keep having cache misses (due to different weak fingerprints) although no changes to the pip have been made.

Note that we populate dependencies and outputs using set, and set order is not guarantee to be stable. 

The proper fix is to not rely on indices when replaying from the cache, but associate which paths where we need to reply from the cache. This requires some changes in the pip cache metadata. 

In anyway, after the change, one needs to bump up the pip fingerprint version to avoid using old cache data.

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -92,7 +92,7 @@ public void ProcessStaticFingerprinting()
                 (fingerprinter, pip) => fingerprinter.ComputeWeakFingerprint(pip));
         }
 
-        [Fact]
+        //[Fact]
```
dannyvv: >//[ [](start = 8, length = 3)

internional? If so, perhaps a:
`[Fact(Skip = ".....here's why...")]`
would be usefull

dannyvv: ignore, should have read pr message first :)

---
In reply to: [301231301](https://github.com/microsoft/BuildXL/pull/552#discussion_r301231301) [](ancestors = 301231301)

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -2851,21 +2851,12 @@ private static int GetCacheLevelForLocality(PublishedEntryRefLocality locality)
...
             {
-                FileArtifactWithAttributes attributedOutput = pip.Outputs[i];
-                if (!attributedOutput.CanBeReferencedOrCached())
-                {
-                    // Skipping non-cacheable outputs (note that StoreContentForProcess does the same).
-                    continue;
-                }
-
+                FileMaterializationInfo materializationInfo = metadata.StaticOutputHashes[i].Info.ToFileMaterializationInfo(pathTable);
+                FileArtifactWithAttributes attributedOutput = pip.Outputs.Where(o => AbsolutePath.Create(pathTable, metadata.StaticOutputHashes[i].AbsolutePath) == o.Path).Single();
```
narasamdya: >pip.Outputs.Where(o => AbsolutePath.Create(pathTable, metadata.StaticOutputHashes[i].AbsolutePath) == o.Path).Single(); [](start = 62, length = 119)

The use of `Where` makes the loop quadratic. You may want to create a transient mapping from paths to output file artifacts before the loop, and thus you can do look-up cheaply.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -2851,21 +2851,12 @@ private static int GetCacheLevelForLocality(PublishedEntryRefLocality locality)
...
-            for (int i = 0; i < pip.Outputs.Length; i++)
+            for (int i = 0; i < metadata.StaticOutputHashes.Count; i++)
             {
-                FileArtifactWithAttributes attributedOutput = pip.Outputs[i];
-                if (!attributedOutput.CanBeReferencedOrCached())
-                {
-                    // Skipping non-cacheable outputs (note that StoreContentForProcess does the same).
-                    continue;
-                }
-
```
narasamdya: Why is the skip output not preserved in the new code?

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -2851,21 +2851,12 @@ private static int GetCacheLevelForLocality(PublishedEntryRefLocality locality)
                 new List<(FileArtifact, FileMaterializationInfo)>(pip.Outputs.Length);
 
             // Outputs should be the same as what was in the metadata section.
-            int outputHashIndex = 0;
-            for (int i = 0; i < pip.Outputs.Length; i++)
+            for (int i = 0; i < metadata.StaticOutputHashes.Count; i++)
```
narasamdya: >metadata.StaticOutputHashes.Count [](start = 32, length = 33)

Please add a unit test where outputs are ordered differently.

Public/Src/Engine/Scheduler/PipExecutor.cs
norahuang: Since I changed the for loop. Instead of looping outputs, I loop the metadata.StaticOutputHashes, which are all cacheable outputs. so I don't need to preserver this skip.

Public/Src/Engine/Scheduler/PipExecutor.cs
norahuang: done

Public/Src/Engine/Scheduler/PipExecutor.cs
norahuang: made the change, thanks.

narasamdya: Can we have the dictionary be placed into ObjectPool? See [http://index/?leftProject=BuildXL.Utilities&leftSymbol=ejggwqg9vtu5&file=Pools.cs](http://index/?leftProject=BuildXL.Utilities&leftSymbol=ejggwqg9vtu5&file=Pools.cs)

---
In reply to: [301331607](https://github.com/microsoft/BuildXL/pull/552#discussion_r301331607) [](ancestors = 301331607)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/OutputReorderTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        [Fact]
+        public void ReorderOutputsCacheHit()
+        {
+            FileArtifact outputFile1 = CreateSourceFile();
+            FileArtifact outputFile2 = CreateSourceFile();
+
+            var builderA = CreatePipBuilder(new Operation[]{});
+            builderA.AddOutputFile(outputFile1, FileExistence.Optional);
```
narasamdya: >FileExistence.Optional [](start = 48, length = 22)

Why optional?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/OutputReorderTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+        }
+
+        [Fact]
+        public void ReorderOutputsCacheHit()
+        {
+            FileArtifact outputFile1 = CreateSourceFile();
+            FileArtifact outputFile2 = CreateSourceFile();
+
+            var builderA = CreatePipBuilder(new Operation[]{});
```
narasamdya: >Operation[]{}) [](start = 48, length = 14)

Can we not have a real write operation that write outputs to files?

norahuang: yes, I have that in my test, but it turns out to be doesn't affect the running the test, so I eliminate it to simplify the case. I will add it back.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/OutputReorderTests.cs
norahuang: fixed.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/OutputReorderTests.cs
norahuang: done

Public/Src/Engine/Scheduler/PipExecutor.cs
norahuang: made the change.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/OutputReorderTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            var builderB = CreatePipBuilder(opsB);
+            builderB.AddOutputFile(outputFile2, FileExistence.Required);
+            builderB.AddOutputFile(outputFile1, FileExistence.Required);
+
+            var pipB = SchedulePipBuilder(builderB).Process;
+            RunScheduler().AssertCacheHit(pipB.PipId);
+        }
+
+
```
narasamdya: Nit: Extra space

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/OutputReorderTests.cs
```diff
@@ -0,0 +1,63 @@
...
+                Operation.ReadFile(inputFile),
+                Operation.WriteFile(outputFile1, doNotInfer: true),
+                Operation.WriteFile(outputFile2, doNotInfer: true),
+            };
+
+            var builderB = CreatePipBuilder(opsB);
+            builderB.AddOutputFile(outputFile2, FileExistence.Required);
+            builderB.AddOutputFile(outputFile1, FileExistence.Required);
+
+            var pipB = SchedulePipBuilder(builderB).Process;
```
narasamdya: I would suggest that the test should also include proper cache replay. I mean, before the second build with `opsB` begins, you should delete `outputFile1` and `outputFile2` from disk. Then, on reply, ensure that the contents you get are what you expect.

norahuang: fixed

PR #553, opened 2019-07-05
QTest pips are being counted twice in PipCounters

nayanshah:
Default tags array was merged twice adding duplicate telemetry tags.
This resulted in QTest pips being counted twice in all counters.


nayanshah:
@dannyvv

narasamdya:
"Default tags array was merged twice adding duplicate telemetry tags" -- should the correct fix be deduping the telemetry tags?

narasamdya:
Approving this change, given that Nayan is going to do the deduping the telemetry tags in BuildXL code.

dannyvv:
Manually running the checks for this PR:

PR #556, opened 2019-07-05
Check if the writer is null

norahuang:
[AB#1563133](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1563133)

### DIFF
--- a/Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs
+++ b/Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs
@@ -134,7 +134,7 @@ namespace BuildXL.Utilities.Tracing
         /// <inheritdoc />
         protected override void Output(EventLevel level, int id, string eventName, EventKeywords eventKeywords, string text, bool doNotTranslatePaths = false)
         {
-            m_writer.WriteLine(level, (m_translator != null && !doNotTranslatePaths) ? m_translator.Translate(text) : text);
+            m_writer?.WriteLine(level, (m_translator != null && !doNotTranslatePaths) ? m_translator.Translate(text) : text);
 
             // At the expense of performance, flush Critical and Error events to the underlying file as they are written
             // so the log can be viewed immediately.
@@ -148,7 +148,7 @@ namespace BuildXL.Utilities.Tracing
         /// <inheritdoc/>
         protected override void UnsynchronizedFlush()
         {
-            m_writer.Flush();
+            m_writer?.Flush();
         }
 
         /// <summary>


Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs
```diff
@@ -134,7 +134,7 @@ public override void Dispose()
         /// <inheritdoc />
         protected override void Output(EventLevel level, int id, string eventName, EventKeywords eventKeywords, string text, bool doNotTranslatePaths = false)
         {
-            m_writer.WriteLine(level, (m_translator != null && !doNotTranslatePaths) ? m_translator.Translate(text) : text);
+            m_writer?.WriteLine(level, (m_translator != null && !doNotTranslatePaths) ? m_translator.Translate(text) : text);
```
narasamdya: > m_writer? [](start = 11, length = 10)

I failed to see how `m_writer` can be null in this class. Can you explain? 

If `m_write` can never be null, then your change doesn't fix anything, no?

dannyvv:
>             m_writer.Dispose();

since we made it optional, should we set it to null here?

---
Refers to: Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs:131 in 0fbff76. [](commit_id = 0fbff767176a1eaf9ee95d515a72c677d26d79c3, deletion_comment = False)

dannyvv:
>             IEventWriter writer,

should we check for null here to find the root why it is null?

---
Refers to: Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs:62 in 0fbff76. [](commit_id = 0fbff767176a1eaf9ee95d515a72c677d26d79c3, deletion_comment = False)

norahuang:
> IEventWriter writer,
> 
> 
> should we check for null here to find the root why it is null?
> 
> Refers to: Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs:62 in 0fbff76. [](commit_id = 0fbff76, deletion_comment = False)

It is checked "Contract.Requires(writer != null);". So I don't think that is the problem form the constructor.

Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs
norahuang: From the stack, the two references in this problematic line are m_writer and m_translator.
m_translator is checked before referencing. So the only thing left is m_writer. What can make m_writer null is the m_writer.Dispose().  And this NullReferenceException happened with the console cancel event. I did try to see how it became null from code. But with all the wrappers, it is hard to see how the dispose() being called before the output(). However, with console cancel, I think there is a chance. I did see that much value to spend a lot of time to really figure out how this happen for a console cancel. If we add a check here the 

> m_writer.Dispose();
> 
> 
> since we made it optional, should we set it to null here?
> 
> Refers to: Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs:131 in 0fbff76. [](commit_id = 0fbff76, deletion_comment = False)

Hmm, I am not sure I see it is optional.

norahuang:
> m_writer.Dispose();
> 
> 
> since we made it optional, should we set it to null here?
> 
> Refers to: Public/Src/Utilities/Utilities/Tracing/TextWriterEventListener.cs:131 in 0fbff76. [](commit_id = 0fbff76, deletion_comment = False)

> m_writer.Dispose();
> 
> 
> since we made it optional, should we set it to null here?
> 


Hmm, I am not sure I see it is optional. And how setting it null is different from Dispose()?

norahuang:
Talked to Iman offline and agree to close it.

PR #561, opened 2019-07-08
Enable the MSBuildResolver to work under dotnet core

smera:
Enable the MSBuild resolver to work under dotnet core, and add the option to run MSBuild full framework vs dotnet core.
The same runtime is used for MSBuild at graph construction and execution time. If dotnetcore is specified, the location of particular dotnet.exe to use is required.

### DIFF
--- a/Public/Sdk/Public/Managed/Shared/types.dsc
+++ b/Public/Sdk/Public/Managed/Shared/types.dsc
@@ -129,12 +129,12 @@ export interface LinkResource {
 }
 
 @@public
-export function isBinary(item: Reference) : item is Binary {
+export function isBinary(item: any) : item is Binary {
     return item["binary"] !== undefined;
 }
 
 @@public
-export function isAssembly(item: Reference) : item is Assembly {
+export function isAssembly(item: any) : item is Assembly {
     return item["name"] !== undefined &&
            (item["compile"] !== undefined || item["runtime"] !== undefined) &&
            item["contents"] === undefined; // Exclude nuget packages
--- a/Public/Sdk/Public/Managed/helpers.dsc
+++ b/Public/Sdk/Public/Managed/helpers.dsc
@@ -3,6 +3,7 @@
 
 import * as Shared from "Sdk.Managed.Shared";
 import * as Csc    from "Sdk.Managed.Tools.Csc";
+import * as Deployment from "Sdk.Deployment";
 
 namespace Helpers {
 
@@ -53,11 +54,13 @@ namespace Helpers {
      * You can control whether to follow the compile or the runtime axis using the optional compile argument.
      */
     @@public
-    export function computeTransitiveReferenceClosure(framework: Shared.Framework, references: Shared.Reference[], compile?: boolean) : Shared.Binary[] {
+    export function computeTransitiveReferenceClosure(framework: Shared.Framework, references: Shared.Reference[], runtimeContentToSkip: Deployment.DeployableItem[], compile?: boolean) : Shared.Binary[] {
         return computeTransitiveClosure([
             ...references,
             ...framework.standardReferences
-        ], compile);
+        ], 
+        runtimeContentToSkip, 
+        compile);
     }
 
     /**
@@ -65,7 +68,7 @@ namespace Helpers {
      * You can control whether to follow the compile or the runtime axis using the optional compile argument.
      */
     @@public
-    export function computeTransitiveClosure(references: Shared.Reference[], compile?: boolean) : Shared.Binary[] {
+    export function computeTransitiveClosure(references: Shared.Reference[], referencesToSkip: Deployment.DeployableItem[], compile?: boolean) : Shared.Binary[] {
         let results = MutableSet.empty<Shared.Binary>();
         let visitedReferences = MutableSet.empty<Shared.Reference>();
 
@@ -75,14 +78,15 @@ namespace Helpers {
 
         for (let ref of allReferences)
         {
-            computeTransitiveReferenceClosureHelper(ref, results, visitedReferences, compile);
+            const referencesToSkipSet = Set.empty<Deployment.DeployableItem>().add(...(referencesToSkip ? referencesToSkip : []));
+            computeTransitiveReferenceClosureHelper(ref, referencesToSkipSet, results, visitedReferences, compile);
         }
 
         return results.toArray();
     }
 
-    function computeTransitiveReferenceClosureHelper(ref: Shared.Reference, results: MutableSet<Shared.Binary>, visitedReferences: MutableSet<Shared.Reference>, compile?: boolean) {
-        if (visitedReferences.contains(ref))
+    function computeTransitiveReferenceClosureHelper(ref: Shared.Reference, referencesToSkip: Set<Deployment.DeployableItem>,results: MutableSet<Shared.Binary>, visitedReferences: MutableSet<Shared.Reference>, compile?: boolean) {
+        if (visitedReferences.contains(ref) || referencesToSkip.contains(ref))
         {
             return;
         }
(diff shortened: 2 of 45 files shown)


Public/Sdk/Public/Managed/helpers.dsc
```diff
@@ -75,14 +78,15 @@ namespace Helpers {
 
         for (let ref of allReferences)
         {
-            computeTransitiveReferenceClosureHelper(ref, results, visitedReferences, compile);
+            const referencesToSkipSet = Set.empty<Deployment.DeployableItem>().add(...(referencesToSkip ? referencesToSkip : []));
```
dannyvv: >nst referencesToSkipSet = S [](start = 14, length = 27)

I beleive in another place we have less recursino by simply exposing the the visited and toskip in one set.

i.e. you'd seed the visisted set with the set to skip

Public/Sdk/Public/Managed/Shared/types.dsc
```diff
@@ -129,12 +129,12 @@ export interface LinkResource {
 }
 
 @@public
-export function isBinary(item: Reference) : item is Binary {
+export function isBinary(item: any) : item is Binary {
```
dannyvv: >any [](start = 31, length = 3)

why do we need this?

Can we work around and keep it strongly typed?

Public/Sdk/SelfHost/Libraries/MSBuild/msbuild.dsc
```diff
@@ -36,3 +31,20 @@ export const msbuildRuntimeContent = [
         importFrom("Microsoft.Build.Runtime").Contents.all.getFile(r`contentFiles/any/net472/MSBuild.exe.config`),
     ],
 ];
+
+function getFrameworkFolder() { 
+    return qualifier.targetFramework === "net472" ? a`net472` : a`dotnetcore`;
```
dannyvv: more rubost way if we switch to 48 at some point

return BuildXLSdk.IsNetCore ? "dotnetcore" : qualifier.targetframework

Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
```diff
@@ -136,12 +136,28 @@ interface MsBuildResolver extends ResolverBase, UntrackingSettings {
...
+     */
+    msBuildRuntime?: "FullFramework" | "DotNetCore";
+
+    /**
+     * Collection of directories to search for dotnet.exe, when DotNetCore is specified as the msBuildRuntime. If not 
+     * specified, locations in %PATH% are used.
+     * Locations are traversed in specification order.
+     * It has no effect if the specified MSBuild runtime is full framework.
+     */
+    dotNetSearchLocations?: Directory[];
```
almiliMSFT: just curious: why couldn't you overload and use the same `msBuildSearchLocations` property for this?

Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildAssemblyLoader.cs
```diff
@@ -31,48 +32,56 @@ public sealed class MsBuildAssemblyLoader : IMsBuildAssemblyLoader
...
-        private readonly ConcurrentDictionary<string, Assembly> m_loadedAssemblies = new ConcurrentDictionary<string, Assembly>(5, s_assemblyNamesToLoad.Length, StringComparer.OrdinalIgnoreCase);
-
-        private static readonly Lazy<MsBuildAssemblyLoader> s_singleton = new Lazy<MsBuildAssemblyLoader>(() => new MsBuildAssemblyLoader());
+        private readonly ConcurrentDictionary<string, Assembly> m_loadedAssemblies;
 
-        private MsBuildAssemblyLoader()
+        public MsBuildAssemblyLoader(bool msBuildRuntimeIsDotNetCore)
         {
-        }
+            m_msbuild = msBuildRuntimeIsDotNetCore ? "MsBuild.dll" : "MsBuild.exe";
```
cdmihai: Actual casing is `MSBuild`, for whenever this needs to run on case sensitive file systems.

Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildAssemblyLoader.cs
```diff
@@ -31,48 +32,56 @@ public sealed class MsBuildAssemblyLoader : IMsBuildAssemblyLoader
         private const string SystemCollectionsImmutable = "System.Collections.Immutable.dll";
         private const string SystemThreadingDataflow = "System.Threading.Tasks.Dataflow.dll";
 
-        // MsBuild.exe is not a required to be loaded but is required to be found
-        private const string MsBuildExe = "MsBuild.exe";
+        // MsBuild is not a required to be loaded but is required to be found
+        // Under DotNetCore there is not an msbuild.exe but a msbuild.dll
+        private string m_msbuild;
```
cdmihai: This is the value that eventually gets passed to the `PipGraphConstructor` constructor right? Couldn't easily find the callsite in the review.

Public/Sdk/Public/Managed/Shared/types.dsc
smera: Needed because the runtime content to skip is a deployable item, with no explicit relation with a Reference. Align as well with the signature of isManagedPackage below, which I think looks like that for similar reasons

---
In reply to: [301268220](https://github.com/microsoft/BuildXL/pull/561#discussion_r301268220) [](ancestors = 301268220)

Public/Sdk/SelfHost/Libraries/MSBuild/msbuild.dsc
smera: Will change

---
In reply to: [301268981](https://github.com/microsoft/BuildXL/pull/561#discussion_r301268981) [](ancestors = 301268981)

Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildAssemblyLoader.cs
smera: Will change

---
In reply to: [301352162](https://github.com/microsoft/BuildXL/pull/561#discussion_r301352162) [](ancestors = 301352162)

Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildAssemblyLoader.cs
smera: Yes, the location of msbuild ends up going into the graph result, which is later serialized and picked up by the graph constructor

---
In reply to: [301352822](https://github.com/microsoft/BuildXL/pull/561#discussion_r301352822) [](ancestors = 301352822)

Public/Sdk/Public/Prelude/Prelude.Configuration.Resolvers.dsc
smera: Not a strong reason... I just thought it was cleaner this way, the locations probably won't match anyway

---
In reply to: [301280165](https://github.com/microsoft/BuildXL/pull/561#discussion_r301280165) [](ancestors = 301280165)

Public/Sdk/Public/Managed/helpers.dsc
smera: Seeding the visited set with the content to skip elements is not enough, because the elements to skip are not only items with type Shared.Reference, but they could be assemblies, or have references themselves. So the content to skip needs to get decomposed into leaf elements, which is already what computeTransitiveReferenceClosureHelper is sort of doing

---
In reply to: [301267844](https://github.com/microsoft/BuildXL/pull/561#discussion_r301267844) [](ancestors = 301267844)

PR #562, opened 2019-07-08
 Pass a CSV mapping when uploading logs to Kusto

almiliMSFT:
Instead of uploading just a CSV file and relying on the order CSV file columns to match the order of columns in the Kusto table, specify a csv mapping explicitly, so that the order of columns in the CSV file doesn't matter.

### DIFF
--- a/Public/Src/Cache/ContentStore/App/Application.cs
+++ b/Public/Src/Cache/ContentStore/App/Application.cs
@@ -418,7 +418,7 @@ namespace BuildXL.Cache.ContentStore.App
                 database: KustoDatabase,
                 table: KustoTable,
                 csvMapping: csvMapping,
-                deleteFilesOnSuccess: false,
+                deleteFilesOnSuccess: true,
                 checkForIngestionErrors: true,
                 log: _consoleLog
                 );
--- a/Public/Src/Cache/ContentStore/App/Application.cs
+++ b/Public/Src/Cache/ContentStore/App/Application.cs
@@ -62,24 +62,23 @@ namespace BuildXL.Cache.ContentStore.App
         /// <summary>
         ///     Target Kusto table for remote telemetry
         /// </summary>
-        private const string KustoTable = "CloudBuildLogEvent";
+        private const string KustoTable = "ContentStoreAppMessage";
 
         /// <summary>
         ///     CSV file schema to be produced by <see cref="CsvFileLog"/>.
         /// </summary>
         private static readonly CsvFileLog.ColumnKind[] KustoTableCsvSchema = new CsvFileLog.ColumnKind[]
         {
-            CsvFileLog.ColumnKind.BuildId,
-            CsvFileLog.ColumnKind.Machine,
             CsvFileLog.ColumnKind.PreciseTimeStamp,
-            CsvFileLog.ColumnKind.LocalPreciseTimeStamp,
-            CsvFileLog.ColumnKind.ThreadId,
-            CsvFileLog.ColumnKind.ProcessId,
             CsvFileLog.ColumnKind.LogLevel,
             CsvFileLog.ColumnKind.LogLevelFriendly,
             CsvFileLog.ColumnKind.Message,
+            CsvFileLog.ColumnKind.ProcessId,
+            CsvFileLog.ColumnKind.ThreadId,
             CsvFileLog.ColumnKind.env_os,
             CsvFileLog.ColumnKind.env_osVer,
+            CsvFileLog.ColumnKind.BuildId,
+            CsvFileLog.ColumnKind.Machine,
         };
 
         private readonly CancellationToken _cancellationToken;
@@ -406,26 +405,20 @@ namespace BuildXL.Cache.ContentStore.App
                 maxFileSize: _csvLogMaxFileSize
                 );
 
-            var indexedColumns = _csvFileLog.FileSchema
-                .Select((col, idx) => new CsvColumnMapping
-                    {
-                    ColumnName = col.ToString(),
-                    Ordinal = idx
-                    });
-            var constColumns = _csvFileLog.ConstSchema
-                .Select(col => new CsvColumnMapping
-                    {
-                    ColumnName = col.ToString(),
-                    ConstValue = _csvFileLog.RenderConstColumn(col)
-                });
+            var indexedColumns = _csvFileLog.FileSchema.Select((col, idx) => new CsvColumnMapping { ColumnName = col.ToString(), Ordinal = idx });
+            var constColumns = _csvFileLog.ConstSchema.Select(col => new CsvColumnMapping { ColumnName = col.ToString(), ConstValue = _csvFileLog.RenderConstColumn(col) });
+            var csvMapping = indexedColumns.Concat(constColumns).ToArray();
+
+            var csvMappingStr = string.Join("", csvMapping.Select(col => $"{Environment.NewLine}  Name: '{col.ColumnName}', ConstValue: '{col.ConstValue}', Ordinal: {col.Ordinal}"));
+            _logger.Always("Using csv mapping:{0}", csvMappingStr);
 
             _kustoUploader = new KustoUploader
                 (
                 kustoConnectionString,
                 database: KustoDatabase,
                 table: KustoTable,
-                csvMapping: indexedColumns.Concat(constColumns),
-                deleteFilesOnSuccess: true,
+                csvMapping: csvMapping,
+                deleteFilesOnSuccess: false,
                 checkForIngestionErrors: true,
                 log: _consoleLog
                 );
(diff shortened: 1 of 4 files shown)


Public/Src/Cache/ContentStore/App/Application.cs
```diff
@@ -406,10 +400,29 @@ private void EnableRemoteTelemetryIfNeeded(string logFilePath)
...
                 severity: _fileLogSeverity,
                 maxFileSize: _csvLogMaxFileSize
                 );
 
+            var indexedColumns = _csvFileLog.FileSchema.Select((col, idx) => new CsvColumnMapping { ColumnName = col.ToString(), Ordinal = idx });
+            var constColumns = _csvFileLog.ConstSchema.Select(col => new CsvColumnMapping { ColumnName = col.ToString(), ConstValue = _csvFileLog.RenderConstColumn(col) });
+            var csvMapping = indexedColumns.Concat(constColumns).ToArray();
+
+            var csvMappingStr = string.Join("", csvMapping.Select(col => $"{Environment.NewLine}  Name: '{col.ColumnName}', ConstValue: '{col.ConstValue}', Ordinal: {col.Ordinal}"));
+            _logger.Always("Using csv mapping:{0}", csvMappingStr);
```
Kr1sso: Nit: space?

almiliMSFT: `csvMappingStr` starts with a new line, so the omission of space there was intentional.

Kr1sso: Just saw it, sorry 👍

Public/Src/Cache/ContentStore/App/Application.cs
```diff
@@ -406,10 +400,29 @@ private void EnableRemoteTelemetryIfNeeded(string logFilePath)
...
+
+            _kustoUploader = new KustoUploader
+                (
+                kustoConnectionString,
+                database: KustoDatabase,
+                table: KustoTable,
+                csvMapping: csvMapping,
+                deleteFilesOnSuccess: false,
```
almiliMSFT:
```suggestion
                deleteFilesOnSuccess: true,
```

Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs
```diff
@@ -152,11 +149,24 @@ public static ColumnKind[] ParseTableSchema(string kustoTableSchema)
                 : defaultValue;
         }
 
+        /// <summary>
+        ///     Whether the value of <paramref name="col"/> is constant over time.
+        ///
+        ///     For example, <code>true</code> is returned for columns <see cref="ColumnKind.Empty"/>,
+        ///     <see cref="ColumnKind.BuildId"/>, <see cref="ColumnKind.Machine"/> etc. because their values
```
almiliMSFT:
```suggestion
        ///     <see cref="ColumnKind.BuildId"/>, <see cref="ColumnKind.Machine"/> etc., because their values
```

Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs
```diff
@@ -152,11 +149,24 @@ public static ColumnKind[] ParseTableSchema(string kustoTableSchema)
...
+        /// </summary>
+        public bool IsConstValueColumn(ColumnKind col) => _constColumns.ContainsKey(col);
+
         /// <summary>
         ///     Constructor.  Initializes this object and does nothing else.
         /// </summary>
         /// <param name="logFilePath">Full path to log file</param>
         /// <param name="schema">CSV schema as a list of columns. Each element in the list denotes a column to be rendered at that position.</param>
+        /// <param name="renderConstColums">
+        ///     When true, const columns (<see cref="IsConstValueColumn"/>) from <paramref name="schema"/> are not rendered to the log file.
```
almiliMSFT:
```suggestion
        ///    When true, const columns (<see cref="IsConstValueColumn"/>) from <paramref name="schema"/> are not
        ///    rendered to the log file (those columns become available through the <see ref="ConstSchema"/> property.
```

mpysson:
>         private string CsvEscape(string str)

maybe a silly question, but how do commas get escaped?

---
Refers to: Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs:300 in 71f9dee. [](commit_id = 71f9dee968ef10ef12de13b8541d2edfa784f780, deletion_comment = False)

almiliMSFT:
> maybe a silly question, but how do commas get escaped?

They don't need to be because every column is always enclosed in `"`, so only `"` needs to be escaped

Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs
```diff
@@ -182,10 +194,31 @@ public CsvFileLog
...
-            Service = CsvEscape(serviceName ?? string.Empty);
-            Machine = CsvEscape(Environment.MachineName);
             BuildId = buildId == default ? Guid.NewGuid() : buildId;
+
+            _constColumns = new Dictionary<ColumnKind, string>
+            {
+                [ColumnKind.Empty] = string.Empty,
+                [ColumnKind.BuildId] = BuildId.ToString(),
+                [ColumnKind.Machine] = CsvEscape(Environment.MachineName),
+                [ColumnKind.Service] = CsvEscape(serviceName ?? string.Empty),
```
dannyvv: shouldnt csvexcape be done at the time of serializing? Not in the internal structure.

And if here, why not osversion data?

dannyvv:
>         private string CsvEscape(string str)

csv escape comma by double-quoting the entire string and have a comma i.e.
3 columns:

`abc,"a,b,c",abc`


---
In reply to: [509713332](https://github.com/microsoft/BuildXL/pull/562#issuecomment-509713332) [](ancestors = 509713332)

---
Refers to: Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs:300 in 71f9dee. [](commit_id = 71f9dee968ef10ef12de13b8541d2edfa784f780, deletion_comment = False)

Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs
almiliMSFT: The assumption is that those will not contain any double quotes, which is the only char that needs escaping. 

Nevertheless, it's a good idea to just call CsvEscape and not think about it.  I'll change.  

As for whether to do it here or every time Render is called, I thought it makes sense to just do it once here.

almiliMSFT:
> > ```
> >     private string CsvEscape(string str)
> > ```
> 
> 
> csv escape comma by double-quoting the entire string and have a comma i.e.
> 
> 3 columns:
> 
> 
> `abc,"a,b,c",abc`
> 
> 
> 
> In reply to: [509713332](https://github.com/microsoft/BuildXL/pull/562#issuecomment-509713332) [](ancestors = 509713332)
> 
> 
> 
> Refers to: Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs:300 in [71f9dee](https://github.com/microsoft/BuildXL/commit/71f9dee968ef10ef12de13b8541d2edfa784f780). [](commit_id = [71f9dee](https://github.com/microsoft/BuildXL/commit/71f9dee968ef10ef12de13b8541d2edfa784f780), deletion_comment = False)

Quotes are added in the `Render` method, but you are right, it's probably better if that's done by CsvEscape.

Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs
almiliMSFT: Changed to now escape everything when rendering.

Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs
```diff
@@ -152,11 +149,25 @@ public static ColumnKind[] ParseTableSchema(string kustoTableSchema)
...
+        /// </summary>
+        public bool IsConstValueColumn(ColumnKind col) => _constColumns.ContainsKey(col);
+
         /// <summary>
         ///     Constructor.  Initializes this object and does nothing else.
         /// </summary>
         /// <param name="logFilePath">Full path to log file</param>
         /// <param name="schema">CSV schema as a list of columns. Each element in the list denotes a column to be rendered at that position.</param>
+        /// <param name="renderConstColums">
+        ///     When true, const columns (<see cref="IsConstValueColumn"/>) from <paramref name="schema"/> are not
```
almiliMSFT:
```suggestion
        ///     When false, const columns (<see cref="IsConstValueColumn"/>) from <paramref name="schema"/> are not
```

Public/Src/Cache/ContentStore/Library/Logging/CsvFileLog.cs
```diff
@@ -152,11 +149,25 @@ public static ColumnKind[] ParseTableSchema(string kustoTableSchema)
...
+        ///     rendered to the log file (those columns become available through the <see ref="ConstSchema"/> property.
```
almiliMSFT:
```suggestion
        ///     rendered to log file (those columns become available through the <see ref="ConstSchema"/> property.
```

PR #564, opened 2019-07-09
Remove FEATURE_MICROSOFT_DIAGNOSTICS_TRACING conditional compilation

almiliMSFT:


Public/Src/App/Bxl/Args.cs
```diff
@@ -5,11 +5,7 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.ContractsLight;
-#if FEATURE_MICROSOFT_DIAGNOSTICS_TRACING
-using Microsoft.Diagnostics.Tracing;
-#else
 using System.Diagnostics.Tracing;
```
mpysson: Please check that the System.Diagnostics.Tracing is as new as the Microsoft.Diagnostics.Tracing package versions we're currently referencing in config.dsc. If they are, you can remove those references. We were using a separate version in order to get some bug fixes. But most likely after updating .net versions the standard System.Diagnostics.Tracing is new enough to have the fixes.

                { id: "Microsoft.Diagnostics.Tracing.EventSource.Redist", version: "1.1.28" },
                { id: "Microsoft.Diagnostics.Tracing.TraceEvent", version: "2.0.30" },

dannyvv: I think that was only for net451, when we moved to 461 we were already okay to remove and definatley after net472

---
In reply to: [301657267](https://github.com/microsoft/BuildXL/pull/564#discussion_r301657267) [](ancestors = 301657267)

PR #566, opened 2019-07-09
Make ImplicitPin configurable in MemoizationStoreAdapterCache

JuanCarlosGI:
It will be needed for DedupStore, since pinning now makes hundreds of calls to VSTS to verify the existence of each chunk of a file, so pinning before each put/place is no longer the "fail fast" route.

### DIFF
--- a/Public/Src/Cache/ContentStore/Interfaces/Stores/ImplicitPin.cs
+++ b/Public/Src/Cache/ContentStore/Interfaces/Stores/ImplicitPin.cs
@@ -1,11 +1,14 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+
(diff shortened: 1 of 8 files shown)


dcwuser:
Won't this change also remove the implicit pin that flows to CaSaaS? Isn't what we want for implicit pin to still flow to CaSaaS, but not to VSTS?

JuanCarlosGI:
> Won't this change also remove the implicit pin that flows to CaSaaS? Isn't what we want for implicit pin to still flow to CaSaaS, but not to VSTS?

We can still leave the implicit pin on for the local CASaaS, while disabling it for VSTS.

JuanCarlosGI:
> Won't this change also remove the implicit pin that flows to CaSaaS? Isn't what we want for implicit pin to still flow to CaSaaS, but not to VSTS?

The way this config works is that the local (CASaaS) and remote (VSTS) caches will have independent flags, so we can disable it in remote but keep it in local.

PR #567, opened 2019-07-09
Grpc copy fixes

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -103,30 +94,43 @@ public async Task<FileExistenceResult> CheckFileExistsAsync(Context context, Con
...
         /// <summary>
         /// Copies content from the server to the stream returned by the factory.
         /// </summary>
-        public async Task<CopyFileResult> CopyToAsync(Context context, ContentHash hash, Func<Stream> streamFactory, CancellationToken ct = default(CancellationToken))
+        public async Task<CopyFileResult> CopyToAsync(Context context, ContentHash hash, Func<Stream> streamFactory, CancellationToken ct)
+        {
+            // Need to track shutdown to prevent invalid operation errors when the instance is used after it was shut down is called.
+            using (var operationContext = TrackShutdown(context, ct))
+            {
+                return await CopyToCoreAsync(operationContext, hash, streamFactory);
```
dannyvv: >await [](start = 23, length = 5)

isn't this await superfluous if you also remove the async?

SergeyTeplyakov: It is very much needed here because it is happening inside the using block.

Without it `operationContext` will be disposed immediately without waiting for the copy to finish.

dcwuser: I make this mistake again and again. So nice to see Danny make it too.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -256,7 +263,7 @@ private async Task<(long Chunks, long Bytes)> StreamContentWithCompressionAsync(
             {
                 using (Stream decompressedStream = new GZipStream(grpcStream, CompressionMode.Decompress, true))
                 {
-                    await decompressedStream.CopyToAsync(targetStream, _bufferSize, ct).ConfigureAwait(false);
```
dcwuser: Minor:  I'm not a fan of your crusade against ConfigureAwait(false). I find it very annoying that BCL choose the wrong default, but I also really don't look forward to putting back on a bunch of ConfigureAwait(false) when someone wants to re-use our code in an app with a UI.

SergeyTeplyakov: I've added an analyzer into ErrorProne.NET that is used in the solution right now. So if we'll change our mind, it will be one line change (for instance [here](http://index/repos/domino?query=%60donotuseconfigureawait&rightProject=BuildXL.Cache.ContentStore.Distributed&file=Properties%5CAssemblyInfo.cs&line=6)) and then we'll run a global fixer to use `ConfigureAwait(false)` everywhere in the entire project.

That's why I think consistency is actually important here: we can either use `ConfigureAwait(false)` everywhere or not use it at all. And this decision is controlled per project level.

PR #568, opened 2019-07-09
Implement post-initialization for MultiplexedContentStore to prevent hangs in shutdown

SergeyTeplyakov:
[AB#1568742](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1568742)

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
+++ b/Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
@@ -229,7 +229,7 @@ namespace BuildXL.Cache.ContentStore.Service
 
             foreach (var store in StoresByName.Values)
             {
-                if (store is IContentStoreWithPostInitialization contentStore)
+                if (store is IContentStore contentStore)
                 {
                     contentStore.PostInitializationCompleted(context, result);
                 }
--- a/Public/Src/Cache/DistributedCache.Host/Service/DistributedCacheServiceFacade.cs
+++ b/Public/Src/Cache/DistributedCache.Host/Service/DistributedCacheServiceFacade.cs
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Threading.Tasks;
 using BuildXL.Cache.ContentStore.Exceptions;
 using BuildXL.Cache.ContentStore.Interfaces.Results;
 using BuildXL.Cache.ContentStore.Interfaces.Tracing;
+using BuildXL.Cache.ContentStore.Service;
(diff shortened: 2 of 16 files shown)


Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
```diff
@@ -229,7 +229,7 @@ protected override async Task<BoolResult> StartupCoreAsync(OperationContext cont
 
             foreach (var store in StoresByName.Values)
             {
-                if (store is IContentStoreWithPostInitialization contentStore)
+                if (store is IContentStore contentStore)
```
dcwuser: Isn't store guaranteed to be IContentStore?

Public/Src/Cache/DistributedCache.Host/Service/DistributedCacheServiceFacade.cs
```diff
@@ -68,5 +71,17 @@ public static async Task RunAsync(DistributedCacheServiceArguments arguments)
                 }
             }
         }
+
+        private static async Task<BoolResult> ShutdownWithTimeout(Context context, LocalContentServer server, TimeSpan timeout)
+        {
+            var shutdownTask = server.ShutdownAsync(context);
+            if (await Task.WhenAny(shutdownTask, Task.Delay(timeout)) != shutdownTask)
```
dcwuser: No cancellation or result processing of Task that lost?

Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
SergeyTeplyakov: `TStore` is `IStartupShutdown` because the same base type is used to store both `IContentStore` and `IMemoizationStore` instances.

Public/Src/Cache/DistributedCache.Host/Service/DistributedCacheServiceFacade.cs
SergeyTeplyakov: Task.Delay can't throw and if `shutdownTask` is not yet finished, then the process will exit anyways. So in this case, it is perfectly fine. I wanted to use our extension method `WithTimeout` but unfortunately this assembly doens't reference Bxl.Utilities and we don't wat to add this reference.

Public/Src/Cache/DistributedCache.Host/Service/Internal/MultiplexedContentStore.cs
```diff
@@ -80,6 +80,8 @@ public Task<BoolResult> StartupAsync(Context context)
             });
         }
 
+
+
         /// <inheritdoc />
```
dannyvv: whoops :)

PR #570, opened 2019-07-09
FileContentInfo - change how length/existence is stored

olkononenko:
[AB#1567427](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1567427)


Public/Src/Utilities/Storage/FileContentInfo.cs
```diff
@@ -18,27 +18,153 @@ namespace BuildXL.Storage
...
+            }
+
+            /// <summary>
+            /// Creates a <see cref="LengthAndExistence"/> with a known length and optional existence value.
+            /// </summary>
+            public LengthAndExistence(long length, PathExistence? existence)
+            {
+                if ((length & LengthValueMask) != length)
+                {
+                    Contract.Assert(false, $"Invalid length value (reserved bits should not be used): {length} ({System.Text.RegularExpressions.Regex.Replace(Convert.ToString(length, 2).PadLeft(64, '0'), ".{4}", "$0 ")})");
```
mpysson: >"Invalid length value (reserved bits should not be used): {length} ({System.Text.RegularExpressions.Regex.Replace(Convert.ToString(length, 2).PadLeft(64, '0'), ".{4}", "$0 ")})"); [](start = 44, length = 179)

Would it be simpler to just have a message that "length cannot be larger than x"?

Public/Src/Engine/Cache/Fingerprints/BondExtensions.cs
```diff
@@ -131,7 +131,7 @@ public static FileMaterializationInfo ToFileMaterializationInfo(this BondFileMat
             Contract.Requires(bondInfo != null);
 
             return new FileMaterializationInfo(
-                new FileContentInfo(bondInfo.Hash.ToContentHash(), bondInfo.Length),
+                new FileContentInfo(bondInfo.Hash.ToContentHash(), FileContentInfo.LengthAndExistence.CreateWithCombinedValue(bondInfo.Length)),
```
mpysson: >Length [](start = 135, length = 6)

So this isn't really Length anymore as it has other data encoded in it. Should you rename the field? Same applies to the data passed into other callsites of CreateWithCombinedValue()

Public/Src/Utilities/Storage/FileContentInfo.cs
```diff
@@ -18,27 +18,153 @@ namespace BuildXL.Storage
...
+
+            private LengthAndExistence(long combinedValue)
+            {
+                m_data = combinedValue;
+            }
+
+            /// <summary>
+            /// Creates a <see cref="LengthAndExistence"/> using a combined existence/length value. Mainly used in deserialization logic.
+            /// </summary>                        
+            public static LengthAndExistence CreateWithCombinedValue(long combinedValue)
```
mpysson: >CreateWithCombinedValue [](start = 45, length = 23)

What do you think about calling this "Deserialize()"?

Public/Src/Utilities/Storage/FileContentInfo.cs
```diff
@@ -18,27 +18,153 @@ namespace BuildXL.Storage
...
+
+            /// <summary>
+            /// Length value. Returns 0 if <see cref="IsKnownLength" /> is false.
+            /// </summary>
+            public long Length => m_data & LengthValueMask;
+
+            /// <summary>
+            /// The underlying value used for storage.
+            /// </summary>
+            public long CombinedValue => m_data;
```
mpysson: >CombinedValue [](start = 24, length = 13)

rename SerializedValue or something similar?

Public/Src/Utilities/Storage/FileContentInfo.cs
```diff
@@ -18,27 +18,153 @@ namespace BuildXL.Storage
...
+        }
+
+        /// <summary>
+        /// Creates a <see cref="FileContentInfo"/> using a combined existence/length value. 
+        /// </summary>
+        /// <remarks>
+        /// Mainly used for deserialization and files with unknown length.
+        /// </remarks>
+        public FileContentInfo(ContentHash hash, LengthAndExistence lengthAndExistence)
+        {
```
mpysson: Please add some tests for these to make sure things round trip correctly through the bit packing of the underlying struct

Public/Src/Utilities/Storage/FileContentInfo.cs
```diff
@@ -18,27 +18,153 @@ namespace BuildXL.Storage
...
         /// <summary>
         /// Creates a <see cref="FileContentInfo"/> with real USN version information.
         /// </summary>
-        public FileContentInfo(ContentHash hash, long length)
+        public FileContentInfo(ContentHash hash, long length) 
         {
-            m_length = length;
+            Contract.Requires(length >= 0);
+            
+            m_lengthAndExistence = new LengthAndExistence(
```
dannyvv: >             [](start = 58, length = 12)

whitespace and end of iles (also line 154)

dannyvv:
>         public static bool IsValidLength(long length, ContentHash hash)

Should this check the file length fits in 54 bits? Rather than the full long?

---
Refers to: Public/Src/Utilities/Storage/FileContentInfo.cs:285 in b76d248. [](commit_id = b76d248eea43da83c65c457a93bc3435430a397e, deletion_comment = False)

Public/Src/Utilities/Storage/FileContentInfo.cs
olkononenko: Well, the main purpose is to check that someone is not passing a previously serialized value. The condition is essentially checking that length is within [0..MaxSupportedLength]

---
In reply to: [301852095](https://github.com/microsoft/BuildXL/pull/570#discussion_r301852095) [](ancestors = 301852095)

PR #571, opened 2019-07-10
Add basic IMemoizationSession constructs for CaChaaS

jbayardo:
There are a few big points that will not go in this PR, and will need to be done in the future:

- Moving the logic away from the ContentLocationDatabase into somewhere more appropriate. Before this happens, several classes have to be created in the Distributed Memoization Store layer, as well as deciding how we are going to share the RocksDB layer.
- Porting of MemoizationSessionTests. These require integration with the content layer and the previous point.
- `IsContentAvailable` requires integration with the content layer, and so will be left as a placeholder. This needs to be implemented, as it is part of the original SQLite implementation.
- Garbage collection and removal of content in general.

Also, please notice that all tests have been copy-pasted from our SQL memoization store tests, modulo minimal changes required to get them to run (i.e. removing async and extra function parameters)

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
@@ -632,14 +632,64 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             context.TraceDebug($"Deleted entry for hash {hash}. Creation Time: '{entry.CreationTimeUtc}', Last Access Time: '{entry.LastAccessTimeUtc}'");
         }
 
-        /// <todoc />
+        /// <summary>
+        /// Load a ContentHashList.
+        /// </summary>
+        /// <param name="context">
+        ///     Tracing context.
+        /// </param>
+        /// <param name="strongFingerprint">
+        ///     Full key for ContentHashList value.
+        /// </param>
+        /// <returns>
+        ///     Result providing the call's completion status.
+        /// </returns>
         public abstract GetContentHashListResult GetContentHashList(OperationContext context, StrongFingerprint strongFingerprint);
 
-        /// <todoc />
+        /// <summary>
+        /// Store a ContentHashList
+        /// </summary>
+        /// <param name="context">
+        ///     Tracing context.
+        /// </param>
+        /// <param name="strongFingerprint">
+        ///     Full key for ContentHashList value.
+        /// </param>
+        /// <param name="contentHashListWithDeterminism">
+        ///     The value, and associated determinism guarantee, to store.
+        /// </param>
+        /// <returns>
+        ///     Result providing the call's completion status.
+        /// </returns>
         public abstract AddOrGetContentHashListResult AddOrGetContentHashList(OperationContext context, StrongFingerprint strongFingerprint, ContentHashListWithDeterminism contentHashListWithDeterminism);
 
-        /// <todoc />
+        /// <summary>
+        /// Gets known selectors for a given weak fingerprint.
+        /// </summary>
+        /// <param name="context">
+        ///     Tracing context.
+        /// </param>
+        /// <param name="weakFingerprint">
+        ///     Weak fingerprint to fetch selectors for
+        /// </param>
+        /// <returns>
+        ///     Result providing the call's completion status.
+        /// </returns>
         public abstract IReadOnlyCollection<GetSelectorResult> GetSelectors(OperationContext context, Fingerprint weakFingerprint);
+
+        /// <summary>
+        /// Enumerates all strong fingerprints currently stored in the cache.
+        /// </summary>
+        /// <remarks>
+        ///     Warning: this function should only ever be used on tests.
+        /// </remarks>
+        /// <param name="context">
+        ///     Tracing context.
+        /// </param>
+        /// <returns>
+        ///     Result providing the call's completion status.
+        /// </returns>
+        public abstract IEnumerable<StructResult<StrongFingerprint>> EnumerateStrongFingerprints(OperationContext context);
 
         private object GetLock(ShortHash hash)
         {
@@ -682,7 +732,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// <remarks>
         /// We explicitly take and pass the instance as parameters in order to avoid lambda capturing.
         /// </remarks>
-        protected byte[] SerializeWithPool<T>(T instance, Action<T, BuildXLWriter> serializeFunc)
+        protected byte[] SerializeCore<T>(T instance, Action<T, BuildXLWriter> serializeFunc)
         {
             using (var pooledWriter = _writerPool.GetInstance())
             {
@@ -698,7 +748,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// <remarks>
         /// Be mindful of avoiding lambda capture when using this function.
         /// </remarks>
-        protected T DeserializeWithPool<T>(byte[] bytes, Func<BuildXLReader, T> deserializeFunc)
+        protected T DeserializeCore<T>(byte[] bytes, Func<BuildXLReader, T> deserializeFunc)
         {
             using (PooledObjectWrapper<StreamBinaryReader> pooledReader = _readerPool.GetInstance())
             {
@@ -712,7 +762,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// </summary>
         protected byte[] SerializeContentLocationEntry(ContentLocationEntry entry)
         {
-            return SerializeWithPool(entry, (instance, writer) => instance.Serialize(writer));
+            return SerializeCore(entry, (instance, writer) => instance.Serialize(writer));
         }
 
         /// <summary>
@@ -720,7 +770,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// </summary>
         protected ContentLocationEntry DeserializeContentLocationEntry(byte[] bytes)
         {
-            return DeserializeWithPool(bytes, ContentLocationEntry.Deserialize);
+            return DeserializeCore(bytes, ContentLocationEntry.Deserialize);
         }
 
         /// <summary>
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
@@ -12,10 +12,13 @@ using BuildXL.Cache.ContentStore.Distributed.Utilities;
 using BuildXL.Cache.ContentStore.Hashing;
 using BuildXL.Cache.ContentStore.Interfaces.Extensions;
 using BuildXL.Cache.ContentStore.Interfaces.Results;
+using BuildXL.Cache.ContentStore.Interfaces.Sessions;
 using BuildXL.Cache.ContentStore.Interfaces.Time;
 using BuildXL.Cache.ContentStore.Tracing;
 using BuildXL.Cache.ContentStore.Tracing.Internal;
 using BuildXL.Cache.ContentStore.Utils;
+using BuildXL.Cache.MemoizationStore.Interfaces.Results;
+using BuildXL.Cache.MemoizationStore.Interfaces.Sessions;
 using BuildXL.Utilities;
 using BuildXL.Utilities.Collections;
 using BuildXL.Utilities.Tasks;
@@ -629,6 +632,15 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             context.TraceDebug($"Deleted entry for hash {hash}. Creation Time: '{entry.CreationTimeUtc}', Last Access Time: '{entry.LastAccessTimeUtc}'");
         }
 
+        /// <todoc />
+        public abstract GetContentHashListResult GetContentHashList(OperationContext context, StrongFingerprint strongFingerprint);
+
+        /// <todoc />
+        public abstract AddOrGetContentHashListResult AddOrGetContentHashList(OperationContext context, StrongFingerprint strongFingerprint, ContentHashListWithDeterminism contentHashListWithDeterminism);
+
+        /// <todoc />
+        public abstract IReadOnlyCollection<GetSelectorResult> GetSelectors(OperationContext context, Fingerprint weakFingerprint);
+
         private object GetLock(ShortHash hash)
         {
             // NOTE: We choose not to use "random" two bytes of the hash because
@@ -665,40 +677,50 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         }
 
         /// <summary>
-        /// Serialize a given <paramref name="entry"/> into a byte stream.
+        /// Uses an object pool to fetch a serializer and feed it into the serialization function.
         /// </summary>
-        protected byte[] Serialize(ContentLocationEntry entry)
+        /// <remarks>
+        /// We explicitly take and pass the instance as parameters in order to avoid lambda capturing.
+        /// </remarks>
+        protected byte[] SerializeWithPool<T>(T instance, Action<T, BuildXLWriter> serializeFunc)
         {
             using (var pooledWriter = _writerPool.GetInstance())
             {
                 var writer = pooledWriter.Instance.Writer;
-                writer.WriteCompact(entry.ContentSize);
-                entry.Locations.Serialize(writer);
-                writer.Write(entry.CreationTimeUtc);
-                long lastAccessTimeOffset = entry.LastAccessTimeUtc.Value - entry.CreationTimeUtc.Value;
-                writer.WriteCompact(lastAccessTimeOffset);
+                serializeFunc(instance, writer);
                 return pooledWriter.Instance.Buffer.ToArray();
             }
         }
 
         /// <summary>
-        /// Deserialize <see cref="ContentLocationEntry"/> from an array of bytes.
+        /// Uses an object pool to fetch a binary reader and feed it into the deserialization function.
         /// </summary>
-        protected ContentLocationEntry Deserialize(byte[] bytes)
+        /// <remarks>
+        /// Be mindful of avoiding lambda capture when using this function.
+        /// </remarks>
+        protected T DeserializeWithPool<T>(byte[] bytes, Func<BuildXLReader, T> deserializeFunc)
         {
             using (PooledObjectWrapper<StreamBinaryReader> pooledReader = _readerPool.GetInstance())
             {
                 var reader = pooledReader.Instance;
-                return reader.Deserialize(new ArraySegment<byte>(bytes), r =>
-                                                                         {
-                                                                             var size = r.ReadInt64Compact();
-                                                                             var locations = MachineIdSet.Deserialize(r);
-                                                                             var creationTimeUtc = r.ReadUnixTime();
-                                                                             var lastAccessTimeOffset = r.ReadInt64Compact();
-                                                                             var lastAccessTime = new UnixTime(creationTimeUtc.Value + lastAccessTimeOffset);
-                                                                             return ContentLocationEntry.Create(locations, size, lastAccessTime, creationTimeUtc);
-                                                                         });
+                return reader.Deserialize(new ArraySegment<byte>(bytes), deserializeFunc);
             }
+        }
+
+        /// <summary>
+        /// Serialize a given <paramref name="entry"/> into a byte stream.
+        /// </summary>
+        protected byte[] SerializeContentLocationEntry(ContentLocationEntry entry)
+        {
+            return SerializeWithPool(entry, (instance, writer) => instance.Serialize(writer));
+        }
+
+        /// <summary>
+        /// Deserialize <see cref="ContentLocationEntry"/> from an array of bytes.
+        /// </summary>
+        protected ContentLocationEntry DeserializeContentLocationEntry(byte[] bytes)
+        {
+            return DeserializeWithPool(bytes, ContentLocationEntry.Deserialize);
         }
 
         /// <summary>
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationEntry.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationEntry.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
 using BuildXL.Cache.ContentStore.Distributed.Utilities;
+using BuildXL.Utilities;
 using BuildXL.Utilities.Collections;
 
 namespace BuildXL.Cache.ContentStore.Distributed.NuCache
@@ -99,6 +100,31 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             }
 
             return FromRedisValue(data, lastAccessTime);
+        }
+
+        /// <summary>
+        /// Serializes an instance into a binary stream.
+        /// </summary>
+        public void Serialize(BuildXLWriter writer)
+        {
+            writer.WriteCompact(ContentSize);
+            Locations.Serialize(writer);
+            writer.Write(CreationTimeUtc);
+            long lastAccessTimeOffset = LastAccessTimeUtc.Value - CreationTimeUtc.Value;
+            writer.WriteCompact(lastAccessTimeOffset);
+        }
+
+        /// <summary>
+        /// Builds an instance from a binary stream.
+        /// </summary>
+        public static ContentLocationEntry Deserialize(BuildXLReader reader)
+        {
+            var size = reader.ReadInt64Compact();
+            var locations = MachineIdSet.Deserialize(reader);
+            var creationTimeUtc = reader.ReadUnixTime();
+            var lastAccessTimeOffset = reader.ReadInt64Compact();
+            var lastAccessTime = new UnixTime(creationTimeUtc.Value + lastAccessTimeOffset);
+            return Create(locations, size, lastAccessTime, creationTimeUtc);
         }
 
         /// <nodoc />
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/InMemory/MemoryContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/InMemory/MemoryContentLocationDatabase.cs
@@ -52,17 +52,31 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache.InMemory
         }
 
         /// <inheritdoc />
-        public override GetContentHashListResult GetContentHashList(OperationContext context, StrongFingerprint strongFingerprint) => null;
+        public override GetContentHashListResult GetContentHashList(OperationContext context, StrongFingerprint strongFingerprint)
+        {
+            throw new NotImplementedException();
+        }
 
         /// <inheritdoc />
         public override AddOrGetContentHashListResult AddOrGetContentHashList(
             OperationContext context,
             StrongFingerprint strongFingerprint,
-            ContentHashListWithDeterminism contentHashListWithDeterminism) =>
-            null;
+            ContentHashListWithDeterminism contentHashListWithDeterminism)
+        {
+            throw new NotImplementedException();
+        }
 
         /// <inheritdoc />
-        public override IReadOnlyCollection<GetSelectorResult> GetSelectors(OperationContext context, Fingerprint weakFingerprint) => null;
+        public override IReadOnlyCollection<GetSelectorResult> GetSelectors(OperationContext context, Fingerprint weakFingerprint)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        public override IEnumerable<StructResult<StrongFingerprint>> EnumerateStrongFingerprints(OperationContext context)
+        {
+            throw new NotImplementedException();
+        }
 
         /// <inheritdoc />
         protected override IEnumerable<ShortHash> EnumerateSortedKeysFromStorage(CancellationToken token)
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/InMemory/MemoryContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/InMemory/MemoryContentLocationDatabase.cs
@@ -11,6 +11,8 @@ using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
 using BuildXL.Cache.ContentStore.Interfaces.Results;
 using BuildXL.Cache.ContentStore.Interfaces.Time;
 using BuildXL.Cache.ContentStore.Tracing.Internal;
+using BuildXL.Cache.MemoizationStore.Interfaces.Results;
+using BuildXL.Cache.MemoizationStore.Interfaces.Sessions;
 
 namespace BuildXL.Cache.ContentStore.Distributed.NuCache.InMemory
 {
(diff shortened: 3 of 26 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -665,42 +677,52 @@ public void ContentTouched(OperationContext context, ShortHash hash, UnixTime ac
...
 
         /// <summary>
-        /// Serialize a given <paramref name="entry"/> into a byte stream.
+        /// Uses an object pool to fetch a serializer and feed it into the serialization function.
         /// </summary>
-        protected byte[] Serialize(ContentLocationEntry entry)
+        /// <remarks>
+        /// We explicitly take and pass the instance as parameters in order to avoid lambda capturing.
+        /// </remarks>
+        protected byte[] SerializeWithPool<T>(T instance, Action<T, BuildXLWriter> serializeFunc)
```
dcwuser: Is this used by methods other than the ContentLocationEntry serialization method? If not, make private.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationEntry.cs
```diff
@@ -101,6 +102,31 @@ public static ContentLocationEntry TryCreateFromRedisValue(byte[] data, UnixTime
...
 
+        /// <summary>
+        /// Serializes an instance into a binary stream.
+        /// </summary>
+        public void Serialize(BuildXLWriter writer)
+        {
+            writer.WriteCompact(ContentSize);
+            Locations.Serialize(writer);
+            writer.Write(CreationTimeUtc);
+            long lastAccessTimeOffset = LastAccessTimeUtc.Value - CreationTimeUtc.Value;
```
dcwuser: What is reason for serializing in this way? Why not just write out both times?

dannyvv:
Is 'CaChaaS' a new service? Or perhaps a typo :)

Public/Src/Cache/ContentStore/Distributed/NuCache/InMemory/MemoryContentLocationDatabase.cs
```diff
@@ -49,6 +51,19 @@ internal override void Persist(OperationContext context, ShortHash hash, Content
             _map.AddOrUpdate(hash, key => entry, (key, old) => entry);
         }
 
+        /// <inheritdoc />
```
dannyvv: Just a generic comment on all the cache interfaces.

I have started seeing a pattern of having a base function and than have one derived class implement them where the others return null.

Does this warrant a capability based design,  or perhaps more specific interfaces?

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,162 @@ private byte[] GetKey(ShortHash hash)
             return hash.ToByteArray();
         }
 
+        #region Memoization
+        // TODO(jubayard): garbage collection / removal in general
+
+        /// <inheritdoc />
+        public override GetContentHashListResult GetContentHashList(OperationContext context, StrongFingerprint strongFingerprint)
+        {
+            var key = GetCaChaaSKey(strongFingerprint);
```
dannyvv: cachaas or cazaas?

SergeyTeplyakov:
It is kind-of "Content Addressable Cache as a Service" it is a code name for the project when we combine content and metadata into one single service.

---
In reply to: [510135565](https://github.com/microsoft/BuildXL/pull/571#issuecomment-510135565) [](ancestors = 510135565)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -629,6 +632,15 @@ private void LogEntryDeletion(OperationContext context, ShortHash hash, ContentL
             context.TraceDebug($"Deleted entry for hash {hash}. Creation Time: '{entry.CreationTimeUtc}', Last Access Time: '{entry.LastAccessTimeUtc}'");
         }
 
+        /// <todoc />
+        public abstract GetContentHashListResult GetContentHashList(OperationContext context, StrongFingerprint strongFingerprint);
+
+        /// <todoc />
```
SergeyTeplyakov: I think we should change it.

I usually use this is a "TODO" before checking things in.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -665,42 +677,52 @@ public void ContentTouched(OperationContext context, ShortHash hash, UnixTime ac
...
+        {
+            return SerializeWithPool(entry, (instance, writer) => instance.Serialize(writer));
+        }
+
+        /// <summary>
+        /// Deserialize <see cref="ContentLocationEntry"/> from an array of bytes.
+        /// </summary>
+        protected ContentLocationEntry DeserializeContentLocationEntry(byte[] bytes)
+        {
+            return DeserializeWithPool(bytes, ContentLocationEntry.Deserialize);
```
SergeyTeplyakov: If you are series about your comment at line 699, then you should change this method because it allocates, at least once, but maybe more than once.

A bit of C# trivia:

Action a = MyClass.FooBar; // always allocates, this is called method group conversion

Action b = () => MyClass.FooBar(); // won't allocate because C# compiler will cache the delegate.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -36,17 +39,36 @@ public sealed class RocksDbContentLocationDatabase : ContentLocationDatabase
...
+            ClusterState,
+            /// <summary>
+            /// Stores mapping from <see cref="StrongFingerprint"/> to a <see cref="ContentHashList"/>. This allows us
+            /// to look up via a <see cref="Fingerprint"/>, or a <see cref="StrongFingerprint"/>. The only reason we
+            /// can look up by <see cref="Fingerprint"/> is that it is stored as a prefix to the
+            /// <see cref="StrongFingerprint"/>.
+            ///
+            /// This serves all of CaChaaS' needs for storage, modulo garbage collection.
+            /// </summary>
+            CaChaaS
```
SergeyTeplyakov: Let's rename it to "Metadata" or "Fingerprints".

"Cachaas" is kind of an internal name that we use during discussion, but I don't think it is a good idea to use it in the code too much.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
SergeyTeplyakov: The same: need to rename this thing.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
jbayardo: It is used in the child class to have access to the serialization pool of the ContentLocationDatabase

---
In reply to: [301862091](https://github.com/microsoft/BuildXL/pull/571#discussion_r301862091) [](ancestors = 301862091)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,162 @@ private byte[] GetKey(ShortHash hash)
...
+            
+            return new GetContentHashListResult(result ?? default);
+        }
+        
+        /// <inheritdoc />
+        public override AddOrGetContentHashListResult AddOrGetContentHashList(OperationContext context, StrongFingerprint strongFingerprint, ContentHashListWithDeterminism contentHashListWithDeterminism)
+        {
+            var key = GetCaChaaSKey(strongFingerprint);
+
+            Possible<AddOrGetContentHashListResult> status = _keyValueStore.Use(
```
SergeyTeplyakov: Let's be consistent (or at least let's try) in terms of var vs explicit types.

I prefer 'var' unless you really want to be explicit, but I see that in this method you have explicit type but in GetContentHashList var is used.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,162 @@ private byte[] GetKey(ShortHash hash)
             return hash.ToByteArray();
         }
 
+        #region Memoization
+        // TODO(jubayard): garbage collection / removal in general
+
+        /// <inheritdoc />
+        public override GetContentHashListResult GetContentHashList(OperationContext context, StrongFingerprint strongFingerprint)
```
SergeyTeplyakov: Please, use 'context.PerformOperation' for all the new methods and add new counter to track the duration.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,162 @@ private byte[] GetKey(ShortHash hash)
...
+                        // hashes exist. Moreover, we can prune content.
+                    }
+                });
+
+            if (!status.Succeeded)
+            {
+                return new GetContentHashListResult(status.Failure.CreateException());
+            }
+            
+            return new GetContentHashListResult(result ?? default);
```
SergeyTeplyakov: Maybe instead of using nullable value you can just use:

`ContentHashListWithDeterminism result = new ContentHashListDeterminism();`

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationEntry.cs
jbayardo: This is not the format I decided upon, but what has been used by the ContentLocationDatabase thus far. Since it can't be changed for backwards-compatibility, nothing can be done except for changing the format, which would require backwards compatibility and hence be more code for the same task.

A wild guess at why this is stored in this way is it may take less space to store the long than the entire timestamp.

---
In reply to: [301862358](https://github.com/microsoft/BuildXL/pull/571#discussion_r301862358) [](ancestors = 301862358)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
SergeyTeplyakov: Or maybe we should be more explicit in case of a miss and use something like:

if (result == null) {
    // miss
    return new GetContentHashListResult(null, CacheDeterminism.None);
}


---
In reply to: [302175997](https://github.com/microsoft/BuildXL/pull/571#discussion_r302175997) [](ancestors = 302175997)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,162 @@ private byte[] GetKey(ShortHash hash)
...
+        
+        /// <inheritdoc />
+        public override AddOrGetContentHashListResult AddOrGetContentHashList(OperationContext context, StrongFingerprint strongFingerprint, ContentHashListWithDeterminism contentHashListWithDeterminism)
+        {
+            var key = GetCaChaaSKey(strongFingerprint);
+
+            Possible<AddOrGetContentHashListResult> status = _keyValueStore.Use(
+                store =>
+                {
+                    int maxAttempts = 5;
```
SergeyTeplyakov: I think this deserves a high level comment about the concurrency and the reason we have multiple attempts.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,162 @@ private byte[] GetKey(ShortHash hash)
...
+        }
+
+        private byte[] GetCaChaaSKey(StrongFingerprint strongFingerprint)
+        {
+            return SerializeStrongFingerprint(strongFingerprint);
+        }
+
+        private StrongFingerprint DeserializeStrongFingerprint(byte[] bytes)
+        {
+            return DeserializeWithPool(bytes, (reader) => new StrongFingerprint(reader));
```
SergeyTeplyakov: I would slightly prefer to call it "DeserializeCore". The fact that we have pool there is not that critical. It is just a helper, imo.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,162 @@ private byte[] GetKey(ShortHash hash)
...
+        }
+
+        #endregion
```
SergeyTeplyakov: And I would prefer to avoid regions...

If the region is collapsed it is hard to assess the size of the file. But if they're expanded, then they're useless.

So my rule of thumb is: if you need regions, you have issues in your code.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,162 @@ private byte[] GetKey(ShortHash hash)
...
+        public override IReadOnlyCollection<GetSelectorResult> GetSelectors(OperationContext context, Fingerprint weakFingerprint)
+        {
+            var result = new List<GetSelectorResult>();
+
+            var status = _keyValueStore.Use(
+                store =>
+                {
+                    var key = SerializeWeakFingerprint(weakFingerprint);
+
+                    // This only works because the strong fingerprint serializes the weak fingerprint first. Hence,
```
SergeyTeplyakov: Do we want to make sure this is always the case?

I can see a few options:
1. Add a test to make sure the behavior is enforced.
2. + change the design in a way that this important decision is explicit in the code and not "accidental".

Public/Src/Cache/ContentStore/Distributed/NuCache/InMemory/MemoryContentLocationDatabase.cs
```diff
@@ -49,6 +51,19 @@ internal override void Persist(OperationContext context, ShortHash hash, Content
...
 
+        /// <inheritdoc />
+        public override GetContentHashListResult GetContentHashList(OperationContext context, StrongFingerprint strongFingerprint) => null;
+
+        /// <inheritdoc />
+        public override AddOrGetContentHashListResult AddOrGetContentHashList(
+            OperationContext context,
+            StrongFingerprint strongFingerprint,
+            ContentHashListWithDeterminism contentHashListWithDeterminism) =>
+            null;
```
SergeyTeplyakov: I would prefer to throw, because it will be more obvious that stuff is not implemented...

Public/Src/Cache/ContentStore/DistributedTest/ContentLocation/NuCache/RocksDbContentLocationDatabaseTests.cs
```diff
@@ -0,0 +1,337 @@
...
+            await database.StartupAsync(operationContext).ShouldBeSuccess();
+            database.SetDatabaseMode(isDatabaseWritable: true);
+
+            action(operationContext, database);
+
+            await database.ShutdownAsync(operationContext).ShouldBeSuccess();
+        }
+
+
+        [Fact]
```
SergeyTeplyakov: redundant new line.

Public/Src/Cache/ContentStore/DistributedTest/ContentLocation/NuCache/RocksDbContentLocationDatabaseTests.cs
```diff
@@ -0,0 +1,337 @@
...
+                contentLocationDatabase.AddOrGetContentHashList(
+                    context, strongFingerprint1, contentHashListWithDeterminism1).ShouldBeSuccess();
+                contentLocationDatabase.AddOrGetContentHashList(
+                    context, strongFingerprint2, contentHashListWithDeterminism2).ShouldBeSuccess();
+
+                List<GetSelectorResult> getSelectorResults = contentLocationDatabase.GetSelectors(context, weakFingerprint).ToList();
+                Assert.Equal(2, getSelectorResults.Count);
+
+                GetSelectorResult r1 = getSelectorResults[0];
+                Assert.True(r1.Succeeded);
```
SergeyTeplyakov: Instead of checking for success explicitely, I encourage using this: `GetSelectorResult r1 = getSelectorResults[0].ShouldBeSuccess()`. This saves space in tests, imo, significantly.

Public/Src/Cache/MemoizationStore/Interfaces/Sessions/ContentHashList.cs
```diff
@@ -67,6 +68,36 @@ public class ContentHashList : IEquatable<ContentHashList>
...
+        {
+            Contract.Requires(reader != null);
+            var numHashes = reader.ReadInt32();
+            if (numHashes < 0)
+            {
+                _contentHashes = null;
+            }
+            else
+            {
+                _contentHashes = new ContentHash[numHashes];
```
SergeyTeplyakov: >_contentHashes [](start = 16, length = 14)

It seems that before your change, _contentHashes was a non-nullable field, but now it can be null.

I think if numHashes is < 0, then _contentHashes should be an empty array:

_contentHashes = CollectionUtils.EmptyArray<T>(); (or something).

SergeyTeplyakov: >_contentHashes [](start = 16, length = 14)

Formatting issue.

SergeyTeplyakov: >_contentHashes [](start = 16, length = 14)

I think it make sense to separate preconditions with an empty line.

SergeyTeplyakov: >_contentHashes [](start = 16, length = 14)

I would strongly encourage to put serialization and deserialization logic close to each other.

It must be symetric, so it is easier to make sure this is the case if the methods are close to each otehr.

SergeyTeplyakov: >_contentHashes [](start = 16, length = 14)

I can't see how you can construct an instance with null in _contentHashes.

SergeyTeplyakov: >_contentHashes [](start = 16, length = 14)

Formatting issue.

Public/Src/Cache/MemoizationStore/Interfaces/Sessions/ContentHashList.cs
```diff
@@ -223,5 +254,35 @@ public string Serialize()
...
+        {
+            Contract.Requires(writer != null);
+
+            if (_contentHashes == null)
+            {
+                writer.Write(-1);
+            }
+            else
+            {
+                writer.Write(_contentHashes.Length);
```
SergeyTeplyakov: I think we have a helper, something like 'WriteList'.

Public/Src/Cache/MemoizationStore/Interfaces/Sessions/ContentHashListWithDeterminism.cs
```diff
@@ -20,6 +22,22 @@ public ContentHashListWithDeterminism(ContentHashList contentHashList, CacheDete
             Determinism = determinism;
         }
 
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="ContentHashListWithDeterminism"/> struct from a binary
+        ///     representation
+        /// </summary>
+        public ContentHashListWithDeterminism(BinaryReader reader)
```
SergeyTeplyakov: It seems that this is slighly inconsistent: in most cases we have serialize/deserialize methods, but here we use constructor.

I would prefer having explicit serialization/deserialization methods to be more consistent.

Public/Src/Cache/MemoizationStore/Interfaces/Sessions/Selector.cs
```diff
@@ -24,6 +26,18 @@ public Selector(ContentHash contentHash, byte[] output = null)
             Output = output;
         }
 
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="Selector" /> struct.
+        /// </summary>
+        public Selector(BinaryReader reader)
```
SergeyTeplyakov: The same: let's use Deserialize method here.

Public/Src/Cache/MemoizationStore/Interfaces/Sessions/Selector.cs
```diff
@@ -24,6 +26,18 @@ public Selector(ContentHash contentHash, byte[] output = null)
...
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="Selector" /> struct.
+        /// </summary>
+        public Selector(BinaryReader reader)
+        {
+            Contract.Requires(reader != null);
+            ContentHash = new ContentHash(reader);
+
+            var length = reader.ReadInt32();
+            Output = length == -1 ? null : reader.ReadBytes(length);
```
SergeyTeplyakov: We should extract method to something: SerializeNullableArray and DeserializeNullableArray.

I've noticed the pattern like this a lot.

Public/Src/Cache/MemoizationStore/InterfacesTest/Sessions/ContentHashListTests.cs
```diff
@@ -136,5 +137,47 @@ public void RandomNullPayload()
         {
             Assert.Null(ContentHashList.Random().Payload);
         }
+
+        [Fact]
+        public void SerializeRoundtrip()
+        {
+            var v = ContentHashList.Random();
```
SergeyTeplyakov: I expect to see similar tests for a bunch of types, so I would say that we should have a helper to test serialization roundtrip.

A helper that will take two callbacks to serailize and deserialize the value and the helper will do all the stream management and assertions in it.

Public/Src/Cache/ContentStore/Distributed/NuCache/InMemory/MemoryContentLocationDatabase.cs
jbayardo: In this particular case, the returning of null is ReSharper filling with null, why doesn't matter because this is a mock class and we actually already have a class that serves an IMemoizationSession from memory elsewhere. I'd check if that's not the case as well for the other interfaces you mentioned.

Regarding your question, I'm not sure what you meant by capability-based design. I'd argue against more interfaces any day, as I feel we have enough of them already, many of which make things more confusing and restraining rather than allow for more concise expression of ideas.

See for example IKeyValueStore, which is only used to represent a RocksDB store; it turns into a constraint: many things that RocksDB supports are not usable because you'd need to create a "generic" interface to those, and you also have to provide string and byte[] methods for every single function call, which is kind of pointless since it just moves the type conversion to the class.

---
In reply to: [302163716](https://github.com/microsoft/BuildXL/pull/571#discussion_r302163716) [](ancestors = 302163716)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: CaChaaS for sure! See Sergey's comment on the Discussion

---
In reply to: [302164434](https://github.com/microsoft/BuildXL/pull/571#discussion_r302164434) [](ancestors = 302164434)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: I am being consistent. All other places use Possible<Unit>; this case is specified only because the type is different than what you would usually expect this to return.

---
In reply to: [302175159](https://github.com/microsoft/BuildXL/pull/571#discussion_r302175159) [](ancestors = 302175159)

Public/Src/Cache/MemoizationStore/Interfaces/Sessions/ContentHashList.cs
jbayardo: Great catch! numHashes can never be <0 because it is -1 if _contentHashes == null and the length of the array otherwise. Since the first thing can never happen, it is always the length of an array. Changed accordingly.

---
In reply to: [302181279](https://github.com/microsoft/BuildXL/pull/571#discussion_r302181279) [](ancestors = 302181279)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationEntry.cs
```diff
@@ -101,6 +102,31 @@ public static ContentLocationEntry TryCreateFromRedisValue(byte[] data, UnixTime
             return FromRedisValue(data, lastAccessTime);
         }
 
+        /// <summary>
+        /// Serializes an instance into a binary stream.
+        /// </summary>
+        public void Serialize(BuildXLWriter writer)
```
SergeyTeplyakov: BTW, I think it make sense to move to BuildXLWriter everywhere, because it has some extra features required for high perf implementaiton.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
SergeyTeplyakov: The question is: do you need to know this type? I think I'm fine with using var here.

---
In reply to: [302228489](https://github.com/microsoft/BuildXL/pull/571#discussion_r302228489) [](ancestors = 302228489,302175159)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: This is not an option for GetSelectors, since it may return a few results and the fail. What do you want me to do in this case?

---
In reply to: [302175307](https://github.com/microsoft/BuildXL/pull/571#discussion_r302175307) [](ancestors = 302175307)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,197 @@ private byte[] GetKey(ShortHash hash)
...
+                                // hashes exist. Moreover, we can prune content.
+                            }
+                        });
+
+                    if (!status.Succeeded)
+                    {
+                        return new GetContentHashListResult(status.Failure.CreateException());
+                    }
+
+                    if (result == null)
```
dcwuser: Minor: Small optimization for null-checking: it's better to write (result is null) than (result == null), particularly on a type that overrides equals. This comment applies to a lot of places.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,197 @@ private byte[] GetKey(ShortHash hash)
...
+        
+        private ContentHashListWithDeterminism CacheMiss(CacheDeterminism determinism)
+        {
+            return new ContentHashListWithDeterminism(null, determinism);
+        }
+
+        /// <inheritdoc />
+        public override IEnumerable<StructResult<StrongFingerprint>> EnumerateStrongFingerprints(OperationContext context)
+        {
+            var result = new List<StructResult<StrongFingerprint>>();
```
dcwuser: Minor: Can you use yield here to return an enumerator without getting the whole list first? This is probably minor because constructing the whole list is probably quite fast.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -507,6 +529,197 @@ private byte[] GetKey(ShortHash hash)
...
+        private bool IsContentAvailable(OperationContext context, ContentHashList contentHashList)
+        {
+            // TODO(jubayard): implement. Since this needs to pin stuff, we need it to happen in a higher layer.
+            return true;
+        }
+
+        /// <inheritdoc />
+        public override IReadOnlyCollection<GetSelectorResult> GetSelectors(OperationContext context, Fingerprint weakFingerprint)
+        {
+            var result = new List<GetSelectorResult>();
```
dcwuser: Minor: Same question as above regarding enumerable vs. list.

Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
```diff
@@ -715,6 +715,73 @@ public IBuildXLKeyValueStore CreateSnapshot()
...
+
+                using (var iterator = m_store.NewIterator(columnFamilyInfo.Handle, readOptions))
+                {
+                    if (prefix == null || prefix.Length == 0)
+                    {
+                        iterator.SeekToFirst();
+                    }
+                    else
+                    {
+                        iterator.Seek(prefix);
```
dcwuser: Do we know the big-O of this Seek? Hopefully O(1) or O(ln N), not O(N).

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: That was my initial response as well. It is doable, but more work without any clear benefit. There are two ways to do it:

1. Implement an iterator that holds the key value store lock throughout until nothing more can be retrieved from the prefix search. This is potentially dangerous, because we have no idea what the user is doing with the enumeration, so potential locks could cause issues with other parts of the system.
2. An iterator that keeps taking and releasing the DB lock when necessary. This may be problematic as well because the iterator could fail in a strange ways. A quick example: if a task is iterating, then goes dark, and before it tries to fetch a new element the database is restored from a different place, we will close it while the iteration is still going on and hence the next fetch for the enumerable may fail in a weird manner.

There doesn't seem to be any good solution, so I just bit the bullet.

---
In reply to: [302767227](https://github.com/microsoft/BuildXL/pull/571#discussion_r302767227) [](ancestors = 302767227)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: This is the same scenario as I described in the other response.

---
In reply to: [302767413](https://github.com/microsoft/BuildXL/pull/571#discussion_r302767413) [](ancestors = 302767413)

Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
jbayardo: If the entire structure was stored in memory, then RocksDB would take O(lg N) to seek.

Given that its not, this is a very hard question to answer, especially because performance is likely dominated by disk seeks, so read amplification, cache friendliness, and others need to be taken into consideration.

In any case, the performance shouldn't be too far from a Get() call, but it is still likely to be the performance bottleneck of GetSelectors(), so if that turns into an issue we may need to spend time thinking how to optimize.

---
In reply to: [302769601](https://github.com/microsoft/BuildXL/pull/571#discussion_r302769601) [](ancestors = 302769601)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationEntry.cs
ljcollins25: Julian's guess is correct. This is for every entry in the db so space savings are very important here.

---
In reply to: [302176571](https://github.com/microsoft/BuildXL/pull/571#discussion_r302176571) [](ancestors = 302176571,301862358)

Public/Src/Cache/MemoizationStore/Interfaces/Sessions/ContentHashList.cs
```diff
@@ -223,5 +225,56 @@ public string Serialize()
 
             return sb.ToString();
         }
+
+        /// <summary>
+        ///     Serialize whole value to a binary writer.
+        /// </summary>
+        public void Serialize(BuildXLWriter writer)
```
ljcollins25: >Serialize [](start = 20, length = 9)

I would move any serialization logic as an extension method in the assembly that uses it provided that the data is available. I don't think these types should be exposing a dependency on BuildXL.Utilities.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -97,7 +120,7 @@ private BoolResult Load(OperationContext context, StoreSlot activeSlot, bool cle
                 Tracer.Info(context, $"Creating rocksdb store at '{storeLocation}'.");
 
                 var possibleStore = KeyValueStoreAccessor.Open(storeLocation,
-                    additionalColumns: new[] { nameof(ClusterState) });
+                    additionalColumns: new[] { nameof(ClusterState), "Metadata" });
```
ljcollins25: >"Metadata" [](start = 69, length = 10)

nameof(Columns.Metadata)

PR #572, opened 2019-07-10
Remove duplicate Pin finished trace from BlobContentSession

JuanCarlosGI:
Since the implementation consists of calling PinBulk, and PinBulk has its own end trace, this results in a duplicate trace.

Example:
PinBulk:     2019-07-09 20:13:34,178 325 DEBUG 3e9892f3-44b4-4039-9224-d98d565e6024 BlobContentSession.PinAsync stop 449.7661ms. result=[Success]. Hashes=[VSO0:CAA07181F68F78C30A56E0AAD64C3D4D541ECA9FDC7EABBAA580B8244FF8543800:Success]
Pin:            2019-07-09 20:13:34,178 325 DEBUG 3e9892f3-44b4-4039-9224-d98d565e6024 BlobContentSession.PinAsync stop 453.5394ms. result=[Success]. input=[VSO0:CAA07181F68F78C30A56E0AAD64C3D4D541ECA9FDC7EABBAA580B8244FF8543800]

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
+++ b/Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
@@ -38,6 +38,9 @@ namespace BuildXL.Cache.ContentStore.Sessions
         public string Name { get; }
 
         /// <nodoc />
+        protected virtual bool TracePinFinished => true;
+
+        /// <nodoc />
         protected ContentSessionBase(string name)
         {
             Name = name;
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
```diff
@@ -80,6 +83,7 @@ protected ContentSessionBase(string name)
                     Tracer,
                     () => PinCoreAsync(operationContext, contentHash, urgencyHint, _counters[ContentSessionBaseCounters.PinRetries]),
                     traceOperationStarted: false,
+                    traceOperationFinished: TracePinFinished,
```
SergeyTeplyakov: Actually, I think we need to stop tracing the start.
Currently, we're adding 'extraEndMessage' so with your change we'll lose the hash altogether.

SergeyTeplyakov: And because we already skipping the start it means that this trace is completely skipped.

Which is fine, but definitely deserves a comment saying that we don't have to trace because of code duplication.

---
In reply to: [302249447](https://github.com/microsoft/BuildXL/pull/572#discussion_r302249447) [](ancestors = 302249447)

JuanCarlosGI: I left a comment stating that in the override of TracePinFinished in BlobContentSession

PR #574, opened 2019-07-10
 Implement Integrated Windows Authentication for cache on CoreCLR

almiliMSFT:
The .NET Core platform cannot ask the OS for the username of the current user, so instead, the username has to be provided explicitly ([source](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Integrated-Windows-Authentication)).  To work around that limitation, when running on .NET Core we p-invoke `GetUserNameExW` to get that info.

[AB#1558017](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1558017)


Public/Src/Cache/ContentStore/Vsts/VssCredentialsFactory.cs
```diff
@@ -91,8 +114,12 @@ public async Task<VssCredentials> CreateVssCredentialsAsync(Uri baseUri, bool us
                 return _helper.GetPATCredentials(_pat);
             }
 
-            return await _helper.GetCredentialsAsync(baseUri, useAad, _credentialBytes, null)
-                .ConfigureAwait(false);
+#if FEATURE_CORECLR
```
almiliMSFT:
```suggestion
// The .NET Core platform cannot ask the OS for the AAD username of the
// current user, so instead, the username has to be provided explicitly.
#if FEATURE_CORECLR
```

dannyvv:
@almili@microsoft.com 
Did you consider PInvoking the windows api that whoami uses?

From reading the docs it seems to match the uderlying api: GetUserNameEx: 
[https://docs.microsoft.com/en-us/windows/win32/api/secext/nf-secext-getusernameexw](https://docs.microsoft.com/en-us/windows/win32/api/secext/nf-secext-getusernameexw) 

I would be great if we can just pinvoke it from C# on windows and don't have to worry at all.

Public/Src/Cache/ContentStore/Vsts/VssCredentialsFactory.cs
```diff
@@ -91,6 +127,13 @@ public async Task<VssCredentials> CreateVssCredentialsAsync(Uri baseUri, bool us
                 return _helper.GetPATCredentials(_pat);
             }
 
+            // If the user name is explicitly provided call a different auth method that's
+            // not going to query the OS for the AAD user name (which is, btw, disallowed on CoreCLR)
```
almiliMSFT:
```suggestion
            // not going to query the OS for the AAD user name (which is, btw, disallowed on CoreCLR).
```

PR #575, opened 2019-07-10
Allow relative paths in ContentStoreApp hashfile verb

nayanshah:


Public/Src/Cache/ContentStore/App/DedupHashContent.cs
```diff
@@ -30,7 +30,7 @@ internal sealed partial class Application
         ///     Hash some files and optionally display their chunks
         /// </summary>
         [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling")]
-        [Verb(Aliases = "hf", Description = "Hash content files")]
+        [Verb(Aliases = "dhf", Description = "Hash content files")]
```
kewilli: >dhf [](start = 25, length = 3)

While this is technically a breaking change, I don't think anyone is using this verb today.

nayanshah: Yeah, but multiple verbs had the same alias so this never worked.

PR #576, opened 2019-07-11
Make sure path is fully canonicalized when doing write in existing file analysis

smera:
Paths like \\?\C:\foo.txt and C:\foo.txt are reported as different paths from detours. This PR changes the logic that deals with OverrideAllowWriteForExistingFiles to handle those paths as the same one.

### DIFF
--- a/Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
+++ b/Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
@@ -27,6 +27,7 @@ namespace Test.BuildXL.Executables.TestProcess
         private const string StdErrMoniker = "stderr";
         private const string WaitToFinishMoniker = "wait";
         private const string AllUppercasePath = "allUpper";
+        private const string UseLongPathPrefix = "useLongPathPrefix";
 
         /// <summary>
         /// Returns an <code>IEnumerable</code> containing <paramref name="operation"/>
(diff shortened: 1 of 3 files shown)


Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -473,11 +474,15 @@ public static Operation CreateDir(FileOrDirectoryArtifact path, bool doNotInfer
         /// Creates a write file operation that appends. The file is created if it does not exist.
         /// Writes random content to file at path if no content is specified.
         /// </summary>
-        public static Operation WriteFile(FileArtifact path, string content = null, bool doNotInfer = false, bool changePathToAllUpperCase = false)
+        public static Operation WriteFile(FileArtifact path, string content = null, bool doNotInfer = false, bool changePathToAllUpperCase = false, bool useLongPathPrefix = false)
         {
+            Contract.Assert(!changePathToAllUpperCase || !useLongPathPrefix, "Cannot specify changePathToAllUpperCase and useLongPathPrefix simultaneously");
+
+            string additionalArgs = changePathToAllUpperCase ? Operation.AllUppercasePath : useLongPathPrefix ? Operation.UseLongPathPrefix : null;
```
dannyvv: would you mind adding parenthesis (or improve multi-line) to indicate if the ? operator is left or right assiciative?

Public/Src/Utilities/UnitTests/Executables/TestProcess/Operation.cs
```diff
@@ -801,6 +806,10 @@ private void DoWriteFile(string file, string content)
                 {
                     file = file.ToUpperInvariant();
                 }
+                if (AdditionalArgs == UseLongPathPrefix)
+                {
+                    file = "\\\\?\\" + file.ToUpperInvariant();
```
dannyvv:
```suggestion 
@"\\?\"
```

PR #579, opened 2019-07-11
Allow Delete to wait for the lock on a hash

kewilli:
Allow Delete to wait for the lock on a hash


Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
```diff
@@ -229,6 +229,7 @@ public void DeleteDirectory(AbsolutePath path, DeleteOptions deleteOptions)
             catch (UnauthorizedAccessException accessException)
             {
                 if ((deleteOptions & DeleteOptions.ReadOnly) != 0 &&
+                    accessException.HResult > 0 &&
```
SergeyTeplyakov: AccessDenied is greater than 0, so why do we need to have an explicit check for > 0?

SergeyTeplyakov:
General question: do we really need to configure the new behavior? It seems that this is the new feature and we can decide to wait all the time or never wait.
Do we really need to increase the complexity by adding an option to control this aspect?

Public/Src/Cache/ContentStore/Library/FileSystem/PassThroughFileSystem.cs
kewilli: Locally, I was seeing an overflow exception in the following cast to `uint`. I don't know how `HResult` became negative.

kewilli:
> General question: do we really need to configure the new behavior? It seems that this is the new feature and we can decide to wait all the time or never wait.
> Do we really need to increase the complexity by adding an option to control this aspect?

The next step is to add a verb to test this out more. After this, we can reduce the code down from two options to one.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -2075,18 +2075,18 @@ public Task<EvictResult> EvictAsync(Context context, ContentHashWithLastAccessTi
...
         }
 
-        private async Task<EvictResult> EvictCoreAsync(Context context, ContentHashWithLastAccessTimeAndReplicaCount contentHashInfo, bool force, bool onlyUnlinked, Action<long> evicted)
+        private async Task<EvictResult> EvictCoreAsync(Context context, ContentHashWithLastAccessTimeAndReplicaCount contentHashInfo, bool force, bool onlyUnlinked, Action<long> evicted, bool ensureSuccess = false)
         {
             ContentHash contentHash = contentHashInfo.ContentHash;
 
             long pinnedSize = 0;
-            using (LockSet<ContentHash>.LockHandle? contentHashHandle = _lockSet.TryAcquire(contentHash))
+            using (LockSet<ContentHash>.LockHandle? contentHashHandle = ensureSuccess ? await _lockSet.AcquireAsync(contentHash) : _lockSet.TryAcquire(contentHash))
```
SergeyTeplyakov: I'm not sure about the name `ensureSuccess` because it is not quite the case here. The flag just blocks the operaiton.

So maybe: `acquireLock`? or something?

kewilli: I mean, sure. Works for me!

PR #580, opened 2019-07-11
Fix crash when we send build request to the dead worker

semihokur:


Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -618,7 +618,24 @@ public async Task SendToRemote(OperationContext operationContext, RunnablePip ru
                 SequenceNumber = Interlocked.Increment(ref m_nextSequenceNumber),
             };
 
-            m_buildRequests.Add(ValueTuple.Create(pipCompletionTask, pipBuildRequest));
+            try
+            {
+                m_buildRequests.Add(ValueTuple.Create(pipCompletionTask, pipBuildRequest));
+            }
+            catch (InvalidOperationException)
```
dannyvv: >InvalidOperationException [](start = 19, length = 25)

Is that the only case this throws invalidoperationexception? Should we check for a stronger signal that this i sthe case and not something else?

semihokur: good point. That was the only case; but it is always better to confirm whether the connection is really lost by checking 'IsConnectionLost' variable.

PR #587, opened 2019-07-12
Enable usage of Azure Blob Storage SAS tokens

jbayardo:
This allows any implementor of IDistributedContentHost to hand out Azure Blob SAS tokens, and sets precedent for further work on implementing new authentication options for CaSaaS.

This change is not backwards compatible, and will require changes in repositories that implement this logic.

[WIP] Figure out a way to test the changes


Public/Src/Cache/ContentStore/App/DistributedService.cs
```diff
@@ -118,9 +118,40 @@ public void OnTeardownCompleted()
...
+                        // Environment variables are null by default. Skip if that's the case.
+                        continue;
+                    }
+
+                    switch (request.Kind)
+                    {
+                        case CredentialsKind.AzureBlobPlainText:
+                            credentials = new AzureBlobStorageCredentials(secret);
+                            break;
+                        case CredentialsKind.EventHubPlainText:
```
SergeyTeplyakov: Do we really need to separate the sources? For instance, why not to use the same set of secret kinds that css provides? Like `PlainText`, `Sas` etc.

Public/Src/Cache/ContentStore/App/DistributedService.cs
```diff
@@ -118,9 +118,40 @@ public void OnTeardownCompleted()
...
+                            credentials = new EventHubCredentials(secret);
+                            break;
+                        case CredentialsKind.RedisPlainText:
+                            credentials = new RedisCredentials(secret);
+                            break;
+                        case CredentialsKind.AzureBlobSASToken:
```
SergeyTeplyakov: Could you explain why this is not supported?

I thought we can easily implement this already here... What prevents us from doing that?

Public/Src/Cache/ContentStore/App/DistributedService.cs
```diff
@@ -118,9 +118,40 @@ public void OnTeardownCompleted()
...
+                            continue;
+                    }
+
+                    result[request.Name] = credentials;
```
SergeyTeplyakov: Two options: assert that `credentials` is not null. Throw in default case in the switch block or (preferrably) do both.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStoreConfiguration.cs
```diff
@@ -280,6 +358,22 @@ public CloudBlobClient CreateCloudBlobClient()
         {
             return CreateCloudStorageAccount().CreateCloudBlobClient();
         }
+
+        /// <inheritdoc />
+        public override bool IsOfKind(CredentialsKind kind)
+        {
+            if (!string.IsNullOrEmpty(ConnectionString))
```
SergeyTeplyakov: I would prefer to compute this as a field in the constructor.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -310,7 +311,7 @@ private async Task ApplySecretSettingsForLlsAsync(RedisContentLocationStoreConfi
 
             var eventStoreConfiguration = new EventHubContentLocationEventStoreConfiguration(
                 eventHubName: "eventhub",
-                eventHubConnectionString: GetRequiredSecret(secrets, _distributedSettings.EventHubSecretName),
+                eventHubConnectionString: (GetRequiredSecret(secrets, _distributedSettings.EventHubSecretName) as EventHubCredentials).ConnectionString,
```
SergeyTeplyakov: Not clear why you cast the result to a specific type in case of redis, but use 'as' cast here.

Public/Src/Cache/DistributedCache.Host/Service/IDistributedCacheServiceHost.cs
```diff
@@ -2,11 +2,30 @@
...
 using System.Threading;
 using System.Threading.Tasks;
+using BuildXL.Cache.ContentStore.Distributed;
 
 namespace BuildXL.Cache.Host.Service
 {
+    /// <summary>
+    /// Type to signal the host which kind of secret is expected to be returned
+    /// </summary>
+    public struct RetrieveSecretsRequest
```
dannyvv: would you mind placing in its own file (or make it a nested type otherwise)

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -320,43 +321,37 @@ private async Task ApplySecretSettingsForLlsAsync(RedisContentLocationStoreConfi
...
             var storageSecretNames = GetAzureStorageSecretNames(errorBuilder);
-            if (storageSecretNames == null)
-            {
-                return null;
-            }
-
-            return storageSecretNames.Select(name => {
-                return new AzureBlobStorageCredentials(GetRequiredSecret(settings, name));
-            }).ToArray();
+            return storageSecretNames?.Select(name => GetRequiredSecret(secrets, name) as AzureBlobStorageCredentials).ToArray();
```
dannyvv: do you need to filter out null here?

dannyvv: I think there is actually an operator for that, let me check...

---
In reply to: [303156470](https://github.com/microsoft/BuildXL/pull/587#discussion_r303156470) [](ancestors = 303156470)

dannyvv: [https://stackoverflow.com/questions/4015930/when-to-use-cast-and-oftype-in-linq](https://stackoverflow.com/questions/4015930/when-to-use-cast-and-oftype-in-linq)

---
In reply to: [303156549](https://github.com/microsoft/BuildXL/pull/587#discussion_r303156549) [](ancestors = 303156549,303156470)

Public/Src/Cache/ContentStore/App/DistributedService.cs
jbayardo: It is required to create the corresponding object for the authentication method, which means that the other side needs to know what is needed in each case; it is also helpful to validate that the other side is returning the expected objects and not others. The alternative is to keep two separate enums, one { EventHub, Redis, AzureBlob } and another { SAS, PAT }

---
In reply to: [303139192](https://github.com/microsoft/BuildXL/pull/587#discussion_r303139192) [](ancestors = 303139192)

Public/Src/Cache/ContentStore/App/DistributedService.cs
```diff
@@ -118,9 +121,60 @@ public void OnTeardownCompleted()
...
+                            credentials = new AzureBlobStorageCredentials(new StorageCredentials(sasToken), account.Credentials.AccountName);
+                            break;
+                        case CredentialsKind.EventHubPlainText:
+                            credentials = new EventHubCredentials(secret);
+                            break;
+                        case CredentialsKind.RedisPlainText:
+                            credentials = new RedisCredentials(secret);
+                            break;
+                        default:
+                            throw new NotImplementedException("It is expected that all supported credential kinds be handled when creating a DistributedService.");
```
SergeyTeplyakov: This should be not supported, but not not implemented.

Public/Src/Cache/DistributedCache.Host/Service/RetrieveSecretsRequest.cs
```diff
@@ -0,0 +1,25 @@
...
+using BuildXL.Cache.ContentStore.Distributed;
+
+namespace BuildXL.Cache.Host.Service
+{
+    /// <summary>
+    /// Type to signal the host which kind of secret is expected to be returned
+    /// </summary>
+    public struct RetrieveSecretsRequest
+    {
+        public string Name { get; }
```
SergeyTeplyakov: /// <nodoc />

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -372,32 +369,51 @@ private List<string> GetAzureStorageSecretNames(StringBuilder errorBuilder)
...
+            {
+                return null;
+            }
+
+            // Validate requests match as expected
+            foreach (var request in retrieveSecretsRequests)
+            {
+                if (secrets.TryGetValue(request.Name, out var credential) && !credential.IsOfKind(request.Kind))
+                {
+                    throw new SecurityException($"The credentials produced by the host for secret named {request.Name} do not match the expected kind");
```
SergeyTeplyakov: I think this if block means that the secret was missing or the kind is different.

But the error message states only one case here.

I'm not sure how exactly your CB implementation would looks like, but DistributedService implementation can skip some secrets if they're unavailable.

SergeyTeplyakov: So maybe it make sense to separate two cases: the secret is missing, and the kind is different.

And I think that the first case is misconfiguration, but the second type of errors is actually a bug.

So we can use exceptions for the first case, but use `Contract.Assert` for the second one.

---
In reply to: [303561809](https://github.com/microsoft/BuildXL/pull/587#discussion_r303561809) [](ancestors = 303561809)

Public/Src/Cache/DistributedCache.Host/Service/IDistributedCacheServiceHost.cs
```diff
@@ -36,6 +37,6 @@ public interface IDistributedCacheServiceHost
         /// <summary>
         /// Retrieves secrets from key vault
         /// </summary>
-        Task<Dictionary<string, string>> RetrieveKeyVaultSecretsAsync(List<string> secrets, CancellationToken token);
+        Task<Dictionary<string, Credentials>> RetrieveKeyVaultSecretsAsync(List<RetrieveSecretsRequest> requests, CancellationToken token);
```
SergeyTeplyakov: Let's discuss this offline, but I do believe that the result should be a secret not credentials.

And I still believe that it make sense to make the interface between bxl and CB as small as possible and it means that I would prefer to deal with type 'SecretKind' = PlaintText | Sas; and the result could be just a string actually. And it is a client's responsibility to parse the string appropriately.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
jbayardo: The two cases are separated. You can only go inside the if block if the secret is present in the returned ones, and the kind does not match. For example, if you requested secret A of kind AzureBlobSAS, and it was found, but the actual kind was AzureBlobPAT, then this error would trigger.

The missing case is handled when the secret itself is actually requested, i.e. all calls to GetRequiredSecret

---
In reply to: [303562197](https://github.com/microsoft/BuildXL/pull/587#discussion_r303562197) [](ancestors = 303562197,303561809)

Public/Src/Cache/DistributedCache.Host/Service/IDistributedCacheServiceHost.cs
jbayardo: I can do this, but the trade-off is we will need to expose an updating token type and will need to implement glue with IUpdatingPlainText/etc in the CloudBuild side.

---
In reply to: [303562855](https://github.com/microsoft/BuildXL/pull/587#discussion_r303562855) [](ancestors = 303562855)

Public/Src/Cache/ContentStore/App/DistributedService.cs
```diff
@@ -118,9 +121,80 @@ public void OnTeardownCompleted()
...
+                    {
+                        case SecretKind.PlainText:
+                            // In this case, the value is expected to be an entire connection string
+                            secret = new PlainTextSecret(secretValue);
+                            break;
+                        case SecretKind.SasToken:
+                            secret = CreateSasTokenSecret(request, secretValue);
+                            break;
+                        default:
+                            throw new NotSupportedException("It is expected that all supported credential kinds be handled when creating a DistributedService.");
```
SergeyTeplyakov: very minor: but could you please put 'request.Kind' into the error message?

Public/Src/Cache/ContentStore/Distributed/NuCache/Secrets.cs
```diff
@@ -0,0 +1,74 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics.ContractsLight;
+
+namespace BuildXL.Cache.ContentStore.Distributed
+{
+    /// <nodoc />
+    public enum SecretKind
```
SergeyTeplyakov: To be honest, I do like this design a lot!

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -320,43 +321,66 @@ private async Task ApplySecretSettingsForLlsAsync(RedisContentLocationStoreConfi
                 value => eventStoreConfiguration.MaxEventProcessingConcurrency = value);
         }
 
-        private AzureBlobStorageCredentials[] GetStorageCredentials(Dictionary<string, string> settings, StringBuilder errorBuilder)
+        private AzureBlobStorageCredentials[] GetStorageCredentials(Dictionary<string, Secret> secrets, StringBuilder errorBuilder)
         {
             var storageSecretNames = GetAzureStorageSecretNames(errorBuilder);
-            if (storageSecretNames == null)
+            // This would have failed earlier otherwise
+            Contract.Requires(storageSecretNames != null);
```
SergeyTeplyakov: nit: Requires should be used for arguments, here we should use Assert.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -320,43 +321,66 @@ private async Task ApplySecretSettingsForLlsAsync(RedisContentLocationStoreConfi
...
-                return null;
+                var secret = GetRequiredSecret(secrets, secretName);
+
+                if (!_distributedSettings.AzureBlobStorageUseSasTokens && secret is PlainTextSecret plainTextSecret)
+                {
+                    credentials.Add(new AzureBlobStorageCredentials(plainTextSecret.Secret));
+                }
+                else if (_distributedSettings.AzureBlobStorageUseSasTokens && secret is UpdatingSasToken updatingSasToken)
+                {
+                    var storageCredentials = new StorageCredentials(sasToken: updatingSasToken.Token.Token);
```
SergeyTeplyakov: nit: I would suggest to move this into a helper/factory method.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -320,43 +321,66 @@ private async Task ApplySecretSettingsForLlsAsync(RedisContentLocationStoreConfi
...
+                    {
+                        storageCredentials.UpdateSASToken(sasToken.Token);
+                    };
+
+                    // The account name should never actually be updated, so its OK to take it from the initial token
+                    credentials.Add(new AzureBlobStorageCredentials(storageCredentials, updatingSasToken.Token.StorageAccount));
+                }
+                else
+                {
+                    throw new NotSupportedException($"Unexpected secret kind retrieved for requested secret {secretName}");
```
SergeyTeplyakov: I think you can actually assert that if AzureBlobStorageUseSasTokens is true then secret **must be** of type UpdatingSasToken.

I think if this is not true, then we have a bug in the code.

And we can use assertions to distinguish bugs from expected exceptions.

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -372,32 +396,67 @@ private List<string> GetAzureStorageSecretNames(StringBuilder errorBuilder)
...
 
+            // Create the credentials requests
+            var retrieveSecretsRequests = new List<RetrieveSecretsRequest>();
+
             var storageSecretNames = GetAzureStorageSecretNames(errorBuilder);
             if (storageSecretNames == null)
             {
                 return null;
             }
+            var azureBlobStorageCredentialsKind = _distributedSettings.AzureBlobStorageUseSasTokens ? SecretKind.SasToken : SecretKind.PlainText;
```
SergeyTeplyakov: very nit: formatting. Missing empty line after the block.

PR #588, opened 2019-07-12
Pin optimizations for DedupContentStore

JuanCarlosGI:
Since pin is a potentially expensive operation in DedupContentStore, optimizations have been made.

Before deciding if a pin is going to be made, there is a quick check of the current keepUntil value of the tree.
Depending on the current keepUntil:
- The pin operation will be ignored if it is greater than some threshold
- The pin operation will be inlined if it is lower than some other threshold
- The pin operation will be done asynchronously and will return success otherwise.

### DIFF
--- a/Public/Src/Cache/ContentStore/Vsts/BackingContentStore.cs
+++ b/Public/Src/Cache/ContentStore/Vsts/BackingContentStore.cs
@@ -39,6 +39,8 @@ namespace BuildXL.Cache.ContentStore.Vsts
         private readonly IAbsFileSystem _fileSystem;
         private readonly IArtifactHttpClientFactory _artifactHttpClientFactory;
         private readonly TimeSpan _timeToKeepContent;
+        private readonly TimeSpan _pinInlineThreshold;
+        private readonly TimeSpan _ignorePinThreshold;
         private IArtifactHttpClient _artifactHttpClient;
         private readonly bool _useDedupStore;
 
@@ -53,12 +55,16 @@ namespace BuildXL.Cache.ContentStore.Vsts
         /// <param name="fileSystem">Filesystem used to read/write files.</param>
         /// <param name="artifactHttpClientFactory">Backing Store HTTP client factory.</param>
         /// <param name="timeToKeepContent">Minimum time-to-live for accessed content.</param>
+        /// <param name="pinInlineThreshold">Maximum time-to-live to inline pin calls.</param>
+        /// <param name="ignorePinThreshold">Minimum time-to-live to ignore pin calls.</param>
         /// <param name="downloadBlobsThroughBlobStore">Flag for BlobStore: If enabled, gets blobs through BlobStore. If false, gets blobs from the Azure Uri.</param>
         /// <param name="useDedupStore">Determines whether or not DedupStore is used for content. Must be used in tandem with Dedup hashes.</param>
         public BackingContentStore(
             IAbsFileSystem fileSystem,
             IArtifactHttpClientFactory artifactHttpClientFactory,
             TimeSpan timeToKeepContent,
+            TimeSpan pinInlineThreshold,
+            TimeSpan ignorePinThreshold,
             bool downloadBlobsThroughBlobStore = false,
             bool useDedupStore = false)
         {
@@ -67,6 +73,8 @@ namespace BuildXL.Cache.ContentStore.Vsts
             _fileSystem = fileSystem;
             _artifactHttpClientFactory = artifactHttpClientFactory;
             _timeToKeepContent = timeToKeepContent;
+            _pinInlineThreshold = pinInlineThreshold;
+            _ignorePinThreshold = ignorePinThreshold;
             _downloadBlobsThroughBlobStore = downloadBlobsThroughBlobStore;
             _useDedupStore = useDedupStore;
         }
@@ -140,7 +148,7 @@ namespace BuildXL.Cache.ContentStore.Vsts
             if (_useDedupStore)
             {
                 return new CreateSessionResult<IReadOnlyContentSession>(new DedupReadOnlyContentSession(
-                    _fileSystem, name, implicitPin, _artifactHttpClient as IDedupStoreHttpClient, _timeToKeepContent));
+                    _fileSystem, name, implicitPin, _artifactHttpClient as IDedupStoreHttpClient, _timeToKeepContent, _pinInlineThreshold, _ignorePinThreshold));
             }
 
             return new CreateSessionResult<IReadOnlyContentSession>(new BlobReadOnlyContentSession(
@@ -154,7 +162,7 @@ namespace BuildXL.Cache.ContentStore.Vsts
             if (_useDedupStore)
             {
                 return new CreateSessionResult<IContentSession>(new DedupContentSession(
-                    context, _fileSystem, name, implicitPin, _artifactHttpClient as IDedupStoreHttpClient, _timeToKeepContent));
+                    context, _fileSystem, name, implicitPin, _artifactHttpClient as IDedupStoreHttpClient, _timeToKeepContent, _pinInlineThreshold, _ignorePinThreshold));
             }
 
             return new CreateSessionResult<IContentSession>(new BlobContentSession(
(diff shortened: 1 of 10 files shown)


Public/Src/Cache/ContentStore/Vsts/BackingContentStore.cs
```diff
@@ -39,6 +39,8 @@ public enum SessionCounters
         private readonly IAbsFileSystem _fileSystem;
         private readonly IArtifactHttpClientFactory _artifactHttpClientFactory;
         private readonly TimeSpan _timeToKeepContent;
+        private readonly TimeSpan _pinInlineThreshold;
```
dcwuser: Comments here to explain how all this should work and what role each parameter plays.

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -155,14 +172,55 @@ public class DedupReadOnlyContentSession : ContentSessionBase
...
             {
-                var pinResult = CheckPinInMemory(contentHash);
-                if (pinResult.Succeeded)
-                {
-                    return pinResult;
-                }
+                return pinResult;
+            }
+
+            VstsBlobIdentifier blobId = ToVstsBlobIdentifier(contentHash.ToBlobIdentifier());
```
dcwuser: Actually, this is the place for a paragraph of comments describing the basic approach.

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -155,14 +172,55 @@ public class DedupReadOnlyContentSession : ContentSessionBase
...
+            var task = PinImplAsync(context, contentHash);
+
+            if (timeLeft < _pinInlineThreshold)
+            {
+                Tracer.Debug(context, $"Pin inlined bacause keepUntil has remaining time [{timeLeft}] that is less than the pinInlineThreshold");
+                _dedupCounters[Counters.PinInlined].Increment();
+                return await task;
+            }
+
+            task.FireAndForget(context);
```
dcwuser: Is there a way to count this case? Say be subtracting the other counts from an overall counter?

JuanCarlosGI: Definitely possible. I wanted to avoid tracing in this case since I expect most calls to fall into it and I don't want to overwhelm the log with this trace. I could add another counter if you think it's necessary, but it could be possible to just calculate it fairly simply

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -312,21 +371,17 @@ public class DedupReadOnlyContentSession : ContentSessionBase
...
             OperationContext context, IReadOnlyList<ContentHash> contentHashes, UrgencyHint urgencyHint, Counter retryCounter, Counter fileCounter)
         {
             try
             {
-                return await Workflows.RunWithFallback(
-                    contentHashes,
-                    hashes => CheckInMemoryCache(hashes),
-                    hashes => UpdateDedupStoreAsync(context, hashes),
-                    result => result.Succeeded);
+                return Task.FromResult(contentHashes.Select(async (contentHash, i) => (await PinAsync(context, contentHash, context.Token, urgencyHint)).WithIndex(i)));
```
dcwuser: Nice to see unnecessary async/awiat removed.

dcwuser:
>         public const int DefaultDaysToKeepContentBags = 7;

Change this to keep 1.5-2x default ignore threshold?

---
Refers to: Public/Src/Cache/MemoizationStore/Vsts/BuildCacheServiceConfiguration.cs:36 in a1a9cce. [](commit_id = a1a9cce5a23005af26aedf0c644557bffc6b9807, deletion_comment = False)

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -118,13 +128,17 @@ public class DedupReadOnlyContentSession : ContentSessionBase
...
+        /// <param name="pinInlineThreshold">Maximum time-to-live to inline pin calls.</param>
+        /// <param name="ignorePinThreshold">Minimum time-to-live to ignore pin calls.</param>
         /// <param name="maxConnections">The maximum number of outboud connections to VSTS.</param>
         public DedupReadOnlyContentSession(
             IAbsFileSystem fileSystem,
             string name,
             ImplicitPin implicitPin,
             IDedupStoreHttpClient dedupStoreHttpClient,
             TimeSpan timeToKeepContent,
+            TimeSpan pinInlineThreshold,
```
SergeyTeplyakov: I'm fine so far, but maybe it's time to add a configuration class, something like DedupConfiguration?

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -155,14 +172,62 @@ public class DedupReadOnlyContentSession : ContentSessionBase
...
+            {
+                // Content not found.
+                return new PinResult(PinResult.ResultCode.ContentNotFound);
+            }
 
+            var timeLeft = keepUntilResult.Value.Value - DateTime.UtcNow;
+
+            if (timeLeft > _ignorePinThreshold)
+            {
+                Tracer.Debug(context, $"Pin ignored bacause keepUntil has remaining time [{timeLeft}] that is greater than the ignorePinThreshold");
```
SergeyTeplyakov: Maybe it make sense to call it "Pin was skipped", instead of "ignored"? I'm fine with the current terms, but I'm curious maybe "skip" here is a better term.

SergeyTeplyakov: I think it make sense to print 'ignorePinThreshold' and 'pinInlineThreshold' here and before the next return on line 217.

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -155,14 +172,62 @@ public class DedupReadOnlyContentSession : ContentSessionBase
...
+                _dedupCounters[Counters.PinIgnored].Increment();
+                return PinResult.Success;
+            }
+
+            var task = PinImplAsync(context, contentHash);
```
SergeyTeplyakov: nit: pinTask?

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -155,14 +172,62 @@ public class DedupReadOnlyContentSession : ContentSessionBase
...
-                }
+                return pinResult;
+            }
+
+            // Since pinning the whole tree can be an expensive operation, we have optimized how we call it. Depending on the current
+            //  keepUntil of the root node, which is unexpensive to check, the operation will behave differently:
+            //      The pin operation will be ignored if it is greater than ignorePinThreshold, to reduce amount of calls
+            //      The pin operation will be inlined if it is lower than pinInlineThreshold, to make sure that we don't try to use
+            //          content that we pin in the background but has expired before we could complete the pin.
+            //      The pin operation will be done asynchronously and will return success otherwise. Most calls should follow this
```
SergeyTeplyakov: I'm a little bit concerned about this pattern, because if the pin will fail, it won't be possible to handle it.
And as you mentioned in this comment, this is the most common case.

Do we know that this is just fine to ignore the results for such a crucial operation?

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
JuanCarlosGI: Right now I don't think there are enough things different in Dedup and Blob constructors, so maybe after some other differences emerge we can reconsider adding something like DedupConfiguration

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
JuanCarlosGI: Discussed in person and was cleared up in the comment

PR #589, opened 2019-07-12
Untrack Detours internal files during pip execution

narasamdya:
When a Detoured pip includes loading Detours (like Vanguard component of code coverage), pip can access internal files that BuildXL engine itself uses. 

For example, the pip can access DetoursServices.pdb in the bin directory or in the server deployment directory of BuildXL, and this causes DFA. 

If a pip indeed depends on those Detours internal files, then the pip should not refer to the ones in the bin directory of the running BuildXL because they are untracked. The proper way is to consume them is from our released Detours package.

This PR pending on repro and testing data from customer.

[AB#1568098](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1568098)

### DIFF
--- a/Public/Src/Engine/Processes/ExternalToolSandboxedProcessExecutor.cs
+++ b/Public/Src/Engine/Processes/ExternalToolSandboxedProcessExecutor.cs
@@ -42,6 +42,10 @@ namespace BuildXL.Processes
             }
 
             string directory = Path.GetDirectoryName(executablePath);
+
+            // If the pip run by this sandboxed process executor tool also executes Detours, then that pip may access
+            // DetoursServices.pdb that is linked with this tool. Thus, we need to tell the pip to untrack the directories
+            // where DetoursServices are located.
             UntrackedScopes = new[] { Path.Combine(directory, "X86"), Path.Combine(directory, "X64") };
         }
 
--- a/Public/Src/Engine/Processes/ExternalToolSandboxedProcessExecutor.cs
+++ b/Public/Src/Engine/Processes/ExternalToolSandboxedProcessExecutor.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
+using System.IO;
+using BuildXL.Utilities;
 
 namespace BuildXL.Processes
 {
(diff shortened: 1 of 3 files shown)


Public/Src/Engine/Processes/Internal/BinaryPaths.cs
```diff
@@ -28,13 +23,16 @@ public BinaryPaths()
...
             VerifyFileExists(DllNameX86);
             VerifyFileExists(DllNameX64);
+
+            DllDirectoryX86 = Path.GetDirectoryName(DllNameX86);
+            DllDirectoryX64 = Path.GetDirectoryName(DllNameX64);
         }
 
         private static void VerifyFileExists(string fileName)
         {
             if (!File.Exists(fileName))
```
dannyvv: Should this use FileUtilities.Exists?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1616,13 +1623,21 @@ private bool PrepareFileAccessMonitoringCommon(LoggingContext loggingContext)
...
-                }               
+                }
 
-                // untrack the original path
-                m_fileAccessManifest.AddScope(path, mask: m_excludeReportAccessMask, values: FileAccessPolicy.AllowAll | FileAccessPolicy.AllowRealInputTimestamps);
+                // Untrack the original path
+                AddUntrackedScopeToManifest(path);
+            }
+
+            if (!OperatingSystemHelper.IsUnixOS)
```
dannyvv: Did you check with the mac team if they perhaps have a similar issue?

Public/Src/Engine/Processes/ExternalToolSandboxedProcessExecutor.cs
```diff
@@ -20,13 +23,26 @@ public class ExternalToolSandboxedProcessExecutor
...
             Contract.Requires(!string.IsNullOrWhiteSpace(executablePath));
             ExecutablePath = executablePath;
+
+            if (!File.Exists(ExecutablePath))
+            {
+                throw new BuildXLException($"Cannot find file '{ExecutablePath}' needed to externally execute process. Did you build all configurations?", rootCause: ExceptionRootCause.MissingRuntimeDependency);
+            }
+
+            string directory = Path.GetDirectoryName(executablePath);
+            UntrackedScopes = new[] { Path.Combine(directory, "X86"), Path.Combine(directory, "X64") };
```
dannyvv: >ne(directory, "X86"), Path.Combine(directory [](start = 48, length = 44)

Can you comment why these are needed?

Public/Src/Engine/Processes/Internal/BinaryPaths.cs
narasamdya: We typically use File.Exists for
- our own files,
- not needed for performance,
- don't care about following symlink.

---
In reply to: [303165327](https://github.com/microsoft/BuildXL/pull/589#discussion_r303165327) [](ancestors = 303165327)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
narasamdya: Mac won't have this issue because
- it's Detours related,
- it's about a pip that also runs Detours during its execution.

---
In reply to: [303165462](https://github.com/microsoft/BuildXL/pull/589#discussion_r303165462) [](ancestors = 303165462)

PR #590, opened 2019-07-12
Make eviction window size configurable

JuanCarlosGI:
Make eviction window size configurable

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreConfiguration.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreConfiguration.cs
@@ -77,6 +77,11 @@ namespace BuildXL.Cache.ContentStore.Distributed.Redis
         public long MaxBlobCapacity { get; set; } = 1024 * 1024 * 1024;
 
         /// <summary>
+        /// Indicates the window size for executing eviction.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreConfiguration.cs
```diff
@@ -91,6 +91,11 @@ public sealed class RedisContentLocationStoreConfiguration : LocalLocationStoreC
         /// </summary>
         public ContentLocationMode ReadMode { get; set; } = ContentLocationMode.Redis;
 
+        /// <summary>
+        /// Indicates the window size for executing eviction.
+        /// </summary>
+        public int EvictionWindowSize { get; set; } = 500;
```
SergeyTeplyakov: We need to change distributed settings to actually make this configurable from cloudbuild configuration.

dannyvv:
>         /// <summary>

maybe place here in the order, since the distributedcontentsettings.cs has it in that place whcih makes it easier to keep the defaults the same values.

Perhaps a comment to keep them in sync iwth eachother would be good.

---
Refers to: Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreConfiguration.cs:79 in e9c4cd2. [](commit_id = e9c4cd22c475fb6ec61ede24458bd5f0e9b87e8c, deletion_comment = False)

PR #595, opened 2019-07-15
Abstract logic for event hub interaction to allow testing most of logic for content location event store.

ljcollins25:
Abstract logic for event hub interaction to allow testing most of logic for content location event store.


Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/AzureEventHubClient.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            return BoolResult.Success;
+        }
+
+        /// <inheritdoc />
+        public BoolResult SuspendProcessing(OperationContext context)
+        {
+            // In unit tests, hangs sometimes occur for this when running multiple tests in sequence.
+            // Adding a timeout to detect when this occurs
+            _partitionReceiver?.CloseAsync().WithTimeoutAsync(TimeSpan.FromMinutes(1)).GetAwaiter().GetResult();
```
SergeyTeplyakov: if `_partitionReceiver` is null this code will fail with NRE.

I think you have to wrap it into an if block.

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/AzureEventHubClient.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            return BoolResult.Success;
+        }
+
+        /// <inheritdoc />
+        protected override async Task<BoolResult> ShutdownCoreAsync(OperationContext context)
+        {
+            SuspendProcessing(context).ThrowIfFailure();
+
+            _partitionSender?.CloseAsync();
```
SergeyTeplyakov: Is it ok that we're no longer awaiting the results here?

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/AzureEventHubClient.cs
```diff
@@ -0,0 +1,113 @@
...
+            _eventHubClient?.CloseAsync();
+
+            return await base.ShutdownCoreAsync(context);
+        }
+
+        /// <inheritdoc />
+        public async Task SendAsync(OperationContext context, EventData eventData)
+        {
+            await _partitionSender.SendAsync(eventData);
```
SergeyTeplyakov: Just curious, should we keep this method async, or we can just return the result of SendAsync method.

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/EventHubContentLocationEventStore.cs
```diff
@@ -93,15 +89,24 @@ public sealed class EventHubContentLocationEventStore : ContentLocationEventStor
                             })
                         .ToArray();
             }
-
-            _extraEventHubClientRetryPolicy = CreateEventHubClientRetryPolicy();
         }
 
-        private RetryPolicy CreateEventHubClientRetryPolicy()
+        /// <summary>
+        /// Factory method for creating an instance of <see cref="ContentLocationEventStore"/> based on <paramref name="configuration"/>.
```
SergeyTeplyakov: instance of <see cref="IEventHubClient"/> ?

PR #596, opened 2019-07-15
Enable dev cache for selfhost builds

almiliMSFT:
[AB#1558017](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1558017)

### DIFF
--- a/Shared/Scripts/bxl.ps1
+++ b/Shared/Scripts/bxl.ps1
@@ -81,10 +81,9 @@ param(
     [Parameter(Mandatory=$false)]
     [switch]$DeployStandaloneTest = $false,
 
-    # Task 544796 to enable this
     [Parameter(Mandatory=$false)]
     [ValidateSet("Disable", "Consume", "ConsumeAndPublish")]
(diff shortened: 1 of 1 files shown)


Shared/Scripts/bxl.ps1
```diff
@@ -81,10 +81,9 @@ param(
     [Parameter(Mandatory=$false)]
     [switch]$DeployStandaloneTest = $false,
 
-    # Task 544796 to enable this
     [Parameter(Mandatory=$false)]
     [ValidateSet("Disable", "Consume", "ConsumeAndPublish")]
-    [string]$SharedCacheMode = "Disable",
+    [string]$SharedCacheMode = "Consume",
```
almiliMSFT: > Do we have to do anything to reenable our dev cache publishing builds running in CloudBuild?

I don't think so, because I never disabled any publishing to the shared cache that is going on in CloudBuild.

almiliMSFT: That said, only net472 bits are currently publishing to shared cache

PR #599, opened 2019-07-15
Add Delete verb to cache GRPC

kewilli:
[AB#1565544](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1565544)

### DIFF
--- a/Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
+++ b/Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
@@ -209,9 +209,10 @@ message DeleteContentRequest {
 }
 
 message DeleteContentResponse {
-    ResponseHeader Header = 1;
-    int64 evictedSize = 4;
-    int64 pinnedSize = 5;
+    ResponseHeader header = 1;
+    int32 result = 2;
+    int64 evictedSize = 3;
+    int64 pinnedSize = 4;
 }
 
 // AddOrGetContentHashList
--- a/Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
+++ b/Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
@@ -27,6 +27,8 @@ service ContentServer {
     rpc CopyFile(CopyFileRequest) returns (stream CopyFileResponse) {}
 
     rpc CheckFileExists(ExistenceRequest) returns (ExistenceResponse) {}
+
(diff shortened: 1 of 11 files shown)


JuanCarlosGI:
Should we make a distinction between ContentNotDeleted and ContentNotFound? ContentNotDeleted could mean that there was an exception on the server side and so the operation could not be completed, and so we might want to retry, where ContentNotFound clearly talls the caller not to retry, and assume a success, even though deleted size will not be available

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
```diff
@@ -199,6 +201,19 @@ message ExistenceResponse {
...
 
+// Delete
+message DeleteContentRequest {
+    string traceId = 1;
+    int32 hashType = 2;
+    bytes contentHash = 3;
+}
+
+message DeleteContentResponse {
+    ResponseHeader Header = 1;
```
dannyvv: >H [](start = 19, length = 1)

casing is off.
Protobuf recommended casing is camlCase. The C# codegen will make it PascalCase....

I see you just copied, but you should probably do a pass before too many folks take dependencies and clean it up.

SergeyTeplyakov:
>                 ? $"Success Hash={ContentHash} Size={EvictedSize} Pinned={PinnedSize}"

Do you want to print the error code here?

---
Refers to: Public/Src/Cache/ContentStore/Interfaces/Results/DeleteResult.cs:94 in d587241. [](commit_id = d587241ec10e97b8279cf1f6a8213461344cfdda, deletion_comment = False)

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
kewilli: I hope that isn't a dangerous change! I'll migrate this entire file to camelCase after this PR is checked in.

---
In reply to: [303659836](https://github.com/microsoft/BuildXL/pull/599#discussion_r303659836) [](ancestors = 303659836)

ljcollins25: I don't see a point in changing this. It's purely cosmetic and more likely to cause problems than it is to help.

---
In reply to: [303673571](https://github.com/microsoft/BuildXL/pull/599#discussion_r303673571) [](ancestors = 303673571,303659836)

PR #603, opened 2019-07-16
[Draft for review] BuildXL as a library design document

narasamdya:
For review only.

### DIFF
--- a/BxlAsLib.txt
+++ b/BxlAsLib.txt
@@ -1,0 +1,377 @@
+BuildXL as a library - aka. BXLLib.
+
+Problem statement
+-----------------
+- Performance
+  - Graph construction takes a long time due to DScript parsing, type checking, and evaluation.
+  - DScript frontend is not scalable to handle Office multi-architecture (mularchy) builds.
+
+- Frontend usability
+  - CMake, MsBuild frontends create their own adhoc JSON formats of build graphs.
+
+- Binary graph fragments are proposed as a way to construct build graphs, but need a way (API)
+  to represent and construct those fragments.
+
+Requirements
+------------
+- API must preserve compatibility, i.e., maintain backward compatibility.
+  - API must be binary compatible.
+  - Graph fragment must be versioned.
+
+- Graph fragment construction must be high performant.
+  - Creating, serializing, and deserializing pips must be performant.
+  - Stitching multiple fragments must be performant.
+
+- Graph fragment interactions:
+  - Fragment can consume artifacts produced by other frontends.
+  - Fragment can produce artifacts that can be consumed by other frontends.
+
+- API for constructing graph fragments that makes sense for customer
+  - Adding basic pips, e.g., process, copy, write, seal directory.
+  - Extensible for adding BuildXL's friends:
+    - Adding QTest pips.
+    - Adding drop and symbol pips.    
+  - Desirable:
+    - Higher-level API to ease migration from existing DScript SDK.
+      - Creating a unique output directory.
+      - Specifying inputs/outputs of pips like in DScript.
+    - API becomes yet another frontend.
+
+- Reuse existing DScript (or other frontends) SDK: Drop, Symbol, QTest, etc.
+
+- Control over BuildXL's friends behavior, e.g., adding a new default to QTest pips.
+
+- Track what files and environment variables used for graph fragment constructions.
+  - This is needed in case the construction is not done in a sandboxed environment.
+
+- BXLLib is applicable to many customers, not only to Office, e.g., WDG can benefit from BXLLib.
+
+Spike
+-----
+- Use BuildXL DScript frontend to evaluate specs and generate specific fragments.
+  - Description:
+    - This approach targets only performance.
+    - In metabuild, besides generating DScript spec, the pip also evaluates the spec (plus the SDK) and generates a fragment.
+    - Instead of using a full-blown BuildXL, we refactor BuildXL frontend to create so-called BXLScriptAnalyzer
+      that is used for evaluation and for generating fragment.
+    - The generated fragments are then fed as inputs to product build.
+    - Performance can be gained because:
+      - The metabuild pip generates a smaller spec that can be evaluated fast (to be proven).
+      - The metabuild pip can be distributed. 
+
+  - Pros:
+    - It's a shortcut and requires moderate efforts.
+    - We can (hopefully) deliver performance without changing DScript SDK, without changing static NMake, without
+      waiting for definite API.
+    - Existing DScript SDK can still be used.
+    - The shipment of BuildXL engine and its DScript SDK doesn't change.
+
+  - Cons:
+    - This may be applicable to Office where their project slices are kind of isolated from one another.
+    - Stitching fragments is non-trivial due to service and IPC pips:
+      - [In progress] Need to unify service and IPC pips:
+        - Instead of having one drop service pip for the whole graph, we now have N drop service pips, where
+          N is the number of generated fragments.
+        - IPC pips are tied to specific service pips through monikers. We need to mess around with IPC pips
+          command-line arguments to unify service and IPC pips.
+
+API
+---
+- Basic API: must be small and not prone to binary incompatibility.
+  
+  static class FragmentFactory 
+  {
+      /* Fragment's name can be thought of as module name */
+      public IGraphFragment InitGraphFragment(string name);
+  }
+
+  interface IGraphFragment 
+  {
+      void AddProcess(ProcessData processData);
+      void AddWriteFile(WriteFileData writeFileData);
+      void AddCopyFile(CopyFileData copyFileData);
+      void AddSealDirectory(SealDirectoryData sealDirectory);
+      ...
+
+      AbsolutePath P(string path);
+      FileArtifact F(AbsolutePath path, int rewriteCount);
+      DirectoryArtifact D(AbsolutePath path);
+  }
+
+- Some utilites need to be exposed for optimization:
+  - AbsolutePath, 
+  - FileArtifact(WithAttaributs), 
+  - DirectoryArtifact,
+  - PathTable
+  - PipData/PipDataBuilder:
+    - PipData builder is needed to construct PipData.
+    - Need to be moved to utilities.
+    - Some renderer that refers to FileContentInfo of BuildXL.Storage needs to use plain string.
+    - Perhaps create a restricted version of PipDataBuilder that user can use easily.
+  - To support reuse, some classes/enums in BuildXL.Pips needs to be moved to a smaller DLL (may be called
+    BuildXL.PipUtilities), e.g., ProcessOptions, EnvironmentVariables, etc.
+ 
+- ProcessData, WriteFileData, ... are the ones that are going to be serialized/deserialized, and thus need to be versioned.
+
+  - Option 1: Use Google protobuf to represent them.
+    - Pros:
+      - Kind of "de-facto" standard.
+      - Serialization and deserialization come out of box.
+      - Binary format is optimized.
+    - Cons:
+      - Cannot use custom serialization:
+        - We want the serialization to leverage pip inlining mechanism.
+      - Generate only classes.
+      - Deserialization can create a lot of objects.
+
+  - Option 2: Use protobuf-net.
+    - Pros:
+      - Serialization and deserialization come out of box.
+      - Doesn't need a separate proto file.
+    - Cons:
+      - Similar to Google protobuf.
+      - Stability, performance, and encoding size need to be investigated.
+
+  - Option 3 (potentially chosen): Classes have serialization & deserialization methods that mimic
+    protobuf, i.e., there are positional fields.
+    Ex.
+
+    class Data 
+    {       
+        /* TBD: We may want to add positional attrbute for each property, like protobuf-net. */
+
+        public AbsolutePath Path { get; set; }
+        public FileArtifact File { get; set; }
+        public void Serialize(Writer writer) 
+        {
+            if (Path.IsValid) { writer.WritePos(1); writer.WritePath(Path); }
+            if (File.IsValid) { writer.WritePos(2); write.WriteFile(File); }
+        }
+        public static Data Deserialize(Reader reader)
+        {
+            Data data;
+            while (!int pos = reader.ReadPos()) != -1) {
+                switch (pos) {
+                    case 1: { data.Path = reader.ReadPath(); break; }
+                    case 2: { data.File = reader.ReadFile(); break; }
+                }
+            }
+            return data;
+        }
+    }
+
+    - Pros: Can use custom serialization/deserialization.
+    - Cons: 
+      - Manually maintained the position.
+      - An intern working on XLG++ may choose protobuf for versioning.
+
+- Excerpt of ProcessData: basically it reflects Process in BuildXL.Pips
+
+  class ProcessData 
+  {
+      /* TBD: We may want to add positional attrbute for each property, like protobuf-net. */
+
+      public FileArtifact Executable { get; set; }
+      public PipData Arguments { get; set; }
+      public AbsolutePath WorkingDirectory { get; set; }
+      public IEnumerable<EnvironmentVariable> EnvironmentVariables { get; set; }
+      public ProcessOptions ProcessOptions { get; set; }
+      public IEnumerable<FileArtifact> Dependencies { get; set; }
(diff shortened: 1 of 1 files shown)


BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+  interface IGraphFragment 
+  {
+      void AddProcess(ProcessData processData);
+      void AddWriteFile(WriteFileData writeFileData);
+      void AddCopyFile(CopyFileData copyFileData);
+      void AddSealDirectory(SealDirectoryData sealDirectory);
+      ...
+
+      AbsolutePath P(string path);
+      FileArtifact F(AbsolutePath path, int rewriteCount);
```
dannyvv: >F [](start = 19, length = 1)

Please make F take no rewrite count to match DScript api.
And have a seperate (non overload) method for output file.

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    - Pros: Can use custom serialization/deserialization.
+    - Cons: 
+      - Manually maintained the position.
+      - An intern working on XLG++ may choose protobuf for versioning.
+
+- Excerpt of ProcessData: basically it reflects Process in BuildXL.Pips
+
+  class ProcessData
```
dannyvv: This should probably be an interface with factory methods on `IGraphFragment`

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+  class ProcessData 
+  {
+      /* TBD: We may want to add positional attrbute for each property, like protobuf-net. */
+
+      public FileArtifact Executable { get; set; }
+      public PipData Arguments { get; set; }
+      public AbsolutePath WorkingDirectory { get; set; }
+      public IEnumerable<EnvironmentVariable> EnvironmentVariables { get; set; }
+      public ProcessOptions ProcessOptions { get; set; }
+      public IEnumerable<FileArtifact> Dependencies { get; set; }
```
dannyvv: >IEnumerable [](start = 13, length = 11)

IEnumerable does not allow on demand construction, i.e. composing the values. So either make this a getter only and have an 'AddDependency' method or make it ICollection

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
+BuildXL as a library - aka. BXLLib.
+
+Problem statement
+-----------------
+- Performance
+  - Graph construction takes a long time due to DScript parsing, type checking, and evaluation.
+  - DScript frontend is not scalable to handle Office multi-architecture (mularchy) builds.
```
dannyvv: >not scalable [](start = 24, length = 12)

Does not have `binary module caching` yet

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
+BuildXL as a library - aka. BXLLib.
+
+Problem statement
+-----------------
+- Performance
+  - Graph construction takes a long time due to DScript parsing, type checking, and evaluation.
+  - DScript frontend is not scalable to handle Office multi-architecture (mularchy) builds.
+
+- Frontend usability
+  - CMake, MsBuild frontends create their own adhoc JSON formats of build graphs.
```
dannyvv: >JSON [](start = 52, length = 4)

Don't see how that is relevant since their problem is that the engine can be written in other langauge and therefore needs either an IPC channel -or- a dataformat. 

Do you expect the format of these to change i.e. for ninja we will write the binary fragment format in Java?

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+
+- Binary graph fragments are proposed as a way to construct build graphs, but need a way (API)
+  to represent and construct those fragments.
+
+Requirements
+------------
+- API must preserve compatibility, i.e., maintain backward compatibility.
+  - API must be binary compatible.
+  - Graph fragment must be versioned.
```
dannyvv: Given your comment about 'frontend usabillity' , the binary format should be language independent so build engines implemented in non .net languages like java or c++ can participate

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+
+- Graph fragment construction must be high performant.
```
dannyvv: higly

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+  - AbsolutePath, 
+  - FileArtifact(WithAttaributs), 
+  - DirectoryArtifact,
+  - PathTable
+  - PipData/PipDataBuilder:
+    - PipData builder is needed to construct PipData.
+    - Need to be moved to utilities.
+    - Some renderer that refers to FileContentInfo of BuildXL.Storage needs to use plain string.
+    - Perhaps create a restricted version of PipDataBuilder that user can use easily.
+  - To support reuse, some classes/enums in BuildXL.Pips needs to be moved to a smaller DLL (may be called
```
dannyvv: >BuildXL.Pip [](start = 44, length = 11)

You probalby mean BuildXL.Utilities...

pips has the classes and the base graph.
PathTable and family are in Utilities

dannyvv: Note if we can't refactor those out into a small reusable library, It could just be a facade assembly with type forwarders to BuildXL.Utiltiies.

---
In reply to: [304077365](https://github.com/microsoft/BuildXL/pull/603#discussion_r304077365) [](ancestors = 304077365)

BxlAsLib.txt
almiliMSFT: Instead of this "not scalable" bullet point, I'd say something like "graph construction is not modular, i.e., there is no way to compose multiple separate (possibly interconnected) graph fragments into a single final graph

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+    - API becomes yet another frontend.
+
+- Reuse existing DScript (or other frontends) SDK: Drop, Symbol, QTest, etc.
+
+- Control over BuildXL's friends behavior, e.g., adding a new default to QTest pips.
+
+- Track what files and environment variables used for graph fragment constructions.
+  - This is needed in case the construction is not done in a sandboxed environment.
+
+- BXLLib is applicable to many customers, not only to Office, e.g., WDG can benefit from BXLLib.
```
almiliMSFT:
```suggestion
- BXLLib is general-purpose and customer-agnostic
```

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+- Use BuildXL DScript frontend to evaluate specs and generate specific fragments.
+  - Description:
+    - This approach targets only performance.
+    - In metabuild, besides generating DScript spec, the pip also evaluates the spec (plus the SDK) and generates a fragment.
+    - Instead of using a full-blown BuildXL, we refactor BuildXL frontend to create so-called BXLScriptAnalyzer
+      that is used for evaluation and for generating fragment.
+    - The generated fragments are then fed as inputs to product build.
+    - Performance can be gained because:
+      - The metabuild pip generates a smaller spec that can be evaluated fast (to be proven).
+      - The metabuild pip can be distributed.
```
almiliMSFT: How can a single pip be distributed?

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+-----
+- Use BuildXL DScript frontend to evaluate specs and generate specific fragments.
+  - Description:
+    - This approach targets only performance.
+    - In metabuild, besides generating DScript spec, the pip also evaluates the spec (plus the SDK) and generates a fragment.
+    - Instead of using a full-blown BuildXL, we refactor BuildXL frontend to create so-called BXLScriptAnalyzer
+      that is used for evaluation and for generating fragment.
+    - The generated fragments are then fed as inputs to product build.
+    - Performance can be gained because:
+      - The metabuild pip generates a smaller spec that can be evaluated fast (to be proven).
```
almiliMSFT: how many fragments are going to be generated by the metabuild?  Is it going to be one fragment per what used to be a .dsc spec?

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+    - It's a shortcut and requires moderate efforts.
+    - We can (hopefully) deliver performance without changing DScript SDK, without changing static NMake, without
+      waiting for definite API.
+    - Existing DScript SDK can still be used.
+    - The shipment of BuildXL engine and its DScript SDK doesn't change.
+
+  - Cons:
+    - This may be applicable to Office where their project slices are kind of isolated from one another.
+    - Stitching fragments is non-trivial due to service and IPC pips:
+      - [In progress] Need to unify service and IPC pips:
```
almiliMSFT: My understanding is that there should be exactly one fragment that defines a service for a moniker `S1`, e.g.,:
```json
"service": {
  "moniker": "S1",
  "start-args": [...],
  "finalization-args": [...],
  "shutdown-args": [...]
}
```
Then, all fragments may define IPC pips that declare their dependency on the service `S1`, e.g.,:
```json
"ipc": {
  "service": "S1",
  "args": [...]
}
```
This service moniker `S1` should be declared, exported, and referenced the same way other inter-fragment symbols/dependencies are declared, exported, and referenced.

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+        - Instead of having one drop service pip for the whole graph, we now have N drop service pips, where
+          N is the number of generated fragments.
+        - IPC pips are tied to specific service pips through monikers. We need to mess around with IPC pips
+          command-line arguments to unify service and IPC pips.
+
+API
+---
+- Basic API: must be small and not prone to binary incompatibility.
```
almiliMSFT: How can an API be prone or not prone to binary incompatibility?  And why in particular the API proposed here is not prone to binary incompatibility?

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+      void AddProcess(ProcessData processData);
+      void AddWriteFile(WriteFileData writeFileData);
+      void AddCopyFile(CopyFileData copyFileData);
+      void AddSealDirectory(SealDirectoryData sealDirectory);
+      ...
+
+      AbsolutePath P(string path);
+      FileArtifact F(AbsolutePath path, int rewriteCount);
+      DirectoryArtifact D(AbsolutePath path);
+  }
```
almiliMSFT: What about exporting/referencing stuff from other fragments?  How is that achieved with this API?

almiliMSFT: Maybe just add a forward reference to the section later in this doc where this is addressed

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+  that can be consumed by other frontends.
+- Even when stitching multiple graphs, we want to estabilish dependency relation between fragments.
+- API:
+  - To expose outputs to other frontends or other fragments:
+
+    interface IGraphFragment
+    {
+        ...
+        /* The implementation may verify that this fragment indeed produces the output file. */
+        AddOutputFileValue(FullSymbol id, FileArtifact outputFile);
```
almiliMSFT:
```suggestion
        Export(FullSymbol id, FileArtifact outputFile);
```

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+        AddOutputDirectoryValue(FullSymbol id, DirectoryArtifact outputDirectory);
```
almiliMSFT:
```suggestion
        Export(FullSymbol id, DirectoryArtifact outputDirectory);
```

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+    {
+        ...
+        /**
+         * Upon deserialization, the entry in these symbolic dependencies are assumed
+         * to exist (by querying FrontEndHost).
+         * Given (module: M, id: x.y.z) as a symbolic dependency, one should not evaluate
+         * "import('M').x.y.z" because the topological order will be violated.
+         */
+        public IEnumerable<SymbolicDependency> SymbolicDependencies { get; set; }
+        public IEnumerable<SymbolicDependency> SymbolicDirectoryDependencies { get; set; }
```
almiliMSFT: why do you need to differentiate between files and directories here?  It seems that just declaring symbolic dependencies should be enough

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+      - Fragment's dependencies must be encoded into the fragment itself.
+      - See example below for details.
+
+  - Other frontends can consume outputs of fragments. For example, one will write the following in
+    DScript:
+    /**
+     * Assume that the fragment is constructed as follows:
+     *   var f = InitGraphFragment("Word.Debug.X64");
+     *   ...
+     *   f.AddOutputFileValue(S("word.exe"), F("Target/Debug/X64/word.exe", 1));
```
almiliMSFT: `S` is not defined

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+       *           new SymbolicDependency { Module: "Utils", Id: IO.common.dll }
+       *       } 
+       *   });
+       *   ----------------------------------
+       *
+       * Liblet fragment is constructed as follows:
+       *   ----------------------------------
+       *   var f = InitGraphFragment("Liblet.Debug.X64");
+       *   ...
+       *   f.AddOutputFileValue(liblet.dll, F("Target/Debug/X64/liblet.dll", 1));
```
almiliMSFT:
```suggestion
       *   f.AddOutputFileValue(S("liblet.dll"), F("Target/Debug/X64/liblet.dll", 1));
```

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+
+      In DScript, one can use the fragment as follows:
+        const f = import("Word.Debug.X64");
+        const o = DropSDK.dropFile(f.word.exe);
+
+BuildXL's friends
+-----------------
+- The above API is the bare minimal one for creating fragments.
+- Dealing with BuildXL's friends is tricky:
+  - For drop, the user needs to add the finalization pip before its dependencies.
```
almiliMSFT: Since drop is already a "friend", the user doesn't need to specify the finalization command at all, because it is always the same for all drops anyway, so BuildXL can always add it.

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+- We still want to have a control over our friends:
+  - QTest/BuildXL teams is improving QTest experience, and want to enable new behavior by
+    changing default values for QTest pips.
+
+- Enrich IGraphFragment API:
+    interface IGraphFragment
+    {
+        ...
+        AddDropServicePip(DropServiceData serviceData);
```
almiliMSFT: This can probably be omitted and and automatically added by BuildXL

BxlAsLib.txt
dannyvv: We have a design to do that wich is the feature I mentioned. Just not implemented yet.

BxlAsLib.txt
dannyvv: Office is composed of many modules (word, excel, outlook, etc) Each can construct their own graph in parralel.

BxlAsLib.txt
dannyvv: if we use a moniker We should use a moniker that is scoped to the module name, we should not introduce globally unique names, i.e. we should have  ("Module1", "localName1") rather than "superunique1"

BxlAsLib.txt
dannyvv: Due to this api needs to be binary compatible over time, I recommend we do not use overloads but more strongly named functions. `ExportFile`, `ExportDirectory` would be a better approach.

BxlAsLib.txt
dannyvv: Per above, Export is fine but we should not use overloads since we want a hihgly compatible api accross versions. It is too easy to introduce a new method overload that will change the behavior when compiling against a new version.

BxlAsLib.txt
```diff
@@ -0,0 +1,377 @@
...
+- BXLLib is applicable to many customers, not only to Office, e.g., WDG can benefit from BXLLib.
+
+Spike
+-----
+- Use BuildXL DScript frontend to evaluate specs and generate specific fragments.
+  - Description:
+    - This approach targets only performance.
+    - In metabuild, besides generating DScript spec, the pip also evaluates the spec (plus the SDK) and generates a fragment.
+    - Instead of using a full-blown BuildXL, we refactor BuildXL frontend to create so-called BXLScriptAnalyzer
+      that is used for evaluation and for generating fragment.
```
mpysson: Do the BXLScriptAnalyzer pips need to run in topological order with respect to the dependencies between graph fragments?

PR #606, opened 2019-07-16
Add BuildXL.Cache.Host.Configuration.dll and BuildXL.Cache.Host.Service.dll to BuildXL.Cache.Libraries nuget package

dannyvv:


SergeyTeplyakov:
I'm sure that the change is required, but I'm just curious to know what is broken exactly?

But the change LGTM.

dannyvv:
> I'm sure that the change is required, but I'm just curious to know what is broken exactly?
> 
> But the change LGTM.

This was request from Erik and Dmitry so they don't have to use the BuildXL.Cache.Tools nuget package as a library, but can use the BuildXL.Cache.Libraries nuget package.

PR #607, opened 2019-07-17
Fix DummyWaiter wrapper in getting executable path

narasamdya:
DummyWaiter executable was copied to a temp location first because creating hardlink to it is not allowed as CAS denied write attributes. CAS should no longer do that now.

Due to this copy, DummyWaiter.exe won't work when run with dotnet.

[AB#1572010](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1572010)

### DIFF
--- a/Public/Src/Utilities/UnitTests/Storage/DummyWaiter.cs
+++ b/Public/Src/Utilities/UnitTests/Storage/DummyWaiter.cs
@@ -57,7 +57,7 @@ namespace Test.BuildXL.Storage
         /// </summary>
         /// <remarks>
         /// In the past we created a copy of the executable because CAS denies write attributes needed for hardlink creation
-        /// to the executable. CAS should no longer denies the write attributes. Also, by copying only the executable, dotnet
+        /// to the executable. CAS no longer denies the write attributes. Also, by copying only the executable, dotnet
         /// won't work.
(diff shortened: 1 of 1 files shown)


Public/Src/Utilities/UnitTests/Storage/DummyWaiter.cs
```diff
@@ -36,35 +36,34 @@ public static DummyWaiter RunAndWait()
...
 
         /// <summary>
-        /// Creates and returns the location of a COPY of the exe deployed with <see cref="ExecutableName"/>.
-        /// The copy can be mutated or have its file permissions changed, unlike the actual exe which will be a hardlink
-        /// into the CAS.
+        /// Returns the location of the exe deployed with <see cref="ExecutableName"/>.
         /// </summary>
+        /// <remarks>
+        /// In the past we created a copy of the executable because CAS denies write attributes needed for hardlink creation
+        /// to the executable. CAS should no longer denies the write attributes. Also, by copying only the executable, dotnet
```
almiliMSFT:
```suggestion
        /// to the executable. CAS no longer denies the write attributes. Also, by copying only the executable, dotnet
```

PR #608, opened 2019-07-17
Increase default timeout for TelemetryShutdown to 2 mins

nayanshah:
There are a few hundred sessions everyday hitting timeout during TelemetryShutdown (DX0476). This results in no final statistics events for those builds. Root cause isn't known but mitigation is to bump the timeout.

### DIFF
--- a/Public/Src/Utilities/Instrumentation/Common/AriaV2StaticState.cs
+++ b/Public/Src/Utilities/Instrumentation/Common/AriaV2StaticState.cs
@@ -19,7 +19,7 @@ namespace BuildXL.Utilities.Instrumentation.Common
         public const int AriaMaxPropertyLength = 100;
 
         private static readonly object s_syncRoot = new object();
-        private static readonly TimeSpan s_defaultShutdownTimeout = TimeSpan.FromSeconds(20);
+        private static readonly TimeSpan s_defaultShutdownTimeout = TimeSpan.FromMinutes(2);
         private static readonly string s_ariaTelemetryDBName = "Aria.db";
 
         private static bool s_hasBeenInitialized;


Public/Src/Utilities/Instrumentation/Common/AriaV2StaticState.cs
```diff
@@ -19,7 +19,7 @@ public static class AriaV2StaticState
         public const int AriaMaxPropertyLength = 100;
 
         private static readonly object s_syncRoot = new object();
-        private static readonly TimeSpan s_defaultShutdownTimeout = TimeSpan.FromSeconds(20);
+        private static readonly TimeSpan s_defaultShutdownTimeout = TimeSpan.FromMinutes(2);
```
almiliMSFT: Can you also update [`AriaLogger.cpp`](https://github.com/microsoft/BuildXL/blob/eebb761925cb7fa986c0ab02f6c52f273ff76090/Public/Src/Utilities/Instrumentation/AriaNative/lib/AriaLogger.cpp#L18) by changing
```c
config[CFG_INT_MAX_TEARDOWN_TIME] = 15;
```
to
```c
config[CFG_INT_MAX_TEARDOWN_TIME] = 2 * 60; // 2 minutes
```

Ideally, the same `s_defaultShutdownTimeout` value should be plumbed all the way through, but that requires recompiling and republishing some binaries for macOS, so I can do that in a separate PR.

nayanshah:
@mpysson : I just wanted to start with

nayanshah:
Covered by https://github.com/microsoft/BuildXL/pull/609/files

PR #609, opened 2019-07-17
Increase telemetry flush timeout in CloudBuild

mpysson:
In a relatively small percentage of builds in CloudBuild we're hitting the telemetry flush timeout. Increase the timeout in CloudBuild where we're ok with a longer build tail for sake of getting the complete telemetry data.

Fixes [AB#1540805](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1540805)

### DIFF
--- a/Public/Src/App/Bxl/BuildXLApp.cs
+++ b/Public/Src/App/Bxl/BuildXLApp.cs
@@ -152,6 +152,9 @@ namespace BuildXL
 
         private readonly CrashCollectorMacOS m_crashCollector;
 
+        // Allow a longer Aria telemetry flush time in CloudBuild since we're more willing to wait at the tail of builds there
+        private TimeSpan TelemetryFlushTimeout => m_configuration.InCloudBuild() ? TimeSpan.FromMinutes(1) : AriaV2StaticState.DefaultShutdownTimeout;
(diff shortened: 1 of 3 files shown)


Public/Src/App/Bxl/BuildXLApp.cs
```diff
@@ -950,8 +950,10 @@ private AppResult RunWithLoggingScope(Func<PerformanceMeasurement, AppResult> ru
                     {
                         Stopwatch sw = Stopwatch.StartNew();
                         Exception telemetryShutdownException;
-
-                        var shutdownResult = AriaV2StaticState.TryShutDown(out telemetryShutdownException);
+                        
+                        // Allow a longer Aria telemetry flush time in CloudBuild since we're more willing to wait at the tail of builds there
+                        TimeSpan telemetryFlushTimeout = m_configuration.InCloudBuild() ? TimeSpan.FromMinutes(1) : AriaV2StaticState.DefaultShutdownTimeout;
```
almiliMSFT: can you move this to a property so that it can easily be passed to `AriaV2StaticState.CreateAriaLogger`?

PR #610, opened 2019-07-17
Discard MAC_LOOKUP reports relative to directory symlinks

almiliMSFT:


Public/Src/Engine/Processes/SandboxedProcess.MacKext.cs
```diff
@@ -493,6 +493,39 @@ private void UpdateAverageTimeSpentInReportQueue(AccessReportStatistics stats)
             m_sumOfReportQueueTimesUs += (long) (stats.DequeueTime - stats.EnqueueTime) / 1000;
         }
 
+        private Dictionary<AbsolutePath, bool> m_isDirSymlinkCache = new Dictionary<AbsolutePath, bool>();
```
dannyvv: >m_isDirSymlinkCache [](start = 47, length = 19)

Okay to be a non concurrent dictionary?

dannyvv: ah okay, found commetn on HandleKextReport :)

---
In reply to: [304694680](https://github.com/microsoft/BuildXL/pull/610#discussion_r304694680) [](ancestors = 304694680)

PR #611, opened 2019-07-17
Plumb teardown timeout all the way down to AriaNative

almiliMSFT:


Public/Src/Utilities/Instrumentation/AriaNative/lib/AriaLogger.cpp
```diff
@@ -9,13 +9,13 @@
...
-AriaLogger::AriaLogger(const char* token, const char *dbPath)
+AriaLogger::AriaLogger(const char* token, const char *dbPath, int teardownTimeoutInSeconds)
 {
     token_ = token;
     dbPath_ = dbPath;
 
     auto& config = LogManager::GetLogConfiguration();
-    config[CFG_INT_MAX_TEARDOWN_TIME] = 60;
+    config[CFG_INT_MAX_TEARDOWN_TIME] = teardownTimeoutInSeconds;
     // config[CFG_STR_CACHE_FILE_PATH] = dbPath; // not necessary
```
Kr1sso: I'd suggest this always was important, the SDK code pushes to DB if there are still events left on timed out tear down.

almiliMSFT: yes, but if that build happened on some VM or somewhere in cloud build, those logs are not going to be picked up in the next run anyway.

PR #612, opened 2019-07-17
Self-healing server deployment

Kr1sso:
This PR introduces self-healing server deployments to address issues we've seen in environments where BuildXL deployment / update is automated and deployment directories get tempered with.

The deployment hashing adds around `80ms` on a SSD to the startup time if the the deployment directory is healthy, otherwise kills the server and redeploys the content to make sure we have a sane environment.

### DIFF
--- a/Public/Src/App/Bxl/ServerDeployment.cs
+++ b/Public/Src/App/Bxl/ServerDeployment.cs
@@ -48,10 +48,8 @@ namespace BuildXL
         /// Throws if the copy fails</exception>
         public static ServerDeployment GetOrCreateServerDeploymentCache(string serverDeploymentRoot, AppDeployment clientApp)
         {
-            var deploymentDir = ComputeDeploymentDir(serverDeploymentRoot);
-
             ServerDeploymentCacheCreated? cacheCreated = null;
-            if (!Directory.Exists(deploymentDir) || clientApp.TimestampBasedHash.ToHex() != GetDeploymentCacheHash(deploymentDir))
+            if (IsServerDeploymentOutOfSync(serverDeploymentRoot, clientApp, out var deploymentDir))
             {
                 cacheCreated = CreateServerDeployment(deploymentDir, clientApp);
             }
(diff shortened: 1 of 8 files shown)


Public/Src/App/Bxl/ServerDeployment.cs
```diff
@@ -174,12 +145,22 @@ public static string ComputeDeploymentDir(string serverDeploymentRoot)
...
         {
             // Check if the main root process (likely bxl.exe) is in use before attempting to delete, so we avoid partially deleting files
-            // Not completey bullet proof (there can be a race) but it is highly unlikely the process starts to be in use right after this check
+            // Not completely bullet proof (there can be a race) but it is highly unlikely the process starts to be in use right after this check
             Assembly rootAssembly = Assembly.GetEntryAssembly();
             Contract.Assert(rootAssembly != null, "Could not look up entry assembly");
 
             string assemblyFullPath = Path.Combine(serverDeploymentRoot, new FileInfo(AssemblyHelper.GetAssemblyLocation(rootAssembly)).Name);
 
+#if FEATURE_CORECLR
```
almiliMSFT: Delete lines 149-152, replace
```csharp
string assemblyFullPath = Path.Combine(serverDeploymentRoot, new FileInfo(AssemblyHelper.GetAssemblyLocation(rootAssembly)).Name);
```
with
```csharp
string assemblyFullPath = Path.Combine(serverDeploymentRoot, new FileInfo(AssemblyHelper.GetThisProgramExeLocation()).Name);
```
then delete this whole block,

Public/Src/App/Bxl/ServerDeployment.cs
```diff
@@ -257,48 +238,18 @@ private static Possible<Unit> TryKillProcess(string assemblyFullPath)
         }
 
         /// <summary>
-        /// Poisons the server deployment to cause it to be redeployed
+        /// Returns a calculated hash of the content of the BuildXL binaries deployment server cache directory.
```
almiliMSFT: something is wrong in this sentence.  Maybe you meant
```suggestion
        /// Calculates a hash of the content of the BuildXL binaries from the server deployment directory.
```

Public/Src/App/Bxl/ServerDeployment.cs
Kr1sso: yay - forgot about that one thanks!

Public/Src/App/Bxl/ServerDeployment.cs
Kr1sso: It's the balkan way of saying what you suggested, come on! 💃

PR #616, opened 2019-07-18
Make deployment of experimental tool conditional temporarily

narasamdya:
Rolling builds are failing because our NuGet packages exceed the size limit.

This change make deployment experimental tools conditional temporarily to unblock rolling builds.

### DIFF
--- a/Public/Src/Deployment/buildXL.dsc
+++ b/Public/Src/Deployment/buildXL.dsc
@@ -62,19 +62,23 @@ namespace BuildXL {
                             }).exe
                         ]
                     },
-                    {
-                        subfolder: r`NinjaGraphBuilder`,
-                        contents: [
-                            importFrom("BuildXL.Tools").NinjaGraphBuilder.exe,
-                            importFrom("BuildXL.Tools.Ninjson").pkg.contents
-                        ]
-                    },
-                    {
-                        subfolder: r`CMakeRunner`,
-                        contents: [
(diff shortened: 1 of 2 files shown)


Public/Src/Deployment/buildXL.dsc
```diff
@@ -62,19 +62,23 @@ namespace BuildXL {
...
-                            importFrom("BuildXL.Tools.Ninjson").pkg.contents
-                        ]
-                    },
-                    {
-                        subfolder: r`CMakeRunner`,
-                        contents: [
-                            importFrom("BuildXL.Tools").CMakeRunner.exe,
-                        ]
-                    },
+                    ...(BuildXLSdk.Flags.deployExperimentalTools
```
Kr1sso: Nit: `...addIfLazy(BuildXLSdk.Flags.deployExperimentalTools, () => [ ...` to avoid evaluation completely.

narasamdya: They should not be evaluated with if-then-else expression, right?

narasamdya:
Rolling builds are failing because our NuGet packages exceed the size limit.

This change make deployment experimental tools conditional temporarily to unblock rolling builds.

Public/Src/Deployment/buildXL.dsc
Kr1sso: AFAIK they are, thats why making it lazy saves some work!

PR #618, opened 2019-07-18
DownloadResolver specifies hash type from config

JuanCarlosGI:


dannyvv:
@imnarasa@micorosft.com: Can you have a look too?

Public/Src/Utilities/Storage/ContentHashingUtilities.cs
```diff
@@ -87,6 +105,8 @@ public static void DisposeAndResetHasher()
         {
             s_hasher?.Dispose();
             s_hasher = HashInfo.CreateContentHasher();
+
+            s_contentHasherByHashType = new ConcurrentDictionary<HashType, IContentHasher>();
```
dannyvv: shoudl we dispose the ones in the dictionary

Public/Src/Utilities/Storage/ContentHashingUtilities.cs
```diff
@@ -22,6 +23,23 @@ public static class ContentHashingUtilities
...
+            {
+                return s_hasher;
+            }
+
+            if (!s_contentHasherByHashType.ContainsKey(hashType))
+            {
+                s_contentHasherByHashType[hashType] = HashInfoLookup.Find(hashType).CreateContentHasher();
+            }
+
+            return s_contentHasherByHashType[hashType];
```
narasamdya: Use `GetOrAdd` method of `ConcurrentDictionary`.

Public/Src/Engine/Dll/FrontEndEngineImplementation.cs
```diff
@@ -274,13 +274,14 @@ public override async Task<ContentHash> GetFileContentHashAsync(string path, boo
 
                 VersionedFileIdentityAndContentInfo? maybeKnownIdentityAndHash = fileContentTable.TryGetKnownContentHash(fs);
 
-                if (maybeKnownIdentityAndHash.HasValue)
+                if (maybeKnownIdentityAndHash.HasValue && 
+                    (hashType == HashType.Unknown || maybeKnownIdentityAndHash.Value.FileContentInfo.Hash.HashType == hashType))
```
narasamdya: Perhaps in `FileContentInfo`, you add the following method:

```
bool MatchHashType(HashType hashType) => hashType == HashType.Unknown /* unknown matches everything */ || HashType == hashType;
```

I believe that method can be useful when needed in other places.

PR #619, opened 2019-07-18
Disable remote telemetry in distributed test

narasamdya:


Public/Src/Utilities/Configuration/ILoggingConfiguration.cs
```diff
@@ -197,7 +197,7 @@ public interface ILoggingConfiguration : IWarningHandling
         /// <summary>
         /// When enabled, sends telemetry information for remote collection. Defaults to false.
         /// </summary>
-        RemoteTelemetry RemoteTelemetry { get; }
+        RemoteTelemetry? RemoteTelemetry { get; }
```
almiliMSFT: I'm surprised to see that `/remoteTelemetry-` wasn't already sufficient

almiliMSFT: What's the difference between this property being `null` and `Disabled`?  I'm still not convinced you had to change this type from `RemoteTelemetry` to `RemoteTelemetry?`

narasamdya: The test runs with /inCloudBuild. In that mode, remote telemetry is enabled.

PR #620, opened 2019-07-18
Fix standalone test deployment

almiliMSFT:
[AB#1573258](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1573258)

### DIFF
--- a/Public/Sdk/SelfHost/BuildXL/StandaloneTest/StandaloneTestSupport.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/StandaloneTest/StandaloneTestSupport.dsc
@@ -21,7 +21,7 @@ namespace StandaloneTestSupport {
             contents: [
                 StandaloneTestUtils.writeFile(a`main.dsc`, specContent),
                 StandaloneTestUtils.writeFile(a`module.config.dsc`, moduleContent),
-                Deployment.createFromDisk(toolDirectory.root, "*", true),
+                toolDirectory,
             ]
         };
     }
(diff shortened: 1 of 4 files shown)


Public/Src/FrontEnd/UnitTests/Core/DsTest.cs
```diff
@@ -17,11 +18,6 @@
 using BuildXL.FrontEnd.Script.Constants;
 using BuildXL.FrontEnd.Workspaces.Core;
 using JetBrains.Annotations;
-#if FEATURE_MICROSOFT_DIAGNOSTICS_TRACING
```
almiliMSFT: `FEATURE_MICROSOFT_DIAGNOSTICS_TRACING` is never being defined any longer

Public/Sdk/SelfHost/BuildXL/StandaloneTest/StandaloneTestSupport.dsc
```diff
@@ -21,7 +21,7 @@ namespace StandaloneTestSupport {
             contents: [
                 StandaloneTestUtils.writeFile(a`main.dsc`, specContent),
                 StandaloneTestUtils.writeFile(a`module.config.dsc`, moduleContent),
-                Deployment.createFromDisk(toolDirectory.root, "*", true),
+                toolDirectory,
```
almiliMSFT: globbing the root of a static directory is not safe because that folder is possibly being populated at build time

PR #627, opened 2019-07-19
Discard reported accesses on non-canonical paths

almiliMSFT:
Concretely, discard observed accesses that have a single MacLookup report on a path that contains directory symlinks.

Reason:
  - observed accesses computed here should only contain fully expanded paths to avoid ambiguity;
  - on Mac, all access reports except for MacLookup report fully expanded paths, so only MacLookup paths need to be curated

[AB#1573235](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1573235)

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -2935,6 +2935,7 @@ namespace BuildXL.Processes
                             firstAccess.ManifestPath.IsValid &&
                             PathContainsSymlinks(firstAccess.ManifestPath.GetParent(m_context.PathTable)))
                         {
+                            Counters.IncrementCounter(SandboxedProcessCounters.DirectorySymlinkPathsDiscardedCount);
                             continue;
                         }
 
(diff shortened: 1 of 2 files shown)


Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2779,6 +2780,44 @@ private bool GetSpecialCaseRulesForSpecialTools(AbsolutePath processPath, Absolu
             return false;
         }
 
+        private Dictionary<AbsolutePath, bool> m_isDirSymlinkCache = new Dictionary<AbsolutePath, bool>();
```
smera: > private Dictionary<AbsolutePath, bool> m_isDirSymlinkCache = new Dictionary<AbsolutePath, bool>(); [](start = 7, length = 99)

Is it worth pooling this instance?

almiliMSFT: not sure, what do you think?

smera: I don't have a good sense of how much churn this is going to get... or how likely is to store stuff on this structure for arbitrary builds.
Feel free to ignore this comment if you don't have a strong feeling about it.

PR #629, opened 2019-07-19
DistributedContentCopier stops trying when copy fails with out of disk space

JuanCarlosGI:
Fixes [AB#1474275](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1474275)

DistributedContentCopier should not retry when copy fails with out of disk space


JuanCarlosGI:
Decided to reclassify errors in the client instead of here

PR #630, opened 2019-07-19
Revert a drop deployment change

olkononenko:
#570 introduced a temporary hack to drop deployment/usage logic -- we started using the drop from the LKG rather than the locally built one. This PR reverts that change since it is no longer needed.


olkononenko:
Closing this PR. AB#1574178 tracks a broader change.

PR #633, opened 2019-07-19
Add Delete verb to ContentStoreApp

kewilli:
[AB#1565544](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1565544)

### DIFF
--- a/Public/Src/Cache/ContentStore/App/DeleteContent.cs
+++ b/Public/Src/Cache/ContentStore/App/DeleteContent.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// ReSharper disable once UnusedMember.Global
+using System;
+using BuildXL.Cache.ContentStore.Hashing;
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/App/DeleteContent.cs
```diff
@@ -0,0 +1,48 @@
...
+using BuildXL.Cache.ContentStore.Interfaces.Utils;
+using BuildXL.Cache.ContentStore.Service.Grpc;
+using CLAP;
+
+namespace BuildXL.Cache.ContentStore.App
+{
+    internal sealed partial class Application
+    {
+        [Verb(Description = "Remove content from the cache")]
+        internal void Delete
```
dannyvv: I've seen a lot of these little verbs with GetAwaiter().GetAsync...

would it be worth adding support to the argument parser for Task methods?

kewilli: We don't own the argument parser, so adding support isn't an option there. We could add something of our own, but I don't think that it would give us much value. These verbs are typically not perf-bound nor highly parallelizable.

PR #637, opened 2019-07-19
Hack hash translations between VSO and Dedup for DownloadResolver

JuanCarlosGI:
This is necessary for testing side by side selfhost builds in both VSO and Dedup hash

### DIFF
--- a/Public/Src/Engine/Dll/Engine.cs
+++ b/Public/Src/Engine/Dll/Engine.cs
@@ -1107,6 +1107,36 @@ namespace BuildXL.Engine
                 mutableConfig.Schedule.EarlyWorkerRelease = false;
             }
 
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/Dll/Engine.cs
```diff
@@ -1107,6 +1107,36 @@ public static void PopulateLoggingAndLayoutConfiguration(CommandLineConfiguratio
...
+                        };
+
+                    foreach (var download in x.Downloads.Cast<DownloadFileSettings>())
+                    {
+                        if (translations.TryGetValue(download.Hash, out var newHash))
+                        {
+                            download.Hash = newHash;
+                        }
+
+                    }
```
dannyvv: nit: double newline

PR #638, opened 2019-07-19
Relax WriteOnAbsentFileProbe check

almiliMSFT:
Don't complain about WriteOnAbsentFileProbe when an explicit dependency already exists

[AB#1577561](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1577561)

### DIFF
--- a/Public/Src/Engine/Scheduler/FileMonitoringViolationAnalyzer.cs
+++ b/Public/Src/Engine/Scheduler/FileMonitoringViolationAnalyzer.cs
@@ -873,6 +873,13 @@ namespace BuildXL.Scheduler
                             return;
                         }
 
+                        // WriteOnAbsentPathProbe message literaly says "declare an explicit dependency between these pips",
+                        // so don't complain if a dependency already exists (i.e., 'pip' must run after 'related').
+                        if (m_graph.IsReachableFrom(pip, related))
+                        {
+                            return;
+                        }
+
                         violationType = DependencyViolationType.WriteOnAbsentPathProbe;
                         break;
                     default:
--- a/Public/Src/Engine/Scheduler/FileMonitoringViolationAnalyzer.cs
+++ b/Public/Src/Engine/Scheduler/FileMonitoringViolationAnalyzer.cs
@@ -875,7 +875,7 @@ namespace BuildXL.Scheduler
 
                         // WriteOnAbsentPathProbe message literaly says "declare an explicit dependency between these pips",
                         // so don't complain if a dependency already exists (i.e., 'pip' must run after 'related').
-                        if (m_graph.IsReachableFrom(from: related, to: pip))
+                        if (m_graph.IsReachableFrom(from: related.PipId, to: pip.PipId))
                         {
                             return;
                         }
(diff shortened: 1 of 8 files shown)


almiliMSFT:
Will add unit tests later...

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/OpaqueDirectoryTests.cs
```diff
@@ -190,7 +190,7 @@ public void ConsumeExplicitFileOutOfOpaque()
             // XAssert.IsFalse(File.Exists(ToString(unconsumedOutputInOpaque)));
         }
 
-        [Fact]
+        //[Fact]
```
almiliMSFT: Will fix the tests instead of disabling them...

Public/Src/Engine/Scheduler/IQueryablePipDependencyGraph.cs
```diff
@@ -111,5 +111,10 @@ public interface IQueryablePipDependencyGraph
         /// Returns the seal directory pip that corresponds to the given directory artifact
         /// </summary>
         Pip GetSealedDirectoryPip(DirectoryArtifact directoryArtifact, PipQueryContext queryContext);
+
+        /// <summary>
+        /// Returns whether pip <paramref name="to"/> is reachable from pip <paramref name="from"/>.
```
olkononenko: >reachable [](start = 57, length = 9)

nit: I'd expand the comment, I find 'reachable' here lacking 'explaining power'. Maybe add the comment you added in the analyzer , i.e., that pip 'from' is executed before pip 'to'?

olkononenko:
>         /// TODO: This will not return correct results w.r.t. meta-pips, e.g. spec file pips. Need to get meta-pips ordered correctly (i.e., add them topologically).

Does this PR need to take into account this comment?  Should we copy-paste this remark into the interface definition?

---
Refers to: Public/Src/Engine/Scheduler/Graph/PipGraph.cs:135 in c8a985f. [](commit_id = c8a985fc4cc234e9973f12aa0a3229eabb18351d, deletion_comment = False)

Public/Src/Engine/Scheduler/FileMonitoringViolationAnalyzer.cs
```diff
@@ -873,6 +873,13 @@ private static bool IsAllowedSameContentDoubleWrite(Process secondWriter, FileMa
                             return;
                         }
 
+                        // WriteOnAbsentPathProbe message literaly says "declare an explicit dependency between these pips",
+                        // so don't complain if a dependency already exists (i.e., 'pip' must run after 'related').
+                        if (m_graph.IsReachableFrom(from: related, to: pip))
```
smera: > if (m_graph.IsR [](start = 23, length = 16)

Wondering if this actually works reliably for the case of exlusive opaques. Exclusive opaques are scrubbed, but right before the writer runs, so the pip that probes may find the write there

smera: In a way, exclusive opaques may suffer from the same problems I was describing on email if the shared opaque scrubber becomes smarter

---
In reply to: [305936799](https://github.com/microsoft/BuildXL/pull/638#discussion_r305936799) [](ancestors = 305936799)

PR #639, opened 2019-07-19
Fix the crash on the worker when master exits before attach

semihokur:
If master sends an exit message before worker completes the attachment, worker crashes due to calling 'Exit' twice.

[AB#1573927](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1573927)

### DIFF
--- a/Public/Src/Engine/Dll/Tracing/Log.cs
+++ b/Public/Src/Engine/Dll/Tracing/Log.cs
@@ -887,13 +887,13 @@ namespace BuildXL.Engine.Tracing
         internal abstract void DistributionDebugMessage(LoggingContext loggingContext, string message);
 
         [GeneratedEvent(
-            (ushort)LogEventId.DistributionWorkerTimeoutAfterMasterExits,
+            (ushort)LogEventId.DistributionWorkerUnexpectedFailureAfterMasterExits,
             EventGenerators = EventGenerators.LocalOnly,
             EventLevel = Level.Verbose,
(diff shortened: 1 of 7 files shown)


Public/Src/Engine/Dll/Tracing/Log.cs
```diff
@@ -886,6 +886,15 @@ public abstract partial class Logger
...
         internal abstract void DistributionDebugMessage(LoggingContext loggingContext, string message);
 
+        [GeneratedEvent(
+            (ushort)LogEventId.DistributionWorkerTimeoutAfterMasterExits,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Verbose,
+            Keywords = (int)Keywords.UserMessage,
+            EventTask = (ushort)Tasks.Distribution,
+            Message = "After we received exit request from the master, there is a timeout for a master-related call (e.g., attach, send response).")]
+        public abstract void DistributionWorkerTimeoutAfterMasterExits(LoggingContext context);
```
mpysson: >DistributionWorkerTimeoutAfterMasterExits [](start = 29, length = 41)

Remove the "timeout" lingo as discussed. Maybe make this more along the lines of "Master completed before worker could attach to build session." ?

semihokur: yes, removed 'timeout' from the comments and log messages as well

PR #645, opened 2019-07-22
MachineName truncates the machine name

kewilli:
`Environment.MachineName` truncates the name of the machine to 15 characters. `Dns.GetHostName()` gives the full machine name.

Found this when attempting to surface content from `TSE-PERFBENCH-02`. Other machines would try to connect to `TSE-PERFBENCH-0`, which doesn't exist.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
@@ -5,11 +5,13 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net.Sockets;
 using System.Text;
 using BuildXL.Cache.ContentStore.Extensions;
 using BuildXL.Cache.ContentStore.Hashing;
 using BuildXL.Cache.ContentStore.Interfaces.Distributed;
 using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Distributed/Utilities/GrpcDistributedPathTransformer.cs
```diff
@@ -21,7 +21,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.Utilities
     public class GrpcDistributedPathTransformer : IAbsolutePathTransformer
     {
         private readonly IReadOnlyDictionary<AbsolutePath, AbsolutePath> _junctionsByDirectory;
-        private static readonly string _localMachineName = Environment.MachineName;
+        private static readonly string _localMachineName = System.Net.Dns.GetHostName();
```
dannyvv: >em.Net.Dns.GetHostName(); [](start = 63, length = 25)

I think this might fail with certain network configuration. We might need some catching & logging around this...

dcwuser: This is a good point. My own worry level is fairly low, but if you put the initialization in a static constructor, and throw or log if an exception is thrown, that's probably enough for us to figure out what's happening.

---
In reply to: [306080394](https://github.com/microsoft/BuildXL/pull/645#discussion_r306080394) [](ancestors = 306080394)

PR #646, opened 2019-07-22
Error-tolerant MaterilizeOutput step on workers

semihokur:
We should not fail the pip when materilizeoutput step cannot succeed in the workers due to connection issues.

### DIFF
--- a/Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
+++ b/Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
@@ -62,8 +62,7 @@ namespace BuildXL.Engine.Distribution
         private bool m_everAvailable;
         public override bool EverAvailable => Volatile.Read(ref m_everAvailable);
 
-        private bool m_isConnectionLost;
-        public bool IsConnectionLost => Volatile.Read(ref m_isConnectionLost);
+        private volatile bool m_isConnectionLost;
 
         private readonly TaskSourceSlim<bool> m_attachCompletion;
         private readonly TaskSourceSlim<bool> m_executionBlobCompletion;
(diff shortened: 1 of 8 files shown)


dcwuser:
This is a very cool idea and I'm very happy to see it. (Even though, by making BXL more tolerant of ring-machine failures, it will put pressure on cache to become equally resilient.)

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -1032,37 +1035,10 @@ private void Stop()
...
-                        {
-                            var result = new ExecutionResult();
-                            result.SetResult(m_appLoggingContext, PipResultStatus.Canceled);
-                            result.Seal();
-                            return result;
-                        };
-                    }
-
-                    pipCompletionTask.Set(executionResultFactory);
+                    FailRemotePip(pipCompletionTask, "No result was received because connection was lost");
```
dcwuser: I'm confused. It looks like we still call FailRemotePip; isn't the idea to re-schedule?

semihokur: The method name is somehow misleading. If you look at the inside of FailRemotePip, we do not set the pip result as failure if the step is MaterializeOutputs. We set the result as NotMaterialized which is treated as Success.

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -759,27 +761,28 @@ private ExecutionResult LogAndGetNetworkFailureResult(PipCompletionTask pipCompl
...
                 result = new ExecutionResult();
                 result.SetResult(operationContext, PipResultStatus.NotMaterialized);
                 result.Seal();
-                pipCompletionTask.Set(result);
+
+                pipCompletionTask.TrySet(result);
+                return;
             }
-            else
-            {
```
mpysson: So I guess the removal of this else is what causes an error to be logged and fixes the crash. Am I following correctly?

mpysson: An error is logged and the pip is marked as NotMaterialized, so not a failure. What ends up happening downstream?

1. Will that still keep our pip result <-> error logged validation out of agreement?
1. How will the consumers of that pip's output behave?

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
semihokur: > This is a very cool idea and I'm very happy to see it. (Even though, by making BXL more tolerant of ring-machine failures, it will put pressure on cache to become equally resilient.)

The changes are to make "metabuild" error-tolerant. If one of the workers fail to materialize the outputs, the master does not fail. The master would continue the product build with the rest of the workers.

PR #647, opened 2019-07-23
Enable local builds using RocksDbMemoizationStore

jbayardo:
This PR adds the required code to allow CASaaS to use RocksDbMemoizationStore when using a LocalCache.


Public/Src/Cache/MemoizationStore/Library/Sessions/LocalCache.cs
```diff
@@ -125,20 +127,39 @@ public class LocalCache : OneLevelCache
...
+            else if (config is RocksDbMemoizationStoreConfiguration rocksDbConfig)
+            {
+                return () => new RocksDbMemoizationStore(
+                    logger,
+                    clock ?? SystemClock.Instance,
+                    rocksDbConfig);
+            }
+            else
+            {
+                throw new NotSupportedException($"Configuration type for memoization store is unhandled.");
```
dcwuser: Minor: Whenever you are constructing an error message, include all the information you reasonably can the first time. Because I guarantee when you see that error message in a log and are trying to figure out what's going on, you will end up going back and putting in more info to debug. In this case, that means identify the type name (or nullity) of config.

Public/Src/Cache/MemoizationStore/Library/Stores/MemoizationStoreConfiguration.cs
```diff
@@ -1,17 +1,41 @@
...
 using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
 using BuildXL.Cache.ContentStore.SQLite;
 using BuildXL.Cache.ContentStore.Stores;
 
 namespace BuildXL.Cache.MemoizationStore.Stores
 {
+    /// <summary>
+    ///     Marker type for children of <see cref="MemoizationStore"/>
+    /// </summary>
+    public class MemoizationStoreConfiguration
```
dcwuser: This is kind of ugly, but I could live with it. I think on balance I would prefer to not have these extra marker/wrapper classes, and instead just have your config factory method take an arbitrary object, which it would then test for being either the new RocksDbContentLocationDatabaseConfiguration or the old SqlLiteMemoizationStoreConfiguration and act accordingly. Yes, having an API signature that accepts an arbitrary object and then fails at runtime if it isn't of the right type is ugly, but this new architecture has almost the same problem (since anyone can extend MemoizationStoreConfiguration) and takes a lot more classes/code to do it.

Public/Src/Cache/VerticalStore/MemoizationStoreAdapter/MemoizationStoreCacheFactory.cs
```diff
@@ -357,33 +356,37 @@ private static void SetDefaultsForStreamCas(Config config)
...
+            if (config.UseRocksDbMemoizationStore)
+            {
+                var rocksDbConfig = new RocksDbMemoizationStoreConfiguration() {
+                    Database = new RocksDbContentLocationDatabaseConfiguration(cacheRoot / "rocksdb"),
+                };
+
+                return rocksDbConfig;
+            }
+
+            var sqliteConfig = new SQLiteMemoizationStoreConfiguration(cacheRoot)
```
dcwuser: Minor: I'm a fan of early return where these is an error case or trivial case being tested for and the early return then allows the "meat" of the implementation to not be buried 6 brackets deep. But in this case the two cases have approximately equal weight, so I would argue for "else" here.

kewilli:
If the PR is WIP, please prefix the PR's title with "[WIP]"

MyCacheConfig.json
```diff
@@ -0,0 +1,25 @@
+{
```
kewilli: Just to check - this won't be checked in, right?

Public/Src/Cache/MemoizationStore/Library/Stores/MemoizationStoreConfiguration.cs
jbayardo: It is true that having a marker type is ugly. However, taking an arbitrary object has a series of issues as well:
- There is no semantic information to aid let the reader know what is going on. Receiving a MemoizationStoreConfiguration automatically conveys that somewhere down the road, a MemoizationStore is going to be created.
- Much better extensibility. If in the future we need to add a piece of configuration common to all memoization stores, it is easy this way. With an object, we would need track down every place where a memoization store is instantiated, generate a config object, and progressively introduce the new type.
- The cost is only 1 extra class (namely this one), so it seems to me a small price to pay for avoiding a suboptimal use of time later.

---
In reply to: [306291727](https://github.com/microsoft/BuildXL/pull/647#discussion_r306291727) [](ancestors = 306291727)

MyCacheConfig.json
jbayardo: Oh no, sorry. This was an unintended side-effect of "git add -A :/"

---
In reply to: [306406002](https://github.com/microsoft/BuildXL/pull/647#discussion_r306406002) [](ancestors = 306406002)

jbayardo:
Just so we are all on the same page:

- The bug where BuildXL in server mode keeps the RocksDB instance open is no longer happening. I re-read the code anyways, and it seems that proper disposal is happening where it should, so nothing to change in that sense for now.
- Preliminary performance measurements are good, when building BXL for example:
image
More on this to come when deployment to test environment happens.
- Configuration will be left as-is for now, as discussed with Lance.

PR #648, opened 2019-07-23
[WIP] Envvars

maneely:
[WIP] Current thoughts on integrating changes from this pending PR in msbuild:
https://github.com/microsoft/msbuild/pull/4461

I assume we'll discuss today. :)

@smera , @cdmihai , @AndyGerlicher

### DIFF
--- a/Public/Src/FrontEnd/MsBuild.Serialization/ProjectGraphWithPredictionsResult.cs
+++ b/Public/Src/FrontEnd/MsBuild.Serialization/ProjectGraphWithPredictionsResult.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
 using Newtonsoft.Json;
@@ -48,12 +49,17 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         /// <nodoc/>
         public bool Succeeded { get; }
 
+        /// <summary>
+        /// A sorted list of the names of environment variables that could affect the build.
+        /// </summary>
+        public IEnumerable<string> EnvironmentVariablesAffectingBuild { get; }
+
         /// <nodoc/>
-        public static ProjectGraphWithPredictionsResult<TPathType> CreateSuccessfulGraph(ProjectGraphWithPredictions<TPathType> projectGraphWithPredictions, IReadOnlyDictionary<string, TPathType> assemblyPathsToLoad, TPathType pathToMsBuild)
+        public static ProjectGraphWithPredictionsResult<TPathType> CreateSuccessfulGraph(ProjectGraphWithPredictions<TPathType> projectGraphWithPredictions, IReadOnlyDictionary<string, TPathType> assemblyPathsToLoad, TPathType pathToMsBuild, IEnumerable<string> environmentVariablesAffectingBuild)
         {
             Contract.Requires(projectGraphWithPredictions != null);
             Contract.Requires(assemblyPathsToLoad != null);
-            return new ProjectGraphWithPredictionsResult<TPathType>(projectGraphWithPredictions, failure: default, msBuildAssemblyPaths: assemblyPathsToLoad, pathToMsBuild: pathToMsBuild, pathToDotNetExe: default(TPathType), succeeded: true);
+            return new ProjectGraphWithPredictionsResult<TPathType>(projectGraphWithPredictions, failure: default, msBuildAssemblyPaths: assemblyPathsToLoad, pathToMsBuild: pathToMsBuild, pathToDotNetExe: default(TPathType), succeeded: true, environmentVariablesAffectingBuild);
         }
 
         /// <nodoc/>
@@ -61,7 +67,7 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         {
             Contract.Requires(failure != null);
             Contract.Requires(assemblyPathsToLoad != null);
-            return new ProjectGraphWithPredictionsResult<TPathType>(default, failure, assemblyPathsToLoad, pathToMsBuild, pathToDotNetExe: default(TPathType), succeeded: false);
+            return new ProjectGraphWithPredictionsResult<TPathType>(default, failure, assemblyPathsToLoad, pathToMsBuild, pathToDotNetExe: default(TPathType), succeeded: false, environmentVariablesAffectingBuild: null);
         }
 
         /// <summary>
@@ -69,11 +75,11 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         /// </summary>
         public ProjectGraphWithPredictionsResult<TPathType> WithPathToDotNetExe(TPathType pathToDotNetExe)
         {
-            return new ProjectGraphWithPredictionsResult<TPathType>(Result, Failure, MsBuildAssemblyPaths, PathToMsBuild, pathToDotNetExe, Succeeded);
+            return new ProjectGraphWithPredictionsResult<TPathType>(Result, Failure, MsBuildAssemblyPaths, PathToMsBuild, pathToDotNetExe, Succeeded, EnvironmentVariablesAffectingBuild);
         }
 
         [JsonConstructor]
-        private ProjectGraphWithPredictionsResult(ProjectGraphWithPredictions<TPathType> result, GraphConstructionError failure, IReadOnlyDictionary<string, TPathType> msBuildAssemblyPaths, TPathType pathToMsBuild, TPathType pathToDotNetExe, bool succeeded)
+        private ProjectGraphWithPredictionsResult(ProjectGraphWithPredictions<TPathType> result, GraphConstructionError failure, IReadOnlyDictionary<string, TPathType> msBuildAssemblyPaths, TPathType pathToMsBuild, TPathType pathToDotNetExe, bool succeeded, IEnumerable<string> environmentVariablesAffectingBuild)
         {
             Result = result;
             Failure = failure;
@@ -81,6 +87,7 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
             PathToMsBuild = pathToMsBuild;
             PathToDotNetExe = pathToDotNetExe;
             MsBuildAssemblyPaths = msBuildAssemblyPaths;
+            EnvironmentVariablesAffectingBuild = environmentVariablesAffectingBuild;
         }
     }
 }
--- a/Public/Src/FrontEnd/MsBuild.Serialization/ProjectGraphWithPredictionsResult.cs
+++ b/Public/Src/FrontEnd/MsBuild.Serialization/ProjectGraphWithPredictionsResult.cs
@@ -50,7 +50,7 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         public bool Succeeded { get; }
 
         /// <summary>
-        /// A sorted list of the names of environment variables that could affect the build.
+        /// A list of the names of environment variables that could affect the build.
         /// </summary>
         public IEnumerable<string> EnvironmentVariablesAffectingBuild { get; }
 
--- a/Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildGraphBuilder.cs
+++ b/Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildGraphBuilder.cs
@@ -187,6 +187,8 @@ namespace MsBuildGraphBuilderTool
                         locatedMsBuildPath);
                 }
 
+                // Get the possible environment variable reads from the logger.
+                // If there was a failure, report the failure message.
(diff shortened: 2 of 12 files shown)


Public/Src/FrontEnd/MsBuild.Serialization/ProjectGraphWithPredictionsResult.cs
```diff
@@ -48,39 +49,46 @@ namespace BuildXL.FrontEnd.MsBuild.Serialization
         /// <nodoc/>
         public bool Succeeded { get; }
 
+        /// <summary>
+        /// A sorted list of the names of environment variables that could affect the build.
```
smera: >sorted list [](start = 14, length = 11)

why the fact that they are sorted is important?

Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildGraphBuilder.cs
```diff
@@ -182,7 +186,11 @@ public static class MsBuildGraphBuilder
                         locatedMsBuildPath);
                 }
 
-                return ProjectGraphWithPredictionsResult.CreateSuccessfulGraph(projectGraphWithPredictions, assemblyPathsToLoad, locatedMsBuildPath);
+                return ProjectGraphWithPredictionsResult.CreateSuccessfulGraph(
+                    projectGraphWithPredictions, 
+                    assemblyPathsToLoad, 
+                    locatedMsBuildPath, 
+                    environmentVariablesLogger.PotentialEnvironmentVariablesReads.OrderBy(s => s));
```
smera: >PotentialEnvironmentVariablesReads [](start = 47, length = 34)

how does the logger inform that the read env variables are not available?

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using BuildXL.Utilities.Collections;
+using Microsoft.Build.Framework;
+
+namespace ProjectGraphBuilder
+{
+    internal class PropertyTrackingLogger : ILogger
```
smera: >class [](start = 13, length = 5)

sealed?

Public/Src/FrontEnd/MsBuild.Serialization/ProjectGraphWithPredictionsResult.cs
maneely: @smera, I didn't want to assume that subsequent builds of a project would always emit log events in the same order (or that they're received in the same order). So for consistency, I was returning them sorted. If the consumer (BuildXL) doesn't care, there's no need for it.

smera:
> using System;

nit: missing headers and comments in general

---
Refers to: Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs:1 in 74a7572. [](commit_id = 74a7572102ceac3b95576908b5b45ab69950a193, deletion_comment = False)

Public/Src/FrontEnd/MsBuild.Serialization/ProjectGraphWithPredictionsResult.cs
smera: There shouldn't be a need for that. Each variable is just registered as 'tracked' so graph construction caching is aware of it.

---
In reply to: [306449982](https://github.com/microsoft/BuildXL/pull/648#discussion_r306449982) [](ancestors = 306449982)

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.Build.Framework;
+
+namespace ProjectGraphBuilder
+{
+    internal class PropertyTrackingLogger : ILogger
+    {
+        private readonly HashSet<string> _variablesRead = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        private bool _nonTrackingSdkResolversExist = false;
+
+        public IEnumerable<string> PotentialEnvironmentVariablesReads
```
smera: >PotentialEnvironmentVariablesReads [](start = 35, length = 34)

Can we add a contract here so this property is only queried when tracking is available?

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+                lock(_variablesRead)
+                {
+                    copy = _variablesRead.ToArray();
+                }
+
+                return copy.AsReadOnlyCollection();
+            }
+        }
+
+        public bool NonTrackingSdkResolversExist => _nonTrackingSdkResolversExist;
```
smera: >NonTrackingSdkResolversExist [](start = 20, length = 28)

Nit: I'd just call this something like 'IsTrackingAvailable'. The Sdk resolver thing sounds like a detail at this level

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+                                                  lock(_variablesRead)
+                                                  {
+                                                      _variablesRead.Add(upArgs.PropertyName);
+                                                  }
+                                              }
+                                              else if (args is BuildWarningEventArgs warnArgs && warnArgs.Code == "MSB4263")
+                                              {
+                                                  // We just want to know whether or not we can trust the above tracking data.
+                                                  _nonTrackingSdkResolversExist = true;
+                                              }
```
smera: So when tracking is not available, can we get a reason for why? So we can print that as a warning on bxl side

Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildGraphBuilder.cs
smera: Never mind. Saw that. But in that case I suggest we not only store the enumeration, but something like a Failure<> or Maybe<> to represent that the prediction is not there, and ideally a reason (like which resolver is to blame)

---
In reply to: [306449763](https://github.com/microsoft/BuildXL/pull/648#discussion_r306449763) [](ancestors = 306449763)

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using BuildXL.Utilities.Collections;
+using Microsoft.Build.Framework;
+
+namespace ProjectGraphBuilder
+{
+    internal class PropertyTrackingLogger : ILogger
+    {
+        private readonly HashSet<string> _variablesRead = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
dannyvv: our coding guideline is to use `m_`

dannyvv: Why the locking and not simply a `ConcurrentSet<string>`?

dannyvv: If we keep locking, we have the pattern to create a seperate object for the lock and are not locking on the set directly. The fact that you don't expose the set makes this case safe, but it is too easy to accidentally leak it and expose issues later.

i.e. `private readonly object m_variablesReadLock =new object()`

---
In reply to: [306899167](https://github.com/microsoft/BuildXL/pull/648#discussion_r306899167) [](ancestors = 306899167)

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+            get
+            {
+                IEnumerable<string> copy;
+
+                lock(_variablesRead)
+                {
+                    copy = _variablesRead.ToArray();
+                }
+
+                return copy.AsReadOnlyCollection();
```
dannyvv: >copy [](start = 23, length = 4)

We use net472 and netcore30, you should not have to use `.AsReadOnlyCollection`. `return m_variablesRead.ToArray()` should be enough

Public/Src/FrontEnd/MsBuild.Serialization/ProjectGraphWithPredictionsResult.cs
maneely: The sorting will be removed in the next commit.

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
maneely: Changed in latest commit.

Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildGraphBuilder.cs
maneely: Changes made in next update. The PropertyTrackingLogger will now surface a Possible<>. If it returns a failure, it'll be reported.

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
maneely: This property was changed. See the new logic in upcoming commit.

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
maneely: The property is removed in the latest commit. It's possible there could be many reasons tracking isn't working/available/reliable. Sdk resolution is just one. Regardless, they'll all be reported as failures to the graph builder.

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
maneely: Added a call to reporter.ReportMessage when there is a failure. Seen in latest commit.

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
maneely: Changed in latest commit.

maneely: @dannyvv, I'm looking for a ConcurrentSet in the codebase and the only thing I see is:
[ConcurrentHashSet](https://github.com/microsoft/BuildXL/blob/master/Private/Wdg/ExecutionLogSdk/ConcurrentHashSet.cs) and [ConcurentBigSet](https://github.com/microsoft/BuildXL/blob/master/Public/Src/Utilities/Collections/ConcurrentBigSet.cs)

The former is in a project not referenced by this one. The latter is geared towards large sets where in this instance it will almost always be <1000 elements.

Am I missing something?

Should I a) bring over the ConcurrentHashSet class above into the current project, duplicating code, b) bring the ConcurrentHashSet class into the Utilities project to be referenced by both the project currently hosting that class and this one? c) Use a standard class like ConcurrentDictionary<string, object> and essentially use it as a hash set by disregarding the value?  I'd think option B to be the best, but I'm certainly not sure of the ramifications of moving such classes around.

Thoughts?

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
maneely: Removed in latest commit.

maneely:
Clarifying note on this PR:
Currently this is just to collaborate on how we want the changes to look/behave. There is more work to do (e.g. UTs) and there is work this is depending on that hasn't been merged into MSBuild yet.

Public/Src/Tools/Tool.MsBuildGraphBuilder/PropertyTrackingLogger.cs
dannyvv: Ah didn't think about referneces....
My personal preference would then be use ConcurrentDictionary<T,U>

---
In reply to: [308304479](https://github.com/microsoft/BuildXL/pull/648#discussion_r308304479) [](ancestors = 308304479)

maneely: Changed to ConcurrentDictionary in latest commit.

olkononenko:
Moved 'WIP' into the tile to prevent an accidental merge.

PR #651, opened 2019-07-23
Dedup pin optimizations only with nodes, not chunks

JuanCarlosGI:
Errors were happening when performing the pin optimization, since it was made with only nodes in mind, and not chunks. Chunks don't need optimizing, so just call the regular implementation with them.

### DIFF
--- a/Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
+++ b/Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
@@ -178,6 +178,15 @@ namespace BuildXL.Cache.ContentStore.Vsts
                 return pinResult;
             }
 
+            VstsBlobIdentifier blobId = ToVstsBlobIdentifier(contentHash.ToBlobIdentifier());
+            VstsDedupIdentifier dedupId = blobId.ToDedupIdentifier();
+
+            if (dedupId.AlgorithmId == Hashing.ChunkDedupIdentifier.ChunkAlgorithmId)
+            {
+                // No need to optimize since pinning a chunk is always a fast operation.
+                return await PinImplAsync(context, contentHash);
+            }
+
             // Since pinning the whole tree can be an expensive operation, we have optimized how we call it. Depending on the current
             //  keepUntil of the root node, which is unexpensive to check, the operation will behave differently:
             //      The pin operation will be ignored if it is greater than ignorePinThreshold, to reduce amount of calls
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -178,6 +178,15 @@ private enum Counters
                 return pinResult;
             }
 
+            VstsBlobIdentifier blobId = ToVstsBlobIdentifier(contentHash.ToBlobIdentifier());
+            VstsDedupIdentifier dedupId = blobId.ToDedupIdentifier();
+
+            if (dedupId.AlgorithmId == Hashing.ChunkDedupIdentifier.ChunkAlgorithmId)
```
dcwuser: This optimization makes sense, but I'd like to understand when it comes into play in practice. Is the situation that, if a piece of content is small, we map it's dedup content hash directly to a single chunk hash, and if it is large, we map it to a root node hash?

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
```diff
@@ -178,6 +178,15 @@ private enum Counters
                 return pinResult;
             }
 
+            VstsBlobIdentifier blobId = ToVstsBlobIdentifier(contentHash.ToBlobIdentifier());
```
dannyvv: Would it make sense to make the optimization more effecient by checking `Contenthash.HashType`?

i.e. 
```
if (contentHash.HashType == HashType.DedupChunk)
```

now we are converting and deserializing just to check and not using blobId or dedupId in the fast path...

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
JuanCarlosGI: Initially I thought that if the content was small, there would always be one root node pointing to a single chunk. However, I saw build failures where the optimization was used on chunks:

ArgumentException: System.ArgumentException: The given hash does not represent a NodeDedupIdentifier
   at Microsoft.VisualStudio.Services.BlobStore.Common.NodeDedupIdentifier.Validate()
   at Microsoft.VisualStudio.Services.BlobStore.Common.DedupIdentifier.CastToNodeDedupIdentifier()

Public/Src/Cache/ContentStore/Vsts/DedupReadOnlyContentSession.cs
JuanCarlosGI: Not sure what we would be saving, since what you're suggesting still requires us to do the conversion every single time.

PR #652, opened 2019-07-23
Revamp incremental nuget spec generation

mpysson:
The nuget spec generation incrementality was based on a "PackageHash" file which was tied to the materialization of outputs on disk but not the actual generation of the spec file. This was problematic for a number of reasons:
- If spec generation was incomplete or never ran, the spec would be locked in the bad state
- If the format of the PackageHash file gets updated in a way that only impacted spec generation, there was nothing to re-emit the file once spec generation ran again.
The result of these was that the specs would sometimes not get generated when they needed to and other times get stuck getting regenerated every build. On my machine 90% of packages were trapped in the always regenerate mode.

This change decouples the state tracking of fetching the package contents from the spec generation. This allows the lifetime of the state tracking file to align with writing out the spec file. It is self-healing on previously downloaded packages and allows me to now not regenerate any spec files when nothing changes. I'm hopeful it will also address the weird things we see every now and then that force us to nuke our generated specs. 

If it doesn't address the weird things I'd like us to just remove the incrementality for spec generation since prior to this change the incremental generation wouldn't have been providing much benefit anyway.

### DIFF
--- a/Public/Src/FrontEnd/Nuget/WorkspaceNugetModuleResolver.cs
+++ b/Public/Src/FrontEnd/Nuget/WorkspaceNugetModuleResolver.cs
@@ -921,11 +921,19 @@ namespace BuildXL.FrontEnd.Nuget
         private bool CanReuseSpecFromDisk(NugetAnalyzedPackage analyzedPackage)
         {
             var packageDsc = GetPackageDscFile(analyzedPackage).ToString(PathTable);
+            
+            // This file contains some state from the last time the spec file was generated. It includes
+            // the fingerprint of the package (name, version, etc) and the version of the spec generator.
+            // It is stored next to the primary generated spec file
+            var (fileFormat, fingerprint) = ReadGeneratedSpecStateFile(packageDsc + SpecGenerationVersionFileSuffix);
 
-            var state = ReadPackageStateFile(packageDsc + SpecGenerationVersionFileSuffix);
-            if (state.fileFormat == SpecGenerationFormatVersion &&
-                state.fingerprint != null &&
-                state.fingerprint == analyzedPackage.PackageOnDisk.PackageDownloadResult.FingerprintHash && 
+            // We can reuse the already generated spec file if all of the following are true:
+            //  * The spec generator is of the same format as when the spec was generated
+            //  * The package fingerprint is the same. This means the binaries are the same
+            //  * Both the generated spec and package config file exist on disk
+            // NOTE: This is not resilient to the specs being modified by other entities than the build engine.
+            if (fileFormat == SpecGenerationFormatVersion &&
+                fingerprint != null && fingerprint == analyzedPackage.PackageOnDisk.PackageDownloadResult.FingerprintHash && 
                 File.Exists(packageDsc) &&
                 File.Exists(GetPackageConfigDscFile(analyzedPackage).ToString(PathTable)))
             {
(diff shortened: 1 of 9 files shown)


Kr1sso:
That fingerprint test sure is flaky :|

Public/Src/FrontEnd/Nuget/WorkspaceNugetModuleResolver.cs
```diff
@@ -918,16 +922,52 @@ private bool CanReuseSpecFromDisk(NugetAnalyzedPackage analyzedPackage)
         {
             var packageDsc = GetPackageDscFile(analyzedPackage).ToString(PathTable);
 
-            if (analyzedPackage.Source == PackageSource.Disk &&
-                analyzedPackage.PackageOnDisk.PackageDownloadResult.SpecsFormatIsUpToDate &&
-                File.Exists(packageDsc))
+            var state = ReadPackageStateFile(packageDsc + SpecGenerationVersionFileSuffix);
```
dannyvv: coudl write

```
var (fileformat, fingerprint) = .....
```

for simpler code

Public/Src/FrontEnd/Nuget/WorkspaceNugetModuleResolver.cs
```diff
@@ -918,16 +922,52 @@ private bool CanReuseSpecFromDisk(NugetAnalyzedPackage analyzedPackage)
...
+                }
+            }
+
+            // Error
+            return (-1, null);
+        }
+
+        private static void WritePackageStateFile(string path, (int fileFormat, string fingerprint) data)
+        {
+            Directory.CreateDirectory(Path.GetDirectoryName(path));
```
dannyvv: do we need try/catch with a warning that the next build the package will be downloaded again? Not sure if the old code, or my attempt and rewrite 18 months ago had that...

Public/Src/FrontEnd/Nuget/WorkspaceNugetModuleResolver.cs
```diff
@@ -918,16 +922,52 @@ private bool CanReuseSpecFromDisk(NugetAnalyzedPackage analyzedPackage)
...
+                File.Exists(packageDsc) &&
+                File.Exists(GetPackageConfigDscFile(analyzedPackage).ToString(PathTable)))
             {
                 return true;
             }
 
             return false;
         }
 
+        private static (int fileFormat, string fingerprint) ReadPackageStateFile(string path)
```
narasamdya: >ReadPackageStateFile [](start = 60, length = 20)

`ReadGeneratedSpecStateFile` perhaps?

narasamdya:
>         private bool CanReuseSpecFromDisk(NugetAnalyzedPackage analyzedPackage)

May be some comments about the file you're using to check up-to-dateness of generated specs and of downloaded package.

---
Refers to: Public/Src/FrontEnd/Nuget/WorkspaceNugetModuleResolver.cs:921 in b6e4f1d. [](commit_id = b6e4f1d055ef9e0f085b943cbc90716e8206c5c7, deletion_comment = False)

Public/Src/FrontEnd/Nuget/WorkspaceNugetModuleResolver.cs
mpysson: added

PR #655, opened 2019-07-24
[macOS Catalina] Data partition filtering and file access adjustments

Kr1sso:
This PR allows for BuildXL to run on macOS Catalina. It adjusts the KEXT to take into account that user folders are now hosted on a readable/writeable partition `/System/Volumes/Data` and fixes a few new DFAs as well as adjusts tests to account for the increased amount of file access reports on directory enumeration.

### DIFF
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -4739,7 +4739,7 @@ namespace BuildXL.Scheduler
                             {
                                 ReportQueueSizeMB = m_configuration.Sandbox.KextReportQueueSizeMb,
                                 EnableReportBatching = m_configuration.Sandbox.KextEnableReportBatching,
-#if FEATURE_CORECLR
+#if PLATFORM_OSX
                                 EnableCatalinaDataPartitionFiltering = OperatingSystemHelper.IsMacOSCatalinaOrHigher,
 #endif
                                 ResourceThresholds = new Sandbox.ResourceThresholds
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -1265,18 +1265,18 @@ namespace BuildXL.Scheduler
             m_serviceManager.Start(loggingContext, OperationTracker);
             m_apiServer?.Start(loggingContext);
             m_chooseWorkerCpu = new ChooseWorkerCpu(
-                loggingContext, 
-                m_configuration.Schedule.MaxChooseWorkerCpu, 
+                loggingContext,
+                m_configuration.Schedule.MaxChooseWorkerCpu,
                 m_workers,
-                m_pipQueue, 
-                PipGraph, 
+                m_pipQueue,
+                PipGraph,
                 m_fileContentManager);
 
             m_chooseWorkerCacheLookup = new ChooseWorkerCacheLookup(
-                loggingContext, 
-                m_configuration.Schedule.MaxChooseWorkerCacheLookup, 
-                m_configuration.Distribution.DistributeCacheLookups, 
-                m_workers, 
+                loggingContext,
+                m_configuration.Schedule.MaxChooseWorkerCacheLookup,
+                m_configuration.Distribution.DistributeCacheLookups,
+                m_workers,
                 m_pipQueue);
 
             ExecutionLog?.DominoInvocation(new DominoInvocationEventData(m_configuration));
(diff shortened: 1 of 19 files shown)


Public/Src/Utilities/Utilities/OperatingSystemHelper.cs
```diff
@@ -68,6 +68,12 @@ public FileSize(long bytes) : this((ulong)bytes)
         /// </summary>
         public static readonly bool IsMacOS = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
 
+        /// <summary>
+        /// Indicates if Catalina (10.15) or a higher macOS version is running on the host
+        /// </summary>
+        public static readonly bool IsMacOSCatalinaOrHigher = GetOSVersionMacOS().CompareTo("macOS 10.15.0") <= 0;
```
dannyvv: >GetOSVersionMacOS [](start = 62, length = 17)

shouldn't this be: >=?

`"macOS 10.16.0".CompareTo("macOS 10.15.0")` 
return 1.

Note: does mac version its bits so string comparisson is okay?

i.e. what about `10.9.0` -or- `10.15.1` vs `10.15.10`

dannyvv: best code is probably split by space and use new Version(part[1]) and compare that against new Version(10,15,0)

---
In reply to: [307053014](https://github.com/microsoft/BuildXL/pull/655#discussion_r307053014) [](ancestors = 307053014)

dannyvv: Now that I look at the code for GetOsVersionMacOS() you already have a version object there... Perhaps we should refactor that method and have it expose a tuple...

---
In reply to: [307053303](https://github.com/microsoft/BuildXL/pull/655#discussion_r307053303) [](ancestors = 307053303,307053014)

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4739,6 +4739,9 @@ protected virtual bool InitSandboxedKextConnection(LoggingContext loggingContext
                             {
                                 ReportQueueSizeMB = m_configuration.Sandbox.KextReportQueueSizeMb,
                                 EnableReportBatching = m_configuration.Sandbox.KextEnableReportBatching,
+#if FEATURE_CORECLR
```
almiliMSFT: Did you mean
```suggestion
#if PLATFORM_OSX
```
instead?

Public/Src/Engine/UnitTests/FingerprintStore/FingerprintStoreTests.cs
```diff
@@ -149,7 +149,7 @@ public void DontOverwriteExistingContentAddressableEntries()
 
             var counters = testHooks.FingerprintStoreTestHooks.Counters;
             XAssert.AreEqual(1, counters.GetCounterValue(FingerprintStoreCounters.NumPathSetEntriesPut));
-            XAssert.AreEqual(1, counters.GetCounterValue(FingerprintStoreCounters.NumDirectoryMembershipEntriesPut));
+            XAssert.IsTrue(counters.GetCounterValue(FingerprintStoreCounters.NumDirectoryMembershipEntriesPut) >= 1);
```
almiliMSFT: I don't know what this test is exactly supposed to do, but it's called "DontOverwriteExisting...Entries", which makes me think that checking `NumDirectoryMembershipEntriesPut == 1` vs `NumDirectoryMembershipEntriesPut >= 1` is significant

Public/Src/Sandbox/MacOs/Sandbox/Src/Kauth/AccessHandler.cpp
```diff
@@ -205,6 +205,22 @@ static bool VNodeMatchesPath(vnode_t vp, vfs_context_t ctx, const char *path)
...
 
+const char* AccessHandler::IgnoreCatalinaDataPartitionPrefix(const char* path)
+{
+    if (!sandbox_->GetConfig().enableCatalinaDataPartitionFiltering)
+    {
+        return path;
+    }
+
+    const char *marker = path;
+    if (strprefix(marker, "/System/Volumes/Data/"))
```
almiliMSFT: Extract `"/System/Volumes/Data/"` into a macro or a const.  Then it will be clearer that here this string has a trailing `/` whereas 2 lines below it doesn't.

Public/Src/Sandbox/MacOs/Sandbox/Src/Kauth/AccessHandler.cpp
```diff
@@ -205,6 +205,22 @@ static bool VNodeMatchesPath(vnode_t vp, vfs_context_t ctx, const char *path)
...
+    {
+        marker = marker + strlen("/System/Volumes/Data");
```
almiliMSFT: How many times is this line going to execute per single build?  If more than once (which seems to be the case), extract `strlen("/System/Volums/Data")` into a constant.

Kr1sso: Extracted it into a const, much cleaner - good point.

Public/Src/Sandbox/MacOs/Sandbox/Src/Kauth/AccessHandler.cpp
Kr1sso: Refactored into consts.

Public/Src/Engine/Scheduler/Scheduler.cs
Kr1sso: Refactored, indeed.

Public/Src/Utilities/Utilities/OperatingSystemHelper.cs
Kr1sso: Refactored, thanks for the suggestions.

PR #659, opened 2019-07-25
Make vscode plugin load when workspace has config.dsc

dannyvv:


Public/Src/IDE/VsCode/client/package.json
```diff
@@ -20,7 +20,8 @@
   "main": "./out/src/extension",
   "activationEvents": [
     "onLanguage:DScript",
-    "onCommand:DScript.reloadWorkspace"
+    "onCommand:DScript.reloadWorkspace",
+    "workspaceContains:config.dsc"
```
smera: maybe 'config.bc' as well

dannyvv: nah, I'll update osgtools one day :)

smera: Sadly it spread out to other places as well (like the CB repo). But that's fine, feel free to ignore my comment.

PR #660, opened 2019-07-25
[WIP] Move cache back to Newtonsoft V10

kewilli:


Public/Src/Cache/DistributedCache.Host/Configuration/BuildXL.Cache.Host.Configuration.dsc
```diff
@@ -3,20 +3,22 @@
 
 namespace Configuration {
 
-    @@public
+    export const references = [
```
dannyvv: why split the references?
If not too much troubpe, would you mind inlining them to be consistent with the rest?

PR #662, opened 2019-07-25
Faster ByteArrayComparer.Equals

erikma:
Based on feedback from the internal Azure datacenter perf team. We remove IList<byte> based overloads that cause code paths to use higher-level boxed enumerators, and add a faster overload for the equality operator, differentiated by NetCore vs. Framework.

Comparative performance (100M iterations, retail build)

- VSO0 33-byte hashes, comparing hashes that differ at byte position 2:
  Old: 548ms
  New-net472: 400ms (-27%)
  New-NetCore: 93ms (-83%, 6X faster)

- VSO0 33-byte same hashes
  Old: 3445ms
  New-net472: 692ms (-80% or 5X faster)
  New-NetCore: 336ms (-90%, 10X faster)

- SHA256 32-byte same hashes:
  Old: 2552ms
  New-net472: 707ms (-72%)
  New-NetCore: 259ms (-90%, 10X faster)

### DIFF
--- a/Public/Src/Cache/ContentStore/Hashing/ByteArrayComparer.cs
+++ b/Public/Src/Cache/ContentStore/Hashing/ByteArrayComparer.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 
 namespace BuildXL.Cache.ContentStore.Interfaces.Utils
@@ -203,13 +204,16 @@ namespace BuildXL.Cache.ContentStore.Interfaces.Utils
         ///     Static comparison method for when the caller does not want to create an instance
         ///     of this class.
         /// </summary>
+#if FEATURE_CORECLR
+        public static bool ArraysEqual(byte[] x, byte[] y)
+        {
+            return x.AsSpan().SequenceEqual(y.AsSpan());
+        }
+#else
         public static unsafe bool ArraysEqual(byte[] x, byte[] y)
         {
             // Adapted from: https://gist.github.com/airbreather/90c5fd3ba9d77fcd7c106db3beeb569b
             // (https://stackoverflow.com/questions/43289/comparing-two-byte-arrays-in-net, Joe Amenta's answer).
-            // TODO: Once we get to NetCore 2.2+, consider ReadOnlyMemory<byte>.SequenceEqual which should be
-            // faster per the SO thread. In testing on net472 at the time of moving to this implementation
-            // this implementation was 50%-100% faster.
             
             // Copyright (c) 2008-2013 Hafthor Stefansson
             // Distributed under the MIT/X11 software license
@@ -266,5 +270,6 @@ namespace BuildXL.Cache.ContentStore.Interfaces.Utils
                 return true;
             }
         }
+#endif
     }
 }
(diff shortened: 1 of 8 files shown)


Public/Src/Cache/ContentStore/Hashing/ByteArrayComparer.cs
```diff
@@ -217,9 +203,68 @@ private static int CompareArrays(IList<byte> x, IList<byte> y)
         ///     Static comparison method for when the caller does not want to create an instance
         ///     of this class.
         /// </summary>
-        public static bool ArraysEqual(IList<byte> x, IList<byte> y)
+        public static unsafe bool ArraysEqual(byte[] x, byte[] y)
         {
-            return CompareArrays(x, y) == 0;
+            // Adapted from: https://gist.github.com/airbreather/90c5fd3ba9d77fcd7c106db3beeb569b
+            // (https://stackoverflow.com/questions/43289/comparing-two-byte-arrays-in-net, Joe Amenta's answer).
+            // TODO: Once we get to NetCore 2.2+, consider ReadOnlyMemory<byte>.SequenceEqual which should be
```
dannyvv: > get to NetCore 2.2+, [](start = 28, length = 21)

We are actually compiling this for:
* netstandard20
* net451
* net461
* net472
* netcoreapp30

BuildXL is slowly moving away from net472 and moving to netcoreapp30. We are close to flipping it in Cloudbuild.

Do you mind measuring on netcore30 and if not improved their, `#if FEATURE_CORECLR`

Public/Src/Cache/ContentStore/Hashing/ByteArrayComparer.cs
```diff
@@ -217,9 +203,68 @@ private static int CompareArrays(IList<byte> x, IList<byte> y)
...
         {
-            return CompareArrays(x, y) == 0;
+            // Adapted from: https://gist.github.com/airbreather/90c5fd3ba9d77fcd7c106db3beeb569b
+            // (https://stackoverflow.com/questions/43289/comparing-two-byte-arrays-in-net, Joe Amenta's answer).
+            // TODO: Once we get to NetCore 2.2+, consider ReadOnlyMemory<byte>.SequenceEqual which should be
+            // faster per the SO thread. In testing on net472 at the time of moving to this implementation
+            // this implementation was 50%-100% faster.
+            
+            // Copyright (c) 2008-2013 Hafthor Stefansson
+            // Distributed under the MIT/X11 software license
```
dannyvv: Don't forget to manually register it to component governance.

Public/Src/Cache/ContentStore/Hashing/ByteArrayComparer.cs
dannyvv: Actually since we are on notecore30 the SO thread refers to SIMD. Should definately be worth trying

---
In reply to: [307537497](https://github.com/microsoft/BuildXL/pull/662#discussion_r307537497) [](ancestors = 307537497)

Public/Src/Cache/ContentStore/Test/Utils/ByteArrayComparerTests.cs
```diff
@@ -61,37 +61,6 @@ public void ArrayEqualToItself()
             Assert.True(_comparer.Equals(b, b));
         }
 
-        [Fact]
-        public void ListEqualsTrue()
```
kewilli: >ListEqualsTrue [](start = 20, length = 14)

Is there any harm in migrating these tests to byte arrays?

Public/Src/Cache/ContentStore/Hashing/ByteArrayComparer.cs
dannyvv: I have a PR #665 to stop building this for net461 (just FYI)

dannyvv: The PR went in.

erikma: Significant speedup with Span.SequenceEqual, added a split in the implementation based on netcore vs. framework.

PR #666, opened 2019-07-26
Exceptions on work retry should not cause contract violation

narasamdya:
[AB#1575540](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1575540)

### DIFF
--- a/Public/Src/Utilities/Native/IO/Helpers.cs
+++ b/Public/Src/Utilities/Native/IO/Helpers.cs
@@ -37,11 +37,8 @@ namespace BuildXL.Native.IO
         /// The <code>work</code> function is invoked a few times; its parameter indicates if the invocation is the last one before
         /// finally giving up; its return value indicates if all work has finished successfully and no further attempt is needed.
         /// </param>
-        /// <param name="logExceptions">
-        /// Whether exceptions unhandled by the caller should be logged. By default, exceptions will be logged.
-        /// </param>
-        /// <param name="rethrowExceptions">
-        /// Whether exceptions unhandled by the caller should be re-thrown, breaking the retry loop. By default, exceptions will be ignored and work() will be retried.
+        /// <param name="onException">
+        /// Handle exception. By default, exceptions will be ignored and work() will be retried.
         /// </param>
         /// <param name="numberOfAttempts">
         /// Total number of attempts to complete when trying an action that may fail.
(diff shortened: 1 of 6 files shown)


Public/Src/Utilities/Native/IO/Helpers.cs
```diff
@@ -83,17 +78,10 @@ public static class Helpers
...
-                    
-                    // Only log exceptions if they are not being re-thrown already
-                    if (logExceptions)
-                    {
-                        Tracing.Logger.Log.RetryOnFailureException(Events.StaticContext, e.ToString());
-                    }
+                    onException?.Invoke(e);
+
+                    // For diagnostic purpose.
+                    Tracing.Logger.Log.RetryOnFailureException(Events.StaticContext, e.ToString());
```
mpysson: Now that it's up the caller to decide if/what to log, should we even bother calling this?

Public/Src/Utilities/UnitTests/Utilities/HelpersTests.cs
```diff
@@ -88,5 +88,76 @@ public void RetryOnFailureAsyncTest()
...
+        {
+            const int MaxAttempts = 3;
+            int attemptCount = 0;
+            bool result = Helpers.RetryOnFailure(
+                lastAttempt =>
+                {
+                    attemptCount++;
+                    throw new BuildXLException(string.Empty);
+                },
+                numberOfAttempts: MaxAttempts, initialTimeoutMs: 100, postTimeoutMultiplier: 1);
```
mpysson: >initialTimeoutMs: 100 [](start = 47, length = 21)

It shouldn't matter to these tests how long is spent between attempts. Can we shrink this down to somethign much smaller to keep our test runtime small? Same comment for all other occurrences in this file

Public/Src/Utilities/Native/IO/Windows/FileUtilities.Win.cs
```diff
@@ -279,7 +279,7 @@ public static string NormalizeDirectoryPath(string path)
                             // Deletion is successful when the child directories/files count matches the expected count
                             return actualRemainingChildCount == remainingChildCount;
                         },
-                        rethrowExceptions: true /* exceptions thrown in work() will occur again on retry, so just rethrow them */);
+                        onException: e => throw e /* exceptions thrown in work() will occur again on retry, so just rethrow them */);
```
dannyvv: Just sharing that this will change the callstack of the error thrown to be shown here, rather than in the original callstack since we used to use `throw`

Public/Src/Utilities/Native/IO/Helpers.cs
narasamdya: For diagnostic purpose.

Public/Src/Utilities/Native/IO/Windows/FileUtilities.Win.cs
narasamdya: Good catch! Yeah, I should've tried to preserved the callstack.

PR #667, opened 2019-07-26
Add RetryIntervalForCopiesMs with a serializable type

kewilli:
Add RetryIntervalForCopiesMs with a serializable type

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentStoreSettings.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentStoreSettings.cs
@@ -130,16 +130,10 @@ namespace BuildXL.Cache.ContentStore.Distributed.Stores
         /// </summary>
         public int ParallelCopyFilesLimit { get; set; } = DefaultParallelCopyFilesLimit;
 
-        private IReadOnlyList<TimeSpan> _retryIntervalForCopies = CacheCopierDefaultRetryIntervals;
-
         /// <summary>
         /// Delays for retries for file copies
         /// </summary>
-        public IReadOnlyList<TimeSpan> RetryIntervalForCopies
-        {
-            get => _retryIntervalForCopies ?? CacheCopierDefaultRetryIntervals;
-            set => _retryIntervalForCopies = value;
-        }
+        public IReadOnlyList<TimeSpan> RetryIntervalForCopies { get; set; } = CacheCopierDefaultRetryIntervals;
 
         /// <summary>
         /// Defines pinning behavior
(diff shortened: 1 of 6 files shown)


Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentStoreSettings.cs
```diff
@@ -130,16 +130,10 @@ public int AssumeAvailableReplicaCount
...
-
         /// <summary>
         /// Delays for retries for file copies
         /// </summary>
-        public IReadOnlyList<TimeSpan> RetryIntervalForCopies
-        {
-            get => _retryIntervalForCopies ?? CacheCopierDefaultRetryIntervals;
-            set => _retryIntervalForCopies = value;
-        }
+        public IReadOnlyList<TimeSpan> RetryIntervalForCopies { get; set; } = CacheCopierDefaultRetryIntervals;
```
dannyvv: Do we have to worry about the behavior chagne where in the old code if someone set it to null, we'd still return the defaults. Now we'll start returning nul...

kewilli: These are all set via settings, so this should not be an issue.

---
In reply to: [307925941](https://github.com/microsoft/BuildXL/pull/667#discussion_r307925941) [](ancestors = 307925941)

Public/Src/Cache/DistributedCache.Host/Configuration/DistributedContentSettings.cs
```diff
@@ -177,6 +177,30 @@ private DistributedContentSettings(IDictionary<string, string> connectionSecretN
         [DataMember]
         public int EvictionWindowSize { get; set; } = 500;
 
+        /// <summary>
+        /// Delays for retries for file copies
+        /// </summary>
+        [DataMember]
+        public int[] RetryIntervalForCopiesMs { get; set; } =
```
jbayardo: Ignorant question here: is it possible for serialization to actually set this to null? if so, things would explode both in the line below and in the code that uses it. There is no `Contract.Requires` in http://index/?leftProject=BuildXL.Cache.ContentStore.Distributed&leftSymbol=bjztwptambty&file=Stores%5CDistributedContentCopier.cs&line=60

kewilli: Not ignorant at all. It could happen. Now that two of you have pointed this out, I'll add some handrails.

---
In reply to: [307927695](https://github.com/microsoft/BuildXL/pull/667#discussion_r307927695) [](ancestors = 307927695)

PR #672, opened 2019-07-29
Allow RocksDB to perform log rotation

jbayardo:
As of today, RocksDB logs are overwritten every time the DB is open and we only keep one file, so that means all logs are lost on every service restart.

This is inadequate for debugging purposes, as the log contains very useful information; for example: compaction rates in the different levels of the LSM tree, file read latencies, read and write amplification, among others.

I'm not sure these are the values we want, the first two are the defaults, and the third I just set to 1 hour, but we may want to rotate at a different cadence.

### DIFF
--- a/Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
+++ b/Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
@@ -140,13 +140,22 @@ namespace BuildXL.Engine.Cache.KeyValueStores
             /// If a store already exists at the given directory, whether any columns that mismatch the the columns that were passed into the constructor
             /// should be dropped. This will cause data loss and can only be applied in read-write mode.
             /// </param>
+            /// <param name="rotateLogs">
+            /// Have RocksDb rotate logs, useful for debugging performance issues. It will rotate logs every 12 hours, 
+            /// up to a maximum of 60 logs (i.e. 30 days). When the maximum amount of logs is reached, the oldest logs
+            /// are overwritten in a circular fashion.
+            /// 
+            /// Every time the RocksDb instance is open, the current log file is truncated, which means that if you
+            /// open the DB more than once in a 12 hour period, you will only have partial information.
+            /// </param>
             public RocksDbStore(
                 string storeDirectory,
                 bool defaultColumnKeyTracked = false,
                 IEnumerable<string> additionalColumns = null,
                 IEnumerable<string> additionalKeyTrackedColumns = null,
                 bool readOnly = false,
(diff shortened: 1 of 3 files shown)


Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
```diff
@@ -197,6 +197,15 @@ private struct DefaultOptions
...
                     .SetPrefixExtractor(SliceTransform.CreateNoOp());
 
+                // Maximum number of information log files
+                m_defaults.DbOptions.SetKeepLogFileNum(1000);
+
+                // Do not rotate information logs based on file size
+                m_defaults.DbOptions.SetMaxLogFileSize(0);
+
+                // How long before we rotate the current information log file
+                m_defaults.DbOptions.SetLogFileTimeToRoll((ulong)TimeSpan.FromHours(1).Seconds);
```
SergeyTeplyakov: Should we rotate every day instead?
Rotating every hour seems a bit too frequent?

Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
```diff
@@ -197,6 +197,15 @@ private struct DefaultOptions
                     .SetBlockBasedTableFactory(blockBasedTableOptions)
                     .SetPrefixExtractor(SliceTransform.CreateNoOp());
 
+                // Maximum number of information log files
+                m_defaults.DbOptions.SetKeepLogFileNum(1000);
```
mpysson: What performance implications does this have on other consumers of RocksDb in the codebase? Is it going to cause BuildXL to keep a zillion rocksDB log files?

jbayardo: I'm afraid it would, yes. This is mainly useful in the case of lab builds; hence, one way to move forward is to add a flag defaulted to false, which we can enable for our use case. Another option is to recycle the log files, so we can just keep at most N log files with at most size S, and RocksDB would proceed in a ring fashion as it runs out of files.

Please let me know which option you think is best.

mpysson: I'd prefer the different flag approach. For the build engine keeping any more than one log file is going to be overkill

Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
jbayardo: Rotating every 12 hours to avoid loosing an entire day out as a result of master changing.

PR #679, opened 2019-07-31
Implement GC stub for RocksDbMemoizationStore

jbayardo:
This PR changes how the `ContentLocationDatabase`'s GC works. From here on, it will always be on; the only difference is that when a machine is demoted from master to slave, only the metadata GC will run, instead of both content and metadata.

The metadata GC is very straightforward and meant to be a temporary implementation to get local builds working in the test environment; the expectation is to change it so it considers both content and metadata in the next few weeks as we progress towards distributed builds.

Briefly, it removes all metadata entries older than some threshold from the current time. The current SQLiteMemoizationStore adopts a different strategy: it keeps only the top K entries in LRU order. This strategy is impractical to implement in RocksDb because obtaining the top K entries requires a column family sorted by LRU order (i.e. with last access time as keys), so this would imply a book keeping cost (including serialization and deserialization) which seems unnecessary for the current purpose.


Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -703,6 +700,64 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
+        protected override void DoGarbageCollectMetadata(OperationContext context)
+        {
+            context.PerformOperation(Tracer, () =>
+            {
+                var cutoffTimeUtc = (Clock.UtcNow - _configuration.MetadataGarbageCollectionProtectionTime).ToFileTimeUtc();
+
+                return _keyValueStore.Use(store => {
+                    Tracer.Debug(context, "Local database's garbage collection for metadata is starting");
+
+                    using (var stopwatch = Counters[ContentLocationDatabaseCounters.GarbageCollectMetadata].Start())
```
dannyvv: >stopwatch [](start = 31, length = 9)

stopwatch not used, you don't have to declare it as a var..

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -696,6 +685,17 @@ private void LogEntryDeletion(OperationContext context, ShortHash hash, ContentL
...
         public abstract IEnumerable<StructResult<StrongFingerprint>> EnumerateStrongFingerprints(OperationContext context);
 
+        /// <summary>
+        /// Perform garbage collection of metadata entries.
+        /// </summary>
+        /// <param name="context">
+        ///     Tracing context.
+        /// </param>
+        /// <returns>
+        ///     Result providing the call's completion status.
```
SergeyTeplyakov: Should it be Result<T>? BoolResult or something similar?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -696,6 +685,17 @@ private void LogEntryDeletion(OperationContext context, ShortHash hash, ContentL
         /// </returns>
         public abstract IEnumerable<StructResult<StrongFingerprint>> EnumerateStrongFingerprints(OperationContext context);
 
+        /// <summary>
+        /// Perform garbage collection of metadata entries.
+        /// </summary>
+        /// <param name="context">
```
SergeyTeplyakov: I vote for removing such kind of comments.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -285,25 +272,27 @@ public void GarbageCollect(OperationContext context)
 
             using (var cancellableContext = TrackShutdown(context))
             {
-                DoGarbageCollect(cancellableContext);
+                var tasks = new List<Task> { Task.Run(() => DoGarbageCollectMetadata(cancellableContext)) };
+                if (_isContentGarbageCollectionEnabled)
```
SergeyTeplyakov: Do you want to have _isMetadataGarbageCollectionEnabled here as well to have the full logic for the entire GC?

SergeyTeplyakov: Another question: do you think it make sense to separate the configuration? Maybe we can do gc for content and metadata if gc is on, without separating them?

This will make the configuration simpler but maybe less flexible. But I can't imagine (for now) the case when we want to GC one thing but not another...

---
In reply to: [309381691](https://github.com/microsoft/BuildXL/pull/679#discussion_r309381691) [](ancestors = 309381691)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -285,25 +272,27 @@ public void GarbageCollect(OperationContext context)
...
+                    tasks.Add(Task.Run(() => DoGarbageCollectContent(cancellableContext)));
+                }
+                Task.WaitAll(tasks.ToArray());
             }
 
             lock (this)
             {
                 if (!ShutdownStarted)
                 {
-                    if (_isGarbageCollectionEnabled)
```
SergeyTeplyakov: It seems that after this change we'll keep setting the timer even when the GC is disabled. Right?

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -703,6 +700,64 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
+
+                        Tracer.Debug(context, $"Metadata Garbage Collection results: ScannedEntries={scannedEntries}, RemovedEntries={removedEntries}, EntriesAfterGc={scannedEntries - removedEntries}, DbSizeInBytesAfterGc={dbSizeInBytesAfterGc}");
+                    }
+
+                    return Unit.Void;
+                }).ToBoolResult();
+            }).ThrowIfFailure();
+        }
+
+        private void WipeMetadata(IBuildXLKeyValueStore store, byte[] strongFingerprint, MetadataEntry metadata)
```
SergeyTeplyakov: RemoveMetadata?

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -703,6 +700,64 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
+                            var lastAccessTimeUtc = DeserializeMetadataLastAccessTimeUtc(kvp.Value);
+
+                            if (lastAccessTimeUtc < cutoffTimeUtc)
+                            {
+                                var metadata = DeserializeMetadataEntry(kvp.Value);
+                                WipeMetadata(store, kvp.Key, metadata);
+
+                                removedEntries++;
+                                Counters[ContentLocationDatabaseCounters.GarbageCollectMetadataEntriesRemoved].Increment();
+                            } else
```
SergeyTeplyakov: Formatting issue.

Public/Src/Cache/MemoizationStore/Test/Sessions/RocksDbMemoizationSessionTests.cs
```diff
@@ -106,5 +117,52 @@ public Task GetSelectorsGivesSelectorsInReverseLruOrderAfterGet()
...
+                    r1.ContentHashList.Should().BeNull();
+                    r1.Determinism.Should().Be(CacheDeterminism.None);
+
+                    var r2 = database.GetContentHashList(ctx, strongFingerprint2).ShouldBeSuccess().ContentHashListWithDeterminism;
+                    r2.Should().BeEquivalentTo(contentHashListWithDeterminism2);
+                },
+                createStoreFunc: CreateStoreInternal);
+
+            // This is needed because type errors arise if you inline
+            IMemoizationStore CreateStoreInternal(DisposableDirectory disposableDirectory)
```
SergeyTeplyakov: lowerCased: createStoreInternal.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
SergeyTeplyakov: Two things here:

I've missed that metadata is collected all the time, but I'm not sure that doing the GC for both metadata and content in parallel is a good idea.

Is it actually better to do this one at a time maybe?

---
In reply to: [309382147](https://github.com/microsoft/BuildXL/pull/679#discussion_r309382147) [](ancestors = 309382147,309381691)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -285,25 +272,27 @@ public void GarbageCollect(OperationContext context)
...
+                    _gcTimer?.Change(_configuration.GarbageCollectionInterval, Timeout.InfiniteTimeSpan);
                 }
             }
         }
 
         /// <summary>
         /// Collect unreachable entries from the local database.
         /// </summary>
-        private void DoGarbageCollect(OperationContext context)
+        private void DoGarbageCollectContent(OperationContext context)
```
ljcollins25: >DoGarbageCollectContent [](start = 21, length = 23)

Can we add a PerformOperation here for tracing?

ljcollins25:
>         public void GarbageCollect(OperationContext context)

Can we add a PerformOperation here for tracing?

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs:263 in 9d980a1. [](commit_id = 9d980a1c432af3637f820d7e28856531108114e1, deletion_comment = False)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -285,25 +272,27 @@ public void GarbageCollect(OperationContext context)
 
             using (var cancellableContext = TrackShutdown(context))
             {
-                DoGarbageCollect(cancellableContext);
+                var tasks = new List<Task> { Task.Run(() => DoGarbageCollectMetadata(cancellableContext)) };
```
ljcollins25: >DoGarbageCollectMetadata [](start = 60, length = 24)

Can we add a PerformOperation here for tracing?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
```diff
@@ -48,30 +48,35 @@ public abstract class ContentLocationDatabaseConfiguration
...
         ///
-        /// Only useful when <see cref="CacheEnabled"/> is true.
+        /// Only useful when <see cref="ContentCacheEnabled"/> is true.
         /// </summary>
         public TimeSpan CacheFlushingMaximumInterval { get; set; } = TimeSpan.FromMinutes(1);
+
+        /// <summary>
+        /// Minimum amount of time to protect a metadata entry since its last usage time.
+        /// </summary>
+        public TimeSpan MetadataGarbageCollectionProtectionTime { get; set; } = TimeSpan.FromHours(2);
```
ljcollins25: >MetadataGarbageCollectionProtectionTime [](start = 24, length = 39)

Can we add this to the cache config that BuildXL reads for the ICacheFactory's?

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -703,6 +700,64 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
+
+                            scannedEntries++;
+                            Counters[ContentLocationDatabaseCounters.GarbageCollectMetadataEntriesScanned].Increment();
+
+                            var strongFingerprint = DeserializeStrongFingerprint(kvp.Key);
+                            var lastAccessTimeUtc = DeserializeMetadataLastAccessTimeUtc(kvp.Value);
+
+                            if (lastAccessTimeUtc < cutoffTimeUtc)
+                            {
+                                var metadata = DeserializeMetadataEntry(kvp.Value);
```
ljcollins25: >metadata [](start = 36, length = 8)

This doesn't appear to be used.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -703,6 +700,64 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
+                var cutoffTimeUtc = (Clock.UtcNow - _configuration.MetadataGarbageCollectionProtectionTime).ToFileTimeUtc();
+
+                return _keyValueStore.Use(store => {
+                    Tracer.Debug(context, "Local database's garbage collection for metadata is starting");
+
+                    using (var stopwatch = Counters[ContentLocationDatabaseCounters.GarbageCollectMetadata].Start())
+                    {
+                        int removedEntries = 0;
+                        int scannedEntries = 0;
+                        ulong dbSizeInBytesAfterGc = 0;
```
ljcollins25: >dbSizeInBytesAfterGc [](start = 30, length = 20)

Can we track how many bytes were removed as well?

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -703,6 +700,64 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
+                            if (lastAccessTimeUtc < cutoffTimeUtc)
+                            {
+                                var metadata = DeserializeMetadataEntry(kvp.Value);
+                                WipeMetadata(store, kvp.Key, metadata);
+
+                                removedEntries++;
+                                Counters[ContentLocationDatabaseCounters.GarbageCollectMetadataEntriesRemoved].Increment();
+                            } else
+                            {
+                                dbSizeInBytesAfterGc += (ulong)kvp.Key.LongLength + (ulong)kvp.Value.LongLength;
```
ljcollins25: >LongLength [](start = 71, length = 10)

Just use Length.

ljcollins25: >LongLength [](start = 101, length = 10)

Just use Length.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
jbayardo: GC is never disabled after this change. Only Content GC may be disabled, and that doesn't affect Metadata GC

---
In reply to: [309383052](https://github.com/microsoft/BuildXL/pull/679#discussion_r309383052) [](ancestors = 309383052)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
jbayardo: Like you said, as it stands, the timer will always be on, and at most what could happen is that it triggers and does nothing; in fact, this is what will indeed happen when this code is deployed, on all nodes in the system, as the metadata column family should be empty)

Since the SQLiteMemoizationStore runs GC independently of its ContentStore, this is what this logic is mimicking. Indeed, it would be better to have 2 separate timers that can be on or off independently, but it also seems overkill for something that ought to be replaced ASAP.

It doesn't change anything to do this serially right now, because both column families are treated separately, so no -new- inconsistencies can arise. Also, right now the ContentLocationDatabase that stores the locations is different than the one that stores metadata


---
In reply to: [309384685](https://github.com/microsoft/BuildXL/pull/679#discussion_r309384685) [](ancestors = 309384685,309382147,309381691)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
jbayardo: Used to,  but was not needed since we trigger this from a timer. Forgot to remove the comment.

---
In reply to: [309381229](https://github.com/microsoft/BuildXL/pull/679#discussion_r309381229) [](ancestors = 309381229)

PR #681, opened 2019-07-31
Add interactive flag to bxl to and hook up to nuget resolver to allow nuget to show auth UI

dannyvv:
…

### DIFF
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -525,6 +525,9 @@ namespace BuildXL
                             "injectCacheMisses",
                             opt => HandleArtificialCacheMissOption(opt, cacheConfiguration)),
                         OptionHandlerFactory.CreateBoolOption(
+                            "interactive",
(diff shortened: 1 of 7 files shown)


dannyvv:
Note: Once we have an LKG publish I'll update bxl.ps1 to start passing the flag in when not in /lab mode.

Public/Src/App/Bxl/Args.cs
```diff
@@ -524,6 +524,9 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                         OptionHandlerFactory.CreateOption(
                             "injectCacheMisses",
                             opt => HandleArtificialCacheMissOption(opt, cacheConfiguration)),
+                        OptionHandlerFactory.CreateBoolOption(
+                            "interactive",
```
mpysson: >interactive [](start = 29, length = 11)

help text

dannyvv: Thanks, forgot about that :)

PR #682, opened 2019-07-31
Reclassify some drop errors as user or infrastructure

almiliMSFT:
[AB#1546948](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1546948)

### DIFF
--- a/Public/Src/Engine/Scheduler/PipExecutor.cs
+++ b/Public/Src/Engine/Scheduler/PipExecutor.cs
@@ -646,6 +646,15 @@ namespace BuildXL.Scheduler
                         connectionString,
                         ipcResult.Payload);
                 }
+                else if (ipcResult.ExitCode == IpcResultStatus.TransmissionError)
+                {
+                    // we separate the 'invalid input' errors here, so they can be classified as 'user errors'
+                    Logger.Log.PipIpcFailedDueToInfrastructureError(
+                        operationContext,
+                        operation.Payload,
+                        connectionString,
+                        ipcResult.Payload);
+                }
                 else
                 {
                     Logger.Log.PipIpcFailed(
--- a/Public/Src/Engine/Scheduler/PipExecutor.cs
+++ b/Public/Src/Engine/Scheduler/PipExecutor.cs
@@ -648,7 +648,7 @@ namespace BuildXL.Scheduler
                 }
                 else if (ipcResult.ExitCode == IpcResultStatus.TransmissionError)
                 {
-                    // we separate the 'invalid input' errors here, so they can be classified as 'user errors'
+                    // we separate transmission errors here, so they can be properly classified as InfrastructureErrors
                     Logger.Log.PipIpcFailedDueToInfrastructureError(
                         operationContext,
                         operation.Payload,
--- a/Public/Src/Engine/Scheduler/Tracing/Log.cs
+++ b/Public/Src/Engine/Scheduler/Tracing/Log.cs
@@ -155,6 +155,19 @@ namespace BuildXL.Scheduler.Tracing
             string message);
 
         [GeneratedEvent(
+            (ushort)EventId.PipIpcFailedDueToInfrastructureError,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Error,
+            Keywords = (int)(Keywords.UserMessage | Keywords.InfrastructureError),
+            EventTask = (ushort)Tasks.PipExecutor,
+            Message = "IPC operation '{operation}' could not be executed via IPC moniker '{moniker}' because of an infrastructure error. Error: {message}")]
+        internal abstract void PipIpcFailedDueToInfrastructureError(
+            LoggingContext loggingContext,
+            string operation,
+            string moniker,
+            string message);
+
+        [GeneratedEvent(
             (ushort)EventId.PipCopyFileFromUntrackableDir,
             EventGenerators = EventGenerators.LocalOnly,
             EventLevel = Level.Error,
--- a/Public/Src/Engine/Scheduler/Tracing/Log.cs
+++ b/Public/Src/Engine/Scheduler/Tracing/Log.cs
@@ -155,7 +155,7 @@ namespace BuildXL.Scheduler.Tracing
             string message);
 
         [GeneratedEvent(
-            (ushort)EventId.PipIpcFailedDueToInfrastructureError,
+            (ushort)LogEventId.PipIpcFailedDueToInfrastructureError,
             EventGenerators = EventGenerators.LocalOnly,
             EventLevel = Level.Error,
             Keywords = (int)(Keywords.UserMessage | Keywords.InfrastructureError),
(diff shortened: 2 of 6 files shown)


Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -646,6 +646,15 @@ private static PipResultStatus FromPossibleResult(Possible<PipResultStatus> poss
                         connectionString,
                         ipcResult.Payload);
                 }
+                else if (ipcResult.ExitCode == IpcResultStatus.TransmissionError)
+                {
+                    // we separate the 'invalid input' errors here, so they can be classified as 'user errors'
```
almiliMSFT:
```suggestion
                    // we separate transmission errors here, so they can be properly classified as InfrastructureErrors
```

mpysson:
>             Keywords = (int)Keywords.UserMessage,

You need to add the "UserError" keyword here if your new event is the one we're going to use for infrastructure/internal errors. Otherwise these will keep getting marked as internal errors which is the default.

---
Refers to: Public/Src/Engine/Scheduler/Tracing/Log.cs:134 in 6611d9a. [](commit_id = 6611d9a3f53284204085ae4f6926e068f4c93c59, deletion_comment = False)

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -646,6 +646,15 @@ private static PipResultStatus FromPossibleResult(Possible<PipResultStatus> poss
...
+                    // we separate transmission errors here, so they can be properly classified as InfrastructureErrors
+                    Logger.Log.PipIpcFailedDueToInfrastructureError(
+                        operationContext,
+                        operation.Payload,
+                        connectionString,
+                        ipcResult.Payload);
+                }
                 else
                 {
                     Logger.Log.PipIpcFailed(
```
mpysson: >PipIpcFailed [](start = 31, length = 12)

So is this catchall still an internal error? I was expecting to see you add more user error patterns to account for the various user errors we had previously been misclassifying as internal errors.

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -154,6 +154,19 @@ internal static string PipWriteFileFailedMessage(string pipDescription, string p
...
             string message);
 
+        [GeneratedEvent(
+            (ushort)EventId.PipIpcFailedDueToInfrastructureError,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Error,
+            Keywords = (int)(Keywords.UserMessage | Keywords.InfrastructureError),
+            EventTask = (ushort)Tasks.PipExecutor,
+            Message = "IPC operation '{operation}' could not be executed via IPC moniker '{moniker}' because of an infrastructure error. Error: {message}")]
+        internal abstract void PipIpcFailedDueToInfrastructureError(
```
almiliMSFT: > You need to add the "UserError" keyword here if your new event is the one we're going to use for infrastructure/internal errors. Otherwise these will keep getting marked as internal errors which is the default.

The idea was the following:
  - `PipIpcFailed` continues to be classified as internal error
  - the existing `PipIpcFailedDueToInvalidInput` is used for IPC user errors
  - this new `PipIpcFailedDueToInfrastructureError` is used for IPC infra errors

Public/Src/Utilities/Utilities/Tracing/EventId.cs
```diff
@@ -312,7 +312,8 @@ public enum EventId
         PipRetryDueToExitedWithAzureWatsonExitCode = 317,
         // Reserved = 318,
 
-        // Free slot 319,
+        PipIpcFailedDueToInfrastructureError = 319,
```
dannyvv: >PipIpcFailedDueToInfrastructureError [](start = 8, length = 36)

Please add this to the Tracing\EventIds.cs in the Scheduler project, not here.

Public/Src/Engine/Scheduler/PipExecutor.cs
mpysson: Nevermind. I see you changed the classification in DropDaemon.cs to route more to InvalidInput

Public/Src/Utilities/Utilities/Tracing/EventId.cs
almiliMSFT: done

Public/Src/Utilities/Utilities/Tracing/EventId.cs
```diff
@@ -313,6 +313,7 @@ public enum EventId
         // Reserved = 318,
 
         // Free slot 319,
+
```
almiliMSFT:
```suggestion
```

Public/Src/Tools/DropDaemon/DropDaemon.cs
```diff
@@ -567,43 +569,74 @@ private async Task ReportStatisticsAsync()
             }
         }
 
+
```
almiliMSFT:
```suggestion
```

Public/Src/Tools/DropDaemon/DropDaemon.cs
```diff
@@ -567,43 +569,73 @@ private async Task ReportStatisticsAsync()
...
 
-        private static async Task<TResult> HandleKnownErrors<TResult>(Func<Task<TResult>> factory, Func<string, TResult> errorValueFactory)
+        private static string RenderIpcStatus(IpcResultStatus status)
+        {
+            return status.ToString();
+        }
+
+        private static IpcResultStatus ParseIpcStatus(string statusString, IpcResultStatus defaultValue = IpcResultStatus.ExecutionError)
+        {
+            return Enum
```
olkononenko: >return Enum [](start = 12, length = 11)

I think calling Enum.TryParse() would look a little bit cleaner. Is there a benefit of doing the parsing/casting this way?

Public/Src/Tools/DropDaemon/DropDaemon.cs
```diff
@@ -567,43 +569,73 @@ private async Task ReportStatisticsAsync()
...
+                .GetValues(typeof(IpcResultStatus))
+                .Cast<IpcResultStatus?>()
+                .Where(status => status.ToString() == statusString)
+                .FirstOrDefault() ?? defaultValue;
```
olkononenko: >?? defaultValue [](start = 34, length = 15)

Should we add a new IpcResultStatus value that would be used specifically for parse failures? I'm afraid that we might be 'hiding' parsing problems here, no?

Public/Src/Tools/DropDaemon/DropDaemon.cs
almiliMSFT: you are right, will change

Public/Src/Tools/DropDaemon/DropDaemon.cs
almiliMSFT: I changed it to use `TryParse`.  I am pretty confident that `ToString`/`TryParse` are the exact inverses of each other in this case, so I'd leave it as is

PR #685, opened 2019-07-31
Create API endpoints to grab event data from RocksDB

sahilmgandhi:
[AB#1578604](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1578604)

Added more api endpoints for grabbing the event data that is stored into RocksDB. I was wondering if I should make this "BXLInvocationAnalyzer" more generic, or if it should be removed entirely?


Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -54,7 +54,7 @@ public sealed class XldbDataStore : IDisposable
         /// Gets all the events of a certain type from the DB
         /// </summary>
         /// <returns>List of event objects recovered from DB </returns>
-        public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
+        private IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
```
sahilmgandhi: I was debating back and forth whether to make this public or private since with the help of the public methods below, a user should never have to call this "GetEventsByType" method (thus private is better), but if they wanted to then the method should be public. What are your thoughts?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/BXLInvocationAnalyzer.cs
```diff
@@ -78,10 +85,16 @@ internal sealed class BXLInvocationAnalyzer : Analyzer
         public override int Analyze()
         {
             var dataStore = new XldbDataStore(storeDirectory: InputDirPath);
-
-            File.WriteAllLines(OutputFilePath, dataStore.GetEventsByType(Xldb.ExecutionEventId.DominoInvocation));
+            File.AppendAllLines(OutputFilePath, dataStore.GetBXLInvocationEvents());
```
dannyvv: Just checking if this is intentional , or just left from debugging.... This will be an ever growing file over time not?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
dannyvv: Private is fine to reduce support surface, if users have a valid use case we can always make it public in the future.

Taking it away from public is always harder since it is a breaking change.

---
In reply to: [309472596](https://github.com/microsoft/BuildXL/pull/685#discussion_r309472596) [](ancestors = 309472596)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+                                storedEvents.Add(StatusReportedEvent.Parser.ParseFrom(kvp.Value).ToString());
+                                break;
+                            case ExecutionEventId.DominoInvocation:
+                                storedEvents.Add(BXLInvocationEvent.Parser.ParseFrom(kvp.Value).ToString());
+                                break;
+                            case ExecutionEventId.PipExecutionDirectoryOutputs:
+                                storedEvents.Add(PipExecutionDirectoryOutputsEvent.Parser.ParseFrom(kvp.Value).ToString());
+                                break;
+                            default:
+                                break;
```
dannyvv: Just for your consideration:
Usually we have an assert if we reach a default case to ensure we catch someone adding a new type and missing it.
Now here we are at an interesting point since we want to have a compatible version / layer :) So failing here goes directly against that :)

Perhaps it is worth to at a minimum to count these and include them in telemetry, or at least log it happened?

I'll leave it for your consideration. No need to hold off the PR for this though.

dannyvv:
>             var storedEvents = new List<string>();

This is okay, just wanted to share that C# has a cool features for returning enumerables. Where the compiler can generate iterator code to to avoid building a list over and over.

See [https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/iterators](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/iterators) for `yield return ...`


---
Refers to: Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs:61 in fa7d61c. [](commit_id = fa7d61caf88fd05637ae133435741abde8a56b6b, deletion_comment = False)

dannyvv:
>             var storedEvents = new List<string>();

This will change the behavior from eagerly looping over the prefixsearch results vs lazy as users iterat over the results of GetEventsByType.

---
In reply to: [517063648](https://github.com/microsoft/BuildXL/pull/685#issuecomment-517063648) [](ancestors = 517063648)

---
Refers to: Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs:61 in fa7d61c. [](commit_id = fa7d61caf88fd05637ae133435741abde8a56b6b, deletion_comment = False)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+                                storedEvents.Add(DependencyViolationReportedEvent.Parser.ParseFrom(kvp.Value).ToString());
+                                break;
+                            case ExecutionEventId.PipExecutionStepPerformanceReported:
+                                storedEvents.Add(PipExecutionStepPerformanceReportedEvent.Parser.ParseFrom(kvp.Value).ToString());
+                                break;
+                            case ExecutionEventId.PipCacheMiss:
+                                storedEvents.Add(PipCacheMissEvent.Parser.ParseFrom(kvp.Value).ToString());
+                                break;
+                            case ExecutionEventId.ResourceUsageReported:
+                                storedEvents.Add(StatusReportedEvent.Parser.ParseFrom(kvp.Value).ToString());
```
dannyvv: The names don't fully line up, intentional?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+                                break;
+                            case ExecutionEventId.DominoInvocation:
+                                storedEvents.Add(BXLInvocationEvent.Parser.ParseFrom(kvp.Value).ToString());
```
dannyvv: So sorry to ask this, but would you mind using `BuildXLInvocationEvent`

We should use BuildXL everywehre in our naming. bxl is just for our executable name.

dannyvv: note: future pr is totally fine.

---
In reply to: [309476346](https://github.com/microsoft/BuildXL/pull/685#discussion_r309476346) [](ancestors = 309476346)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+        /// Gets all the Pip Cache Miss Events
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<string> GetPipCacheMissEvents() => GetEventsByType(ExecutionEventId.PipCacheMiss);
+
+        /// <summary>
+        /// Gets all the BXL Invocation Events
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<string> GetBXLInvocationEvents() => GetEventsByType(ExecutionEventId.DominoInvocation);
```
dannyvv: >BXL [](start = 38, length = 3)

Like above BuildXL.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+        /// Gets all the Pip Execution Step Performance Reported Events
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<string> GetPipExecutionStepPerformanceReportedEvents() => GetEventsByType(ExecutionEventId.PipExecutionStepPerformanceReported);
+
+        /// <summary>
+        /// Gets all the Status Reported Events
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<string> GetStatusReportedEvents() => GetEventsByType(ExecutionEventId.ResourceUsageReported);
```
dannyvv: >ts() => GetEventsByType(ExecutionEventId.Resourc [](start = 56, length = 48)

name mismatch as well.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: Sure sounds good, I'll keep it private.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: So this naming mismatch is due to the BXL (in the corresponding enum) being DominoInvocation. I can change it, but I left it there for similarity purposes

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: The names didn't line up in our BXL code as well for that enum, and I kept the same names for the protobuf enums. It is a bit confusing but I'm assuming we can't change it in the BXL codebase due to telemetry related reasons (and so I'm not sure if I should change it here either).

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/BXLInvocationAnalyzer.cs
sahilmgandhi: This is intentional, and the file should not be growing over time as you should only be calling this "analyzer" once.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: As discussed offline, the mismatch will be kept to mirror what we have in the BXL codebase

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: Changing it in the next commit.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
                 })
             );
 
             return storedEvents;
         }
 
+        /// <summary>
+        /// Gets all the File Artifact Content Decided Events
+        /// </summary>
+        /// <returns></returns>
```
almiliMSFT:
```suggestion
```

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+        public IEnumerable<string> GetFileArtifactContentDecidedEvents() => GetEventsByType(ExecutionEventId.FileArtifactContentDecided);
+
+        /// <summary>
+        /// Gets all the Worker List Events
+        /// </summary>
+        /// <returns></returns>
```
almiliMSFT:
```suggestion
```

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+        public IEnumerable<string> GetWorkerListEvents() => GetEventsByType(ExecutionEventId.WorkerList);
+
+        /// <summary>
+        /// Gets all the Pip Execution Performance Events
+        /// </summary>
+        /// <returns></returns>
```
almiliMSFT:
```suggestion
```

...and so on.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
almiliMSFT: @dannyvv: are you 100% sure it's safe to rename this event?  I vaguely remember a comment somewhere in the code saying that it could not be renamed because of a Kusto table that others depend on (or something along this line).   I'll try to find this comment.

almiliMSFT: This code in [`IExecutionLogTarget`](https://github.com/microsoft/BuildXL/blob/e72fa8841719728183443808432059a92ac0397f/Public/Src/Engine/Scheduler/Tracing/ExecutionLog.Events.cs#L108-L109) is what I meant:
```csharp
        // $Rename: Due to telemetry backend scripts this cannot be renamed to BuildXL
        void DominoInvocation(DominoInvocationEventData data);
```

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
                 {
                     foreach (var kvp in database.PrefixSearch(eventQuery.ToByteArray()))
                     {
-                        storedEvents.Add(BXLInvocationEvent.Parser.ParseFrom(kvp.Value).ToString());
+                        switch (eventTypeID)
```
almiliMSFT: There is too much repetition going on here.  Can we replace this whole `switch` with something like:
```csharp
if (m_eventParsersDictionary.TryGetValue(eventTypeID, out var parser)
{
    sotredEvents.Add(parser.ParseFrom(kvp.Value).ToString());
}  
```
Then you just need to initialize this `m_eventParsersDictionary` once in the constructor or wherever.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -68,14 +68,133 @@ public IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
             );
 
             return storedEvents;
         }
 
+        /// <summary>
+        /// Gets all the File Artifact Content Decided Events
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<string> GetFileArtifactContentDecidedEvents() => GetEventsByType(ExecutionEventId.FileArtifactContentDecided);
```
almiliMSFT: Do you really need all these helper properties?  I'd guess that just exposing `GetEventsByType` is enough.

sahilmgandhi: Initial plan was to just expose GetEventsByType and not have any of these other methods, but then we later thought that these specific endpoints would allow users more easily get the information they were looking for (instead of trying to fiddle around with the right enum values to pass into GetEventsByType)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: Good idea, ill go ahead and implement that.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/BXLInvocationAnalyzer.cs
```diff
@@ -78,10 +85,16 @@ internal sealed class BXLInvocationAnalyzer : Analyzer
...
-            File.WriteAllLines(OutputFilePath, dataStore.GetEventsByType(Xldb.ExecutionEventId.DominoInvocation));
+            File.AppendAllLines(OutputFilePath, dataStore.GetBXLInvocationEvents());
 
             return 0;
         }
+
+        /// <inheritdoc/>
+        protected override bool ReadEvents()
+        {
+            // Do nothing. This analyzer does not read events.
```
olkononenko: >/ Do nothing. [](start = 13, length = 13)

so why do you need this method then?

sahilmgandhi: Because otherwise, it would invoke "Input.ReadExecutionLog" which we do not want to do anymore.

---
In reply to: [309831355](https://github.com/microsoft/BuildXL/pull/685#discussion_r309831355) [](ancestors = 309831355)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
olkononenko: While we cannot rename it in the BXL codebase, I think we can rename it in the Xldb codebase (so users of the lib won't see 'domino'). We might need to ad a comment in 'write db analyzer' explaining why we are using a different name.

---
In reply to: [309813123](https://github.com/microsoft/BuildXL/pull/685#discussion_r309813123) [](ancestors = 309813123)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
olkononenko: I think it's valid point. I also think that we need to differentiate between "I looked for X and did not find it" and "I looked for X, found it, but I don't know who to deserialize it"

---
In reply to: [309475293](https://github.com/microsoft/BuildXL/pull/685#discussion_r309475293) [](ancestors = 309475293)

PR #687, opened 2019-08-01
Fix machineKbitsPerSecSent/Received overflow.

olkononenko:
[AB#1575551](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1575551)

### DIFF
--- a/Public/Src/Utilities/Utilities/PerformanceCollector.cs
+++ b/Public/Src/Utilities/Utilities/PerformanceCollector.cs
@@ -233,7 +233,7 @@ namespace BuildXL.Utilities
         private static double BytesToKbits(long bytes)
         {
             // Convert to Kbits
-            return bytes / 1024.0 * 8;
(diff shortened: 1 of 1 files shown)


Public/Src/Utilities/Utilities/PerformanceCollector.cs
```diff
@@ -224,15 +230,10 @@ private void CollectOnce()
...
         {
-            if (bytes.HasValue)
-            {
-                // Convert to Kbits
-                return 8 * bytes.Value / 1024;
-            }
-
-            return (double)bytes;
+            // Convert to Kbits
+            return bytes / 1024.0 * 8;
```
narasamdya: >bytes / 1024.0 * 8; [](start = 19, length = 19)

For clarity, can you add parenthesis.

PR #691, opened 2019-08-02
Custom encoding of build context data in session id guid bytes for better searching in Kusto.

ljcollins25:


Public/Src/App/Bxl/BuildXLApp.cs
```diff
@@ -979,6 +981,47 @@ private AppResult RunWithLoggingScope(Func<PerformanceMeasurement, AppResult> ru
...
+
+            // 06-07: InCloudBuild = FFFF, !InCloudBuild = 0000
+            var inCloudBuildSpecifier = inCloudBuild ? byte.MaxValue : (byte)0;
+            bytes[6] = inCloudBuildSpecifier;
+            bytes[7] = inCloudBuildSpecifier;
+
+            // 08-09: executionEnvironment
+            bytes[8] = (byte)(((int)executionEnvironment >> 8) & 0xFF);
+            bytes[9] = (byte)((int)executionEnvironment & 0xFF);
+
```
mpysson: double checking low legit it is to steal this many bits:
[https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions) 

Updating that from using 122 bits down to 98 yields  
1/2 + (1/4 + 2 + ln(2) * 2^98)^.5 = 4x10^14 guids needing to be generated for probability of collision being 50%. So it seems fine to me at our scale

Public/Src/Utilities/Configuration/ExecutionEnvironment.cs
```diff
@@ -19,99 +22,134 @@ public enum ExecutionEnvironment
...
+        ProductBuild = 0x0030,
+        Tests        = 0x0040,
+        WrapItUp     = 0x0050,
+        Tools        = 0x0060,
+        NightlyPerf  = 0x0070,
+
+        // Location 0x___X
+        NoLocation = 0x0000,
+        Lab        = 0x0007,
+        PrimeLab   = 0x0009,
```
mpysson: >     PrimeLab   = 0x0009, [](start = 3, length = 25)

No need to differentiate this. It doesn't appear that anyone's passing the OsgPrimeLab flag anymore. Don't delete the enum for sake of not breaking back compat, but i think we can keep the locations simple

PR #692, opened 2019-08-02
Log an error as soon as ApiServer.MaterializeFile fails

almiliMSFT:
When a file fails to be materialized in response to such a request from an IPC pip, the IPC pip ends up failing with an internal error, and the root cause of this failure (i.e., the "ErrorBucket") becomes "IpcPipFailed".

With this PR, an error message is logged as soon as a file fails to be materialized, so the ErrorBucket will end up pointing to the real root cause of this particular problem.

### DIFF
--- a/Public/Src/Engine/Scheduler/ApiServer.cs
+++ b/Public/Src/Engine/Scheduler/ApiServer.cs
@@ -161,7 +161,7 @@ namespace BuildXL.Scheduler
             string absoluteFilePath = cmd.File.Path.ToString(m_context.PathTable);
 
             // if file materialization failed, log an error here immediately, so that this errors gets picked up as the root cause 
-            // (which is evetually used to compute the "ErrorBucket") instead of whatever fallout ends up happening (e.g., IPC pip fails)
+            // (i.e., the "ErrorBucket") instead of whatever fallout ends up happening (e.g., IPC pip fails)
             if (!succeeded)
             {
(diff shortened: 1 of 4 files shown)


Public/Src/Engine/Scheduler/ApiServer.cs
```diff
@@ -156,8 +156,21 @@ private async Task<IIpcResult> ExecuteMaterializeFile(MaterializeFileCommand cmd
...
             }
 
-            bool succeeded = await m_fileContentManager.TryMaterializeFile(cmd.File);
-            Tracing.Logger.Log.ApiServerMaterializeFileExecuted(m_loggingContext, cmd.File.Path.ToString(m_context.PathTable), succeeded);
+            var result = await m_fileContentManager.TryMaterializeFileAsync(cmd.File);
+            bool succeeded = result == ArtifactMaterializationResult.Succeeded;
+            string absoluteFilePath = cmd.File.Path.ToString(m_context.PathTable);
+
+            // if file materialization failed, log an error here immediately, so that this errors gets picked up as the root cause 
+            // (which is evetually used to compute the "ErrorBucket") instead of whatever fallout ends up happening (e.g., IPC pip fails)
```
almiliMSFT:
```suggestion
            // (i.e., the "ErrorBucket") instead of whatever fallout ends up happening (e.g., IPC pip fails)
```

PR #693, opened 2019-08-02
Use public Azure DevOps feed for Lkg in favor of our old myget feed

dannyvv:


sahilmgandhi:
Was there a reason we were using this other feed instead of the public azure one?

dannyvv:
> Was there a reason we were using this other feed instead of the public azure one?
Public feeds went in public preview on July 19th: 
https://twitter.com/ElijahBatkoski/status/1152281496029454343
So we couldn't use them in our public repo yet :)

PR #694, opened 2019-08-02
Add dbeventcount verification unit test and fix proto bug

sahilmgandhi:
[AB#1574154](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1574154) 

Creates a unit test to check that the count of events (per event type) that we log is as expected per the data in the .xlg file. Also fixed a bug in our protobuf key since default values are not serialized in protobuf (by design) so if we had a key of value 0, it would not be serialized (and prefixmatch for this would then just match everything). As per @dannyvv we made the default enum value (0) to be invalid for "ExecutionEventId".

### DIFF
--- a/Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/Events.proto
+++ b/Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/Events.proto
@@ -43,23 +43,28 @@ message EventTypeQuery{
 // ie: ChooseWorkerCpu ---> PipExecutionStep_ChooseWorkerCpu
 
 enum ExecutionEventId{
-    FileArtifactContentDecided = 0;
-    WorkerList = 1;
-    ExecutionEventId_PipExecutionPerformance = 2;
-    DirectoryMembershipHashed = 3;
+    // Since we are using ExecutionEventId as in the key and protobuf 
+    // does not serialize default values, the pattern matching was bugging out. 
+    // A solution is to have an "invalid" enum constant take position 0. 
+    ExecutionEventId_Invalid = 0;
+
+    FileArtifactContentDecided = 1;
+    WorkerList = 2;
+    ExecutionEventId_PipExecutionPerformance = 3;
+    DirectoryMembershipHashed = 4;
 
     // Deprecated in favor of ProcessFingerprintComputation (enum 10)
-    ObservedInputs = 4; 
+    ObservedInputs = 5; 
 
-    ProcessExecutionMonitoringReported = 5;
-    ExecutionEventId_ExtraEventDataReported = 6;
-    DependencyViolationReported = 7;
-    PipExecutionStepPerformanceReported = 8;
-    ResourceUsageReported = 9;
-    ProcessFingerprintComputation = 10;
-    PipCacheMiss = 11;
-    PipExecutionDirectoryOutputs = 12;
-    BxlInvocation = 13;
+    ProcessExecutionMonitoringReported = 6;
+    ExecutionEventId_ExtraEventDataReported = 7;
+    DependencyViolationReported = 8;
+    PipExecutionStepPerformanceReported = 9;
+    ResourceUsageReported = 10;
+    ProcessFingerprintComputation = 11;
+    PipCacheMiss = 12;
+    PipExecutionDirectoryOutputs = 13;
+    BxlInvocation = 14;
 }
 
 enum PipOutputOrigin{
--- a/Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
+++ b/Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
+using System.Text;
 using BuildXL.Engine.Cache.KeyValueStores;
 using BuildXL.Execution.Analyzer.Xldb;
 using BuildXL.Utilities;
@@ -18,6 +19,7 @@ namespace BuildXL.Analyzers.Core.XLGPlusPlus
         /// </summary>
         private KeyValueStoreAccessor Accessor { get; set; }
         private Dictionary<ExecutionEventId, MessageParser> m_eventParserDictionary = new Dictionary<ExecutionEventId, MessageParser>();
+        public static string EventCountKey = "EventCount";
 
         /// <summary>
         /// Open the datastore and populate the KeyValueStoreAccessor for the XLG++ DB
@@ -125,6 +127,11 @@ namespace BuildXL.Analyzers.Core.XLGPlusPlus
         public IEnumerable<string> GetProcessExecutionMonitoringReportedEvents() => GetEventsByType(ExecutionEventId.ProcessExecutionMonitoringReported);
 
         /// <summary>
+        /// Gets all the Process Execution Monitoring Reported Events
+        /// </summary>
+        public IEnumerable<string> GetProcessFingerprintComputationEvents() => GetEventsByType(ExecutionEventId.ProcessFingerprintComputation);
+
+        /// <summary>
         /// Gets all the Extra Event Data Reported Events
         /// </summary>
         public IEnumerable<string> GetExtraEventDataReportedEvents() => GetEventsByType(ExecutionEventId.ExtraEventDataReported);
@@ -158,6 +165,24 @@ namespace BuildXL.Analyzers.Core.XLGPlusPlus
         /// Gets all the Pip Execution Directory Outputs Events
         /// </summary>
         public IEnumerable<string> GetPipExecutionDirectoryOutputsEvents() => GetEventsByType(ExecutionEventId.PipExecutionDirectoryOutputs);
+
+        /// <summary>
+        /// Gets the total number of stored xlg events in the database
+        /// </summary>
+        public uint GetEventCount()
+        {
+            uint eventCount = 0;
+
+            Analysis.IgnoreResult(
+                Accessor.Use(database =>
+                {
+                    database.TryGetValue(Encoding.ASCII.GetBytes(EventCountKey), out var eventCountObj);
+                    eventCount = EventCount.Parser.ParseFrom(eventCountObj).Value;
+                })
+            );
+
+            return eventCount;
+        }
 
         /// <summary>
         /// Closes the connection to the DB
--- a/Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
+++ b/Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
@@ -80,23 +80,26 @@ namespace BuildXL.Analyzers.Core.XLGPlusPlus
                 EventTypeID = eventTypeID,
             };
 
-            Analysis.IgnoreResult(
-                Accessor.Use(database =>
+            var maybeFound = Accessor.Use(database =>
+            {
+                foreach (var kvp in database.PrefixSearch(eventQuery.ToByteArray()))
                 {
-                    foreach (var kvp in database.PrefixSearch(eventQuery.ToByteArray()))
+                    if (m_eventParserDictionary.TryGetValue(eventTypeID, out var parser))
                     {
-                        if (m_eventParserDictionary.TryGetValue(eventTypeID, out var parser))
-                        {
-                            storedEvents.Add(parser.ParseFrom(kvp.Value).ToString());
-                        }
-                        else
-                        {
-                            // We will never reach here since this is a private method and we explicitly control which ExecutionEventIDs are passed in (ie. the public facing helper methods below)
-                            _ = Contract.AssertFailure("Invalid Execution Event ID passed in. Exiting");
-                        }
+                        storedEvents.Add(parser.ParseFrom(kvp.Value).ToString());
                     }
-                })
-            );
+                    else
+                    {
+                        // We will never reach here since this is a private method and we explicitly control which ExecutionEventIDs are passed in (ie. the public facing helper methods below)
+                        _ = Contract.AssertFailure("Invalid Execution Event ID passed in. Exiting");
+                    }
+                }
+            });
+
+            if (!maybeFound.Succeeded)
+            {
+                maybeFound.Failure.Throw();
+            }
 
             return storedEvents;
         }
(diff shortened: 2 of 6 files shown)


Public/Src/Tools/UnitTests/Analyzers/XLGToDBAnalyzerTests.cs
```diff
@@ -69,5 +74,62 @@ public void DBCreatedAndPopulated()
...
+            var buildA = RunScheduler().AssertSuccess();
+
+            var analyzerRes = RunAnalyzer(buildA).AssertSuccess();
+
+            var dataStore = new XldbDataStore(storeDirectory: OutputDirPath.ToString(Context.PathTable));
+
+            XAssert.AreEqual(dataStore.GetFileArtifactContentDecidedEvents().Count(), 5);
+            XAssert.AreEqual(dataStore.GetWorkerListEvents().Count(), 0);
+            XAssert.AreEqual(dataStore.GetPipExecutionPerformanceEvents().Count(), 4);
+            XAssert.AreEqual(dataStore.GetDirectoryMembershipHashedEvents().Count(), 0);
```
mpysson: >XAssert.AreEqual(dataStore.GetDirectoryMembershipHashedEvents().Count(), 0); [](start = 12, length = 76)

I'm surprised this event isn't logged since pipD enumerates a directory. This should cause BuildXL to need to compute a directory fingerprint when storing the pip to cache after it executes

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/Events.proto
```diff
@@ -43,23 +43,28 @@ message EventTypeQuery{
 // ie: ChooseWorkerCpu ---> PipExecutionStep_ChooseWorkerCpu
 
 enum ExecutionEventId{
```
olkononenko: > ExecutionEventId{ [](start = 4, length = 18)

why do we need to this 'invalid' member? if it's a protobuf requirement, should it be added to all the enums? Also, I think we might want to have a helper function, so we do not forget that we need to +1 when converting from xlg -> proto

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -18,6 +19,7 @@ public sealed class XldbDataStore : IDisposable
         /// </summary>
         private KeyValueStoreAccessor Accessor { get; set; }
         private Dictionary<ExecutionEventId, MessageParser> m_eventParserDictionary = new Dictionary<ExecutionEventId, MessageParser>();
+        public static string EventCountKey = "EventCount";
```
olkononenko: >static [](start = 15, length = 6)

const instead of a static? you are not writing into that field, right?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
 using System.Linq;
```
olkononenko: nit: remove, it looks like you don't need it

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/Events.proto
sahilmgandhi: It's not a protobuf requirement, its just since if our key contains {eventID = 0}, protobuf will serialize that as {} since it discards default values (0 is default value for int). This will cause problems for searching for the key. 

Also at no point do we convert xlg -> proto for this enum (the only place it is used is the key and where is no conversion being done here).

---
In reply to: [310239009](https://github.com/microsoft/BuildXL/pull/694#discussion_r310239009) [](ancestors = 310239009)

Public/Src/Tools/UnitTests/Analyzers/XLGToDBAnalyzerTests.cs
olkononenko: yeah, this is puzzling... @Sahil can you add some dummy source (existing prior build starts) files under that directory? Maybe there is some weird optimization is going on, and we report nothing for an empty dir?

---
In reply to: [310238006](https://github.com/microsoft/BuildXL/pull/694#discussion_r310238006) [](ancestors = 310238006)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: yea was not going to write into it. Would const or static readonly be better?

---
In reply to: [310239344](https://github.com/microsoft/BuildXL/pull/694#discussion_r310239344) [](ancestors = 310239344)

Public/Src/Tools/UnitTests/Analyzers/XLGToDBAnalyzerTests.cs
olkononenko: Other thoughts (no need to investigate/address in this PR) - I bet `Operation.EnumerateDir` calls into some .Net API for enumeration, and we saw in the past that in some really rare cases Detours had troubles with catching the accesses (I remember switching some operation(s) from calling into .net to invoking winapi directly)

---
In reply to: [310242591](https://github.com/microsoft/BuildXL/pull/694#discussion_r310242591) [](ancestors = 310242591,310238006)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
olkononenko: I'd use const

---
In reply to: [310243113](https://github.com/microsoft/BuildXL/pull/694#discussion_r310243113) [](ancestors = 310243113,310239344)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/Events.proto
olkononenko: Hmmm... I might be missing something here. But don't we need to have "Invalid = 0" on all of our enums?

---
In reply to: [310241033](https://github.com/microsoft/BuildXL/pull/694#discussion_r310241033) [](ancestors = 310241033,310239009)

sahilmgandhi: Yes, but if its not a key then it won't make a difference. The defacto standard in protobuf is that a missing or non-serialized field is the default value, so the consumer of the protoobject when seeing {} should handle it as being 0.

---
In reply to: [310244883](https://github.com/microsoft/BuildXL/pull/694#discussion_r310244883) [](ancestors = 310244883,310241033,310239009)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
almiliMSFT: yes, either const or `readonly` static

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -159,6 +166,24 @@ private IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
         public IEnumerable<string> GetPipExecutionDirectoryOutputsEvents() => GetEventsByType(ExecutionEventId.PipExecutionDirectoryOutputs);
 
+        /// <summary>
+        /// Gets the total number of stored xlg events in the database
+        /// </summary>
+        public uint GetEventCount()
+        {
+            uint eventCount = 0;
+
+            Analysis.IgnoreResult(
```
almiliMSFT: Can't you replace this whole method with
```csharp
    return Accessor.Use(database =>
	                {
	                    database.TryGetValue(Encoding.ASCII.GetBytes(EventCountKey), out var eventCountObj);
                        return EventCount.Parser.ParseFrom(eventCountObj).Value;
	                })
```

Public/Src/Tools/UnitTests/Analyzers/XLGToDBAnalyzerTests.cs
```diff
@@ -69,5 +74,62 @@ public void DBCreatedAndPopulated()
...
+        public void DBEventCountVerification()
+        {
+            Configuration.Logging.LogExecution = true;
+
+            var fileOne = FileArtifact.CreateOutputFile(Combine(TestDirPath, "foo.txt"));
+            var fileTwo = FileArtifact.CreateOutputFile(Combine(TestDirPath, "bar.txt"));
+            var dirOne = CreateOutputDirectoryArtifact(Combine(TestDirPath, "baz").ToString(Context.PathTable));
+            Directory.CreateDirectory(ArtifactToString(dirOne));
+
+            var pipA = CreateAndSchedulePipBuilder(new Operation[]
```
almiliMSFT: Just FYI, `new Operation[]` can be shortened to `new[]`

Public/Src/Tools/UnitTests/Analyzers/XLGToDBAnalyzerTests.cs
```diff
@@ -69,5 +74,62 @@ public void DBCreatedAndPopulated()
...
+                Operation.WriteFile(CreateOutputFileArtifact())
+            }).Process;
+
+            var buildA = RunScheduler().AssertSuccess();
+
+            var analyzerRes = RunAnalyzer(buildA).AssertSuccess();
+
+            var dataStore = new XldbDataStore(storeDirectory: OutputDirPath.ToString(Context.PathTable));
+
+            XAssert.AreEqual(dataStore.GetFileArtifactContentDecidedEvents().Count(), 5);
```
almiliMSFT: Why is 5 expected here?  I see only 4 output files in this test.

More generally speaking, instead of using a constant number here, I'd like to use some expression instead.  For example, after calling `RunScheduler()`, pip graph is created and saved in `SchedulerIntegrationTestBase.m_lastGraph` field.  That field is currently private, but you could add a protected getter property for it, and then here instead of `5` say something like `LastGraph.AllFiles.Count()` (or whatever the logic should be)

Public/Src/Tools/UnitTests/Analyzers/XLGToDBAnalyzerTests.cs
```diff
@@ -69,5 +74,62 @@ public void DBCreatedAndPopulated()
...
+            XAssert.AreEqual(dataStore.GetWorkerListEvents().Count(), 0);
+            XAssert.AreEqual(dataStore.GetPipExecutionPerformanceEvents().Count(), 4);
```
almiliMSFT: If you take my advice, here instead of `4` you could for sure say `LastGraph.RetrievePipReferencesOfType(PipType.Process).Count()`.  That would explicitly make it clear that we expect to get one PipExecutionPerformanceEvent per process pip.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: The Accessor.Use method returns a possible int value, so I'm not sure if we can do a simplification like this.

almiliMSFT: in that case:
```csharp
var maybeValue = Accessor.Use(...);
if (!maybeValue.Succeeded) maybeValue.Failure.Throw(); // or something like that
return maybeValue.Result;
```

sahilmgandhi: In other uses of the rocksDb store by say FingerprintStore.cs or the cache team seem to be doing it the same way as me (using the IgnoreResult, and having a var outside that is assigned to within the .Use( ... ) function. I think I should also follow this pattern then right?

almiliMSFT: You are right that in other places this same pattern is used; for example, I found this in FingerprintStore.cs
```csharp
            var keyFound = false;
            Analysis.IgnoreResult(
                Accessor.Use(store =>
                {
                    keyFound = store.Contains(key, columnFamilyName);
                })
            );
            return keyFound;
```
That, however, doesn't necessarily make it right.  At face value, it looks ridiculous ignore explicitly returned result and instead from within a closure mutate some local var defined outside of the closure.  But there is more.  `Accessor.Use` can fail, in which case it returns a `Failure` object; when that happens, the failure is completely ignored here and instead the initial value assigned to `keyFound` (which is `false`) is returned.  Maybe the intention here is to exactly do that, i.e., if an error happens just ignore it and return the default value of `false`.  If so, this intention is completely obscured by this pattern; instead, if this code were written as
```csharp
var maybeFound = Accessor.Use(store => store.Contains(key, columnFamilyName));
return maybeFound.Succeeded 
    ? maybeFound.Result
    : false; // ignore the failure and return false
```
the intention would be 100% clear.

Code like this is pretty much the definition of spaghetti code.

Public/Src/Tools/UnitTests/Analyzers/XLGToDBAnalyzerTests.cs
```diff
@@ -69,5 +74,64 @@ public void DBCreatedAndPopulated()
...
+
+            var buildA = RunScheduler().AssertSuccess();
+
+            var analyzerRes = RunAnalyzer(buildA).AssertSuccess();
+
+            var dataStore = new XldbDataStore(storeDirectory: OutputDirPath.ToString(Context.PathTable));
+
+            XAssert.AreEqual(dataStore.GetFileArtifactContentDecidedEvents().Count(), LastGraph.AllFiles.Count());
+            XAssert.AreEqual(dataStore.GetWorkerListEvents().Count(), 0);
+            XAssert.AreEqual(dataStore.GetPipExecutionPerformanceEvents().Count(), LastGraph.RetrievePipReferencesOfType(PipType.Process).Count());
```
almiliMSFT: can you extract `LastGraph.RetrievePipReferencesOfType(PipType.Process).Count());` into a local variable (e.g., `processPipCount` and use that variable here and elsewhere?  You can then use that value for `GetPipCacheMissEvents` as well.

PR #696, opened 2019-08-02
Reduce Sqlite nuget dependencies

dannyvv:


config.dsc
```diff
@@ -121,10 +121,7 @@ config({
                 { id: "Newtonsoft.Json", version: "11.0.2" },
                 { id: "Newtonsoft.Json", version: "10.0.3", alias: "Newtonsoft.Json.v10" },
                 { id: "Newtonsoft.Json.Bson", version: "1.0.1" },
-                { id: "System.Data.SQLite", version: "1.0.109.2" },
```
almiliMSFT: how come that all of a sudden we don't need this?

Kr1sso: I guess it’s because you disabled DB init for aria 😀

Public/Src/App/Bxl/Bxl.dsc
```diff
@@ -75,7 +75,6 @@ namespace Main {
             importFrom("BuildXL.FrontEnd").CMake.dll,
             importFrom("BuildXL.FrontEnd").Sdk.dll,
             importFrom("Newtonsoft.Json").pkg,
-            importFrom("System.Data.SQLite.Core").pkg,
```
almiliMSFT: note that this package has the `.Core` suffix which is the one you didn't remove from `config.dsc` (I'm not sure if this comment carries any significance though :))

config.dsc
almiliMSFT: only on Windows; on macOS it should be just the same as before

dannyvv: It is not used at all, there were only references to the .Core pacakge

---
In reply to: [310282058](https://github.com/microsoft/BuildXL/pull/696#discussion_r310282058) [](ancestors = 310282058)

dannyvv: Note used, see other comment

---
In reply to: [310281164](https://github.com/microsoft/BuildXL/pull/696#discussion_r310281164) [](ancestors = 310281164)

Public/Src/App/Bxl/Bxl.dsc
dannyvv: our main exe doesn't need the dll to compile, so I removed it.

---
In reply to: [310281736](https://github.com/microsoft/BuildXL/pull/696#discussion_r310281736) [](ancestors = 310281736)

PR #698, opened 2019-08-02
Fix SAS tokens NullReferenceException when updating token

jbayardo:


Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -354,11 +354,12 @@ private AzureBlobStorageCredentials[] GetStorageCredentials(Dictionary<string, S
             return credentials.ToArray();
         }
 
-        private static AzureBlobStorageCredentials CreateAzureBlobCredentialsFromSasToken(UpdatingSasToken updatingSasToken)
+        private AzureBlobStorageCredentials CreateAzureBlobCredentialsFromSasToken(string secretName, UpdatingSasToken updatingSasToken)
         {
             var storageCredentials = new StorageCredentials(sasToken: updatingSasToken.Token.Token);
-            updatingSasToken.TokenUpdated += (token, sasToken) =>
+            updatingSasToken.TokenUpdated += (updatingToken, sasToken) =>
```
SergeyTeplyakov: I think the first argument here is a sender, so I think it make sense to rename the argument.

SergeyTeplyakov:
>         public event EventHandler<SasToken> TokenUpdated;

btw, you may use Action here to avoid confusion in the handling logic.

But this is minor...

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/Secrets.cs:58 in 1f80de9. [](commit_id = 1f80de9ac80baa7f99b026144b2e3abdfa233ea2, deletion_comment = False)

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
dannyvv: Sine you are not using the variable you can use `_` i.e.

```suggestion
updatingSasToken.TokenUpdated += (_, sasToken) =>
```

---
In reply to: [310297477](https://github.com/microsoft/BuildXL/pull/698#discussion_r310297477) [](ancestors = 310297477)

PR #699, opened 2019-08-02
Optimize AST conversion

narasamdya:
This change introduces an option to optimize conversion. This optimization includes:
- Disabling some analyses (lint?) during conversion
- Skip conversion of some constructs, like types, that are not necessary during evaluation.
The latter prevents us from allocating a lot of lists.

This optimization is in general unsafe, but when the specs are machine generated, this optimization should be fine.

Testing it on OfficeClientSetup_1907_retail_C2RUniversal (3587 modules) shows 40% improvement; 57s -> 33s.

This optimization may also ameliorate the long conversion time that has hit Office several times (to be tested)

This optimization is useful, not only for the current Office build, but also for the binary graph effort.

[AB#1568050](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1568050)


Public/Src/App/Bxl/Args.cs
```diff
@@ -990,6 +990,11 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                             "unsafe_IgnoreSetFileInformationByHandle",
                             sign => sandboxConfiguration.UnsafeSandboxConfigurationMutable.IgnoreSetFileInformationByHandle = sign,
                             isUnsafe: true),
+                        OptionHandlerFactory.CreateBoolOption(
+                            "unsafe_IgnoreUndeclaredAccessesUnderSharedOpaques",
```
SergeyTeplyakov: Just in case this is accidental... was this change intentional? (I'm 100% fine with it, just want to double check that this change didn't sneak in by an accident).

Public/Src/App/Bxl/Args.cs
```diff
@@ -1014,6 +1019,11 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                             sign =>
                             sandboxConfiguration.UnsafeSandboxConfigurationMutable.MonitorFileAccesses = sign,
                             isUnsafe: true),
+                        OptionHandlerFactory.CreateBoolOption(
+                            "unsafe_OptimizedAstConversion",
```
dannyvv: I through tyou needed it for BxlScriptAnalayzer app, just curious why doing this for bxl as well.

Public/Src/App/Bxl/Args.cs
narasamdya: I reordered it alphabetically. We used to maintain that order, but the check was turned off.

Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/AstConverter.cs
```diff
@@ -1364,7 +1364,7 @@ private TypeParameter ConvertTypeParameter(ITypeParameterDeclaration typeParamet
                 var name = SymbolAtom.Create(StringTable, nameStr);
 
                 // It is ok if type is null here
-                var type = ConvertType(declaration.Type, currentQualifierSpaceId);
+                Type type = m_conversionConfiguration.UnsafeOptions.SkipTypeConversion ? null : ConvertType(declaration.Type, currentQualifierSpaceId);
```
dannyvv: I'm actually curious why we do the type conversion at all....

Typecheck has passed so why do the runtime one?

Public/Src/App/Bxl/Args.cs
narasamdya: There has been a thread that AST conversion takes a long time. While the binary fragment is not there yet, this can be a remedy. But the stuff implemented here is beneficial for constructing the binary fragment that we're doing currently.

Public/Src/FrontEnd/Script/RuntimeModel/AstBridge/AstConverter.cs
narasamdya: At that time there's only 1 AST, eval AST. I guess you guys tried to maintain the structure as much as possible.

Public/Src/Utilities/Utilities/Tracing/EventId.cs
```diff
@@ -570,6 +570,7 @@ public enum EventId
         ConfigUnsafeAllowMissingOutput = 928,
         ConfigIgnoreValidateExistingFileAccessesForOutputs = 929,
         ConfigUnsafeIgnoreUndeclaredAccessesUnderSharedOpaques = 930,
+        ConfigUnsafeOptimizedAstConversion = 931,
```
dannyvv: Please don't ass new ones to Utilities.Tracing. Please add to the Tracing\log.cs file in the project that raises the event... In this case `Public/src/Engine/Dll/Tracing/log.cs`

narasamdya: We should do a clean sweep in a separate PR.

Public/Src/App/Bxl/Strings.resx
```diff
@@ -1009,6 +1009,9 @@ Example: ad2d42d2ec5d2ca0c0b7ad65402d07c7ef40b91e</value>
   <data name="Args_AdminRequiredProcessExecutionMode_NotSupportedOnNonWindows" xml:space="preserve">
     <value>Admin-required process execution mode '{0}' is not supported on non-Windows OS</value>
   </data>
+  <data name="HelpText_DisplayHelp_OptimizedAstConversion" xml:space="preserve">
+    <value>When enabled, optimized AST conversion by disabling some analyses and skipping some AST constructs. Defauts to off.</value>
```
mpysson: >disabling some analyses and skipping some AST constructs. [](start = 53, length = 57)

Could you add a note as to what the impact of doing this is?

PR #701, opened 2019-08-02
[macOS] compute job accounting info

almiliMSFT:
This PR adds job accounting information for process pips when executing on macOS.

This info includes processor system and user time, as well as peak memory usage.

This is achieved by probing the process usage information every 2 seconds and aggregating the results.

Computing this information is currently disabled by default, and can be enabled by passing `/kextMeasureProcessCpuTimes`.   This default is likely to change soon.

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcess.MacKext.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcess.MacKext.cs
@@ -7,16 +7,15 @@ using System.Diagnostics.ContractsLight;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
+using BuildXL.Interop;
+using BuildXL.Interop.MacOS;
 using BuildXL.Native.IO;
 using BuildXL.Utilities;
-using BuildXL.Utilities.Collections;
 using BuildXL.Utilities.Tasks;
 using JetBrains.Annotations;
-using BuildXL.Interop.MacOS;
 using static BuildXL.Interop.MacOS.Sandbox;
 using static BuildXL.Processes.SandboxedProcessFactory;
 using static BuildXL.Utilities.FormattableStringEx;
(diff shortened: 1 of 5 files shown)


Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -229,13 +229,14 @@ public virtual async Task<SandboxedProcessResult> GetResultAsync()
 
             var reportFileAccesses = ProcessInfo.FileAccessManifest?.ReportFileAccesses == true;
             var fileAccesses = reportFileAccesses ? (reports?.FileAccesses ?? s_emptyFileAccessesSet) : null;
-
+            var processTimes = GetProcessTimes();
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Processes/UnSandboxedProcess.cs
```diff
@@ -247,7 +248,7 @@ public virtual async Task<SandboxedProcessResult> GetResultAsync()
                 MessageProcessingFailure            = reports?.MessageProcessingFailure,
                 DumpCreationException               = m_dumpCreationException,
                 DumpFileDirectory                   = ProcessInfo.TimeoutDumpDirectory,
-                PrimaryProcessTimes                 = GetProcessTimes(),
+                PrimaryProcessTimes                 = processTimes,
```
almiliMSFT:
```suggestion
                PrimaryProcessTimes                 = GetProcessTimes(),
```

Public/Src/Engine/Processes/SandboxedProcess.MacKext.cs
```diff
@@ -165,11 +207,8 @@ private async Task OnProcessStartedAsync()
 
             if (MeasureCpuTime)
             {
-                // Allow read access for /usr/bin/time
-                ProcessInfo.FileAccessManifest.AddPath(
-                    AbsolutePath.Create(PathTable, TimeUtil),
-                    mask: FileAccessPolicy.MaskNothing,
-                    values: FileAccessPolicy.AllowReadAlways);
+                // start the timer now
+                m_perfTimer.Change(TimeSpan.FromSeconds(0), PerfProbeInternal);
```
mpysson: >m_perfTimer.Change(TimeSpan.FromSeconds(0), PerfProbeInternal); [](start = 16, length = 63)

A bit less intuitive but oftentimes the safer pattern is to only ever set the due time and not the period on the timer. Then when it fires you call Change() with a new due time in your timer callback. That prevents the timer from firing again while its still in the callback which can cause some unintended concurrency issues in the callback

almiliMSFT: will change

Public/Src/Engine/Processes/SandboxedProcessMac.cs
```diff
@@ -136,6 +177,10 @@ protected override System.Diagnostics.Process CreateProcess()
             process.StartInfo.Arguments = string.Empty;
             process.StartInfo.RedirectStandardInput = true;
 
+            // Stop the perf timer when the process exits
+            // NOTE: even if the timer fires after the process has exited, nothing bad should happen
+            process.Exited += (o, e) => m_perfTimer.Change(Timeout.InfiniteTimeSpan, PerfProbeInternal);
```
almiliMSFT:
```suggestion
            process.Exited += (o, e) => m_perfTimer.Change(Timeout.InfiniteTimeSpan, Timeout.InfiniteTimeSpan);
```

PR #705, opened 2019-08-05
Implement proactive copy

JuanCarlosGI:
Proactive copy has been implemented. When a piece of content is put into the distributed cache, it checks if it is the first one with this piece of content. If it is, it does a proactive copy asynchronously.

Implementation:
- Machine A selects a random MachineLocation (Machine B) from already known locations.
- Machine A sends machine B a gRPC message requesting that it copies a file from machine A.
- Machine B will then use whatever FileCopier implementation it is using to copy the file from machine A, via a Pin.


Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -22,9 +26,11 @@ namespace BuildXL.Cache.ContentStore.Distributed.Sessions
     /// A content location based content session with an inner content session for storage.
     /// </summary>
     /// <typeparam name="T">The content locations being stored.</typeparam>
-    public class DistributedContentSession<T> : ReadOnlyDistributedContentSession<T>, IContentSession
+    public class DistributedContentSession<T> : ReadOnlyDistributedContentSession<T>, IContentSession, IFileCopyingSession
         where T : PathBase
     {
+        private static Random Random = new Random();
```
kewilli: >Random Random = new Random(); [](start = 23, length = 29)

Consider using `ThreadSafeRandom` instead

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +181,32 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
 
             return putResult;
         }
+
+        private async Task<BoolResult> RequestProactiveCopyIfNeededAsync(OperationContext context, ContentHash hash, UrgencyHint urgencyHint)
+        {
+            var hashArray = new[] { hash };
+
+            // First check in local location store, then global if failed.
+            var getLocationsResult = await ContentLocationStore.GetBulkAsync(context, hashArray, context.Token, UrgencyHint.Nominal, GetBulkOrigin.Local);
```
SergeyTeplyakov: Need to trace all of this and add counters.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -284,6 +287,82 @@ private async Task CopyFileAsync(CopyFileRequest request, IServerStreamWriter<Co
...
+        /// Implements a request copy file request
+        /// </summary>
+        private async Task<RequestCopyFileResponse> RequestCopyFileAsync(RequestCopyFileRequest request, CancellationToken cancellationToken)
+        {
+            OperationStarted();
+
+            var startTime = DateTime.UtcNow;
+            var cacheContext = new OperationContext(new Context(new Guid(request.TraceId), _logger), cancellationToken);
+
+            var sessionResult = await GetRequestCopySessionAsync(cacheContext, request.CacheName);
```
ljcollins25: >GetRequestCopySessionAsync [](start = 38, length = 26)

I don't think this should be on the session level. I think it should be IFileCopyingStore instead. Doesn't make sense to create a session just to initiate a copy.

JuanCarlosGI: Is it possible to do a Put without a session? Because technically this is not just a copy, its a copy and then a put

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +181,32 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
 
             return putResult;
         }
+
+        private async Task<BoolResult> RequestProactiveCopyIfNeededAsync(OperationContext context, ContentHash hash, UrgencyHint urgencyHint)
```
dcwuser: We don't use UrgencyHints. Let's stop propagating them in new methods.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
dcwuser: Minor: Can we add a non-bulk ContentLocationStore.Get? Even if for now if is implemented by calling the bulk method, this keeps all the array allocations in one place that we can eliminate later when we add a real non-bulk path.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +181,32 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
...
+            // First check in local location store, then global if failed.
+            var getLocationsResult = await ContentLocationStore.GetBulkAsync(context, hashArray, context.Token, UrgencyHint.Nominal, GetBulkOrigin.Local);
+            if (!getLocationsResult.Succeeded || getLocationsResult.ContentHashesInfo[0].Locations.Count == 0)
+            {
+                getLocationsResult = await ContentLocationStore.GetBulkAsync(context, hashArray, context.Token, UrgencyHint.Nominal, GetBulkOrigin.Global);
+            }
+
+            if (getLocationsResult.Succeeded)
+            {
+                if (getLocationsResult.ContentHashesInfo[0].Locations.Count == 1)
```
dcwuser: Use <= 1, to deal with (admitted unexpected) case that no records are found.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +181,32 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
...
+                {
+                    var locations = ContentLocationStore.GetKnownMachineLocations();
+                    var location = locations[ThreadSafeRandom.Generator.Next(locations.Length)];
```
dcwuser: Deal with corner case that randomly chosen location is self. Suggest while (location == self) { get new location } so that we never exit with location == self.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +181,32 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
...
+            if (!getLocationsResult.Succeeded || getLocationsResult.ContentHashesInfo[0].Locations.Count == 0)
+            {
+                getLocationsResult = await ContentLocationStore.GetBulkAsync(context, hashArray, context.Token, UrgencyHint.Nominal, GetBulkOrigin.Global);
+            }
+
+            if (getLocationsResult.Succeeded)
+            {
+                if (getLocationsResult.ContentHashesInfo[0].Locations.Count == 1)
+                {
+                    var locations = ContentLocationStore.GetKnownMachineLocations();
```
dcwuser: This returns a cached singleton list, right? We don't want to re-construct list every time.

dcwuser: Corner case: if locations.Length <= 2, break off.

Public/Src/Cache/ContentStore/Distributed/NuCache/ClusterState.cs
```diff
@@ -130,5 +130,10 @@ public bool TryResolveMachineId(MachineLocation machineLocation, out MachineId m
         {
             return _idByLocationMap.TryGetValue(machineLocation, out machineId);
         }
+
+        /// <summary>
+        /// Gets all locations that are currently known.
+        /// </summary>
+        public MachineLocation[] GetKnownLocations() => _idByLocationMap.Keys.ToArray();
```
dcwuser: Yeah, it looks like this does create a new array on every call. To avoid this, use a read-only wrapper or maybe just create a GetRandomLocation API that operates on existing location map.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
JuanCarlosGI: While it would be useful, I think this is something we should do in a separate PR; This PR is starting to be too big and doing refactors might make it confusing.

Public/Src/Cache/ContentStore/Distributed/ContentLocations/IContentLocationStore.cs
```diff
@@ -114,6 +114,11 @@ public interface IContentLocationStore : IStartupShutdown
         /// Reports about new reputation for a given location.
         /// </summary>
         void ReportReputation(MachineLocation location, MachineReputation reputation);
+
+        /// <summary>
+        /// Returns a random machine location, excluding the specified location. Returns default if operation is not possible.
+        /// </summary>
+        MachineLocation GetRandomMachineLocation(MachineLocation except);
```
SergeyTeplyakov: Just curious: should it be 'Result<MachineLocation>'?

SergeyTeplyakov: Do you want to name it 'currentMachine'? Because I suspect we just want to exclude the current machine.

If we'll use 'currentMachine' then the method will be slightly less generic, but imo it would be clearer.

Public/Src/Cache/ContentStore/Distributed/NuCache/ClusterState.cs
```diff
@@ -130,5 +131,25 @@ public bool TryResolveMachineId(MachineLocation machineLocation, out MachineId m
         {
             return _idByLocationMap.TryGetValue(machineLocation, out machineId);
         }
+
+        /// <summary>
+        /// Gets a random locations excluding the specified location. Returns default if operation is not possible.
```
SergeyTeplyakov: you may use <inheritdoc /> here to avoid comments duplication, if you want.

Very minor.

Public/Src/Cache/ContentStore/Distributed/NuCache/ClusterState.cs
```diff
@@ -130,5 +131,25 @@ public bool TryResolveMachineId(MachineLocation machineLocation, out MachineId m
...
+        /// </summary>
+        public MachineLocation GetRandomMachineLocation(MachineLocation except)
+        {
+            if (_idByLocationMap.Keys.Any(location => !location.Equals(except)))
+            {
+                MachineLocation location;
+                do
+                {
+                    location = _idByLocationMap.Keys.ElementAt(ThreadSafeRandom.Generator.Next(_idByLocationMap.Keys.Count));
```
SergeyTeplyakov: YOu may consider getting all the keys into a local because the Keys property is very expensive:

It creates a copy of all the keys and it acquires all the locks.

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStore.cs
```diff
@@ -1306,6 +1306,24 @@ public Task<Result<byte[]>> GetBlobAsync(OperationContext context, ContentHash h
...
+            {
+                MachineLocation location;
+                do
+                {
+                    location = _idsByLocation.Keys.ElementAt(ThreadSafeRandom.Generator.Next(_idsByLocation.Keys.Count));
+                }
+                while (location.Equals(except));
+
+                return location;
+            }
```
SergeyTeplyakov: If we have two similar implementations we can add a helper method that will give us machine location from a collection of machine locations to avoid code duplication.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -25,6 +25,14 @@ namespace BuildXL.Cache.ContentStore.Distributed.Sessions
     public class DistributedContentSession<T> : ReadOnlyDistributedContentSession<T>, IContentSession
         where T : PathBase
     {
+        private enum Counters
+        {
+            GetLocationsSatisfiedFromLocal,
+            GetLocationsSatisfiedFromRemote
+        }
+
+        private CounterCollection<Counters> _counters = new CounterCollection<Counters>();
```
SergeyTeplyakov: nit: readonly.

Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -223,6 +226,17 @@ void traceCopyFailed(Context c)
             }
         }
 
+        /// <summary>
+        /// Requests another machine to copy from the current machine.
+        /// </summary>
+        public Task<BoolResult> RequestCopyFileAsync(OperationContext context, ContentHash hash, MachineLocation targetLocation)
```
SergeyTeplyakov: Should we wrap implementation into PerformOperationAsync?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1235,6 +1235,11 @@ public Task<Result<byte[]>> GetBlobAsync(OperationContext context, ContentHash h
             return GlobalStore.GetBlobAsync(context, hash);
         }
 
+        /// <summary>
+        /// Gets a random machine location, excluding a specified location.
+        /// </summary>
+        public MachineLocation GetRandomMachineLocation(MachineLocation except) => ClusterState.GetRandomMachineLocation(except);
```
SergeyTeplyakov: nit: you may use <inheritdoc /> if you want to avoid comments duplication.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +180,51 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
...
             return putResult;
         }
+
+        private async Task<BoolResult> RequestProactiveCopyIfNeededAsync(OperationContext context, ContentHash hash)
+        {
+            var hashArray = new[] { hash };
+
+            // First check in local location store, then global if failed.
+            var getLocationsResult = await ContentLocationStore.GetBulkAsync(context, hashArray, context.Token, UrgencyHint.Nominal, GetBulkOrigin.Local);
+            if (!getLocationsResult.Succeeded || getLocationsResult.ContentHashesInfo[0].Locations.Count == 0)
```
SergeyTeplyakov: Is it possible that ContentHashesInfo is empty?

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +180,51 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
 
             return putResult;
         }
+
+        private async Task<BoolResult> RequestProactiveCopyIfNeededAsync(OperationContext context, ContentHash hash)
```
SergeyTeplyakov: Need to wrap the entire thing into context.PerformOperartion.

SergeyTeplyakov: With its own counter for tracking the operation duration.

---
In reply to: [312225644](https://github.com/microsoft/BuildXL/pull/705#discussion_r312225644) [](ancestors = 312225644)

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +180,51 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
...
+                getLocationsResult = await ContentLocationStore.GetBulkAsync(context, hashArray, context.Token, UrgencyHint.Nominal, GetBulkOrigin.Global);
+
+                if (getLocationsResult.Succeeded)
+                {
+                    _counters[Counters.GetLocationsSatisfiedFromRemote].Increment();
+                }
+            }
+            else
+            {
+                _counters[Counters.GetLocationsSatisfiedFromLocal].Increment();
```
SergeyTeplyakov: nit, but I would suggest to reverse the if and do:

if (getLocationsResult && getLocationsResult.LocationsCount != 0) {
  counters[Local].Increment();
}

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStore.cs
JuanCarlosGI: They ended up being slightly different because in one case we get an array which can contain the 'default' value, and we need to make sure that we don't return that.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -165,5 +180,51 @@ private async Task<PutResult> RegisterPutAsync(Context context, CancellationToke
...
+            }
+
+            if (getLocationsResult.Succeeded)
+            {
+                // Only request copy when this is the only machine where the content is located.
+                if (getLocationsResult.ContentHashesInfo[0].Locations.Count <= 1)
+                {
+                    var location = ContentLocationStore.GetRandomMachineLocation(except: LocalCacheRootMachineLocation);
```
SergeyTeplyakov: We need another counter to track the following more explicitely (in the trace message for sure and, maybe in the counters):

The result of this operation is:
we either:
- requested a copy because there is only N copies available (this threshold should be configurable)
- did not requested a copy because we have enough copies
- did not requested a copy because we failed to find a machine in the stamp.

I would suggest to restructure this in the fllowing way (-ish):

var getLocations = GetLocationsFromMultipleLevels(context, hash);

if (!getLocations) {
  // We failed to get locations, but we still (potentially) may try to do a proactive copy if we want!
  return new BoolResult($"Failed to proactively copy the hash '{hash}'. {getLocations}");
}

if (getLocations.Origin == Local) {
  incrementCounter for locals;
}
else {
  incrmenent counter for global;
}

if (getLocations.Locations.Count < ConfiguredThreashold) {
  // assuming GetRandomMachineLocations returns BoolResult.
   var targetLocation = GetRandomMachineLocation();
  if (!targetLocation) {
    return new BoolResult($"Failed to proactively copy the hash '{hash}'. {targetLocation});
  }

  return await DistributedCopier.RequestCopyFileAsync();
}

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
JuanCarlosGI: If it is, it is a bug and we're right to blow up, IMO

Public/Src/Cache/ContentStore/Distributed/ContentLocations/IContentLocationStore.cs
JuanCarlosGI: I think "except" better conveys the behavior or the function, whereas "currentMachine" doesn't tell me anything about what the implementation will do with it

JuanCarlosGI: I think "except" better conveys the behavior or the function, whereas "currentMachine" doesn't tell me anything about what the implementation will do with it

---
In reply to: [312215220](https://github.com/microsoft/BuildXL/pull/705#discussion_r312215220) [](ancestors = 312215220)

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStore.cs
JuanCarlosGI: They ended up being slightly different because in one case we get an array which can contain the 'default' value, and we need to make sure that we don't return that.

---
In reply to: [312219858](https://github.com/microsoft/BuildXL/pull/705#discussion_r312219858) [](ancestors = 312219858)

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
JuanCarlosGI: While it would be useful, I think this is something we should do in a separate PR; This PR is starting to be too big and doing refactors might make it confusing.

---
In reply to: [311186928](https://github.com/microsoft/BuildXL/pull/705#discussion_r311186928) [](ancestors = 311186928)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
JuanCarlosGI: Is it possible to do a Put without a session? Because technically this is not just a copy, its a copy and then a put

---
In reply to: [311172395](https://github.com/microsoft/BuildXL/pull/705#discussion_r311172395) [](ancestors = 311172395)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
JuanCarlosGI: This does not inherit from IContentLocationStore

---
In reply to: [312225091](https://github.com/microsoft/BuildXL/pull/705#discussion_r312225091) [](ancestors = 312225091)

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
JuanCarlosGI: If it is, it is a bug and we're right to blow up, IMO

---
In reply to: [312225497](https://github.com/microsoft/BuildXL/pull/705#discussion_r312225497) [](ancestors = 312225497)

Public/Src/Cache/ContentStore/Distributed/NuCache/ClusterState.cs
```diff
@@ -130,5 +132,28 @@ public bool TryResolveMachineId(MachineLocation machineLocation, out MachineId m
...
         }
+
+        /// <summary>
+        /// Gets a random locations excluding the specified location. Returns default if operation is not possible.
+        /// </summary>
+        public Result<MachineLocation> GetRandomMachineLocation(MachineLocation except)
+        {
+            using (_lock.AcquireReadLock())
+            {
+                if (_locationByIdMap.Any(location => !location.Equals(except)))
```
ljcollins25: >_locationByIdMap [](start = 20, length = 16)

Exclude inactive machines.

Public/Src/Cache/ContentStore/Distributed/NuCache/ClusterState.cs
```diff
@@ -130,5 +132,28 @@ public bool TryResolveMachineId(MachineLocation machineLocation, out MachineId m
...
+                {
+                    MachineLocation location;
+                    do
+                    {
+                        location = _locationByIdMap[ThreadSafeRandom.Generator.Next(_locationByIdMap.Length)];
+                    }
+                    while (location.Equals(default) || location.Equals(except));
```
ljcollins25: >location [](start = 55, length = 8)

Exclude inactive machines.

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStore.cs
ljcollins25: Reactivating. I think we can unify these. Let's chat.

---
In reply to: [312661031](https://github.com/microsoft/BuildXL/pull/705#discussion_r312661031) [](ancestors = 312661031,312219858)

Public/Src/Cache/ContentStore/Distributed/NuCache/ClusterState.cs
```diff
@@ -130,5 +132,28 @@ public bool TryResolveMachineId(MachineLocation machineLocation, out MachineId m
...
+        public Result<MachineLocation> GetRandomMachineLocation(MachineLocation except)
+        {
+            using (_lock.AcquireReadLock())
+            {
+                if (_locationByIdMap.Any(location => !location.Equals(except)))
+                {
+                    MachineLocation location;
+                    do
+                    {
+                        location = _locationByIdMap[ThreadSafeRandom.Generator.Next(_locationByIdMap.Length)];
```
ljcollins25: >_locationByIdMap.Length [](start = 84, length = 23)

Use MaxMachineId + 1.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
ljcollins25: Indeed this would involve a non-trivial amount of shuffling. That said, we still need to shutdown the session.

---
In reply to: [312661540](https://github.com/microsoft/BuildXL/pull/705#discussion_r312661540) [](ancestors = 312661540,311172395)

JuanCarlosGI: Decided to complete PR as is and worry about this at a later time

---
In reply to: [312988309](https://github.com/microsoft/BuildXL/pull/705#discussion_r312988309) [](ancestors = 312988309,312661540,311172395)

PR #706, opened 2019-08-05
Add BuildXLRuntimeCacheMissAllPips env var

semihokur:


Public/Src/Engine/Scheduler/Tracing/RuntimeCacheMissAnalyzer.cs
```diff
@@ -227,7 +227,8 @@ private bool IsCacheMissEligible(PipId pipId)
                 return false;
             }
 
-            if (m_changedPips.WasVisited(pipId.ToNodeId()))
+            if (!EngineEnvironmentSettings.RuntimeCacheMissAllPips
+                && m_changedPips.WasVisited(pipId.ToNodeId()))
```
olkononenko: >&& [](start = 16, length = 2)

should it be "||" ?

PR #708, opened 2019-08-06
  Exclude pips in inner filter for requiredfor filter

norahuang:
[AB#1577773](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1577773)


dannyvv:
Just general comment that your commit history is 'funky' again with old commits..

norahuang:
> Just general comment that your commit history is 'funky' again with old commits..
Sorry for letting the old commits in again. I will make sure it doesn't happen next time.

dannyvv:
> > Just general comment that your commit history is 'funky' again with old commits..
> > Sorry for letting the old commits in again. I will make sure it doesn't happen next time.

not a problem :) Just observing in case it makes things easier on your end :)

Public/Src/Engine/Scheduler/Filter/ClosureFunctionFilter.cs
```diff
@@ -163,7 +168,7 @@ public enum ClosureMode
         /// <summary>
         /// Gets only the direct neighbors (excludes the matching pips for the inner filter).
         /// NOTE: If pip is a matching neighbor of one of the matching pips for the inner filter
-        /// it will be included.
+        /// it will be excluded.
```
mpysson: >/// it will be excluded. [](start = 8, length = 24)

I hope there's also some user facing documentation to update for this. I know there's a big block of help text for filtering. If requiredfor, copydpt, dpc, dpt, and these special filters aren't already in there, could you add?

Public/Src/Engine/UnitTests/Scheduler/PipFilterTest.cs
```diff
@@ -407,7 +407,30 @@ public void TestRequiredInputsFilter()
...
+        {
+            // P0 -> P1 -> P2 -> Seal -> P3 (P3 doesn't matter in this test case, just to reuse the existing SetupDependenciesFilterGraph funcion)
+            FileArtifact o0, o1, o2;
+            Process p1, p2;
+            PipGraph graph;
+            SetupDependenciesFilterGraph(out o0, out o1, out o2, out p1, out p2, out graph);
+
+            var filter =
+                new RootFilter(
+                    new DependenciesFilter(
```
mpysson: >DependenciesFilter [](start = 24, length = 18)

I don't see any other tests for DependenciesFilter operating on a set of more than 1 pip. Could you add test cases to cover those scenarios where the inner filter yields a nonzero set of pips but the ClosureMode is TransitiveIncludingSelf?

Public/Src/Engine/Scheduler/Filter/ClosureFunctionFilter.cs
norahuang: Added to helptext

Public/Src/Engine/UnitTests/Scheduler/PipFilterTest.cs
norahuang: Made the test case a theory and add the TransitiveIncludingSelf to it.

PR #712, opened 2019-08-07
Stop the build after the first materialization error

semihokur:
The motivation is to save the machine hours. After we try 'stop on the first CTMIS' for a week, I will generalize this concept to all internal errors.

### DIFF
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -2347,17 +2347,20 @@ namespace BuildXL.Scheduler
                     // We stop on the first error only on the master or single-machine builds.
                     // During cancellation, master coordinates with workers to stop the build.
 
+                    // ErrorsLoggedById is a ConcurrentBag. Its Contains() isn't particularly performant. It copies everything to a new list and then enumerates that.
                     bool hasMaterializationErrorHappened = m_executePhaseLoggingContext.ErrorsLoggedById.Contains((ushort)EventId.PipMaterializeDependenciesFromCacheFailure)
                         || m_executePhaseLoggingContext.ErrorsLoggedById.Contains((ushort)EventId.PipMaterializeDependenciesFailureUnrelatedToCache);
 
-                    // Early terminate the build if StopOnFirstError is enabled or a materialization error is occurred. 
-                    bool earlyTerminate = m_scheduleConfiguration.StopOnFirstError || hasMaterializationErrorHappened;
+                    // Early terminate the build if 
+                    // (1) StopOnFirstError is enabled or 
+                    // (2) a materialization error is occurred in a distributed build.
+                    bool earlyTerminate = m_scheduleConfiguration.StopOnFirstError || (hasMaterializationErrorHappened && IsDistributedMaster);
 
                     if (!IsTerminating && earlyTerminate)
                     {
                         Logger.Log.ScheduleTerminatingDueToPipFailure(m_executePhaseLoggingContext, pipDescription);
 
-                        RequestTermination(false);
+                        RequestTermination(cancelQueue: false);
                     }
 
                     Contract.Assert(m_executePhaseLoggingContext.ErrorWasLogged, I($"Should have logged error for pip: {pipDescription}"));
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -2342,19 +2342,22 @@ namespace BuildXL.Scheduler
                 // This happens before we traverse the dependents of this failed pip; since SchedulePipIfReady
                 // checks m_scheduleTerminating this means that there will not be 'Skipped' pips in this mode
                 // (they remain unscheduled entirely).
-                if (!succeeded && m_scheduleConfiguration.StopOnFirstError)
+                if (!succeeded && !IsDistributedWorker)
                 {
-                    if (!m_scheduleTerminating)
+                    // We stop on the first error only on the master or single-machine builds.
+                    // During cancellation, master coordinates with workers to stop the build.
+
+                    bool hasMaterializationErrorHappened = m_executePhaseLoggingContext.ErrorsLoggedById.Contains((ushort)EventId.PipMaterializeDependenciesFromCacheFailure)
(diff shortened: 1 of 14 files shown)


dcwuser:
How does this interact with the new "it's not CTMIS if there are also worker communication errors" logic? Wouldn't this mean that, if the CTMIS came first, we wouldn't see the worker communication errors that would follow, and therefore we would again mis-classify the failure as CTMIS?

semihokur:
> How does this interact with the new "it's not CTMIS if there are also worker communication errors" logic? Wouldn't this mean that, if the CTMIS came first, we wouldn't see the worker communication errors that would follow, and therefore we would again mis-classify the failure as CTMIS?

Those communication errors are logged before CTMIS, so we would still see them

dcwuser:
Sorry, I don't understand this. There are definitely cases where a materialization error message, even a lot of them, appear in Domino.err before worker communication error messages. That's the entire reason that the bad CTMIS errors appeared in batmon and BXL error fields, because they both based their reports on the first error. So saying this won't happen because communication errors should be logged first seems to be contradicted by experience: it occurs very often that materialization errors are logged first.

---
In reply to: [519239888](https://github.com/microsoft/BuildXL/pull/712#issuecomment-519239888) [](ancestors = 519239888)

mpysson:
I share David's concern. Historically we had seen connectivity problems first encountered by a worker failing to materialize from an unavailable worker before the master realized it too was also unable to communicate with the unavailable worker for sake of scheduling operations. There's some special error handling that gives precedence to master/worker connectivity failures when determining the overall failure for the build session.

Depending on how we shut down the session things may just work out though. Let's say the master stops scheduling all pips after the first internal failure and goes down the road to terminate the session. As long as it still goes down the codepath for the end of build handshake between master and all workers, that handshake could still fail with the communication error that causes us to classify the build as having connectivity problems rather than materialization. So the big question is will a DistributionExecutePipFailedNetworkFailure still get logged during that teardown? If so, we should be good.

---
In reply to: [519251043](https://github.com/microsoft/BuildXL/pull/712#issuecomment-519251043) [](ancestors = 519251043,519239888)

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2342,19 +2342,22 @@ public virtual async Task OnPipCompleted(RunnablePip runnablePip)
...
                         Logger.Log.ScheduleTerminatingDueToPipFailure(m_executePhaseLoggingContext, pipDescription);
 
-                        // This flag prevents normally-scheduled pips (i.e., by refcount) from starting (thus m_numPipsQueuedOrRunning should
-                        // reach zero quickly). But we do allow further pips to run inline (see RunPipInline); that's safe from an error
-                        // reporting perspective since m_hasFailures latches to false.
-                        m_scheduleTerminating = true;
-
-                        // A build that has been canceled should return failure
-                        m_hasFailures = true;
+                        RequestTermination(false);
```
mpysson: >false [](start = 43, length = 5)

nit - name parameter so we know what intent is

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2342,19 +2342,22 @@ public virtual async Task OnPipCompleted(RunnablePip runnablePip)
                 // This happens before we traverse the dependents of this failed pip; since SchedulePipIfReady
                 // checks m_scheduleTerminating this means that there will not be 'Skipped' pips in this mode
                 // (they remain unscheduled entirely).
-                if (!succeeded && m_scheduleConfiguration.StopOnFirstError)
+                if (!succeeded && !IsDistributedWorker)
                 {
-                    if (!m_scheduleTerminating)
+                    // We stop on the first error only on the master or single-machine builds.
```
mpysson: >single-machine builds. [](start = 72, length = 22)

I wouldn't apply this to single machine builds to begin with

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2342,19 +2342,22 @@ public virtual async Task OnPipCompleted(RunnablePip runnablePip)
...
                 // checks m_scheduleTerminating this means that there will not be 'Skipped' pips in this mode
                 // (they remain unscheduled entirely).
-                if (!succeeded && m_scheduleConfiguration.StopOnFirstError)
+                if (!succeeded && !IsDistributedWorker)
                 {
-                    if (!m_scheduleTerminating)
+                    // We stop on the first error only on the master or single-machine builds.
+                    // During cancellation, master coordinates with workers to stop the build.
+
+                    bool hasMaterializationErrorHappened = m_executePhaseLoggingContext.ErrorsLoggedById.Contains((ushort)EventId.PipMaterializeDependenciesFromCacheFailure)
```
mpysson: >ErrorsLoggedById.Contains [](start = 88, length = 25)

ConcurrentBag.Contains() isn't particularly performant. It copies everything to a new list and then enumerates that. I think it's smart in the empty case and we're only doing this for failed pips so it's probably ok. It might be worth a comment to that effect. Your call

semihokur: added a comment. I would not worry about it as this is only for failed pips and we iterate through the errors.

semihokur:
> I share David's concern. Historically we had seen connectivity problems first encountered by a worker failing to materialize from an unavailable worker before the master realized it too was also unable to communicate with the unavailable worker for sake of scheduling operations. There's some special error handling that gives precedence to master/worker connectivity failures when determining the overall failure for the build session.
> 
> Depending on how we shut down the session things may just work out though. Let's say the master stops scheduling all pips after the first internal failure and goes down the road to terminate the session. As long as it still goes down the codepath for the end of build handshake between master and all workers, that handshake could still fail with the communication error that causes us to classify the build as having connectivity problems rather than materialization. So the big question is will a DistributionExecutePipFailedNetworkFailure still get logged during that teardown? If so, we should be good.
> 
> Signing off assuming you work address this concern.
> 
> In reply to: [519251043](https://github.com/microsoft/BuildXL/pull/712#issuecomment-519251043) [](ancestors = 519251043,519239888)

Added ProblematicWorkerExit that is logged during the teardown. I updated your logic in BuildXLApp accordingly.

PR #715, opened 2019-08-07
Fix VSS auth on .NET Core when credential provider is specified

almiliMSFT:


Public/Src/Cache/VerticalStore/BuildCacheAdapter/BuildCacheUtils.cs
```diff
@@ -87,7 +92,7 @@ private static string GetAadUserNameUpn()
             }
             else
             {
-                throw new ArgumentException("PAT must be supplied when running with CoreCLR");
```
dannyvv: >with CoreCLR [](start = 79, length = 12)

Nice catch :)

Note: Let me know in IM/Email if you'd like me to rush rollout of this commit to CB.

PR #723, opened 2019-08-12
Update Prelude for incrementalTools

semihokur:


Public/Sdk/Public/Prelude/Prelude.Configuration.dsc
```diff
@@ -415,6 +415,9 @@ interface Configuration {
 
     searchPathEnumerationTools?: RelativePath[];
     
+    /** List of incremental tools for special observed file access handling */
+    incrementalTools?: RelativePath[];
```
dannyvv: I really wish this was per-pip.....

semihokur: We discussed about it; but we decided on the global level. one good reason is that it is so similar to search path enumeration tools, which is global as well.

PR #724, opened 2019-08-13
Populate RocksDB with Static Graph information

sahilmgandhi:
[AB#1581025](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1581025)

[Update]
No longer a WIP, so removed the tag. 

With this PR, we can now ingest all [immediately] relevant information (events, piptable, pipgraph) to the rocksdb instance and I have ported over the eventstats and dumppip analyzers as a proof of concept. Next steps include flushing out more of the public facing apis and inserting other "creative" indexes and information to the DB.


Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/BXLInvocationAnalyzer.cs
```diff
@@ -79,13 +81,22 @@ internal sealed class BXLInvocationAnalyzer : Analyzer
         /// </summary>
         public string OutputFilePath;
 
-        public BXLInvocationAnalyzer(AnalysisInput input) : base(input) { }
+        private Stopwatch m_stopWatch;
+
+        public BXLInvocationAnalyzer(AnalysisInput input) : base(input)
+        {
+            m_stopWatch = new Stopwatch();
+            m_stopWatch.Start();
```
almiliMSFT: Better start this stopwatch in `Analyze()`.  If you do that, a local variable in `Analyze()` (instead of the `m_stopWatch` field) would suffice

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/EnumsAndHelperStructs.proto
```diff
@@ -0,0 +1,1226 @@
...
+//
+// Enums
+//
+
+// Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, 
+// not children of it. Therefore, enum values must be unique within "BuildXL.Execution.Analyzer", not 
+// just within a particular enum. If there are conflicting enum values, append the name of the enum in front.
+// ie: ChooseWorkerCpu ---> PipExecutionStep_ChooseWorkerCpu
+
+enum ExecutionEventId{
```
almiliMSFT: It really looks to me that this file ought to be auto-generated.  Otherwise, whenever someone adds any new enum constant this file will have to be manually updated.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/NewDumpPipAnalyzer.cs
```diff
@@ -0,0 +1,256 @@
...
+                //dataStore.GetPipsOfType(Xldb.PipType.CopyFile).ToList().ForEach(p => Console.WriteLine(p.ToString()));
+
+                var pip = dataStore.GetPipBySemiStableHash(m_semiStableHash, out var pipType);
+                if (pip == null)
+                {
+                    Console.WriteLine($"Pip with the SemiStableHash {m_semiStableHash} was not found. Exiting Analyzer");
+                    return 1;
+                }
+
+                dynamic testVar = null;
```
almiliMSFT: I don't see any uses of this var

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/NewEventStatsAnalyzer.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        /// <inheritdoc/>
+        public override int Analyze()
+        {
+            using (var dataStore = new XldbDataStore(storeDirectory: InputDirPath))
+            {
+                using (var outputStream = File.OpenWrite(OutputFilePath))
+                {
+                    using (var writer = new StreamWriter(outputStream))
```
almiliMSFT: you don't need braces for nested `using` statements, i.e., you should write it like:
```csharp
using (var dataStore = new XldbDataStore(storeDirectory: InputDirPath))
using (var outputStream = File.OpenWrite(OutputFilePath))
using (var writer = new StreamWriter(outputStream))
{
    ...
}
```

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
```diff
@@ -389,24 +470,155 @@ public override void DominoInvocation(DominoInvocationEventData data)
...
+                WriteBatchToDb(pipSemistableKeys, pipSemistableValues, XldbDataStore.PipColumnFamilyName);
+                WriteBatchToDb(pipPipIdKeys, pipValues, XldbDataStore.PipColumnFamilyName);
+            });
         }
 
         /// <summary>
         /// Write a key/value pair to the db
         /// </summary>
-        public void WriteToDb(byte[] key, byte[] value)
+        public void WriteToDb(byte[] key, byte[] value, string columnFamilyName = null)
```
jbayardo: Noticed you have some usages like:

for (...) {
  ...
  WriteToDb(..)
}

This means you will take the lock in the accessor once for every write. I'd take it only once and pass the IBuildXLKeyValueStore around instead. Keep in mind that the accessor's lock is there only to avoid incorrect disposal of the DB.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
```diff
@@ -389,24 +470,155 @@ public override void DominoInvocation(DominoInvocationEventData data)
...
+        /// <summary>
+        /// Write batches of key/value pairs to the db
+        /// </summary>
+        public void WriteBatchToDb(IEnumerable<byte[]> keys, IEnumerable<byte[]> values, string columnFamilyName = null)
         {
             Analysis.IgnoreResult(
                 m_accessor.Use(database =>
                 {
-                    database.Put(key, value);
+                    database.ApplyBatch(keys, values, columnFamilyName);
```
jbayardo: IKeyValueStore no longer has the [contravariance](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/in-generic-modifier) requirement on `TKey`, so `ApplyBatch` can be changed to take a `Dictionary<byte[], byte[]>`, which would at the same time let you to avoid using so many lists in the method that calls this.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
```diff
@@ -389,24 +470,155 @@ public override void DominoInvocation(DominoInvocationEventData data)
...
+
+                for (int j = start; j < end; j++)
+                {
+                    var pipId = pipIds[j];
+                    pipsIngested++;
+
+                    if (pipsIngested % 100000 == 0)
+                    {
+                        Console.Write(".");
+                        WriteBatchToDb(pipSemistableKeys, pipSemistableValues, XldbDataStore.PipColumnFamilyName);
```
jbayardo: You are doing 2 separate transactions here, may want to do a single one, or do the 2 in parallel. Also, if you are anyhow able to feed the keys in sorted order, that should speed up insertion.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/StaticGraph.proto
```diff
@@ -0,0 +1,151 @@
...
+
+message PipGraphStaticFingerprints{
+    map<uint32, ContentFingerprint> PipStaticFingerprints = 1;
+}
+
+//
+//  Graph Data
+//
+
+message PipTable{
```
ljcollins25: >PipTable [](start = 8, length = 8)

Pip table is just the list of pips. There should be no need to have a separate data structure for this.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/EnumsAndHelperStructs.proto
ljcollins25: I agree with Aleks. That said, I would just  add a TODO here since I think that's out of scope for this PR.

---
In reply to: [313452127](https://github.com/microsoft/BuildXL/pull/724#discussion_r313452127) [](ancestors = 313452127)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/NewDumpPipAnalyzer.cs
```diff
@@ -0,0 +1,214 @@
...
+using System.Linq;
+using BuildXL.Analyzers.Core.XLGPlusPlus;
+using BuildXL.Execution.Analyzer.Xldb;
+using BuildXL.ToolSupport;
+
+namespace BuildXL.Execution.Analyzer
+{
+    internal partial class Args
+    {
+        public Analyzer InitializeNewDumpPipAnalyzer()
```
ljcollins25: >InitializeNewDumpPipAnalyzer [](start = 24, length = 28)

Can we refer to the "new" analyzers as {Name}XldbAnalyzer instead of New{Name}Analyzer.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/EnumsAndHelperStructs.proto
sahilmgandhi: I also agree with this, the problem though is that Protobuf does not support things like inheritance or fancy maps with objects (messages in Protobuf) as keys and what not. Thus the auto-generator would have to still be maintained when things are added, but we can discuss this more. Another idea that Newton and I discussed was to instead have some kind of "validator" that could run and check if fields in the files are present here, and a user could suppress fields that should/shouldn't be in the proto file with some annotations. This way the developer making the change is more aware of what has been changed/added (and it won't mess up the numbering of items in an enum or in a message). Newton mentioned the Roslyn compiler can provide some object representation or string representation of fields so maybe we can use that. 

I'll put this as a todo for a future PR. 

---
In reply to: [314339170](https://github.com/microsoft/BuildXL/pull/724#discussion_r314339170) [](ancestors = 314339170,313452127)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
```diff
@@ -162,7 +182,8 @@ public XLGToDBAnalyzerInner(AnalysisInput input) : base(input)
         /// <inheritdoc/>
         public override void Prepare()
         {
-            var accessor = KeyValueStoreAccessor.Open(storeDirectory: OutputDirPath);
+            string[] additionalColumns = { XldbDataStore.EventColumnFamilyName, XldbDataStore.PipColumnFamilyName, XldbDataStore.StaticGraphColumnFamilyName };
+            var accessor = KeyValueStoreAccessor.Open(storeDirectory: OutputDirPath, additionalColumns: additionalColumns);
```
jbayardo: In this use-case, where you're only writing in bulk to the DB, you may benefit from PrepareForBulkLoad(). See the [RocksDb FAQ](https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/StaticGraph.proto
sahilmgandhi: This just contains metadata about the piptable such as the number of reads/writes, the different stablekeys, and more. Perhaps I can rename it to PipTableMetaData or something similar

---
In reply to: [314336956](https://github.com/microsoft/BuildXL/pull/724#discussion_r314336956) [](ancestors = 314336956)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+                IsDisposed = pipTable.IsDisposed,
+                Reads = pipTable.Reads,
+                Writes = pipTable.Writes,
+                Count = pipTable.Count,
+                PageStreamsCount = pipTable.PageStreamsCount,
+                Size = pipTable.Size,
+                WritesMilliseconds = pipTable.WritesMilliseconds,
+                ReadsMilliseconds = pipTable.ReadsMilliseconds,
+                Used = pipTable.Used,
+                Alive = pipTable.Alive
```
mpysson: I don't think these internal implementation details of the PipTable structure are particularly useful to put into the database. Do you even need the PipTable structure at all?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+
+            xldbPipTable.StableKeys.AddRange(pipTable.StableKeys.Select(stableKey => stableKey.Value));
+            xldbPipTable.Keys.AddRange(pipTable.Keys.Select(key => key.Value));
+            xldbPipTable.DeserializationContexts.AddRange(pipTable.DeserializationContexts.Select(
+                context => new PipDeserializationContext() { Key = (PipQueryContext)context.Key, Value = context.Value }));
+            return xldbPipTable;
+        }
+
+        /// <nodoc />
+        public static Xldb.PipData ToPipData(this PipData pipData)
```
mpysson: >Xldb.PipData  [](start = 22, length = 13)

PipData is a bit of an internal implementation as well. We are abstracting file paths as strings. Should we do the same for PipData?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+        }
+
+        /// <nodoc />
+        public static Xldb.Pip ToPip(this Pip pip, DirectedGraph directedGraph)
+        {
+            var xldbPip = new Xldb.Pip()
+            {
+                SemiStableHash = pip.SemiStableHash,
+                FormattedSemiStableHash = pip.FormattedSemiStableHash,
+                ProcessAllowsUndeclaredSourceReads = pip.ProcessAllowsUndeclaredSourceReads,
```
mpysson: >ProcessAllowsUndeclaredSourceReads [](start = 16, length = 34)

This seems very specific to process pips. I wouldn't include it on Xldb.Pip

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+            {
+                xldbFileOrDirectoryArtifact.IsFile = true;
+                xldbFileOrDirectoryArtifact.FileArtifact = artifact.FileArtifact.ToFileArtifact(pathTable);
+            }
+
+            return xldbFileOrDirectoryArtifact;
+        }
+
+        /// <nodoc />
+        public static Xldb.Pip ToPip(this Pip pip, DirectedGraph directedGraph)
```
mpysson: >Xldb.Pip [](start = 22, length = 8)

I see there are specific pip objects for all relevant pip types (Process, CopyFile, ...) How do you see Xldb.Pip being used?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+        public static Xldb.CopyFile ToCopyFile(this CopyFile pip, PathTable pathTable, Xldb.Pip parentPip)
+        {
+            var xldbCopyFile = new Xldb.CopyFile
+            {
+                ParentPipInfo = parentPip,
+                Source = pip.Source.ToFileArtifact(pathTable),
+                Destination = pip.Destination.ToFileArtifact(pathTable),
+                OutputsMustRemainWritable = pip.OutputsMustRemainWritable,
+                Provenance = pip.Provenance.ToPipProvenance(),
+                PipType = (PipType)pip.PipType
```
mpysson: >PipType [](start = 16, length = 7)

Isn't this redundant based on the type?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+            if (pip.Tags.IsValid)
+            {
+                xldbProcessPip.Tags.AddRange(pip.Tags.Select(key => key.ToString()));
+            }
+
+            return xldbProcessPip;
+        }
+
+        /// <nodoc />
+        public static Xldb.HashSourceFile ToHashSourceFile(this HashSourceFile pip, PathTable pathTable, Xldb.Pip parentPip)
```
mpysson: >HashSourceFile [](start = 27, length = 14)

I'm not sure it's worth exposing HSF pips in the graph

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
sahilmgandhi: It is used as the ParentPipInfo object for each of the specific pips (since Protobuf doesn't support inheritance, but we still want this information stored in the specific pips)

---
In reply to: [314446991](https://github.com/microsoft/BuildXL/pull/724#discussion_r314446991) [](ancestors = 314446991)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/Events.proto
```diff
@@ -19,10 +20,22 @@ message EventCount{
...
 
+message EventCountByTypeKey{
+    ExecutionEventId EventTypeID = 1;
+}
+
+message EventCountByTypeValue{
+    map<uint32, int32> WorkerToCountMap = 1;
+
+    map<uint32, int32> WorkerToPayloadMap = 2;
+}
```
mpysson: Explicitly storing the event counts in the db feels very specific for that one analyzer. Is it worth storing it in the db? It doesn't feel very generic.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
sahilmgandhi: Good point, ill go ahead and remove it. 

---
In reply to: [314445602](https://github.com/microsoft/BuildXL/pull/724#discussion_r314445602) [](ancestors = 314445602)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/EnumsAndHelperStructs.proto
mpysson: Another thing to consider is that the proto schema should be pretty static so we don't make breaking changes for things like renaming fields or refactoring. Having an automatic tool to update the schema could be counter productive to that goal in some circumstances

---
In reply to: [314401935](https://github.com/microsoft/BuildXL/pull/724#discussion_r314401935) [](ancestors = 314401935,314339170,313452127)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
mpysson: Then we should make sure none of the fields like Provenance are overlapping

It doesn't promote consistency but I wonder if it would be a nicer experience to just push the useful fields from Pip onto Process and the othe rspecific pip types. I think some of those fields won't be to all pip types

---
In reply to: [314450315](https://github.com/microsoft/BuildXL/pull/724#discussion_r314450315) [](ancestors = 314450315,314446991)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+            return xldbFileOrDirectoryArtifact;
+        }
+
+        /// <nodoc />
+        public static Xldb.Pip ToPip(this Pip pip, DirectedGraph directedGraph)
+        {
+            var xldbPip = new Xldb.Pip()
+            {
+                SemiStableHash = pip.SemiStableHash,
+                FormattedSemiStableHash = pip.FormattedSemiStableHash,
```
mpysson: These are redundant with each other. Let's just store the numeric

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
sahilmgandhi: Are they not used anywhere else at all (any analyzers) or important at all? I put them in there since they are relatively lightweight and could have the potential of being used?(but maybe not)

---
In reply to: [314448106](https://github.com/microsoft/BuildXL/pull/724#discussion_r314448106) [](ancestors = 314448106)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+            {
+                ParentPipInfo = parentPip,
+                Module = pip.Module.Value.Value,
+                Identity = pip.Identity.ToString(),
+                ResolverKind = pip.ResolverKind.ToString(),
+                ResolverName = pip.ResolverName.ToString(),
+                Version = pip.Version.ToString(),
+                Location = new Xldb.LocationData()
+                {
+                    IsValid = pip.Location.IsValid,
```
mpysson: >IsValid [](start = 20, length = 7)

If protobuf generates classes we shouldn't need these IsValid bools to be serialized since we can just use null for unpopulated references

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
```diff
@@ -558,5 +548,479 @@ public static Xldb.Fingerprint ToFingerprint(this Fingerprint fingerprint)
...
+            if (pip.Tags.IsValid)
+            {
+                xldbWriteFile.Tags.AddRange(pip.Tags.Select(key => key.ToString()));
+            }
+
+            return xldbWriteFile;
+        }
+
+        /// <nodoc />
+        public static ProcessPip ToProcessPip(this Process pip, PathTable pathTable, Xldb.Pip parentPip)
```
mpysson: >ToProcessPip [](start = 33, length = 12)

General comment: You did a great job of faithfully of replicating every piece of data on the Process pip. This is great for completeness but I'm wondering if we should be more selective about exposing a smaller subset of properties for the sake of making it easier to keep the schema consistent.

An incomplete list of things I'm not sure we should lead with:
Metapips (Spec, module, value, HashSourceFile)
Priority
Warning/Error regex
ProducedPathIndependentOutputs
Container related data
...
Basically things that are more about the inner workings of the scheduler or obscure features


This would be a good thing to discuss with Oleksii. We can always add things later.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
sahilmgandhi: As discussed offline this can be done in a different PR since several places/functions would need to be changed that use ApplyBatch

---
In reply to: [314126225](https://github.com/microsoft/BuildXL/pull/724#discussion_r314126225) [](ancestors = 314126225)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
sahilmgandhi: Will look into this for a separate "OptimizeDB" pr :) 

---
In reply to: [314403390](https://github.com/microsoft/BuildXL/pull/724#discussion_r314403390) [](ancestors = 314403390)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
sahilmgandhi: It is relatively cheap (to insert and storage wise) and a single entry into the db. The most important things here are the stablekeys and keys that stored since this allows you to iterate over all pips. Thus I think some, but not all of this metadata could be immediately useful, and the others are "nicetohaves" 

---
In reply to: [314443512](https://github.com/microsoft/BuildXL/pull/724#discussion_r314443512) [](ancestors = 314443512)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
sahilmgandhi: Actually you are right. Wherever there is invalid, I can just not generate that class. Let me go ahead and change those.

---
In reply to: [314454672](https://github.com/microsoft/BuildXL/pull/724#discussion_r314454672) [](ancestors = 314454672)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/Events.proto
sahilmgandhi: It is not generic, and it really just helps 1 analyzer, but at the same time, its 13 elements stored into a db so it is "cheap" and could fall under the "interesting indexing" for the db :) 

---
In reply to: [314450465](https://github.com/microsoft/BuildXL/pull/724#discussion_r314450465) [](ancestors = 314450465)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
sahilmgandhi: Ordinarily this would be an overriding attribute due to inheritance. I'm going to go ahead and just remove it from the "parent pip" object since we don't need it in there. 

---
In reply to: [314447532](https://github.com/microsoft/BuildXL/pull/724#discussion_r314447532) [](ancestors = 314447532)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
sahilmgandhi: I removed overlapping (overridden) fields so that the specific pips contain the information and not the "parentPip" object. However I think this abstraction of having info like the semistable hash/pipid and the edges stored in this ParentObject make it easier to understand that these are fields found in pips of any type

---
In reply to: [314453819](https://github.com/microsoft/BuildXL/pull/724#discussion_r314453819) [](ancestors = 314453819,314450315,314446991)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbProtobufExtensions.cs
sahilmgandhi: Based on offline chat we have removed some things from the proto structs

---
In reply to: [314456761](https://github.com/microsoft/BuildXL/pull/724#discussion_r314456761) [](ancestors = 314456761)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/EnumsAndHelperStructs.proto
sahilmgandhi: Making a user story so we can discuss this at length for a future PR. 

---
In reply to: [314451213](https://github.com/microsoft/BuildXL/pull/724#discussion_r314451213) [](ancestors = 314451213,314401935,314339170,313452127)

dannyvv: One thing that might make it a bit more maintainable long term would be to split each enum/message in its own file.
For future archeolygy if we have to see why feelds were added / removed it will be easier to look in the history.

---
In reply to: [314551183](https://github.com/microsoft/BuildXL/pull/724#discussion_r314551183) [](ancestors = 314551183,314451213,314401935,314339170,313452127)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/Events.proto
dannyvv: I suspect we will get more examples where we can benefit from storing information in the database, either for a single analysis or for future. I.e. critical path computation for each node. Having that computed once and then stored int he db can benefit many analyzers.

Perhaps if we label these somehow as derived tables (either by name prefix, or placing them in a shared location, different assembly etc, don't have a good idea of what the best longterm model is for that.

We can set the trend and guidelines for others to follow.


---
In reply to: [314472749](https://github.com/microsoft/BuildXL/pull/724#discussion_r314472749) [](ancestors = 314472749,314450465)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -104,80 +117,270 @@ private IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+                database.TryGetValue(Encoding.ASCII.GetBytes(EventCountKey), out var eventCountObj);
+                return EventCount.Parser.ParseFrom(eventCountObj).Value;
+            });
+
+            if (!maybeFound.Succeeded)
+            {
+                maybeFound.Failure.Throw();
+            }
+
+            return maybeFound.Result;
```
dannyvv: Not needed now, but I suspect we are going to get a bunch more point queries or list queries.

Perhaps it would be usefull to create helper methods that contain this logic and the serialization and error handling so we don't have to do it over and over i.e.

int QueryInt32<T>(T message) where T : IMessage { ... }

and/or TryQueryInt32

dannyvv: Maybe worth doing this PR as I see `GetPipTableMetaData` :)

---
In reply to: [314812200](https://github.com/microsoft/BuildXL/pull/724#discussion_r314812200) [](ancestors = 314812200)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/EnumsAndHelperStructs.proto
sahilmgandhi: This is an interesting idea, but then there will be a bunch of import statements whenever we need to import 1 or more enums. The question then stands, would we also want to split each message into its own file, and would that lead to too much clutter with a user having to [possibly] maintain a bunch of files rather than a small handful?

---
In reply to: [314809335](https://github.com/microsoft/BuildXL/pull/724#discussion_r314809335) [](ancestors = 314809335,314551183,314451213,314401935,314339170,313452127)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
sahilmgandhi: Talked offline -> moving to separate PR.

---
In reply to: [314812365](https://github.com/microsoft/BuildXL/pull/724#discussion_r314812365) [](ancestors = 314812365,314812200)

Public/Src/Tools/Execution.Analyzer/BxlAnalyzer.dsc
```diff
@@ -43,6 +44,7 @@ namespace Execution.Analyzer {
             importFrom("BuildXL.Ide").Generator.dll,
             importFrom("BuildXL.Utilities").dll,
             importFrom("BuildXL.Utilities").Branding.dll,
+            importFrom("BuildXL.Utilities").Ipc.dll,
```
olkononenko: >mportFrom("BuildXL.Utilities").Ipc.dll, [](start = 13, length = 39)

why do you need this?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/BXLInvocationAnalyzer.cs
```diff
@@ -79,13 +81,22 @@ internal sealed class BXLInvocationAnalyzer : Analyzer
         /// </summary>
         public string OutputFilePath;
 
-        public BXLInvocationAnalyzer(AnalysisInput input) : base(input) { }
+        private Stopwatch m_stopWatch;
```
olkononenko: >private Stopwatch m_stopWatch; [](start = 8, length = 30)

this looks like some debug/test code - I don't think it should be a part of the codebase

Public/Src/Tools/Execution.Analyzer/BxlAnalyzer.dsc
sahilmgandhi: Good question, I think this may have been needed for some field before we nuked because it does not seem to be needed anymore.

---
In reply to: [314835632](https://github.com/microsoft/BuildXL/pull/724#discussion_r314835632) [](ancestors = 314835632)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/EnumsAndHelperStructs.proto
```diff
@@ -0,0 +1,1061 @@
...
+    // Since we are using ExecutionEventId as in the key and protobuf 
+    // does not serialize default values, the pattern matching was bugging out. 
+    // A solution is to have an "invalid" enum constant take position 0.
```
olkononenko: I think I commented re this in past but I want to bring it up again:
I think we might want to add that 'invlid = 0' to all objects. If one day we decide that we want to use something as a key, we'll need to move all the fields, and I'll be a breaking change. Alternatively, maybe create objects that are exclusively used as keys?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/BXLInvocationAnalyzer.cs
sahilmgandhi: I agree, i'll go ahead and remove it

---
In reply to: [314836322](https://github.com/microsoft/BuildXL/pull/724#discussion_r314836322) [](ancestors = 314836322)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/NewDumpPipAnalyzer.cs
```diff
@@ -0,0 +1,210 @@
...
+                        }
+                    }
+
+                    writer.WriteLine("Geting directory dependency information for Process Pip");
+
+                    var pipGraph = dataStore.GetPipGraphMetaData();
+                    var directories = new Stack<(DirectoryArtifact artifact, string path)>(
+                    ((ProcessPip)castedPip).DirectoryDependencies
+                        .Select(d => (artifact: d, path: d.Path.Value))
+                        .OrderByDescending(tupple => tupple.path));
```
olkononenko: indentation

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/StaticGraph.proto
```diff
@@ -0,0 +1,121 @@
...
+    repeated uint32 StableKeys = 6;
```
olkononenko: >6 [](start = 33, length = 1)

nit

olkononenko: >repeated uint32 StableKeys = 6; [](start = 4, length = 31)

consider moving this into PipGraph

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -19,23 +21,27 @@ public sealed class XldbDataStore : IDisposable
         /// </summary>
         private KeyValueStoreAccessor Accessor { get; set; }
         private Dictionary<ExecutionEventId, MessageParser> m_eventParserDictionary = new Dictionary<ExecutionEventId, MessageParser>();
+        private Dictionary<PipType, MessageParser> m_pipParserDictionary = new Dictionary<PipType, MessageParser>();
         public const string EventCountKey = "EventCount";
```
olkononenko: >p [](start = 8, length = 1)

add empty line

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -19,23 +21,27 @@ public sealed class XldbDataStore : IDisposable
         /// </summary>
         private KeyValueStoreAccessor Accessor { get; set; }
         private Dictionary<ExecutionEventId, MessageParser> m_eventParserDictionary = new Dictionary<ExecutionEventId, MessageParser>();
+        private Dictionary<PipType, MessageParser> m_pipParserDictionary = new Dictionary<PipType, MessageParser>();
         public const string EventCountKey = "EventCount";
 
+        public const string EventColumnFamilyName = "Event";
```
olkononenko: nuke line

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -65,28 +71,35 @@ public sealed class XldbDataStore : IDisposable
...
         }
 
         /// <summary>
         /// Gets all the events of a certain type from the DB
         /// </summary>
-        private IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
+        /// <returns>List of events, empty if no such event exists</returns>
+        private IEnumerable<IMessage> GetEventsByType(ExecutionEventId eventTypeID)
         {
             Contract.Assert(Accessor != null, "XldbStore must be initialized via OpenDatastore first");
```
olkononenko: >Assert [](start = 21, length = 6)

Requires

olkononenko:
>                     }

move to the outside of the loop

---
Refers to: Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs:108 in f0977ab. [](commit_id = f0977ab40b408e58b98bef62790ba0a4c53e7dad, deletion_comment = False)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -104,80 +117,270 @@ private IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+            {
+                EventTypeID = eventTypeID,
+                PipId = pipID
+            };
+
+            var maybeFound = Accessor.Use(database =>
+            {
+                foreach (var kvp in database.PrefixSearch(eventQuery.ToByteArray(), EventColumnFamilyName))
+                {
+                    if (m_eventParserDictionary.TryGetValue(eventTypeID, out var parser))
```
olkononenko: >  if (m_eventP [](start = 18, length = 14)

same comment, pls go through other similar places

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XldbDataStore.cs
```diff
@@ -104,80 +117,270 @@ private IEnumerable<string> GetEventsByType(ExecutionEventId eventTypeID)
...
+            Contract.Assert(Accessor != null, "XldbStore must be initialized via OpenDatastore first");
+
+            IMessage foundPip = null;
+
+            var pipQuery = new PipQuerySemiStableHash()
+            {
+                SemiStableHash = semiStableHash
+            };
+
+            var outPipType = (PipType) 0;
```
olkononenko: >var outPipType = (PipType) 0; [](start = 12, length = 29)

`PipType outPipType;`

Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections;
```
olkononenko: ???

PR #729, opened 2019-08-13
Fix logging for cache

JuanCarlosGI:
The implementation of EtwFileLog overrides WriteLine.

Not calling WriteLine meant that EtwFileLog implementation, on which cache telemetry depends on would never be called.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Logging/FileLog.cs
+++ b/Public/Src/Cache/ContentStore/Library/Logging/FileLog.cs
@@ -179,7 +179,7 @@ namespace BuildXL.Cache.ContentStore.Logging
                     message
                 );
 
-            WriteLineInternal(severity, line);
+            WriteLine(severity, severity.ToString(), line);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Library/Logging/FileLog.cs
```diff
@@ -179,7 +179,7 @@ public virtual void Write(DateTime dateTime, int threadId, Severity severity, st
                     message
                 );
 
-            WriteLineInternal(severity, line);
+            WriteLine(severity, severity.ToString(), line);
```
dannyvv: just FYI on full framework this does extra boxing. But since you are diong advanced string formatting already I assume this is not too perf-critical

almiliMSFT: Maybe leave a comment saying that it's important to call `WriteLine` and not `WriteLineInternal` because subclasses may override `WriteLine` to add some extra processing.

JuanCarlosGI: Just realized that were already capturing the severityName in a non-boxing way. Nice catch!

---
In reply to: [313590070](https://github.com/microsoft/BuildXL/pull/729#discussion_r313590070) [](ancestors = 313590070)

PR #731, opened 2019-08-13
Add timed full range compaction for RocksDbContentLocationDatabase

jbayardo:
We have observed in production that database size grows over time even when the number of keys decrease or stay the same, seeing dramatic effects such as a 2x or 3x size increase. Investigation has shown this to be likely caused to deleted keys/old versions that sink into lower levels of the LSM tree and don't ever get compacted.

This PR adds a timer to the `RocksDbContentLocationDatabase` that runs a full compaction over the entire key range every few hours, thereby forcing the size of the DB to stay relatively small.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
@@ -120,5 +120,11 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// Gets whether the database is cleared on initialization. (Defaults to true because the standard use case involves restoring from checkpoint after initialization)
         /// </summary>
         public bool CleanOnInitialize { get; set; } = true;
+
+        /// <summary>
+        /// Time between full range compactions. These help keep the size of the DB instance down to a minimum.
+        /// Required because of our workload tends to generate a lot of cruft.
+        /// </summary>
+        public TimeSpan FullRangeCompactionInterval { get; set; } = TimeSpan.FromHours(6);
     }
 }
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
@@ -123,7 +123,9 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 
         /// <summary>
         /// Time between full range compactions. These help keep the size of the DB instance down to a minimum.
-        /// Required because of our workload tends to generate a lot of cruft.
+        /// 
+        /// Required because of our workload tends to generate a lot of short-lived entries, which clutter the deeper
+        /// levels of the RocksDB LSM tree.
         /// </summary>
         public TimeSpan FullRangeCompactionInterval { get; set; } = TimeSpan.FromHours(6);
     }
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
@@ -40,6 +40,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         private StoreSlot _activeSlot = StoreSlot.Slot1;
         private string _storeLocation;
         private readonly string _activeSlotFilePath;
+        private Timer _compactionTimer;
 
         private static readonly byte[] EmptyBytes = CollectionUtilities.EmptyArray<byte>();
 
@@ -90,13 +91,28 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// <inheritdoc />
         protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
         {
+            lock (this)
+            {
+                _compactionTimer?.Dispose();
+            }
+
             _keyValueStore?.Dispose();
+
             return base.ShutdownCoreAsync(context);
         }
 
         /// <inheritdoc />
         protected override BoolResult InitializeCore(OperationContext context)
         {
+            if (_configuration.FullRangeCompactionInterval != Timeout.InfiniteTimeSpan)
+            {
+                _compactionTimer = new Timer(
+                    _ => FullRangeCompaction(context),
+                    null,
+                    IsGarbageCollectionEnabled ? _configuration.FullRangeCompactionInterval : Timeout.InfiniteTimeSpan,
+                    Timeout.InfiniteTimeSpan);
+            }
+
             var result = Load(context, GetActiveSlot(context.TracingContext), clean: _configuration.CleanOnInitialize);
             if (result && _configuration.TestInitialCheckpointPath != null)
             {
@@ -104,6 +120,49 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             }
 
             return result;
+        }
+
+        /// <inheritdoc />
+        public override void SetDatabaseMode(bool isDatabaseWriteable)
+        {
+            if (IsGarbageCollectionEnabled != isDatabaseWriteable)
+            {
+                var nextCompactionTimeSpan = isDatabaseWriteable ? _configuration.FullRangeCompactionInterval : Timeout.InfiniteTimeSpan;
+                _compactionTimer?.Change(nextCompactionTimeSpan, Timeout.InfiniteTimeSpan);
+            }
+
+            base.SetDatabaseMode(isDatabaseWriteable);
+        }
+
+        private void FullRangeCompaction(OperationContext context)
+        {
+            lock (this)
+            {
+                if (ShutdownStarted)
+                {
+                    return;
+                }
+            }
+
+            _keyValueStore.Use(store =>
+            {
+                foreach (var columnFamilyName in new[] { "default", nameof(Columns.ClusterState), nameof(Columns.Metadata) })
+                {
+                    var result = context.PerformOperation(Tracer, () =>
+                    {
+                        store.CompactRange((byte[])null, null, columnFamilyName: columnFamilyName);
+                        return BoolResult.Success;
+                    }, extraStartMessage: $"ColumnFamily={columnFamilyName}").ThrowIfFailure();
+                }
+            }).ThrowOnError();
+
+            lock (this)
+            {
+                if (!ShutdownStarted)
+                {
+                    _compactionTimer?.Change(_configuration.FullRangeCompactionInterval, Timeout.InfiniteTimeSpan);
+                }
+            }
         }
 
         private BoolResult Load(OperationContext context, StoreSlot activeSlot, bool clean = false)
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
@@ -21,7 +21,6 @@ using BuildXL.Engine.Cache;
 using BuildXL.Engine.Cache.KeyValueStores;
 using BuildXL.Native.IO;
 using BuildXL.Utilities;
-using BuildXL.Utilities.Collections;
 using BuildXL.Utilities.Threading;
 using AbsolutePath = BuildXL.Cache.ContentStore.Interfaces.FileSystem.AbsolutePath;
 using Unit = BuildXL.Utilities.Tasks.Unit;
@@ -41,8 +40,6 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         private string _storeLocation;
         private readonly string _activeSlotFilePath;
         private Timer _compactionTimer;
-
-        private static readonly byte[] EmptyBytes = CollectionUtilities.EmptyArray<byte>();
 
         private enum StoreSlot
         {
@@ -91,7 +88,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// <inheritdoc />
         protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
         {
-            lock (this)
+            lock (ShutdownLock)
             {
                 _compactionTimer?.Dispose();
             }
@@ -136,7 +133,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 
         private void FullRangeCompaction(OperationContext context)
         {
-            lock (this)
+            lock (ShutdownLock)
             {
                 if (ShutdownStarted)
                 {
@@ -144,6 +141,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                 }
             }
 
+            // We can safely ignore the result, as any relevant errors will be logged by the inner operations
             _keyValueStore.Use(store =>
             {
                 foreach (var columnFamilyName in new[] { "default", nameof(Columns.ClusterState), nameof(Columns.Metadata) })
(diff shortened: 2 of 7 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
```diff
@@ -120,5 +120,11 @@ public RocksDbContentLocationDatabaseConfiguration(AbsolutePath storeLocation)
         /// Gets whether the database is cleared on initialization. (Defaults to true because the standard use case involves restoring from checkpoint after initialization)
         /// </summary>
         public bool CleanOnInitialize { get; set; } = true;
+
+        /// <summary>
+        /// Time between full range compactions. These help keep the size of the DB instance down to a minimum.
+        /// Required because of our workload tends to generate a lot of cruft.
```
dannyvv: >cruft [](start = 72, length = 5)

short lived entries !?!? :D

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -90,13 +91,28 @@ public RocksDbContentLocationDatabase(IClock clock, RocksDbContentLocationDataba
         /// <inheritdoc />
         protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
         {
+            lock (this)
```
dannyvv: >this [](start = 18, length = 4)

In .net it is highly discourage to ever lock on this or any other publically exposed object (since anybody can lock on it) 

[https://stackoverflow.com/questions/251391/why-is-lockthis-bad](https://stackoverflow.com/questions/251391/why-is-lockthis-bad)

Would you mind adding a local object 
`private object m_lock = new object()`
and lock on that?

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -106,6 +122,49 @@ protected override BoolResult InitializeCore(OperationContext context)
...
+                var nextCompactionTimeSpan = isDatabaseWriteable ? _configuration.FullRangeCompactionInterval : Timeout.InfiniteTimeSpan;
+                _compactionTimer?.Change(nextCompactionTimeSpan, Timeout.InfiniteTimeSpan);
+            }
+
+            base.SetDatabaseMode(isDatabaseWriteable);
+        }
+
+        private void FullRangeCompaction(OperationContext context)
+        {
+            lock (this)
```
dannyvv: similar to above

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -106,6 +122,49 @@ protected override BoolResult InitializeCore(OperationContext context)
...
+                {
+                    var result = context.PerformOperation(Tracer, () =>
+                    {
+                        store.CompactRange((byte[])null, null, columnFamilyName: columnFamilyName);
+                        return BoolResult.Success;
+                    }, extraStartMessage: $"ColumnFamily={columnFamilyName}").ThrowIfFailure();
+                }
+            }).ThrowOnError();
+
+            lock (this)
```
dannyvv: >this [](start = 18, length = 4)

similar to above

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -90,13 +91,28 @@ public RocksDbContentLocationDatabase(IClock clock, RocksDbContentLocationDataba
...
             return base.ShutdownCoreAsync(context);
         }
 
         /// <inheritdoc />
         protected override BoolResult InitializeCore(OperationContext context)
         {
+            if (_configuration.FullRangeCompactionInterval != Timeout.InfiniteTimeSpan)
+            {
+                _compactionTimer = new Timer(
+                    _ => FullRangeCompaction(context),
```
dannyvv: Should we add a logger for when this fails?
Not sure if the timer kills the entire process, if the timer stops functioning on errors and not perform cleanup anymore, etc...

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -90,13 +88,28 @@ public RocksDbContentLocationDatabase(IClock clock, RocksDbContentLocationDataba
...
+                    null,
+                    IsGarbageCollectionEnabled ? _configuration.FullRangeCompactionInterval : Timeout.InfiniteTimeSpan,
```
SergeyTeplyakov: This looks strange to me, why we need to instantiate timer if gc is disabled? In this case we'll create a timer that will never be fired.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -106,6 +119,55 @@ protected override BoolResult InitializeCore(OperationContext context)
...
+            if (IsGarbageCollectionEnabled != isDatabaseWriteable)
+            {
+                var nextCompactionTimeSpan = isDatabaseWriteable ? _configuration.FullRangeCompactionInterval : Timeout.InfiniteTimeSpan;
+                _compactionTimer?.Change(nextCompactionTimeSpan, Timeout.InfiniteTimeSpan);
+            }
+
+            base.SetDatabaseMode(isDatabaseWriteable);
+        }
+
+        private void FullRangeCompaction(OperationContext context)
```
SergeyTeplyakov: Need to trace this.

We need to know the duration of such a critical operation like compaction.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -106,6 +119,55 @@ protected override BoolResult InitializeCore(OperationContext context)
             return result;
         }
 
+        /// <inheritdoc />
+        public override void SetDatabaseMode(bool isDatabaseWriteable)
+        {
+            if (IsGarbageCollectionEnabled != isDatabaseWriteable)
```
SergeyTeplyakov: I don't really like the code that deals with _compactionTimer.

For instance, why in this case there is no checks for IsGarbageCollectionEnabled?
Should the logic be in initializeCore or only in SetDatabaseMode?

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -106,6 +119,55 @@ protected override BoolResult InitializeCore(OperationContext context)
...
+                        break;
+                    }
+                }
+            }).ToBoolResult().IgnoreFailure();
+
+            lock (ShutdownLock)
+            {
+                if (!ShutdownStarted)
+                {
+                    _compactionTimer?.Change(_configuration.FullRangeCompactionInterval, Timeout.InfiniteTimeSpan);
```
SergeyTeplyakov: I don't think you need to use locks here, because ShutdownStarted is backed by cancellation token source that is thread safe.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: This is exactly the same logic we use for all other timers. The reason for it is that a machine could be promoted to master, in which case it needs to start using the timer.

---
In reply to: [313643144](https://github.com/microsoft/BuildXL/pull/731#discussion_r313643144) [](ancestors = 313643144)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
SergeyTeplyakov: The same is true for both checks: at the beginning and at the end of the method.

---
In reply to: [313644899](https://github.com/microsoft/BuildXL/pull/731#discussion_r313644899) [](ancestors = 313644899)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: Because we use _isDatabaseWriteable directly. IsGarbageCollectionEnabled is exactly the same as "Is the database in writeable mode". This logic needs to be here, because this is called whenever a machine is elected as master

---
In reply to: [313644408](https://github.com/microsoft/BuildXL/pull/731#discussion_r313644408) [](ancestors = 313644408)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
SergeyTeplyakov: Timer restart should be inside try/finally block.

We don't want to stop the collection if the exception will occur here.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: This is being traced below, when effectively compacting each one of the column families.

---
In reply to: [313643384](https://github.com/microsoft/BuildXL/pull/731#discussion_r313643384) [](ancestors = 313643384)

SergeyTeplyakov: Nope. It wasn't. It was traced per family. We need overall time.

---
In reply to: [313645611](https://github.com/microsoft/BuildXL/pull/731#discussion_r313645611) [](ancestors = 313645611,313643384)

SergeyTeplyakov: I don't think it is that important to know per family. But I rather have one global time.

---
In reply to: [313645851](https://github.com/microsoft/BuildXL/pull/731#discussion_r313645851) [](ancestors = 313645851,313645611,313643384)

jbayardo: Overall time is meaningless for this operation, as compaction time varies widely between each of the column families

---
In reply to: [313645851](https://github.com/microsoft/BuildXL/pull/731#discussion_r313645851) [](ancestors = 313645851,313645611,313643384)

SergeyTeplyakov: why it meaningless? For casaas with specific configuration it would be roughtly the same. No?

---
In reply to: [313646240](https://github.com/microsoft/BuildXL/pull/731#discussion_r313646240) [](ancestors = 313646240,313645851,313645611,313643384)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
SergeyTeplyakov: This is not entirely true. In some cases we actually check that a feature is disable as well:

if (_configuration.ContentCacheEnabled && _configuration.CacheFlushingMaximumInterval != Timeout.InfiniteTimeSpan)
            {
  /// timer creation
}


---
In reply to: [313644959](https://github.com/microsoft/BuildXL/pull/731#discussion_r313644959) [](ancestors = 313644959,313643144)

SergeyTeplyakov: And the logic is actually pretty inconsistent.

I would say that we should follow one pattern or another. But right now it is not clear why one configuration option is more important than another.

---
In reply to: [313646774](https://github.com/microsoft/BuildXL/pull/731#discussion_r313646774) [](ancestors = 313646774,313644959,313643144)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: According to documentation, Timer.Change can only throw if the timer has already been disposed of or if parameters are wrong. The first condition should never happen due to the check for shutdown, the others would cause a failure much earlier.

---
In reply to: [313645457](https://github.com/microsoft/BuildXL/pull/731#discussion_r313645457) [](ancestors = 313645457)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
SergeyTeplyakov: Here is the case:

IsGarbagleCollectionEnabled is false.
We call SetDatabaseMode(true) for master.
In this case we'll enable the timer even though the GC is disabled.

---
In reply to: [313645345](https://github.com/microsoft/BuildXL/pull/731#discussion_r313645345) [](ancestors = 313645345,313644408)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: In this particular piece of code, the same thing happens: if the full range compaction interval is infinite, it will automatically be disabled. The end result is equivalent to disabling the feature, and is the same thing we do with garbage collection.

---
In reply to: [313646774](https://github.com/microsoft/BuildXL/pull/731#discussion_r313646774) [](ancestors = 313646774,313644959,313643144)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
dannyvv: The problem that I see is that `_keyvaluestore.Use` can throw, or the compation itself

---
In reply to: [313647176](https://github.com/microsoft/BuildXL/pull/731#discussion_r313647176) [](ancestors = 313647176,313645457)

jbayardo: if _keyValueStore.Use (which is just taking a lock) throws, everything will fail, not just this function, so there's nothing we can do.

If compaction fails, the inner operation ("result") will fail, which means we will trace and break.

---
In reply to: [313649456](https://github.com/microsoft/BuildXL/pull/731#discussion_r313649456) [](ancestors = 313649456,313647176,313645457)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
```diff
@@ -120,5 +120,13 @@ public RocksDbContentLocationDatabaseConfiguration(AbsolutePath storeLocation)
...
         /// </summary>
         public bool CleanOnInitialize { get; set; } = true;
+
+        /// <summary>
+        /// Time between full range compactions. These help keep the size of the DB instance down to a minimum.
+        /// 
+        /// Required because of our workload tends to generate a lot of short-lived entries, which clutter the deeper
+        /// levels of the RocksDB LSM tree.
+        /// </summary>
+        public TimeSpan FullRangeCompactionInterval { get; set; } = TimeSpan.FromHours(6);
```
ljcollins25: >FullRangeCompactionInterval [](start = 24, length = 27)

Maybe add a compaction epoch so we can force this.

PR #732, opened 2019-08-13
Add support for Azure devops Optimized output

dannyvv:
This will add a flag to BuildXL that will trigger it to improve the console output to include special Azure Dev Ops markers to give hints in the log outputs. I.e. markers like:
[Logging commands](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash)

It also reduces the number of multiline progress reports to make the text easier to read as flat-text.

### DIFF
--- a/Public/Src/App/Bxl/ConsoleEventListener.cs
+++ b/Public/Src/App/Bxl/ConsoleEventListener.cs
@@ -173,7 +173,7 @@ namespace BuildXL
         /// Maximum number of concurrently executing pips to render in Fancy Console view.
         /// </param>
         /// <param name="optimizeForAzureDevOps">
-        /// Whether console output should e optimized for Azure DevOps output.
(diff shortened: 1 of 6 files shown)


Public/Src/App/Bxl/ConsoleEventListener.cs
```diff
@@ -156,6 +172,9 @@ public sealed class ConsoleEventListener : FormattingEventListener
         /// <param name="maxStatusPips">
         /// Maximum number of concurrently executing pips to render in Fancy Console view.
         /// </param>
+        /// <param name="optimizeForAzureDevOps">
+        /// Whether console output should e optimized for Azure DevOps output.
```
almiliMSFT:
```suggestion
        /// Whether console output should be optimized for Azure DevOps output.
```

PR #733, opened 2019-08-14
Add fast path to put file which avoids locking.

ljcollins25:


Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -884,6 +886,48 @@ private static bool ShouldAttemptHardLink(AbsolutePath contentPath, FileAccessMo
...
+                                        realizationMode,
+                                        contentHash,
+                                        fileInfo,
+                                        fastPath: true);
+
+                    if (placeLinkResult == CreateHardLinkResult.Success)
+                    {
+                        return new PutResult(contentHash, fileInfo.FileSize)
+                        {
+                            Diagnostics = "FastPath"
```
SergeyTeplyakov: I'm not sure we'll print this for successful case.

Public/Src/Cache/ContentStore/Test/Stores/FileSystemContentStoreInternalPutFileTests.cs
```diff
@@ -110,6 +112,61 @@ public Task PutFilePins()
             });
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public Task PutSameContentManyTimesTest(bool useRedundantPutFileShortcut)
```
SergeyTeplyakov: How long does it take to run this test? Just curious, because in some cases stress tests are taking relatively long time.

ljcollins25: A few seconds.

PR #736, opened 2019-08-14
Add gating for PutFile in DistributedContentSession

jbayardo:


Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -83,10 +89,25 @@ private enum Counters
...
+            });
+        }
+
+        private Task<TResult> PerformPutFileGatedOperationAsync<TResult>(OperationContext operationContext, Func<Task<TResult>> func)
+        {
+            return _putFileGate.GatedOperationAsync((timeWaiting) =>
+            {
+                if (timeWaiting > Settings.PutFileWaitWarning)
+                {
+                    Tracer.Warning(operationContext, $"Spent {timeWaiting} waiting for PutFile gate, exceeding deadline of {Settings.PutFileWaitWarning}");
```
SergeyTeplyakov: I think it should be Info, not Warning, IMO.
I'm afraid of polluting the warning logs with all this...

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -83,10 +89,25 @@ private enum Counters
             UrgencyHint urgencyHint,
             Counter retryCounter)
         {
-            return PutCoreAsync(
-                operationContext,
-                (decoratedStreamSession, wrapStream) => decoratedStreamSession.PutFileAsync(operationContext, path, contentHash, realizationMode, operationContext.Token, urgencyHint, wrapStream),
-                session => session.PutFileAsync(operationContext, contentHash, path, realizationMode, operationContext.Token, urgencyHint));
+            return PerformPutFileGatedOperationAsync(operationContext, () => {
+                return PutCoreAsync(
```
SergeyTeplyakov: >PutCoreAsync [](start = 23, length = 12)

I think it make sense to have a comment here explaining that we intentionally not gating PutStream operations because we don't expect a high number of them at the same time.

PR #738, opened 2019-08-15
XLG Debugger for VsCode

almiliMSFT:
[AB#1581992](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1581992)


Public/Src/IDE/Debugger/DScriptDebugerRenderer.cs
```diff
@@ -0,0 +1,270 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
almiliMSFT: This is pretty much DScript-specific stuff extracted from the old `Renderer.cs` into a separate file (i.e., no effective changes here)

Public/Src/IDE/Debugger/Renderer.cs
```diff
@@ -9,21 +9,17 @@
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
-using BuildXL.FrontEnd.Script.Ambients;
```
almiliMSFT: Modulo some trivial refactoring, this file is also pretty much unchanged except that DScript-specific bits were extracted into `DScriptDebuggerRenderer.cs`.

Public/Src/IDE/Debugger/RendererAST.cs
```diff
@@ -4,10 +4,12 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
+using System.Linq;
```
almiliMSFT: Just making some things lazy here...

sahilmgandhi:
Public/Src/FrontEnd/Script/Evaluator/IDisplayStackTraceEntry.cs seems to be an empty file. Any reason for this?

Public/Src/IDE/Debugger/DebuggerState.cs
```diff
@@ -66,7 +82,7 @@ public DebuggerState(PathTable pathTable, LoggingContext loggingContext, Logger
         /// <returns>List of removed ThreadId -> ThreadState key value pairs.</returns>
         public IReadOnlyDictionary<int, ThreadState> ClearStoppedThreads()
         {
-            lock (this)
+            lock (m_lock)
```
sahilmgandhi: >lock (m_lock) [](start = 12, length = 13)

What is the reason to lock this "object" m_lock instead of locking "this" as was done before?

Public/Src/IDE/VsCode/client/src/xlg-status.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+function splitColumns(line: string): string[] {
+    return line.split(",").map(s => s.trim());
+}
+
+function convertBxlTimeToSeconds(bxlTimeValue: string) {
+    if (bxlTimeValue.startsWith("["))
+        bxlTimeValue = bxlTimeValue.substring(1);
+    if (bxlTimeValue.endsWith("]"))
```
sahilmgandhi: It seems like in other places we have been surrounding if/else statements with brackets even if there is only 1 line in the statement (ie line 88 or 121) . Should we standardize that everywhere or does TS have some different formatting rules?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/XlgDebuggerState.cs
```diff
@@ -0,0 +1,386 @@
...
+
+        /// <nodoc />
+        public ObjectInfo Render(Renderer renderer, object ctx, object obj)
+        {
+
+            switch (obj)
+            {
+                case AbsolutePath p:               return new ObjectInfo(p.ToString(PathTable));
+                case Result result:                return renderer.GetObjectInfo(ctx, result.IsScalar ? result.Value.First() : result.Value.ToArray());
+                case Function func:                return new ObjectInfo("fun");
```
sahilmgandhi: Nit: Should these be all on new lines like how the "return null" is on a new line on line 207

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/JPath.cs
```diff
@@ -0,0 +1,138 @@
...
+            var lexer = new JPathLexer(new AntlrInputStream(str));
+            var parser = new JPathParser(new CommonTokenStream(lexer));
+            var listener = new JPathListener();
+            parser.AddErrorListener(listener);
+            var exprCst = parser.expr();
+
+            if (listener.HasErrors)
+            {
+                return new Failure<string>("Syntex error: " + listener.FirstError);
+            }
```
sahilmgandhi: nit: Syntax*. Same with Line 110

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/JPath.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            var exprAst = exprCst.Accept(new AstConverter());
+            var value = evaluator.Eval(exprAst);
+
+            return new[] { "hi", exprAst.Print(), value.Value.ToString() };
+        }
```
sahilmgandhi: Why are we returning the string "hi" here?

almiliMSFT: Very good question :)

This whole method is unused and should be deleted.

Thanks for spotting.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/XlgDebuggerState.cs
almiliMSFT: I like it this way better.  And this is what I used in other switch statements.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/JPath.cs
almiliMSFT: will fix...

Public/Src/IDE/Debugger/RendererAST.cs
sahilmgandhi: Just so I'm on the same track, the Lazyness enabled here was what was demoed yesterday where not everything loaded if there were thousands of pips or files that were in the logs right?

Public/Src/IDE/Debugger/DebuggerState.cs
almiliMSFT: Best practices recommend to use a private object for such locking.  The idea is that if you lock on `this`, and the `this` object is accessible to others (which it always is), then someone else can also use it for locking and potentially cause deadlocks.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/gen.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+java \
```
dannyvv: Since we do not use antlr as a nuget or npm package the automated dependency tracker is not aware.

Could you manually register usage of antlr in component covernance before checking in?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/JPath.g4
```diff
@@ -0,0 +1,149 @@
...
+    : Token=(CONCAT | INTERSECT) ;
+
+anyBinaryOp
+    : intBinaryOp
+    | boolBinaryOp
+    | logicBinaryOp
+    | arrayBinaryOp
+    ;
+
+intExpr
```
dannyvv: surprised you haven't needed string expressions yet, how do you search for filenames?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/JPath.g4
```diff
@@ -0,0 +1,149 @@
...
+expr
+    : '$'                                             #RootExpr
+    | Var=VarId                                       #VarExpr
+    | Sub=selector                                    #SelectorExpr
+    | Lit=literal                                     #LiteralExpr
+    | Lhs=expr '.' Selector=selector                  #MapExpr
+    | Lhs=expr '[' Filter=logicExpr ']'               #FilterExpr
+    | Lhs=expr '[' Index=intExpr ']'                  #IndexExpr
+    | Lhs=expr '[' Begin=intExpr '..' End=intExpr ']' #RangeExpr
+    | '#' Sub=expr                                    #CardinalityExpr
```
dannyvv: >CardinalityExpr [](start = 55, length = 15)

This one might need a bit more explanation

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/JPath.g4
```diff
@@ -0,0 +1,149 @@
...
+    | Func=expr '(' Args+=expr (',' Args+=expr)* ')'  #FuncAppExprParen
+    | Func=expr Arg=expr                              #FuncAppExpr
```
dannyvv: Having space as function application seems very fragile and hard to learn and read. I still get confused about powershells sometimes parens, sometimes space argument separation. Should we just keep it consistent and straight-forward with just parens?

almiliMSFT: sounds good, will change

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/JPath.g4
almiliMSFT: I'll add a comment

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/JPath.g4
almiliMSFT: there are string literals and regex literals and there is the `~` operator which is a string/regex match.

But yes, in principle there should be either some string operators or the same can be done with string functions

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XlgDebugger/JPath/gen.sh
almiliMSFT: We still use the antlr runtime from Nuget.   I'll double check if that's the same jar file as the one used for generating parser/lexer and if it is not I'll register this one manually

Public/Src/IDE/Debugger/RendererAST.cs
almiliMSFT: Yes, it (among other things) contributes to that effect.

Public/Src/IDE/VsCode/client/src/xlg-status.ts
almiliMSFT: I don't think we have a strict standard for writing .ts files, so I'll leave it as is.

PR #741, opened 2019-08-15
Strengthen the pattern condition for identifying RC temp files

narasamdya:
DFA occurs on `d:\dbs\el\esi\obj\x86fre\termsrv\rdms\server\service\objfre\i386\rdms.lce`

The directory `d:\dbs\el\esi\obj\x86fre\termsrv\rdms\server\service\objfre\i386` is a shared opaque output directory. 

The write to that file by RC was ignored by sandboxed process reporting because of a weak identification pattern. Thus, subsequent reads/probes to that file by the same pip are considered access violation.

[AB#1584975](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1584975)

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -2571,7 +2571,8 @@ namespace BuildXL.Processes
             CCRewrite = 8,
             WinDbg = 9,
             XAMLWrapper = 11,
-            Mt = 12
+            Mt = 12,
+            BuildXLTest = 13,
         }
 
         private static readonly Dictionary<string, SpecialProcessKind> s_specialTools = new Dictionary<string, SpecialProcessKind>(StringComparer.OrdinalIgnoreCase)
@@ -2591,6 +2592,10 @@ namespace BuildXL.Processes
             ["windbg.exe"] = SpecialProcessKind.WinDbg,
             ["tool.xamlcompilerwrapper"] = SpecialProcessKind.XAMLWrapper,
             ["tool.xamlcompilerwrapper.exe"] = SpecialProcessKind.XAMLWrapper,
+
+            // Tool use for testing.
+            ["test.buildxl.executables.testprocess"] = SpecialProcessKind.BuildXLTest,
+            ["test.buildxl.executables.testprocess.exe"] = SpecialProcessKind.BuildXLTest,
 
             // TODO: deprecate this.
             ["cccheck"] = SpecialProcessKind.CCCheck,
@@ -2619,6 +2624,41 @@ namespace BuildXL.Processes
             return s_specialTools.TryGetValue(toolName.ToLowerInvariant(), out SpecialProcessKind kind) ? kind : SpecialProcessKind.NotSpecial;
         }
 
+        private static bool StringLooksLikeBuildXLTestTempFile(string fileName)
+        {
+            if (!fileName.EndsWith(".tmp", StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            int beginCharIndex = fileName.LastIndexOf('\\');
+
+            if (beginCharIndex == -1 || beginCharIndex + 3 >= fileName.Length)
+            {
+                return false;
+            }
+
+            char c1 = fileName[beginCharIndex + 1];
+            if (c1.ToUpperInvariantFast() != 'B')
+            {
+                return false;
+            }
+
+            char c2 = fileName[beginCharIndex + 2];
+            if (c2.ToUpperInvariantFast() != 'X')
+            {
+                return false;
+            }
+
+            char c3 = fileName[beginCharIndex + 3];
+            if (c3.ToUpperInvariantFast() != 'L')
+            {
+                return false;
+            }
+
+            return true;
+        }
+
         private static bool StringLooksLikeRCTempFile(string fileName)
         {
             int len = fileName.Length;
@@ -2645,6 +2685,13 @@ namespace BuildXL.Processes
                 return false;
             }
 
+            char c4 = fileName[len - 4];
+            if (c4.ToUpperInvariantFast() == '.')
+            {
+                // RC's temp files have no extension.
+                return false;
+            }
+
             return true;
         }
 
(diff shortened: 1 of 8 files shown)


Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2619,6 +2624,41 @@ private SpecialProcessKind GetProcessKind(AbsolutePath processOverride)
...
+            char c3 = fileName[beginCharIndex + 3];
+            if (c3.ToUpperInvariantFast() != 'L')
+            {
+                return false;
+            }
+
+            return true;
+        }
+
         private static bool StringLooksLikeRCTempFile(string fileName)
```
mpysson: >StringLooksLikeRCTempFile [](start = 28, length = 25)

I'm really not a fan of these special cased exceptions in SandboxedProcessPipExecutor. You're going the right direction in making this one less likely to be exercised but I'd love to move more in the direction of not utilizing them at all. For shared opaques, it seems like we shouldn't really even need them

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2619,6 +2624,41 @@ private SpecialProcessKind GetProcessKind(AbsolutePath processOverride)
             return s_specialTools.TryGetValue(toolName.ToLowerInvariant(), out SpecialProcessKind kind) ? kind : SpecialProcessKind.NotSpecial;
         }
 
+        private static bool StringLooksLikeBuildXLTestTempFile(string fileName)
```
mpysson: >StringLooksLikeBuildXLTestTempFile [](start = 28, length = 34)

why is this addition necessary?

mpysson:
> bool StringLooksLikeRCTempFile(PCPathChar str, size_t str_length)

Can we pinvoke the native code instead of duplicating it in C#?

---
Refers to: Public/Src/Sandbox/Windows/DetoursServices/StringOperations.cpp:238 in f45b785. [](commit_id = f45b785446a7724a533cfef4acf873268063a251, deletion_comment = False)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs
```diff
@@ -1257,6 +1257,36 @@ public void AbsentPathProbeInUndeclaredOpaquesUnsafeModeCachedPip(bool forceDepe
...
+            AbsolutePath sharedOpaqueDirPath = AbsolutePath.Create(Context.PathTable, sharedOpaqueDir);
+            DirectoryArtifact sharedOpaqueDirArtifact = DirectoryArtifact.CreateWithZeroPartialSealId(sharedOpaqueDirPath);
+            FileArtifact outputInSharedOpaqueDir = CreateOutputFileArtifact(root: sharedOpaqueDir, prefix: "sod-file");
+            FileArtifact sourceFile = CreateSourceFile();
+
+            FileArtifact whitelistFile = CreateSourceFile();
+            global::BuildXL.Utilities.Configuration.Mutable.FileAccessWhitelistEntry entry = new global::BuildXL.Utilities.Configuration.Mutable.FileAccessWhitelistEntry()
+            {
+                Value = "testValue",
+                PathFragment = ArtifactToString(whitelistFile),
```
mpysson: >whitelistFile [](start = 48, length = 13)

It doesn't look like the test pip is interacting with this file. Maybe I'm missing the point of the test?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
narasamdya: I know you're not a big fan of this exception. We can add a field in the configuration file describing the tool and the paths to ignore. But I guess that's beyond the purpose of this PR. We can add a user story to have a proper implementation.

Yes, shared opaque directories do not need this change in Detours. This change is for completeness.

---
In reply to: [314459791](https://github.com/microsoft/BuildXL/pull/741#discussion_r314459791) [](ancestors = 314459791)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
narasamdya: There's no unit test for this special exception. I just want to add one to exercise the feature.

---
In reply to: [314459941](https://github.com/microsoft/BuildXL/pull/741#discussion_r314459941) [](ancestors = 314459941)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs
narasamdya: Good catch. At first I thought that the accesses were not reported because I didn't have whitelist entries. I'm removing it.

---
In reply to: [314461083](https://github.com/microsoft/BuildXL/pull/741#discussion_r314461083) [](ancestors = 314461083)

narasamdya:
> bool StringLooksLikeRCTempFile(PCPathChar str, size_t str_length)

We can. But I'm not sure about the pinvoke on Mac side. Let's create a user story for that as such a change is beyond the purpose of this PR.

---
In reply to: [521764276](https://github.com/microsoft/BuildXL/pull/741#issuecomment-521764276) [](ancestors = 521764276)

---
Refers to: Public/Src/Sandbox/Windows/DetoursServices/StringOperations.cpp:238 in f45b785. [](commit_id = f45b785446a7724a533cfef4acf873268063a251, deletion_comment = False)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
mpysson: I'm not following. This adds a new exception rule for a new process into the product code. Why does adding a unit test necessitate that?

---
In reply to: [314472403](https://github.com/microsoft/BuildXL/pull/741#discussion_r314472403) [](ancestors = 314472403,314459941)

mpysson:
> bool StringLooksLikeRCTempFile(PCPathChar str, size_t str_length)

None of these rules apply for mac builds since they are just special cases for windows tools

---
In reply to: [521778803](https://github.com/microsoft/BuildXL/pull/741#issuecomment-521778803) [](ancestors = 521778803,521764276)

---
Refers to: Public/Src/Sandbox/Windows/DetoursServices/StringOperations.cpp:238 in f45b785. [](commit_id = f45b785446a7724a533cfef4acf873268063a251, deletion_comment = False)

PR #743, opened 2019-08-15
Don't wait after the final copy attempt

kewilli:
Without these changes:
* Try to copy from all locations. If any succeed, return
* Wait, then loop

This leads to us waiting after the final attempt.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
@@ -132,24 +132,10 @@ namespace BuildXL.Cache.ContentStore.Distributed.Stores
                 PutResult putResult = null;
                 var badContentLocations = new HashSet<MachineLocation>();
                 var missingContentLocations = new HashSet<MachineLocation>();
-                int attemptCount = -1;
+                int attemptCount = 0;
 
                 while (attemptCount < _retryIntervals.Count && (putResult == null || !putResult))
                 {
-                    if (attemptCount >= 0)
-                    {
-                        long waitTicks = _retryIntervals[attemptCount].Ticks;
-
-                        // Randomize the wait delay to `[0.5 * delay, 1.5 * delay)`
-                        TimeSpan waitDelay = TimeSpan.FromTicks((long)((waitTicks / 2) + (waitTicks * ThreadSafeRandom.Generator.NextDouble())));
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -169,17 +183,6 @@ protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
...
-                    long waitTicks = _retryIntervals[attemptCount].Ticks;
-
-                    // Randomize the wait delay to `[0.5 * delay, 1.5 * delay)`
-                    TimeSpan waitDelay = TimeSpan.FromTicks((long)((waitTicks / 2) + (waitTicks * ThreadSafeRandom.Generator.NextDouble())));
-
-                    Tracer.Warning(operationContext, $"{AttemptTracePrefix(attemptCount)} All replicas {hashInfo.Locations.Count} failed. Retrying for hash {hashInfo.ContentHash.ToShortString()} in {waitDelay.TotalMilliseconds}ms...");
-
-                    attemptCount++;
-
-                    await Task.Delay(waitDelay, cts);
```
SergeyTeplyakov: if (attemptCount != _retryIntervals.Count) {
  await Task.Delay();
}

SergeyTeplyakov: And the rest stays the same

---
In reply to: [314512295](https://github.com/microsoft/BuildXL/pull/743#discussion_r314512295) [](ancestors = 314512295)

dcwuser:
>                         break;

Log something like "success at attempt {attemptCount}".

---
Refers to: Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs:152 in 80c78a8. [](commit_id = 80c78a8540c887267c3d735688ddfa1229baf8ef, deletion_comment = False)

dcwuser:
>                 while (attemptCount < _retryIntervals.Count && (putResult == null || !putResult))

Willing to live with this, but still suggest while (true) is cleaner now.

---
Refers to: Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs:137 in 80c78a8. [](commit_id = 80c78a8540c887267c3d735688ddfa1229baf8ef, deletion_comment = False)

PR #747, opened 2019-08-16
[MsBuild] Upgrade to 16.3.0-preview-19415-01

dannyvv:
This updates the MsBuild version used by BuildXL to the latest preview build.


dannyvv:
@maneely With These changes you should be able to use your locally build version. I did not hit the issue with the localization files, so maybe that is a difference between debug and release.

The core issue you were having is that BuildXL still uses System.Memory 4.5.1 and MsBuild moved to 4.5.3 (and its dependents). What this change does is to have private versions of those packages for the msbuild deployment so it can be ahead.

PR #748, opened 2019-08-16
Change-based code coverage affected input computation

norahuang:
Implementation of computing the change-based affected inputs of a pip.

The algorithm consists of:
1. When a pip completed, checks if the outputs of the pip affected by the source code change. If the outputs are affected, report it to the scheduler. This will only happen on master.
2. When sending file hash to workers, query the global affected output list and mark those ones affected.
3. When the worker reports inputs, reports to its own fileContentManager if encounter a change affected file, so the worker will have the synced state of the master.
4. Before running a pip, compute the change affected input lis and a file containing the affected input list will be written.

[AB#1579736](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1579736)
[AB#1579468](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1579468)


Public/Src/Pips/Dll/Operations/Pip.cs
```diff
@@ -388,5 +389,18 @@ public string GetShortDescription(PipExecutionContext context)
 
             return $"{moduleName} - {valueName}{toolName} [{qualifierName}]";
         }
+
+        /// <summary>
+        /// Check if there is any output of this pip are affected by the source change
+        /// </summary>
+        public virtual bool IsOutputAffectedBySourceChange(
```
narasamdya: >IsOutputAffectedBySourceChange( [](start = 28, length = 31)

The pattern that I see in the code is pips are treated as data. Yes, your approach is more object oriented, but given that pips are data, adding a method here may not be proper. 

All the information you need for pips' inputs/outputs can be retrieved using our `PipArtifacts.cs`. Can you use that class instead?

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -220,6 +223,14 @@ public sealed partial class Process : Pip
         [PipCaching(FingerprintingRole = FingerprintingRole.Semantic)]
         public ReadOnlyArray<AbsolutePath> UntrackedScopes { get; }
 
+        /// <summary>
+        /// File Path of which the Source Change Affected Inputs are written into.
+        /// </summary>
+        /// <remarks>
+        /// </remarks>
```
narasamdya: Remove `<remarks/>` if none is provided.

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -220,6 +223,14 @@ public sealed partial class Process : Pip
         [PipCaching(FingerprintingRole = FingerprintingRole.Semantic)]
         public ReadOnlyArray<AbsolutePath> UntrackedScopes { get; }
 
+        /// <summary>
+        /// File Path of which the Source Change Affected Inputs are written into.
```
narasamdya: >File Path of which the Source Change Affected Inputs are written into. [](start = 12, length = 70)

No need for excessive letter capitalization.

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -331,6 +342,18 @@ public sealed partial class Process : Pip
...
+        /// Source Change Affected Input Files
+        /// </summary>
+        [PipCaching(FingerprintingRole = FingerprintingRole.None)]
+        public ReadOnlyArray<AbsolutePath> SourceChangeAffectedOutputFiles { get; set; }
+
+        /// <summary>
+        /// Source Change Affected Input Directories
+        /// </summary>
+        [PipCaching(FingerprintingRole = FingerprintingRole.None)]
+        public ReadOnlyArray<AbsolutePath> SourceChangeAffectedOutputDirectories { get; set; }
```
narasamdya: Why do we need these two? Also we don't allow you to modify the Process once it is created. So, the property `set` method should not be there.

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -220,6 +223,14 @@ public sealed partial class Process : Pip
         [PipCaching(FingerprintingRole = FingerprintingRole.Semantic)]
         public ReadOnlyArray<AbsolutePath> UntrackedScopes { get; }
 
+        /// <summary>
+        /// File Path of which the Source Change Affected Inputs are written into.
+        /// </summary>
+        /// <remarks>
+        /// </remarks>
+        [PipCaching(FingerprintingRole = FingerprintingRole.None)]
```
narasamdya: >FingerprintingRole.None [](start = 41, length = 23)

Why is it not `FingerprintingRole.Semantic`? I think it should be `Semantic`, right?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -673,6 +673,23 @@ public async Task<SandboxedProcessPipExecutionResult> RunAsync(CancellationToken
                     string arguments = m_pip.Arguments.ToString(m_pipDataRenderer);
                     m_timeout = GetEffectiveTimeout(m_pip.Timeout, m_sandboxConfig.DefaultTimeout, m_sandboxConfig.TimeoutMultiplier);
 
+                    // If ChangeAffectedInputListWrittenFilePath is set, write the change affected inputs of the pip to the file.
+                    if (m_pip.ChangeAffectedInputListWrittenFilePath != default)
```
narasamdya: >m_pip.ChangeAffectedInputListWrittenFilePath [](start = 24, length = 44)

This should be `m_pip.ChangeAffectedInputListWrittenFilePath.IsValid` .

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -673,6 +673,23 @@ public async Task<SandboxedProcessPipExecutionResult> RunAsync(CancellationToken
                     string arguments = m_pip.Arguments.ToString(m_pipDataRenderer);
                     m_timeout = GetEffectiveTimeout(m_pip.Timeout, m_sandboxConfig.DefaultTimeout, m_sandboxConfig.TimeoutMultiplier);
 
+                    // If ChangeAffectedInputListWrittenFilePath is set, write the change affected inputs of the pip to the file.
+                    if (m_pip.ChangeAffectedInputListWrittenFilePath != default)
+                    {
+                        var changeAffectedInputs = m_pip.GetChangeAffectedInputNames(m_pathTable);
```
narasamdya: The affected inputs must be passed from pip executor to this sandboxed process pip executor. That information is not available with the process.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -673,6 +673,23 @@ public async Task<SandboxedProcessPipExecutionResult> RunAsync(CancellationToken
                     string arguments = m_pip.Arguments.ToString(m_pipDataRenderer);
                     m_timeout = GetEffectiveTimeout(m_pip.Timeout, m_sandboxConfig.DefaultTimeout, m_sandboxConfig.TimeoutMultiplier);
 
+                    // If ChangeAffectedInputListWrittenFilePath is set, write the change affected inputs of the pip to the file.
+                    if (m_pip.ChangeAffectedInputListWrittenFilePath != default)
+                    {
+                        var changeAffectedInputs = m_pip.GetChangeAffectedInputNames(m_pathTable);
+                        if (changeAffectedInputs != ReadOnlyArray<string>.Empty)
```
narasamdya: > if (changeAffectedInputs != ReadOnlyArray<string>.Empty) [](start = 23, length = 57)

Does array not have `.Length` property?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -673,6 +673,23 @@ public async Task<SandboxedProcessPipExecutionResult> RunAsync(CancellationToken
                     string arguments = m_pip.Arguments.ToString(m_pipDataRenderer);
                     m_timeout = GetEffectiveTimeout(m_pip.Timeout, m_sandboxConfig.DefaultTimeout, m_sandboxConfig.TimeoutMultiplier);
 
+                    // If ChangeAffectedInputListWrittenFilePath is set, write the change affected inputs of the pip to the file.
+                    if (m_pip.ChangeAffectedInputListWrittenFilePath != default)
+                    {
+                        var changeAffectedInputs = m_pip.GetChangeAffectedInputNames(m_pathTable);
+                        if (changeAffectedInputs != ReadOnlyArray<string>.Empty)
+                        {
```
narasamdya: >{ [](start = 24, length = 1)

If you only write the file when the change inputs are non-empty, then the pip would fail if it expects change input file (by specifying some command line argument), but no such file is created by BuildXL.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -673,6 +673,23 @@ public async Task<SandboxedProcessPipExecutionResult> RunAsync(CancellationToken
                     string arguments = m_pip.Arguments.ToString(m_pipDataRenderer);
                     m_timeout = GetEffectiveTimeout(m_pip.Timeout, m_sandboxConfig.DefaultTimeout, m_sandboxConfig.TimeoutMultiplier);
 
+                    // If ChangeAffectedInputListWrittenFilePath is set, write the change affected inputs of the pip to the file.
```
narasamdya: >// If ChangeAffectedInputListWrittenFilePath is set, write the change affected inputs of the pip to the file. [](start = 20, length = 109)

Refactor this part into a method. See `PrepareResponseFile` for a reference. You just have to create a similar method to that.

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -334,6 +334,18 @@ public void ReportInputContent(FileArtifact artifact, in FileMaterializationInfo
...
+        /// Report the change affected outputs the pip. Only report those produced by the pip.
+        /// </summary>
+        public void ReportSourceChangeAffectedOutputs(AbsolutePath output, PipOutputOrigin origin, bool isFile)
+        {
+            if (origin == PipOutputOrigin.Produced)
+            {
+                // Notify the host with content that was reported
+                m_host.ReportSourceChangeAffectedOutputs(output, origin, isFile);
+            }
+        }
```
narasamdya: Why does this reporting have to go through `FileContentManager`?

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -934,6 +934,40 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
...
+            if (status == PipResultStatus.NotMaterialized)
+            {
+                overrideOutputOrigin = PipOutputOrigin.NotMaterialized;
+            }
+
+            if (pip.IsOutputAffectedBySourceChange(dynamicallyObservedFiles, dynamicallyObservedEnumerations, environment.Context.PathTable, sourceChangeAffectedOutputFiles, sourceChangeAffectedOutputDirectroies))
+            {
+                foreach (var output in outputContent ?? ReadOnlyArray<(FileArtifact, FileMaterializationInfo, PipOutputOrigin)>.Empty)
+                {
+                    environment.State.FileContentManager.ReportSourceChangeAffectedOutputs(output.Item1, overrideOutputOrigin ?? output.Item3, true);
```
narasamdya: >environment.State.FileContentManager [](start = 20, length = 36)

Why doe this reporting have to go through `FileContentManager`? `PipExecutor` is basically our scheduler, so can we not report directly to the scheduler? Also, the `FileContentManager` does not hold any state regarding this report.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -934,6 +934,40 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
             }
         }
 
+        /// <summary>
+        /// Report change affected ouputs to file content manager
+        /// </summary>
+        internal static void ReportSourceChangeAffectedOutputs(
```
narasamdya: >ReportSourceChangeAffectedOutputs [](start = 29, length = 33)

Who's calling this method?

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -934,6 +934,40 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
...
+            IReadOnlyCollection<(FileArtifact, FileMaterializationInfo, PipOutputOrigin)> outputContent = null,
+            IReadOnlyCollection<DirectoryArtifact> directoryOutputs = null,
+            IReadOnlyCollection<AbsolutePath> sourceChangeAffectedOutputFiles = null,
+            IReadOnlyCollection<AbsolutePath> sourceChangeAffectedOutputDirectroies = null)
+        {
+            PipOutputOrigin? overrideOutputOrigin = null;
+            if (status == PipResultStatus.NotMaterialized)
+            {
+                overrideOutputOrigin = PipOutputOrigin.NotMaterialized;
+            }
```
narasamdya: Can you provide a comment for this `overrideOutputOrigin` ?

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -934,6 +934,40 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
...
+
+            if (pip.IsOutputAffectedBySourceChange(dynamicallyObservedFiles, dynamicallyObservedEnumerations, environment.Context.PathTable, sourceChangeAffectedOutputFiles, sourceChangeAffectedOutputDirectroies))
+            {
+                foreach (var output in outputContent ?? ReadOnlyArray<(FileArtifact, FileMaterializationInfo, PipOutputOrigin)>.Empty)
```
narasamdya: >?? [](start = 53, length = 2)

We can always have `if (outputContent != null) { foreach (...) ... }`

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -289,6 +289,19 @@ public partial class Scheduler : IPipScheduler, IPipExecutionEnvironment, IFileC
         private int m_maxUnresponsivenessFactor = 0;
         private DateTime m_statusLastCollected = DateTime.MaxValue;
 
+        /// <summary>
+        /// Output file contents which affected by the source change of this build. 
+        /// </summary>
+        /// <remarks>
+        /// This list should be initialized with the source change list provided by GBR.
+        /// </remarks>
+        private readonly ConcurrentBigSet<AbsolutePath> m_sourceChangeAffectedOutputFiles = new ConcurrentBigSet<AbsolutePath>();
```
narasamdya: Should `AbsolutePath` be `FileArtifact`?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -289,6 +289,19 @@ public partial class Scheduler : IPipScheduler, IPipExecutionEnvironment, IFileC
...
+        /// </summary>
+        /// <remarks>
+        /// This list should be initialized with the source change list provided by GBR.
+        /// </remarks>
+        private readonly ConcurrentBigSet<AbsolutePath> m_sourceChangeAffectedOutputFiles = new ConcurrentBigSet<AbsolutePath>();
+
+        /// <summary>
+        /// Output file contents which affected by the source change of this build. 
+        /// </summary>
+        private readonly ConcurrentBigSet<AbsolutePath> m_sourceChangeAffectedOutputDirectories = new ConcurrentBigSet<AbsolutePath>();
```
narasamdya: >AbsolutePath [](start = 42, length = 12)

Should `AbsolutePath` be `DirectoryArtifact`? Also, can the set be joined with `m_sourceChangeAffectedOutputFiles` ?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -289,6 +289,19 @@ public partial class Scheduler : IPipScheduler, IPipExecutionEnvironment, IFileC
         private int m_maxUnresponsivenessFactor = 0;
         private DateTime m_statusLastCollected = DateTime.MaxValue;
 
+        /// <summary>
+        /// Output file contents which affected by the source change of this build. 
+        /// </summary>
+        /// <remarks>
+        /// This list should be initialized with the source change list provided by GBR.
```
narasamdya: >GBR [](start = 84, length = 3)

Not necessarily `GBR`. I think what you should say is it is initialized by the input change list provided from the configuration option that I introduced last time.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3512,6 +3525,9 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
                         }
 
                         processRunnable.Executed = true;
+                        processRunnable.Process.SourceChangeAffectedOutputFiles = GetSourceChangeAffectedOutputs().ToReadOnlyArray();
+                        processRunnable.Process.SourceChangeAffectedOutputDirectories = GetSourceChangeAffectedOutputDirectories().ToReadOnlyArray();
+
```
narasamdya: We should not update process pip. You can think of it as immutable structure once it is created.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3653,7 +3669,23 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
...
                         // master which called worker to execute post process.
-                        PipExecutor.ReportExecutionResultOutputContent(operationContext, environment, processRunnable.Description, executionResult, processRunnable.Process.DoubleWritePolicy.ImpliesDoubleWriteIsWarning());
+                        PipExecutor.ReportExecutionResultOutputContent(
+                            operationContext,
+                            environment,
+                            processRunnable.Description,
+                            executionResult,
+                            processRunnable.Process.DoubleWritePolicy.ImpliesDoubleWriteIsWarning());
+
+                        PipExecutor.ReportSourceChangeAffectedOutputs(
```
narasamdya: >PipExecutor.ReportSourceChangeAffectedOutputs( [](start = 24, length = 46)

Does the worker need to know about the affected outputs? Or should the master send that information to the worker?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4861,6 +4922,25 @@ private void ProcessFileChanges(LoggingContext loggingContext, SchedulerState sc
...
+                // The Algorithm can't deal with removal currently, so disable change-based code coverage and have to run full code coverage if has removal. 
+                if (!inputChangeList.ChangedPaths.Where(p => p.PathChanges == PathChanges.Removed).Any())
+                {
+                    foreach (var changePath in inputChangeList.ChangedPaths)
+                    {
+                        switch (changePath.PathChanges)
+                        {
+                            case PathChanges.DataOrMetadataChanged:
+                            case PathChanges.NewlyPresentAsFile:
+                                m_sourceChangeAffectedOutputFiles.GetOrAdd(AbsolutePath.Create(Context.PathTable, changePath.Path));
```
narasamdya: Why not `TryAdd` ?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4861,6 +4922,25 @@ private void ProcessFileChanges(LoggingContext loggingContext, SchedulerState sc
                     m_configuration.Schedule.InputChanges.ToString(Context.PathTable),
                     m_configuration.Layout.SourceDirectory.ToString(Context.PathTable),
                     DirectoryTranslator);
+
+                // The Algorithm can't deal with removal currently, so disable change-based code coverage and have to run full code coverage if has removal. 
+                if (!inputChangeList.ChangedPaths.Where(p => p.PathChanges == PathChanges.Removed).Any())
```
narasamdya: >inputChangeList.ChangedPaths.Where [](start = 21, length = 34)

`inputChangeList.ChangedPaths.Any(p => p.PathChanges == PathChanges.Removed)`.

Public/Src/Engine/Scheduler/Scheduler.cs
narasamdya: Can we have these two structures in a separate class?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4861,6 +4922,25 @@ private void ProcessFileChanges(LoggingContext loggingContext, SchedulerState sc
                     m_configuration.Schedule.InputChanges.ToString(Context.PathTable),
                     m_configuration.Layout.SourceDirectory.ToString(Context.PathTable),
                     DirectoryTranslator);
+
+                // The Algorithm can't deal with removal currently, so disable change-based code coverage and have to run full code coverage if has removal.
```
narasamdya: >// The Algorithm can't deal with removal currently, so disable change-based code coverage and have to run full code coverage if has removal [](start = 16, length = 139)

If you have `m_sourceChangeAffectedOutputFiles` and `m_sourceChangeAffectedOutputDirectories` in a separate class, you can use that class to refactor this part of the code.

narasamdya:

Is there any change to this file?

---
Refers to: Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SchedulerIntegrationTestBase.cs:1 in 000cec3. [](commit_id = 000cec362680291275915f7f3c7fabcdcce5de43, deletion_comment = False)

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -765,6 +793,119 @@ public bool TryComputePipUniqueOutputHash(PathTable pathTable, out long pipUniqu
...
+        /// <inheritdoc />
+        public override bool IsOutputAffectedBySourceChange(
+            ReadOnlyArray<AbsolutePath> dynamicallyObservedFiles,
+            ReadOnlyArray<AbsolutePath> dynamicallyObservedEnumerations,
+            PathTable pathTable,
+            IReadOnlyCollection<AbsolutePath> sourceChangeAffectedOutputFiles = null,
+            IReadOnlyCollection<AbsolutePath> sourceChangeAffectedOutputDirectroies = null)
+        {
+            // sourceChangeAffectedOutputFiles and sourceChangeAffectedOutputDirectroies 
+            // are initilized with the source change list from GBR.
```
narasamdya: >GBR [](start = 63, length = 3)

Not necessarily GBR; see my previous comment.

Public/Src/Pips/Dll/Operations/Process.cs
norahuang: This is the global affected list of the whole build. It represent the state of the build by the time this pip execute. It shouldn't be in the fingerprint.

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -765,6 +793,119 @@ public bool TryComputePipUniqueOutputHash(PathTable pathTable, out long pipUniqu
...
+            // If it is null, that means no file or directory change for this build.
+            // Thus no output is affected by the change.
+            if (!sourceChangeAffectedOutputFiles.Any() && !sourceChangeAffectedOutputDirectroies.Any())
```
narasamdya: This `if` condition will blow up with null reference exception if they are `null`.

Public/Src/Pips/Dll/Operations/Process.cs
dannyvv: Yes, the Process object should indeed be immutable, it can also be paged out during builds or analysis and then these fields aren't saved, only the fields added during graph construction are saved to the byte array and then the pips are hydrated as needed (potentially many times)

---
In reply to: [314804613](https://github.com/microsoft/BuildXL/pull/748#discussion_r314804613) [](ancestors = 314804613)

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -934,6 +934,40 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
...
+            IReadOnlyCollection<AbsolutePath> sourceChangeAffectedOutputFiles = null,
+            IReadOnlyCollection<AbsolutePath> sourceChangeAffectedOutputDirectroies = null)
+        {
+            PipOutputOrigin? overrideOutputOrigin = null;
+            if (status == PipResultStatus.NotMaterialized)
+            {
+                overrideOutputOrigin = PipOutputOrigin.NotMaterialized;
+            }
+
+            if (pip.IsOutputAffectedBySourceChange(dynamicallyObservedFiles, dynamicallyObservedEnumerations, environment.Context.PathTable, sourceChangeAffectedOutputFiles, sourceChangeAffectedOutputDirectroies))
```
narasamdya: >pip.IsOutputAffectedBySourceChange [](start = 16, length = 34)

Why are we doing it here? If you look the `OnPipCompleted` method of `Scheduler`, we update the incremental scheduling state with dynamically observed observations there. Why can't we do it in `OnPipCompleted`?

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -765,6 +793,119 @@ public bool TryComputePipUniqueOutputHash(PathTable pathTable, out long pipUniqu
...
+            {
+                return false;
+            }
+
+            // Check if any dynamic and static file dependency is sourceChangeAffectedOutputFiles
+            var hasAffected = sourceChangeAffectedOutputFiles.Intersect(dynamicallyObservedFiles).Any()
```
narasamdya: >sourceChangeAffectedOutputFiles.Intersect(dynamicallyObservedFiles).Any() [](start = 30, length = 73)

My gut feeling says that, due to, `sourceChangeAffectedOutputFiles.Intersect(dynamicallyObservedFiles).Any()`, this method doesn't belong in this `Process.cs`. The reason is that the expression does some check on something that is not pip's static property.

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -765,6 +793,119 @@ public bool TryComputePipUniqueOutputHash(PathTable pathTable, out long pipUniqu
...
+                || sourceChangeAffectedOutputFiles.Intersect(Dependencies.Select(d => d.Path)).Any();
+
+            if (hasAffected)
+            {
+                return true;
+            }
+
+            if (dynamicallyObservedEnumerations.Any() || DirectoryDependencies.Any())
+            {
+                // Check if dynamic and static directory denpendencis under any directory of sourceChangeAffectedOutputDirectroies
```
narasamdya: typo

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -765,6 +793,119 @@ public bool TryComputePipUniqueOutputHash(PathTable pathTable, out long pipUniqu
             return true;
         }
 
+        /// <summary>
+        /// Compute the intersection of the pip's dependencies and global affected SourceChangeAffectedOutputs of the build
+        /// </summary>
+        public ReadOnlyArray<string> GetChangeAffectedInputNames(PathTable pathTable)
+        {
+            var changeAffectedInputs = SourceChangeAffectedOutputFiles.Intersect(Dependencies.Select(d => d.Path)).Select(i => i.GetName(pathTable).ToString(pathTable.StringTable)).ToHashSet();
```
dannyvv: >. [](start = 114, length = 1)

nit: Small formatting suggestion since there are so many parenthesis:

I usually find it easier to read code when deep exressions like this are left-aligh the larger dots and then indent i.e. something like
```
SourceChangeAffectedOutputFiles
    .Intersect(Dependencies.Select(d => d.Path))
    .Select(i => i.GetName(pathTable).ToStringpathTable.StringTable))
    .ToHashSet();
```

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -765,6 +793,119 @@ public bool TryComputePipUniqueOutputHash(PathTable pathTable, out long pipUniqu
...
+        /// </summary>
+        public ReadOnlyArray<string> GetChangeAffectedInputNames(PathTable pathTable)
+        {
+            var changeAffectedInputs = SourceChangeAffectedOutputFiles.Intersect(Dependencies.Select(d => d.Path)).Select(i => i.GetName(pathTable).ToString(pathTable.StringTable)).ToHashSet();
+
+            foreach (var file in SourceChangeAffectedOutputFiles)
+            {
+                foreach (var directory in DirectoryDependencies)
+                {
+                    if (directory.Path.TryGetRelative(pathTable, file, out var relativePath))
```
dannyvv: we have a helper that just performs the check called `isWithin` which should be faster since it doesn't have to allocate and compute the result

Public/Src/Pips/Dll/Operations/Process.cs
dannyvv: I noticed a few similar typo's, you might want to do a full text search. 

---
In reply to: [314827920](https://github.com/microsoft/BuildXL/pull/748#discussion_r314827920) [](ancestors = 314827920)

dannyvv:
>         internal static Process InternalDeserialize(PipReader reader)

Don't forget to add ChangeAffectedInputListWrittenFilePath to serialize and deserialize methods

---
Refers to: Public/Src/Pips/Dll/Operations/Process.cs:912 in 000cec3. [](commit_id = 000cec362680291275915f7f3c7fabcdcce5de43, deletion_comment = False)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
dannyvv: >GetChangeAffectedInputNames [](start = 57, length = 27)

Did I infer correctly that this does not yet have the transitive pattern?

Public/Src/Pips/Dll/Operations/Process.cs
norahuang: This two are set at the time when master distribute the pip. Master will retrieve the global impacted outputs with the current state of the build. But the process is created earlier than the distribution, that's why I need the set here.

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -334,6 +334,18 @@ public void ReportInputContent(FileArtifact artifact, in FileMaterializationInfo
             }
         }
 
+        /// <summary>
+        /// Report the change affected outputs the pip. Only report those produced by the pip.
+        /// </summary>
+        public void ReportSourceChangeAffectedOutputs(AbsolutePath output, PipOutputOrigin origin, bool isFile)
+        {
+            if (origin == PipOutputOrigin.Produced)
```
narasamdya: >origin == PipOutputOrigin.Produced [](start = 16, length = 34)

Is it only for pip that gets a cache miss.

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -334,6 +334,18 @@ public void ReportInputContent(FileArtifact artifact, in FileMaterializationInfo
             }
         }
 
+        /// <summary>
+        /// Report the change affected outputs the pip. Only report those produced by the pip.
+        /// </summary>
+        public void ReportSourceChangeAffectedOutputs(AbsolutePath output, PipOutputOrigin origin, bool isFile)
```
narasamdya: >ReportSourceChangeAffectedOutputs [](start = 20, length = 33)

Can you please consult Lance about this change in the `FileContentManager`?

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -765,6 +793,119 @@ public bool TryComputePipUniqueOutputHash(PathTable pathTable, out long pipUniqu
...
+                        }
+                    }
+
+                    foreach (var staticDir in DirectoryDependencies)
+                    {
+                        if (staticDir.Path.TryGetRelative(pathTable, affectedFile, out var relativePath))
+                        {
+                            return true;
+                        }
+                    }
```
narasamdya: We cannot statically check if a path is relative to another. We need to consult to `FileContentManager`. Thus, I don't think this method belongs here.

For example, suppose that your pip depends on a seal directory `D` with explicitly specified members `D\a` and `D\b`. If the affected input contains only `D\c`, then, with your method, this pip is affected, but it is actually not.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: I will generate a file with empty content in this case. However, there are some other cases, for example removing file/directory, that I will not generate a file which indicate qtest to instrument all load files. I talked to Aniruth and he agree to add a check in qtest to see if the file(by specifying some command line argument) doesn't exist, qtest will instrument all.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: It calls the function GetChangeAffectedInputNames() which implemented in Process.cs to compute the affected inputs here. And the Process has the needed information.

Public/Src/Engine/Scheduler/PipExecutor.cs
norahuang: In the scheduler, postprocess and copyfile.

Public/Src/Engine/Scheduler/Scheduler.cs
norahuang: I don't see neccesarity of FileArtifact.

Public/Src/Engine/Scheduler/Scheduler.cs
norahuang: I still think AbsolutePath is enough.
The reason I seperate the set to sourceChangeAffectedOutputFiles and sourceChangeAffectedOutputFiles is that the computing algorithm interact with directoryDenpendencies and Denpendencies as well as dynamicallyObservedFiles and dynamicallyObservedEnumerations. Making it two seperate set will make the computation algorithm easier.

Public/Src/Pips/Dll/Operations/Process.cs
norahuang: I changed the implementation. I compute the change affected input in the master and pass the result in the processRunnable instead of process.

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
norahuang: moved to IPipExecutionEnvironment.

Public/Src/Engine/Scheduler/PipExecutor.cs
norahuang: moved to IPipExecutionEnvironment.

Public/Src/Engine/Scheduler/Scheduler.cs
norahuang: The global affected outputs are only maintained in master. When master distribute the runnerablepip, the change affected inputs of the pip was computed in the master and set in the runnerablepip. The worker will write out this affected inputs to a file.

Public/Src/Engine/Scheduler/Scheduler.cs
norahuang: There is no TryAdd in ConcurrentBigSet.

Public/Src/Engine/Scheduler/PipExecutor.cs
norahuang: I moved the reporting to the PipExecutionStep.ExecuteProcess and only do the reporting on the master. I can't do that in the OnPipCompleted, since the master would not has OnPipCompleted for all pips. But master has ExecuteProcess for all pips.

Public/Src/Pips/Dll/Operations/Process.cs
norahuang: Moved it to PipArtifacts.cs

Public/Src/Pips/Dll/Operations/Process.cs
norahuang: Thanks

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
norahuang: I removed this from FileContentManager.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: It includes the transitive pattern.

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
norahuang: I think it should only includes the cache miss pip. If a pip is cache hit, I don't see any possible that its output is change affected.

Public/Src/Pips/Dll/Operations/CopyFile.cs
```diff
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
+using System.Linq;
```
narasamdya: Revert this.

Public/Src/Pips/Dll/Operations/Pip.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
```
narasamdya: Revert this.

Public/Src/Utilities/Configuration/ISandboxConfiguration.cs
```diff
@@ -262,5 +262,12 @@ public interface ISandboxConfiguration
         /// Temporary flag to use tool incremental behavior when preserve outputs is enabled.
         /// </summary>
         bool PreserveOutputsForIncrementalTool { get; }
+
+        /// <summary>
+        /// Flag indicate if change based code coverage is enable.
+        /// This will be set as true when inputChanges commmand line option is presented.
+        /// If the change list contain any removal, it will fall back to false.
+        /// </summary>
+        bool EnableChangeBasedCodeCoverage { get; set; }
```
narasamdya: BuildXL doesn't have code coverage. Also, don't mention QTest here. We should not be specific to QTest. 

Name the feature based on its behavior.

Public/Src/Pips/Dll/Artifacts/PipArtifacts.cs
```diff
@@ -230,5 +233,102 @@ public static bool ForEachOutput(Pip pip, Func<FileOrDirectoryArtifact, bool> ou
 
             return result;
         }
+
+        /// <summary>
+        /// Check if there is any output of this pip are affected by the source change
+        /// </summary>
+        public static bool IsOutputAffectedBySourceChange(
```
narasamdya: >IsOutputAffectedBySourceChange [](start = 27, length = 30)

Move it to separate class.

Public/Src/Pips/Dll/Artifacts/PipArtifacts.cs
```diff
@@ -230,5 +233,102 @@ public static bool ForEachOutput(Pip pip, Func<FileOrDirectoryArtifact, bool> ou
...
+                case PipType.CopyFile:
+                    return sourceChangeAffectedOutputFiles == null ? false : sourceChangeAffectedOutputFiles.Contains(((CopyFile)pip).Source.Path);
+                default:
+                    return false;
+            }
+        }
+
+
+        /// <inheritdoc />
+        private static bool IsOutputOfProcessAffectedBySourceChange(
```
narasamdya: >IsOutputOfProcessAffectedBySourceChange [](start = 28, length = 39)

Move to a separate class.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
narasamdya: We should not make any specific assumption about QTest here.

---
In reply to: [314851206](https://github.com/microsoft/BuildXL/pull/748#discussion_r314851206) [](ancestors = 314851206)

Public/Src/Engine/Scheduler/ProcessRunnablePip.cs
```diff
@@ -57,7 +57,12 @@ public sealed class ProcessRunnablePip : RunnablePip
         /// </summary>
         public CacheableProcess CacheableProcess { get; private set; }
 
-        private readonly int m_weightBasedOnHistoricCpuUsage;
+        /// <summary>
+        /// Source change affected input of the pip
+        /// </summary>
+        public ReadOnlyArray<string> ChangeAffectedInputNames { get; set; }
```
narasamdya: >ChangeAffectedInputNames [](start = 37, length = 24)

string?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3512,8 +3528,30 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
                         }
 
                         processRunnable.Executed = true;
+
+                        // Since the source change affected outputs are only maintained on the master, 
+                        // The Affected Inputs of the pip can only be computed on master.
+                        // The result will be set in the processRunnable and passed along to the worker who execute the process
+                        if (!IsDistributedWorker)
+                        {
+                            processRunnable.ChangeAffectedInputNames = PipExecutor.GetChangeAffectedInputNames(environment, processRunnable.Process);
```
narasamdya: >ChangeAffectedInputNames [](start = 44, length = 24)

Does this field get serialized to the worker?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2558,6 +2563,48 @@ private async Task<bool> PrepareResponseFile()
...
+                        () => Path.GetDirectoryName(destination),
+                        ex => { throw new BuildXLException("Cannot get directory name", ex); });
+
+                    PreparePathForOutputFile(m_pip.ChangeAffectedInputListWrittenFilePath);
+                    FileUtilities.CreateDirectory(directoryName);
+                    await FileUtilities.WriteAllTextAsync(
+                       destination,
+                       string.Join(
+                           Environment.NewLine,
+                           changeAffectedInputs.Select(i => i.GetName(m_pathTable).ToString(m_pathTable.StringTable)).ToHashSet()
```
mpysson: >changeAffectedInputs.Select(i => i.GetName(m_pathTable).ToString(m_pathTable.StringTable)).ToHashSet() [](start = 27, length = 102)

I think we should ensure the items in this file have consistent order

Public/Src/Engine/Processes/Tracing/Log.cs
```diff
@@ -829,6 +829,21 @@ internal abstract partial class Logger
             int errorCode,
             string message);
 
+        [GeneratedEvent(
+            (int)LogEventId.PipProcessChangeAffectedInputsWrittenFileCreationFailed,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Error,
+            Keywords = (int)Keywords.UserMessage,
+            EventTask = (int)Tasks.PipExecutor,
+            Message = EventConstants.PipSpecPrefix + "Process response file could not be prepared, path '{2}', error code {3:X8}: {4}")]
```
mpysson: >Process response file could not be prepared, path '{2}', error code {3:X8}: {4}")] [](start = 54, length = 82)

Update message

Public/Src/Engine/Processes/Tracing/LogEventId.cs
```diff
@@ -59,7 +59,8 @@ public enum LogEventId
         PipProcessStandardInputException = 86,
 
         PipProcessInvalidErrorRegex = 89,
-        //// Reserved  = 90,
```
mpysson: >//// Reserved  = 90, [](start = 8, length = 20)

I suspect this is marked "reserved" because it is used in some other eventid file. So it probably isn't ok to reuse. If my suspicion is correct there should also be a unit test that fails due to the conflict.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4857,10 +4907,17 @@ private void ProcessFileChanges(LoggingContext loggingContext, SchedulerState sc
...
             {
                 inputChangeList = InputChangeList.CreateFromFile(
-                    loggingContext, 
+                    loggingContext,
                     m_configuration.Schedule.InputChanges.ToString(Context.PathTable),
                     m_configuration.Layout.SourceDirectory.ToString(Context.PathTable),
                     DirectoryTranslator);
+
+                // Only master maintain the list
+                if (!IsDistributedWorker && "DynamicCodeCov".Equals(Environment.GetEnvironmentVariable("[Sdk.BuildXL]qCodeCoverageEnumType")))
```
mpysson: >"DynamicCodeCov".Equals(Environment.GetEnvironmentVariable("[Sdk.BuildXL]qCodeCoverageEnumType") [](start = 44, length = 96)

This is icky. Is it just temporary for testing? If no we need to figure out a better way to pass this configuration

mpysson:
> using System.Collections.Generic;

Copyright header

---
Refers to: Public/Src/Engine/Scheduler/ChangeAffectedOutput/AffectedOutputList.cs:1 in 9542e81. [](commit_id = 9542e8191c8ca4d8a5eff79e42cf67137edc9870, deletion_comment = False)

Public/Src/Engine/Scheduler/ChangeAffectedOutput/AffectedOutputList.cs
```diff
@@ -0,0 +1,258 @@
...
+
+namespace BuildXL.Scheduler.ChangeAffectedOutput
+{
+    /// <summary>
+    /// Class representing source change affected output list.
+    /// </summary>
+    public sealed class AffectedOutputList
+    {
+
+        /// <summary>
```
mpysson: >         [](start = 0, length = 8)

nit - extra newline

Public/Src/Engine/Scheduler/ChangeAffectedOutput/AffectedOutputList.cs
```diff
@@ -0,0 +1,258 @@
...
+            m_fileContentManager = fileContentManager;
+        }
+
+        /// <summary>
+        /// Initial affected output list with the external source change list
+        /// </summary>
+        public bool InitialAffectedOutputList(InputChangeList inputChangeList, PathTable pathTable)
+        {
+            var allInputsAreAffected = false;
+            // The Algorithm can't deal with removal currently, so disable change-based code coverage and have to run full code coverage if has removal.
```
mpysson: >change-based code coverage and have to run full code coverage [](start = 75, length = 61)

nit - change based code coverage is the first consumer of this but it is meant to be a generic feature. so make the comments generic

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4857,10 +4907,17 @@ private void ProcessFileChanges(LoggingContext loggingContext, SchedulerState sc
...
+                {
+                    m_affectedOutputList = new AffectedOutputList(Context.PathTable, m_fileContentManager);
+                    m_configuration.Sandbox.AreAllInputsAffected = m_affectedOutputList.InitialAffectedOutputList(inputChangeList, Context.PathTable);
```
mpysson: >m_configuration.Sandbox.AreAllInputsAffected [](start = 20, length = 44)

Typically we use the global configuration object for things that are set on the command line for how the overall app/engine/scheduler are supposed to operate. This doesn't look to be user configurable and is only consumed within Scheduler.cs. It seems like it would be a better fit to just have it as local state within the scheduler.

Public/Src/Engine/Scheduler/Scheduler.cs
mpysson: Should this just be based on any process in the build graph having ChangeAffectedInputListWrittenFilePath set?

---
In reply to: [317211659](https://github.com/microsoft/BuildXL/pull/748#discussion_r317211659) [](ancestors = 317211659)

Public/Src/Engine/Processes/Tracing/LogEventId.cs
norahuang: It seems it is the opposite. The one marked as reserved is available and those ones not marked may have been used. I did a full text search for some instance and always find the reserved ones only appear as reserved.

Public/Src/Engine/Scheduler/Scheduler.cs
norahuang: I will remove this unnecessary check. m_configuration.Schedule.InputChanges and [Sdk.BuildXL]qCodeCoverageEnumType" should present or absent together.

Public/Src/Engine/Scheduler/ChangeAffectedOutput/AffectedOutputList.cs
norahuang: done

Public/Src/Pips/Dll/Artifacts/PipArtifacts.cs
```diff
@@ -230,5 +235,31 @@ public static bool ForEachOutput(Pip pip, Func<FileOrDirectoryArtifact, bool> ou
...
+        /// </summary>
+        public static IReadOnlyCollection<AbsolutePath> GetAllInputs(Process process, PathTable pathTable)
+        {
+            var inputs = new HashSet<AbsolutePath>();
+
+            inputs.AddRange(process.Dependencies.Select(d => d.Path));
+
+            foreach (var inDir in process.DirectoryDependencies)
+            {
+                FileUtilities.EnumerateDirectoryEntries(
```
ljcollins25: >EnumerateDirectoryEntries [](start = 30, length = 25)

I don't think this is correct. Pips don't consume every file in their directory dependencies.

PR #749, opened 2019-08-16
Use inactive machines list as an indicator in machine reputation.

ljcollins25:


Public/Src/Cache/ContentStore/Distributed/ContentLocations/MachineReputationTracker.cs
```diff
@@ -93,6 +101,15 @@ public virtual void ReportReputation(MachineLocation location, MachineReputation
                 return;
             }
 
+            if (reputation == MachineReputation.Good
+                && _clusterState != null
+                && _clusterState.TryResolveMachineId(location, out var machineId)
+                && _clusterState.IsMachineMarkedInactive(machineId))
+            {
+                _context.Debug($"Marked machine {machineId}='{location}' active due to report of good reputation.");
```
SergeyTeplyakov: very nit: but is the message should be something like "Marking machine ... active", because we're about to make it. The operation is not finished yet.

PR #753, opened 2019-08-16
Add support for parallel graph loading using topological sorting

erickulcyk:
Each pip that is on the same topological level can be loaded in parallel.  Edges are constructed using producer and consumer files/directories.  Modules/specs/values are all serialized at the start.  Special IPC pips are serialized in serial as they need to be added in the correct order.


Public/Src/Pips/Dll/Operations/PipGraphFragmentSerializer.cs
```diff
@@ -127,19 +144,31 @@ public void Serialize(AbsolutePath filePath, IReadOnlyCollection<Pip> pipsToSeri
...
-                writer.Write(pipsToSerialize.Count);
-
-                foreach (var pip in pipsToSerialize)
+                writer.Write(totalPipCount);
+                m_serializedPipCount = 0;
+                foreach (var pipGroup in pipsToSerialize)
                 {
-                    pip.Serialize(writer);
+                    int i = 0;
+                    foreach (var pip in pipGroup)
```
ljcollins25: >pipGroup [](start = 40, length = 8)

Make this an IReadOnlyList and use BuildXLWriter.WriteeReadOnlyList

Public/Src/Engine/Scheduler/Graph/GraphFragmentBuilder.cs
```diff
@@ -29,6 +30,10 @@ public class GraphFragmentBuilder : IPipGraph
         private readonly PipExecutionContext m_pipExecutionContext;
         private readonly SealedDirectoryTable m_sealDirectoryTable;
         private readonly ConcurrentQueue<Pip> m_pips = new ConcurrentQueue<Pip>();
+        private readonly ConcurrentDictionary<PipId, IList<Pip>> m_pipDependents = new ConcurrentDictionary<PipId, IList<Pip>>();
+        private readonly ConcurrentBigMap<FileArtifact, PipId> m_fileProducers = new ConcurrentBigMap<FileArtifact, PipId>();
+        private readonly ConcurrentBigMap<DirectoryArtifact, PipId> m_opaqueDirectoryProducers = new ConcurrentBigMap<DirectoryArtifact, PipId>();
```
ljcollins25: >ConcurrentBigMap [](start = 25, length = 16)

Assuming its not a big deal, I would move this logic to a wrapping IPipGraph.

narasamdya:
I expect to have the changes rather isolated. See further comments.

narasamdya:
>     public class GraphFragmentBuilder : IPipGraph

Here, we can create a wrapper for `GraphFragmentBuilder` where the builder does more that this `GraphFragmentBuilder`. Then, depending on the option you pass to the `GraphFragmentGenerator`, either you choose this builder or the wrapper one. 

---
Refers to: Public/Src/Engine/Scheduler/Graph/GraphFragmentBuilder.cs:27 in 0434e6d. [](commit_id = 0434e6d9032ecf1f019dd865ae0af54a1870b00f, deletion_comment = False)

narasamdya:
>         public bool Deserialize(

I'd expect to have a separate method. One for standard one, and the other for the top-sort one. 

The way you do that is to write a `bool` value whether the fragment should be read in standard way or in top sort way.

---
Refers to: Public/Src/Pips/Dll/Operations/PipGraphFragmentSerializer.cs:74 in 0434e6d. [](commit_id = 0434e6d9032ecf1f019dd865ae0af54a1870b00f, deletion_comment = False)

Public/Src/Tools/BxlScriptAnalyzer/GraphFragment/GraphFragmentGenerator.cs
```diff
@@ -95,7 +105,17 @@ public override bool FinalizeAnalysis()
...
+                var pips = PipGraph.RetrieveScheduledPips().ToList();
+                if (m_topSort)
+                {
+                    var finalPipList = TopSort(pips);
+                    serializer.Serialize(m_absoluteOutputPath, finalPipList, pips.Count, m_description);
+                }
+                else
+                {
+                    serializer.Serialize(m_absoluteOutputPath, pips.Select(x => new List<Pip>() { x }).ToList(), pips.Count, m_description);
+                }
```
narasamdya: Preserve the old way of serialization. See previous comment where you can write a bool indicating if this is serialized in top-sort or in standard way.

erickulcyk:
>         public bool Deserialize(

Why use a separate method when I can share code?

---
In reply to: [522166281](https://github.com/microsoft/BuildXL/pull/753#issuecomment-522166281) [](ancestors = 522166281)

---
Refers to: Public/Src/Pips/Dll/Operations/PipGraphFragmentSerializer.cs:74 in 0434e6d. [](commit_id = 0434e6d9032ecf1f019dd865ae0af54a1870b00f, deletion_comment = False)

Public/Src/Tools/BxlScriptAnalyzer/GraphFragment/GraphFragmentGenerator.cs
erickulcyk: Why?  The old way will likely never see production as it is not as efficient.

---
In reply to: [314905882](https://github.com/microsoft/BuildXL/pull/753#discussion_r314905882) [](ancestors = 314905882)

Public/Src/Pips/Dll/Operations/PipGraphFragmentSerializer.cs
```diff
@@ -127,23 +169,50 @@ public void Serialize(AbsolutePath filePath, IReadOnlyCollection<Pip> pipsToSeri
...
+        /// Serializes list of pips to a file.
+        /// </summary>
+        public void SerializeTopSort(AbsolutePath filePath, IReadOnlyCollection<IReadOnlyList<Pip>> pipsToSerialize, int totalPipCount, string fragmentDescription = null)
+        {
+            Contract.Requires(filePath.IsValid);
+            Contract.Requires(pipsToSerialize != null);
+
+            string fileName = filePath.ToString(m_pipExecutionContext.PathTable);
+            using (var stream = new FileStream(fileName, FileMode.CreateNew, FileAccess.Write, FileShare.None))
+            using (var writer = new PipRemapWriter(m_pipExecutionContext, m_pipGraphFragmentContext, stream))
```
narasamdya: Part of this method can be refactored with SerializedSerially.

Public/Src/Tools/BxlScriptAnalyzer/GraphFragment/GraphFragmentGenerator.cs
```diff
@@ -46,6 +50,12 @@ public override bool HandleOption(CommandLineUtilities.Option opt)
                 return true;
             }
 
+            if (m_topSortOption.Match(opt.Name))
+            {
+                m_topSort = CommandLineUtilities.ParseBooleanOption(opt);
+                return base.HandleOption(opt);
+            }
+
```
narasamdya: Is this code reachable? The one in `Args.cs` seems to shadow this code.

Public/Src/Tools/BxlScriptAnalyzer/Args.cs
```diff
@@ -33,6 +33,11 @@ internal sealed partial class Args : CommandLineUtilities
         /// </summary>
         public readonly bool Help;
 
+        /// <summary>
+        /// Serialize using top sort so that loading the graph is faster
+        /// </summary>
+        public readonly bool TopSort;
```
narasamdya: This should be property of Analyzer; just like the Engine phases. One use top-sort-capable pip graph if any of the analyzer set the top sort property to true.

PR #754, opened 2019-08-16
Various reliability improvements for unhandled exception handler

mpysson:
- Increase timeout to block other threads for thread with first unhandled exception to complete processing
- Fix some crashes in the crash handler
- Reorder crash handler to put most important operations first
- Add console logging for status of crash handler

Fixes [AB#1519100](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1519100)

### DIFF
--- a/Public/Src/App/Bxl/BuildXLApp.cs
+++ b/Public/Src/App/Bxl/BuildXLApp.cs
@@ -120,7 +120,9 @@ namespace BuildXL
     /// </summary>
     internal sealed class BuildXLApp : IDisposable
     {
-        private const int FailureCompletionTimeoutMs = 30 * 1000;
+        // We give the failure completion logic a generous 60 seconds to complete since in some cases taking a crash dump
+        // can take quite a while
+        private const int FailureCompletionTimeoutMs = 60 * 1000;
 
         // 24K buffer size means that internally, the StreamWriter will use 48KB for a char[] array, and 73731 bytes for an encoding byte array buffer --- all buffers <85000 bytes, and therefore are not in large object heap
         private const int LogFileBufferSize = 24 * 1024;
@@ -1170,6 +1172,11 @@ namespace BuildXL
             /// </summary>
             public readonly string LogPath;
 
+            /// <summary>
+            /// The path to the log directory
+            /// </summary>
+            public string RootLogDirectory { get; private set; }
+
             public AppLoggers(
                 DateTime startTime,
                 IConsole console,
@@ -1229,15 +1236,13 @@ namespace BuildXL
 
             public TrackingEventListener TrackingEventListener { get; private set; }
 
-            private string rootLogDirectory = null;
-
             public void ConfigureLogging(LoggingContext loggingContext)
             {
                 lock (m_lock)
                 {
                     Contract.Assume(!m_disposed);
 
-                    rootLogDirectory = m_configuration.LogsDirectory.IsValid ? m_configuration.LogsDirectory.ToString(m_pathTable) : null;
+                    RootLogDirectory = m_configuration.LogsDirectory.IsValid ? m_configuration.LogsDirectory.ToString(m_pathTable) : null;
                     ConfigureTrackingListener();
 
                     if (m_configuration.FileVerbosity != VerbosityLevel.Off && m_configuration.LogsDirectory.IsValid)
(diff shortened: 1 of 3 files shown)


Public/Src/App/Bxl/BuildXLApp.cs
```diff
@@ -1736,14 +1742,14 @@ internal bool OnConsoleCancelEvent(bool isTermination)
                         break;
                     default:
                         Logger.Log.CatastrophicFailure(pm.LoggingContext, failureMessage, s_buildInfo?.CommitId ?? string.Empty, s_buildInfo?.Build ?? string.Empty);
+                        WriteToConsole(Strings.App_LogsDirectory, loggers.RootLogDirectory);
+                        WriteToConsole("Collecting some information about this crash...");
+
                         break;
```
sahilmgandhi: Nit: empty line

Public/Src/App/Bxl/BuildXLApp.cs
```diff
@@ -1229,15 +1236,13 @@ private static WarningManager CreateWarningManager(IWarningHandling configuratio
...
-            private string rootLogDirectory = null;
-
             public void ConfigureLogging(LoggingContext loggingContext)
             {
                 lock (m_lock)
                 {
                     Contract.Assume(!m_disposed);
 
-                    rootLogDirectory = m_configuration.LogsDirectory.IsValid ? m_configuration.LogsDirectory.ToString(m_pathTable) : null;
+                    RootLogDirectory = m_configuration.LogsDirectory.IsValid ? m_configuration.LogsDirectory.ToString(m_pathTable) : null;
```
olkononenko: >RootLogDirectory = m_configuration.LogsDirectory.IsValid ? m_configuration.LogsDirectory.ToString(m_pathTable) : null; [](start = 20, length = 118)

nit: should this be happening in the constructor?  We are initializing `LogPath` there, but `RootLogDirectory` here...

Public/Src/App/Bxl/BuildXLApp.cs
```diff
@@ -1790,6 +1795,13 @@ internal bool OnConsoleCancelEvent(bool isTermination)
 
                     if (!OperatingSystemHelper.IsUnixOS)
                     {
+                        string[] filesToAttach = m_configuration != null
```
olkononenko: >m_configuration != null [](start = 49, length = 23)

i know that you only moved that code, but I do not see how `m_configuration` can be null, i.e., I think this check is not necessary

Public/Src/App/Bxl/BuildXLApp.cs
```diff
@@ -1815,9 +1827,11 @@ internal bool OnConsoleCancelEvent(bool isTermination)
                 Environment.Exit(ExitCode.FromExitKind(effectiveExitKind));
             }
 #pragma warning disable ERP022 // Unobserved exception in generic exception handler
-            catch (Exception)
+            catch (Exception ex)
             {
                 // Oh my, this isn't going very well.
+                WriteErrorToConsole("Unhandled exception in exception handler");
+                WriteErrorToConsole(ex.ToString());
```
olkononenko: >ToString [](start = 39, length = 8)

should we use `DemistifyToString()` here? `ToString()` might not give us useful results

Public/Src/App/Bxl/BuildXLApp.cs
mpysson: It can in fact be null due to the wacky configuration swapping that happens during Engine.Run(). But your point is still valid since this block no longer requires it to be non-null since I am working off of loggers instead of m_configuration.Logging

---
In reply to: [316293989](https://github.com/microsoft/BuildXL/pull/754#discussion_r316293989) [](ancestors = 316293989)

Public/Src/App/Bxl/BuildXLApp.cs
mpysson: Good call. I didn't look particularly closely at it but it seems rootLogDirectory was previously dead code before I added this new reference on it. I moved to the constructor and also changed it to be based on the parent of the log file instead of the configuration which may or may not actually be set

---
In reply to: [316291646](https://github.com/microsoft/BuildXL/pull/754#discussion_r316291646) [](ancestors = 316291646)

PR #755, opened 2019-08-16
Update ApplyBatch with dictionary and open write only DB connection

sahilmgandhi:
[AB#1585618](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1585618)

Based on @jbayardo earlier comments, I created a Dictionary version of apply batch to store the db records and also opened the DB in a write-only mode that allows us to set up the DB for "bulkLoading" in the db options. While I do not have any large logs to test this on, the smaller logs led to about 10% improvement: 3.2s -> 2.9s and 1.75s -1.6s.


Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
```diff
@@ -247,6 +243,16 @@ public override int Analyze()
         /// <inheritdoc/>
         public override void Dispose()
         {
+            Analysis.IgnoreResult(
```
jbayardo: You should probably move this inside the RocksDbStore (i.e. when it is done and working in bulk load mode, issue a manual compression on Dispose). There's no scenario in which this isn't needed, and seems like easy to miss.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
```diff
@@ -182,8 +179,7 @@ public XLGToDBAnalyzerInner(AnalysisInput input) : base(input)
         /// <inheritdoc/>
         public override void Prepare()
         {
-            string[] additionalColumns = { XldbDataStore.EventColumnFamilyName, XldbDataStore.PipColumnFamilyName, XldbDataStore.StaticGraphColumnFamilyName };
-            var accessor = KeyValueStoreAccessor.Open(storeDirectory: OutputDirPath, additionalColumns: additionalColumns);
+            var accessor = KeyValueStoreAccessor.Open(storeDirectory: OutputDirPath, additionalColumns: m_additionalColumns);
```
jbayardo: You don't seem to be opening for bulk load, is this correct?

Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
```diff
@@ -148,14 +148,18 @@ private struct DefaultOptions
...
             public RocksDbStore(
                 string storeDirectory,
                 bool defaultColumnKeyTracked = false,
                 IEnumerable<string> additionalColumns = null,
                 IEnumerable<string> additionalKeyTrackedColumns = null,
                 bool readOnly = false,
                 bool dropMismatchingColumns = false,
-                bool rotateLogs = false)
+                bool rotateLogs = false,
+                bool openWriteOnly = false)
```
jbayardo: I wouldn't call this `openWriteOnly`, but `openBulkLoad` or similar (the db can be read when prepared for bulk load, so the name's misleading)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
```diff
@@ -493,105 +489,92 @@ public void IngestAllPips()
...
-
-                for (int j = start; j < end; j++)
-                {
-                    var pipId = pipIds[j];
-                    pipsIngested++;
+                var pipSemistableMap = new Dictionary<byte[], byte[]>();
+                var pipIdMap= new Dictionary<byte[], byte[]>();
 
-                    if (pipsIngested % 100000 == 0)
+                Analysis.IgnoreResult(
```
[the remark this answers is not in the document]
sahilmgandhi: Would the best way be to surround the Use with a try-catch, or check if the return of the Use was a "Failure"?

olkononenko:
>     internal sealed class XLGToDBAnalyzerInner : Analyzer

pls add a comment explaining why we need this

---
Refers to: Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs:140 in 07ee0f5. [](commit_id = 07ee0f5376a4f4dba99b7e572d34850b221ff715, deletion_comment = False)

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/XLGPlusPlus/XLGToDBAnalyzer.cs
```diff
@@ -493,107 +492,99 @@ public void IngestAllPips()
...
-                        pipSemistableValues.Clear();
-                        pipPipIdKeys.Clear();
-                        pipValues.Clear();
-                    }
+                        var pipId = pipIds[j];
+                        pipsIngested++;
 
-                    var hydratedPip = CachedGraph.PipTable.HydratePip(pipId, Pips.PipQueryContext.PipGraphRetrieveAllPips);
-                    var pipType = hydratedPip.PipType;
+                        if (pipsIngested % 100000 == 0)
```
olkononenko: >100000 [](start = 43, length = 6)

nit: make it a const

PR #756, opened 2019-08-16
Add statistics to ContentLocationDatabase's cache

jbayardo:
Adds some statistics over how flushing went to the logging. The counter reporting granularity is way too large vs how many times flushing actually happens.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
@@ -591,14 +591,21 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                     try
                     {
                         await Task.Yield();
-                        await _inMemoryCache.FlushAsync(context);
-                        return BoolResult.Success;
+                        return Result.Success(await _inMemoryCache.FlushAsync(context));
                     }
                     finally
                     {
                         Interlocked.Exchange(ref _cacheUpdatesSinceLastFlush, 0);
                         ResetFlushTimer();
                     }
+                }, extraEndMessage: maybeStatistics => {
+                    if (!maybeStatistics.Succeeded)
+                    {
+                        return string.Empty;
+                    }
+
+                    var statistics = maybeStatistics.Value;
+                    return $"Persisted={statistics.Persisted} Leftover={statistics.Leftover} Growth={statistics.Growth} FlushingTimeMs={statistics.FlushingTime.TotalMilliseconds} CleanupTimeMs={statistics.CleanupTime.TotalMilliseconds}";
                 }).ThrowIfFailure();
         }
 
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
@@ -598,14 +598,15 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                         Interlocked.Exchange(ref _cacheUpdatesSinceLastFlush, 0);
                         ResetFlushTimer();
                     }
-                }, extraEndMessage: maybeStatistics => {
+                }, extraEndMessage: maybeStatistics =>
+                {
                     if (!maybeStatistics.Succeeded)
                     {
                         return string.Empty;
                     }
 
                     var statistics = maybeStatistics.Value;
-                    return $"Persisted={statistics.Persisted} Leftover={statistics.Leftover} Growth={statistics.Growth} FlushingTimeMs={statistics.FlushingTime.TotalMilliseconds} CleanupTimeMs={statistics.CleanupTime.TotalMilliseconds}";
+                    return $"Persisted={statistics.Persisted} Leftover={statistics.Leftover} Growth={statistics.Growth} FlushingTime={statistics.FlushingTime.TotalMilliseconds}ms CleanupTime={statistics.CleanupTime.TotalMilliseconds}ms";
                 }).ThrowIfFailure();
         }
 
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
@@ -598,15 +598,15 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                         Interlocked.Exchange(ref _cacheUpdatesSinceLastFlush, 0);
                         ResetFlushTimer();
                     }
-                }, extraEndMessage: maybeStatistics =>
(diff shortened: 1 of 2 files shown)


dannyvv:
/azp run 
Public BuildXL PR Validation

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -591,14 +591,21 @@ private Task ForceCacheFlushAsync(OperationContext context)
...
-                        await _inMemoryCache.FlushAsync(context);
-                        return BoolResult.Success;
+                        return Result.Success(await _inMemoryCache.FlushAsync(context));
                     }
                     finally
                     {
                         Interlocked.Exchange(ref _cacheUpdatesSinceLastFlush, 0);
                         ResetFlushTimer();
                     }
+                }, extraEndMessage: maybeStatistics => {
```
SergeyTeplyakov: The format is weird:)

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
```diff
@@ -87,23 +89,35 @@ public bool TryGetEntry(ShortHash hash, out ContentLocationEntry entry)
...
             // AcquireAsync is used so as to avoid multiple concurrent tasks just waiting; this way we return the
             // task to the thread pool in between.
             using (await _flushMutex.AcquireAsync())
             {
-                PerformFlush(context);
+                return PerformFlush(context);
             }
         }
 
+        public struct FlushStatistics
```
SergeyTeplyakov: Could you please move this at the end of the file?

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
```diff
@@ -114,6 +128,8 @@ private void PerformFlush(OperationContext context)
                     _cache = new ConcurrentBigMap<ShortHash, ContentLocationEntry>();
                 }
 
+                stopwatch.Start();
```
SergeyTeplyakov: Just curious why you've decided to exclude locking from the duration?

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
```diff
@@ -157,8 +177,15 @@ private void PerformFlush(OperationContext context)
                     }
                 }
 
+                statistics.CleanupTime = stopwatch.Elapsed;
+                statistics.Leftover = _flushingCache.Count;
+
                 _database.Counters[ContentLocationDatabaseCounters.TotalNumberOfCompletedCacheFlushes].Increment();
             }
+
+            statistics.Growth = _cache.Count;
```
SergeyTeplyakov: Should it be a diff between the old count and the new count?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -591,14 +591,21 @@ private Task ForceCacheFlushAsync(OperationContext context)
...
+                    if (!maybeStatistics.Succeeded)
+                    {
+                        return string.Empty;
+                    }
+
+                    var statistics = maybeStatistics.Value;
+                    return $"Persisted={statistics.Persisted} Leftover={statistics.Leftover} Growth={statistics.Growth} FlushingTimeMs={statistics.FlushingTime.TotalMilliseconds} CleanupTimeMs={statistics.CleanupTime.TotalMilliseconds}";
```
SergeyTeplyakov: Minor: I think I would prefer FlushingDuration=1234ms (the same for all the other durations.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
SergeyTeplyakov: I mean: the opening brace should be on the next line.

---
In reply to: [315389444](https://github.com/microsoft/BuildXL/pull/756#discussion_r315389444) [](ancestors = 315389444)

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
```diff
@@ -137,6 +153,10 @@ private void PerformFlush(OperationContext context)
                     actionBlock.CompletionAsync().Wait();
                 }
 
+                statistics.FlushingTime = stopwatch.Elapsed;
+                statistics.Persisted = _flushingCache.Count;
+                stopwatch.Restart();
+
                 _database.Counters[ContentLocationDatabaseCounters.NumberOfPersistedEntries].Add(_flushingCache.Count);
```
SergeyTeplyakov: I think we're mixing two concepts together here: on one hand we have counters already, but on the other hand we track everything separately once again.

I'm thinking, does it make sense to compute one thing based on another?

For instance, I had a similar challenge in ContentLcoationEventStore, and what I did is that I've created new CounterCollection instance and used it instead of using separate stats class.

SergeyTeplyakov: In this case, I think, we'll be using one concept instead of two.

---
In reply to: [315392700](https://github.com/microsoft/BuildXL/pull/756#discussion_r315392700) [](ancestors = 315392700)

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
jbayardo: Because I am only interested in the actual transaction time and RocksDb writing time. Also, the lock:
* Is required for correctness, I couldn't optimize it out even if it was an issue.
* Although it is contended, I'm relying on it being fair; if it wasn't, we'd probably never see a flush happen.
* If really needed (i.e. flush time + cleanup time don't add up to the total time taken by the flush operation), then it can be added later.

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
jbayardo: No, the old cache is cleared as soon as we perform the exchange, and "Growth" is meant to measure "how many entries have been added since the flush started".

This gives us an idea on how much we load we had between flushes, and allows us to infer if we are doing many back-to-back flushes.

Public/Src/Cache/ContentStore/Distributed/NuCache/FlushableCache.cs
jbayardo: I'm not sure I understand: so you're saying I should create a new CounterCollection? Isn't that overkill for a few, transient counters like these?

If you instead mean that I should compute these values based on the counters we already have, this is not possible right now because those don't track the values I'm interested in.

I could certainly add the necessary ones, but I don't see how that helps to log these values at the end of the flush.

PR #758, opened 2019-08-16
Remove BondRPC

semihokur:


dannyvv:
is it now possible to remove the pacakges from config.microsfotinternal.dsc as well i.e. `{ id: "Bond.Rpc.NET", version: "3.2.0" },`

dannyvv:
as well as the 'fake' ones for the public build: `Private\InternalSdk\Bond.Net\module.config.dsc`

---
In reply to: [522182236](https://github.com/microsoft/BuildXL/pull/758#issuecomment-522182236) [](ancestors = 522182236)

dannyvv:
I just did a search and it seems we can, as well as `Bond.Core.NET`. It seems we still need Bond.CSharp and Bond.CSharp.osx-x64 for the cache.

---
In reply to: [522182321](https://github.com/microsoft/BuildXL/pull/758#issuecomment-522182321) [](ancestors = 522182321,522182236)

sahilmgandhi:
🥇

Public/Src/App/Bxl/Args.cs
```diff
@@ -371,9 +371,6 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                                     cacheConfiguration.UseDedupStore = true;
                                 }
                             }),
-                        OptionHandlerFactory.CreateBoolOption(
-                            "enableGrpc",
```
mpysson: >enableGrpc [](start = 29, length = 10)

I think you'll need to keep this option around and have it do nothing since there will be builds lingering around that are configured to pass various values to it. Those can and should be cleaned up but making bxl.exe not blow up on the unrecognized argument will make that process easier

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -273,19 +273,21 @@ private FileContentTable FileContentTable
...
 
-            if (configuration.Distribution.IsGrpcEnabled)
-            {
-                bool grpcHandlerInliningEnabled = GrpcSettings.HandlerInliningEnabled;
+            bool grpcHandlerInliningEnabled = GrpcSettings.HandlerInliningEnabled;
 
 #if NET_CORE
-                // Handler inlining causing deadlock on the mac platform.
-                grpcHandlerInliningEnabled = false;
+            // Handler inlining causing deadlock on the mac platform.
```
mpysson: >// Handler inlining causing deadlock on the mac platform. [](start = 12, length = 57)

This will apply to windows builds too now that we switched them over to NetCore. Can we just remove the directive for simplicity?

Kr1sso: We should go and vet the code base for `#if NET_CORE` now anyway, and use `PLATFORM_OSX` where macOS specific adjustments were done (and `PLATFORM_LINUX` later).

semihokur:
Thanks. deleted more :) 

---
In reply to: [522182674](https://github.com/microsoft/BuildXL/pull/758#issuecomment-522182674) [](ancestors = 522182674,522182321,522182236)

Public/Src/App/Bxl/Args.cs
semihokur: good catch! 

---
In reply to: [315296455](https://github.com/microsoft/BuildXL/pull/758#discussion_r315296455) [](ancestors = 315296455)

Public/Src/Engine/Dll/Engine.cs
semihokur: switched to PLATFORM_OSX

---
In reply to: [315298936](https://github.com/microsoft/BuildXL/pull/758#discussion_r315298936) [](ancestors = 315298936)

semihokur: switched to PLATFORM_OSX

---
In reply to: [315296948](https://github.com/microsoft/BuildXL/pull/758#discussion_r315296948) [](ancestors = 315296948)

PR #764, opened 2019-08-19
Change PutFile to always log the gate's waiting time

jbayardo:


Public/Src/Cache/ContentStore/Interfaces/Results/PutResult.cs
```diff
@@ -107,5 +107,13 @@ public override string ToString()
...
                 : GetErrorString();
         }
+
+        internal class OptionalMetadata
+        {
+            public TimeSpan GateWaitTime;
+            public int GateOccupiedCount;
+        }
+
+        internal OptionalMetadata Metadata;
```
SergeyTeplyakov: Could you please property instead of using a field?

Public/Src/Cache/ContentStore/Interfaces/Results/PutResult.cs
```diff
@@ -107,5 +107,13 @@ public override string ToString()
                 ? $"Success Hash={ContentHash.ToShortString()} Size={ContentSize}{this.GetDiagnosticsMessageForTracing()}"
                 : GetErrorString();
         }
+
+        internal class OptionalMetadata
```
SergeyTeplyakov: I don't think that the type name should reflect that the data is optional.

When the property is mutable it kind of implies that the data is optional.

I think you can name it ExtraMetadata.

Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
```diff
@@ -211,7 +211,16 @@ protected ContentSessionBase(string name)
                     Tracer,
                     () => PutFileCoreAsync(operationContext, hashType, path, realizationMode, urgencyHint, _counters[ContentSessionBaseCounters.PutFileRetries]),
                     extraStartMessage: $"({path},{realizationMode},{hashType}) trusted=false",
-                    extraEndMessage: _ => "trusted=false",
+                    extraEndMessage: (result) => {
```
SergeyTeplyakov: Formatting issue.
The open brace should be on the new line if the lambda is not a single line one.

PR #767, opened 2019-08-19
Bug 1586418: InvalidOperationException: Cannot exceed capacity set to…

kewilli:
Bug 1586418: InvalidOperationException: Cannot exceed capacity set to 0 bytes.

[AB#1586418](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1586418)


Public/Src/Cache/ContentStore/Distributed/Sessions/RecordingStream.cs
```diff
@@ -14,21 +14,36 @@ internal class RecordingStream : Stream
     {
         private readonly byte[] _recordedBytes;
         private readonly Stream _inner;
-        private readonly long _capacity;
+        private readonly long? _capacity;
```
dcwuser: Minor and arguable, but in this case, where nullity is simple and fixed, I would just use _capacity = -1 as a signal, and avoid the Nullable layer.

PR #751, opened 2019-08-16
[WIP] Add metadata store to CASaaS server

jbayardo:
This PR is WIP. I have a few questions, which I have left as comments in the relevant places.

PR #770, opened 2019-08-20
Enable fully remote cache for BuildXL

jbayardo:
This PR enables BuildXL to obtain both content and metadata from CASaaS's process via gRPC.

It depends on https://github.com/microsoft/BuildXL/pull/751 , and has been validated locally by running appropriately configured builds.


Public/Src/Cache/ContentStore/Interfaces/Results/CreateSessionResult.cs
```diff
@@ -78,5 +78,19 @@ public override int GetHashCode()
         {
             return base.GetHashCode() ^ Session.Name.GetHashCode();
         }
+
+        /// <nodoc />
+        public CreateSessionResult<B> Map<B>(Func<T, B> transform)
```
dannyvv: Nit: In .Net ended with 'Select' for 'map' and 'Where' for 'filter'....

Public/Src/Cache/MemoizationStore/Library/Service/ServiceClientCache.cs
```diff
@@ -69,5 +71,30 @@ public new CreateSessionResult<ICacheSession> CreateSession(Context context, str
         {
             return AsyncEnumerable.Empty<StructResult<StrongFingerprint>>();
         }
+
+        /// <inheritdoc />
+        public CreateSessionResult<IReadOnlyMemoizationSession> CreateReadOnlySession(Context context, string name)
+        {
+            return CreateReadOnlySession(context, name, ImplicitPin.None).Map((session) => (IReadOnlyMemoizationSession)session);
```
dannyvv: >( [](start = 78, length = 1)

nit: no need for parens for single arg lambda's.

PR #771, opened 2019-08-20
Fix inconsistency in VSIX naming for marketplace

dannyvv:
This is the last step that will allow us to programatically publish the vscode plugin to the marketplace in our CI pipeline.

### DIFF
--- a/Public/Src/IDE/VsCode/client/package.json
+++ b/Public/Src/IDE/VsCode/client/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "buildxldscript",
+  "name": "dscript",
   "displayName": "BuildXL",
   "description": "Visual Studio Code Extension for BuildXL",
   "icon": "BuildXL.png",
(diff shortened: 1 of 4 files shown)


Public/Src/IDE/VsCode/client/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "buildxldscript",
+  "name": "dscript",
```
dannyvv: >dscript [](start = 11, length = 7)

It turns out that the vscode and the publish scripts look at this name as the identity, but the backend marketplace looks at the extension.vsixmanifest 'id' field in [https://github.com/microsoft/BuildXL/blob/master/Public/Src/IDE/VsCode/pluginTemplate/extension.vsixmanifest](https://github.com/microsoft/BuildXL/blob/master/Public/Src/IDE/VsCode/pluginTemplate/extension.vsixmanifest)

Since that one was already baked into the market place I opted to use that id here rather than update the old, we'd have to ask everybody to uninstall the plugin and then reinstall the plugin again...

PR #772, opened 2019-08-20
Create new OperationContext to avoid cancelling Proactive Copy prematurely

JuanCarlosGI:
Operation was being cancelled >99% of the time because operation context was being cancelled while operation ran asynchronously.


Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -179,7 +179,9 @@ private Task<TResult> PerformPutFileGatedOperationAsync<TResult>(OperationContex
 
             if (putResult.Succeeded && Settings.EnableProactiveCopy)
             {
-                RequestProactiveCopyIfNeededAsync(context, putResult.ContentHash).FireAndForget(context);
+                // Since the rest of the operation is done asynchronously, create new context to stop cancelling operation prematurely.
+                var operationContext = new OperationContext(context.TracingContext);
```
ljcollins25: >new OperationContext(context.TracingContext) [](start = 39, length = 44)

Use OperationContext method from StartupShutdownSlimBase since it will cancel on shutdown.

PR #773, opened 2019-08-20
Add parent to CounterCollection

JuanCarlosGI:
Add notion of parent to CounterCollection. 

Since sessions have their own counters, this is a way for stores to keep track of their sessions, while allowing sessions to have counters independent of other sessions.

This PR's scope is only to allow this scenario to be fulfilled, but actual use of parent CounterCollections will be left for future PRs.

### DIFF
--- a/Public/Src/Utilities/Utilities/Tracing/CounterCollection.cs
+++ b/Public/Src/Utilities/Utilities/Tracing/CounterCollection.cs
@@ -78,7 +78,7 @@ namespace BuildXL.Utilities.Tracing
         /// Creates a collection with the specified number of counters. Each counter in <c>[0, numberOfCounters - 1]</c>
         /// may be accessed with <see cref="AddToCounterInternal"/> or <see cref="GetCounterValueInternal"/>.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Public/Src/Utilities/Utilities/Tracing/CounterCollection.cs
```diff
@@ -349,8 +359,8 @@ static CounterCollection()
         /// Creates a collection with a counter for every value of <typeparamref name="TEnum"/>.
         /// Note that the enum should be dense, since this creates <c>MaxEnumValue - MinEnumValue + 1</c> counters.
         /// </summary>
-        public CounterCollection()
-            : base((ushort)s_counterTypes.Length)
+        public CounterCollection(CounterCollection<TEnum> parent = null)
```
dannyvv: This PR doesn't have any users yet. Totally fine just checking in case you expected there to be.

JuanCarlosGI: That's right. Currently there are no uses for this, but we do have plans for it in the future.

PR #775, opened 2019-08-20
Generate separate DLL for xldb and xldb proto

sahilmgandhi:
[AB#1581030](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1581030)

[AB#1581031](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1581031)

Generate separate DLL for xldb and xldb proto, and also create a nuget package for the xldb dll for developers to easily consume. Furthermore, this PR also contains the "helloworld" app that contains the two previous analyzers that were ported over.


Public/Src/Deployment/NugetPackages.dsc
```diff
@@ -122,6 +122,54 @@ namespace NugetPackages {
         }).deployment
     });
 
+    const xldbnetcorequalifier : BuildXLSdk.DefaultQualifier = {
```
dannyvv: to other reviewers: We have a little bit of cleanup to do here and can reuse it with others. I'll give it some thought and organlize later.

Public/Src/Tools/Execution.Analyzer/BxlAnalyzer.dsc
```diff
@@ -13,13 +13,8 @@ namespace Execution.Analyzer {
...
         rootNamespace: "BuildXL.Execution.Analyzer",
         skipDocumentationGeneration: true,
-        sources: [
-            ...globR(d`.`, "*.cs"),
-            ...GrpcSdk.generate({
-                proto: globR(d`.`, "*.proto"),
-                includes: [importFrom("Google.Protobuf.Tools").Contents.all],
-            }).sources,
-        ],
+        sources: [...globR(d`.`, "*.cs")],
```
dannyvv: No need to create an array and then to a single ... (addRange). 
you can simply write:

```sources: globR(d`.`, "*.cs"),```

File: Public/Src/Tools/Xldb/Xldb.dsc [hunk 3]
```diff
@@ -0,0 +1,26 @@
...
+import * as GrpcSdk from "Sdk.Protocols.Grpc";
+
+namespace Xldb {
+    export declare const qualifier: BuildXLSdk.DefaultQualifier;
+    @@public
+    export const dll = BuildXLSdk.library({
+        assemblyName: "Xldb",
+        rootNamespace: "BuildXL.Xldb",
+        skipDocumentationGeneration: true,
+        sources: [...globR(d`.`, "*.cs")],
```
dannyvv: >[...globR(d`. [](start = 17, length = 13)

same, no need for array and then addrange with an array :_)

Public/Src/Tools/Xldb/XldbDataStore.cs
```diff
@@ -55,7 +53,7 @@ public sealed class XldbDataStore : IDisposable
             else
             {
                 Accessor = null;
-                Console.Error.WriteLine("Could not create an accessor for RocksDB. Accessor is null");
+                Console.Error.WriteLine("Could not create an accessor for RocksDB. Accessor is null!!! " + accessor.Failure.DescribeIncludingInnerFailures());
```
dannyvv: Now that we fixed, we don't need to shout so much :)

Perhpas a single . will be enough ;)

Public/Src/Tools/Xldb.Analyzer/Xldb.Analyzer.dsc
```diff
@@ -0,0 +1,26 @@
...
+import * as GrpcSdk from "Sdk.Protocols.Grpc";
+
+namespace Xldb.Analyzer {
+    export declare const qualifier: BuildXLSdk.DefaultQualifier;
+    @@public
+    export const exe = BuildXLSdk.executable({
+        assemblyName: "xldbanalyzer",
+        rootNamespace: "BuildXL.Xldb.Analyzer",
+        skipDocumentationGeneration: true,
+        sources: [...globR(d`.`, "*.cs")],
```
dannyvv: just `globR` is enough

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
```diff
@@ -0,0 +1,314 @@
...
+    /// <summary>
+    /// Xldb Analyzers that use the rocksDb instanceand the Xldb API to analyze BXL logs
+    /// </summary>
+    public class Program
+    {
+
+        private const string s_eventStatsAnalyzer = "eventstats";
+        private const string s_dumpPipAnalyzer = "dumppip";
+
+        private Dictionary<string, string> m_commandLineOptions = new Dictionary<string, string>();
```
dannyvv: You might want to pass `StringComparer.OrdinalIgnoreCase`

that way `/m` and `/M` work both

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
```diff
@@ -0,0 +1,314 @@
...
+
+            if (!p.ProcessArguments(args))
+            {
+                Console.WriteLine("Invalid arguments passed in, exiting analyzer ...");
+                return 1;
+            }
+
+            if (p.m_commandLineOptions.TryGetValue("/m", out var mode))
+            {
+                if (mode == s_eventStatsAnalyzer)
```
dannyvv:
```suggestion
                if (string.Equals(s_eventStatsAnalyzer, mode, StringComparer.OrdinalIgnoreCase))
```

to handle different casing on the commandline

dannyvv: alternatively if you prefer you could write a switch statement because C# supports a switch on strings like:

```
switch (mode.ToLowerInvariant())
{
    case s_eventSTatsAnalyzer:
        return p.AnalyzeEventStats();
    case s_dumpPipAnalyzer:
        return p.AnalyzeDumpPip();
     default:
           Console.Write($"Valid modes are: /m:{s_eventStatsAnalyzer}, /m:{s_dumpPipAnalyzer}");
           return 1;
}
```

---
In reply to: [315939965](https://github.com/microsoft/BuildXL/pull/775#discussion_r315939965) [](ancestors = 315939965)

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
```diff
@@ -0,0 +1,314 @@
...
+
+            return true;
+        }
+
+        /// <summary>
+        /// Parses the long representation of the semistable hash from the string passed in (eg. PipC623BCE303738C69 -> 14277462926896108000)
+        /// </summary>
+        public bool ParseSemistableHash(string pipHash, out long parsedHash)
+        {
+            var adjustedOption = pipHash.ToUpper().Replace("PIP", "");
```
dannyvv: I recall that someone wrote a helper for this. Let me check...

dannyvv: Yes!!! `Pip.TryParseSemiStableHash`

Now I realize this might be problematic for dependencies, but perhaps a helper for other analyzers would be usefull.

---
In reply to: [315940683](https://github.com/microsoft/BuildXL/pull/775#discussion_r315940683) [](ancestors = 315940683)

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
```diff
@@ -0,0 +1,314 @@
...
+
+        /// <summary>
+        /// Analyzes the event stats from the Xldb instance
+        /// </summary>
+        public int AnalyzeEventStats()
+        {
+            if (m_commandLineOptions.ContainsKey("/h"))
+            {
+                Console.WriteLine("\nEvent Stats Analyzer");
+                Console.WriteLine("Generates stats on the aggregate size and count of execution log events, but uses the RocksDB database as the source of truth");
```
dannyvv: >tes stats on the aggregate size and  [](start = 41, length = 36)

cute how to doe /h with a mode!!! I like it!

Public/Src/Tools/Xldb.Proto/Xldb.Proto.dsc
```diff
@@ -0,0 +1,18 @@
...
+namespace Xldb.Proto {
+
+    @@public
+    export const dll = BuildXLSdk.library({
+        assemblyName: "Xldb.Proto",
+        rootNamespace: "BuildXL.Xldb.Proto",
+        skipDocumentationGeneration: true,
+        sources: [
+            ...GrpcSdk.generate({proto: globR(d`.`, "*.proto"), 
+            includes: [importFrom("Google.Protobuf.Tools").Contents.all]}).sources],
```
dannyvv: would you mind updating the formatting slightly?

```suggestion
        sources: GrpcSdk.generate({
            proto: globR(d`.`, "*.proto"), 
            includes: [importFrom("Google.Protobuf.Tools").Contents.all]
        }).sources,
```

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
sahilmgandhi: The reason i don't have Pip.TryParseSemiStableHash is because I don't want to be dependent on any more Bxl code than the minimal set that I currently am using. 

---
In reply to: [315941026](https://github.com/microsoft/BuildXL/pull/775#discussion_r315941026) [](ancestors = 315941026,315940683)

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
sahilmgandhi: Sure, i can use a switch statement. I left the "valid modes" print after the if statement though so that I would not have to repeat as much code and let both the errors (bad mode, or missing mode) get caught with 1 print statement.

---
In reply to: [315940523](https://github.com/microsoft/BuildXL/pull/775#discussion_r315940523) [](ancestors = 315940523,315939965)

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
sahilmgandhi: I convert everything to lower in ProcessArguments.

---
In reply to: [315939713](https://github.com/microsoft/BuildXL/pull/775#discussion_r315939713) [](ancestors = 315939713)

Public/Src/Deployment/buildXL.dsc
```diff
@@ -22,6 +22,7 @@ namespace BuildXL {
             importFrom("BuildXL.Tools").Execution.Analyzer.exe,
             importFrom("BuildXL.Tools").BxlScriptAnalyzer.exe,
             importFrom("BuildXL.Cache.VerticalStore").Analyzer.exe,
+            importFrom("BuildXL.Tools").Xldb.Analyzer.exe,
```
mpysson: >importFrom("BuildXL.Tools").Xldb.Analyzer.exe, [](start = 12, length = 46)

This is more like a demo than a tool that we want to distribute. Let's put it in out\bin\{release/debug}\tools instead of main deployment

Public/Src/Deployment/NugetPackages.dsc
```diff
@@ -122,6 +122,54 @@ namespace NugetPackages {
...
+        id: `${packageNamePrefix}.Xldb`,
+        deployment: {
+            contents: [ 
+                Nuget.createAssemblyLayout(importFrom("BuildXL.Tools").Xldb.withQualifier(xldbnetcorequalifier).dll),
+                Nuget.createAssemblyLayout(importFrom("BuildXL.Tools").Xldb.Proto.withQualifier(xldbnetcorequalifier).dll),
+                Nuget.createAssemblyLayout(importFrom("BuildXL.Utilities").withQualifier(xldbnetcorequalifier).dll),
+                Nuget.createAssemblyLayout(importFrom("BuildXL.Utilities").KeyValueStore.withQualifier(xldbnetcorequalifier).dll),
+                Nuget.createAssemblyLayout(importFrom("BuildXL.Utilities").Collections.withQualifier(xldbnetcorequalifier).dll),
+                Nuget.createAssemblyLayout(importFrom("BuildXL.Utilities").Native.withQualifier(xldbnetcorequalifier).dll),
+                Nuget.createAssemblyLayout(importFrom("BuildXL.Utilities").Storage.withQualifier(xldbnetcorequalifier).dll),
```
mpysson: Its a bummer to include these and the primary BuildXL.Utilities dlls in the public surface area of this dll. Is it possible to avoid this?

sahilmgandhi: Danny and I tried to create a minimal set of dlls required for this package. Since we use the key-value store accessor, it relies on the native and storage dlls along with several others (the ones here). Removing any of them will render the nuget package unusable (we tried with a separate playground app that imported the new nuget package)

---
In reply to: [316293225](https://github.com/microsoft/BuildXL/pull/775#discussion_r316293225) [](ancestors = 316293225)

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
```diff
@@ -0,0 +1,312 @@
...
+
+        public static int Main(string[] args)
+        {
+            var p = new Program();
+
+            if (!p.ProcessArguments(args))
+            {
+                Console.WriteLine("Invalid arguments passed in, exiting analyzer ...");
+                return 1;
+            }
```
mpysson: We probably don't need to be fancy with multiple mode support. I thought our goal here was to do a demonstration HelloWorld for consuming xlgdb, rather than providing scaffolding for replicating the existing analyzer model on top of a different backend. Is that still the idea?

Public/Src/Deployment/buildXL.dsc
sahilmgandhi: Could you clarify what you mean by this? The exe is generated in out/bin/debug/tools 

---
In reply to: [316292295](https://github.com/microsoft/BuildXL/pull/775#discussion_r316292295) [](ancestors = 316292295)

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
sahilmgandhi: I think this is a demonstration with just 2 "modes" that serve as 2 demos for people to follow/use. It is up to them if they want to add their new analyzers/queries here, or keep it completely separate from our code base. We do not enforce anything or suggest it either. 

---
In reply to: [316299007](https://github.com/microsoft/BuildXL/pull/775#discussion_r316299007) [](ancestors = 316299007)

Public/Src/Tools/Execution.Analyzer/Args.cs
```diff
@@ -634,9 +624,6 @@ private static void WriteHelp()
             //writer.WriteLine("");
             //WriteXLGToDBHelp(writer);
```
olkononenko: - there will be no analyzers here, right? I think we can remove that `todo`
- uncomment the `WriteXLGToDBHelp` ?

sahilmgandhi: Correct, there are no other analyzers. I was going to uncomment out the xllgtodb helper after the "interesting indexing" step

---
In reply to: [316308804](https://github.com/microsoft/BuildXL/pull/775#discussion_r316308804) [](ancestors = 316308804)

Public/Src/Tools/Xldb.Proto/Events.proto
```diff
@@ -7,9 +7,9 @@ import "tools/google/protobuf/timestamp.proto";
 import "tools/google/protobuf/duration.proto";
 import "EnumsAndHelperStructs.proto";
 
-package BuildXL.Execution.Analyzer;
+package Xldb.Proto;
 
-option csharp_namespace = "BuildXL.Execution.Analyzer.Xldb";
+option csharp_namespace = "Xldb.Proto";
```
olkononenko: >Xldb.Proto [](start = 27, length = 10)

I don't like that the namespace is just `Xldb.*`... I think it must be under the `BuildXL`, i.e., `BuildXL.Xldb`

Public/Src/Tools/Xldb/XldbDataStore.cs
```diff
@@ -7,12 +7,11 @@
...
 using BuildXL.Engine.Cache.KeyValueStores;
-using BuildXL.Execution.Analyzer.Xldb;
-using BuildXL.Utilities;
 using Google.Protobuf;
-using PipType = BuildXL.Execution.Analyzer.Xldb.PipType;
+using Xldb.Proto;
+using PipType = Xldb.Proto.PipType;
 
-namespace BuildXL.Analyzers.Core.XLGPlusPlus
+namespace Xldb
```
olkononenko: >Xldb [](start = 10, length = 4)

same comment re namespace

Public/Src/Tools/Xldb.Analyzer/XldbAnalyzer.cs
```diff
@@ -0,0 +1,312 @@
...
+
+namespace Xldb.Analyzer
+{
+    /// <summary>
+    /// Xldb Analyzers that use the rocksDb instanceand the Xldb API to analyze BXL logs
+    /// </summary>
+    public class Program
+    {
+
+        private const string s_eventStatsAnalyzer = "eventstats";
```
olkononenko: nit: nuke the line

Public/Src/Tools/Xldb.Proto/Events.proto
sahilmgandhi: Should we rename the entire dll as well then to BuildXL.Xldb.dll? and similarly the nuget package?

---
In reply to: [316311200](https://github.com/microsoft/BuildXL/pull/775#discussion_r316311200) [](ancestors = 316311200)

PR #776, opened 2019-08-20
Change default timed compaction setting to disabled

jbayardo:
Since this setting is in minutes, it can't be disabled from config right now. Thus, I'm setting it to disabled by default to avoid an (unwanted) release to prod.


Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabaseConfiguration.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Threading;
```
SergeyTeplyakov: Is it needed?

jbayardo: Yes, for the timeout.

PR #778, opened 2019-08-21
Remove the old static VisualzationModel with a new instance based ViewModel that we can use to extend summary information for BuildXL

dannyvv:


Public/Src/Engine/ViewModel/BuildViewModel.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using BuildXL.Pips;
+using BuildXL.Utilities;
+
+namespace BuildXL.ViewModel
+{
+    /// <summary>
+    /// Provides access to internals of the build engine used to provide the user experience.
+    /// This can be for the fancy console to provide progress or to emit azure devops markers.
```
mpysson: >fancy console to provide progress or to emit azure devops markers. [](start = 28, length = 66)

Not the content of this PR but this comment reminds me that /ado mode should disable /fancyconsole

dannyvv: It already does :)

PR #785, opened 2019-08-22
Allow macOS sandbox implementation to be configurable

Kr1sso:
This PR introduces a new sandbox kind value and enables users to chose which macOS sandbox implementation they want to execute with. This will allow for easier transitioning with the release of macOS Catalina. Also a deployment issue for tooling has been fixed.

### DIFF
--- a/Public/Src/Deployment/tools.dsc
+++ b/Public/Src/Deployment/tools.dsc
@@ -24,8 +24,7 @@ namespace Tools {
             ]
         };
     
-        @@public
-        export const deployed = BuildXLSdk.DeploymentHelpers.deploy({
+        const deployed = Context.getCurrentHost().os !== "win" ? undefined : BuildXLSdk.DeploymentHelpers.deploy({
             definition: deployment,
             targetLocation: r`${qualifier.configuration}/tools/XldbAnalyzer/${qualifier.targetRuntime}`,
         });
--- a/Public/Src/Deployment/tools.dsc
+++ b/Public/Src/Deployment/tools.dsc
@@ -16,15 +16,11 @@ namespace Tools {
 
         export const deployment : Deployment.Definition = {
             contents: [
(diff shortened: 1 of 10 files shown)


Public/Src/Deployment/tools.dsc
```diff
@@ -24,8 +24,7 @@ namespace Tools {
             ]
         };
     
-        @@public
-        export const deployed = BuildXLSdk.DeploymentHelpers.deploy({
+        const deployed = Context.getCurrentHost().os !== "win" ? undefined : BuildXLSdk.DeploymentHelpers.deploy({
```
mpysson: >Context.getCurrentHost().os !== "win" ? undefined :  [](start = 25, length = 52)

I would expect this to work on macOS. Is there a reason we aren't deploying it on that platform?

mpysson: @sahilmgandhi

Kr1sso: This builds the detours code, which needs the WindowsSDK dependencies and invokes cl.exe to compile, then link windows binaries. This fails on macOS.

Kr1sso: I've adjusted it to build for all targets, it compiles for and on macOS, so no need to force the qualifier and export the deployment as its also not packaged up. @sahilmgandhi I'm not sure what you intended to do here.

sahilmgandhi: Ahh yea i think we didn't realize/forgot to compile it for mac as well. Thanks for catching it :) 

---
In reply to: [316770978](https://github.com/microsoft/BuildXL/pull/785#discussion_r316770978) [](ancestors = 316770978)

sahilmgandhi:
@sahilmgandhi is added to the review. #Closed

PR #786, opened 2019-08-22
macOS Catalina Beta 6 ES adjustments and improvements

Kr1sso:
This PR adjusts the existing ES sandbox implementation to work with the changes introduced in macOS Catalina Beta 6. It also lays down the foundations for multi ES client support and improves the performance due to improvements around the number of observed ES events.

### DIFF
--- a/Public/Src/Sandbox/MacOs/Interop/Sandbox/ESSandbox.cpp
+++ b/Public/Src/Sandbox/MacOs/Interop/Sandbox/ESSandbox.cpp
@@ -12,19 +12,11 @@
 // initialized below in InitializeEndpointSecuritySandbox (which is called once by the host process)
 static ESSandbox *sandbox;
 
-void processEndpointSecurityEvent(es_client_t *client, const es_message_t *msg, pid_t host)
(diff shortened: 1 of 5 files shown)


Public/Src/Sandbox/MacOs/Interop/Sandbox/ESSandbox.cpp
```diff
@@ -158,10 +150,26 @@ extern "C"
         
         // Subsribe and activate the ES client
         
-        es_return_t status = es_subscribe(client, sandbox->GetSubscibedESEvents(), sandbox->GetSubscribedESEventsCount());
+        es_return_t status = es_subscribe(client, sandbox->GetEventsForType(EventType::ProcessEvent), sandbox->GetEventCountForType(EventType::ProcessEvent));
```
dannyvv: >ndbox->GetEventCountForType(EventType:: [](start = 104, length = 39)

Just curious: Would it make sense to store this in a variable and use in the error messages to help diagnose some of the errors below?

PR #790, opened 2019-08-22
Implement GC strategy more similar to SQLite's for RocksDb Metadata Store

jbayardo:


Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -769,51 +774,66 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
+                var timeToRemove = stopwatch.Elapsed;
+
+                var liveDbSizeInBytesAfterGc = int.Parse(store.GetProperty(
+                    "rocksdb.estimate-live-data-size",
+                    columnFamilyName: nameof(Columns.Metadata)));
+
+                // NOTE(jubayard): since we report the live DB size, it is possible it may increase after GC, because
+                // new tombstones have been added. However, there is way to compute how much we added/removed that
+                // doesn't involve either keeping track of the values, or doing two passes over the column family.
+                Tracer.Debug(context, $"Metadata Garbage Collection results: ScannedEntries={sortedEntries.Count}, RemovedEntries={cutoffIndex}, EntriesAfterGc={sortedEntries.Count - cutoffIndex}, LiveDbSizeInBytesBeforeGc={liveDbSizeInBytesBeforeGc}, LiveDbSizeInBytesAfterGc={liveDbSizeInBytesAfterGc}, TimeToEnumerate={timeToEnumerate.TotalMilliseconds}ms, TimeToSort={timeToSort.TotalMilliseconds}ms, TimeToRemove={timeToRemove.TotalMilliseconds}");
```
ljcollins25: >" [](start = 450, length = 1)

Add units (ms)

Public/Src/Cache/VerticalStore/MemoizationStoreAdapter/CloudStoreLocalCacheServiceFactory.cs
```diff
@@ -159,8 +159,8 @@ private sealed class Config
             [DefaultValue(60 * 60)]
             public int RocksDbMemoizationStoreGarbageCollectionIntervalInSeconds { get; set; }
 
-            [DefaultValue(60 * 60 * 2)]
-            public int RocksDbMemoizationStoreGarbageCollectionProtectionTimeInSeconds { get; set; }
+            [DefaultValue(500_000)]
+            public int RocksDbMemoizationStoreGarbageCollectionMaximumNumberOfEntries { get; set; }
```
ljcollins25: >RocksDbMemoizationStoreGarbageCollectionMaximumNumberOfEntries [](start = 23, length = 62)

Be sure to delete this value from the config in cbtest so we don't break things when we deploy.

jbayardo: I'll make sure to disable the RocksDb in CBTest when we are ready to deploy instead. Otherwise, the GC time will force evictions before a build completes

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -765,55 +771,70 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
+                var liveDbSizeInBytesBeforeGc = int.Parse(store.GetProperty(
+                    "rocksdb.estimate-live-data-size",
+                    columnFamilyName: nameof(Columns.Metadata)));
+
+                var scannedEntries = 0;
+                var removedEntries = 0;
+
+                var entries = new PriorityQueue<(long fileTimeUtc, byte[] strongFingerprint)>(
+                    capacity: _configuration.MetadataGarbageCollectionMaximumNumberOfEntries + 1,
+                    comparer: new SortMetadataByFileTimeUtcComparer());
```
ljcollins25: >new SortMetadataByFileTimeUtcComparer() [](start = 30, length = 39)

Comparer<T>.Create instead of creating a custom type.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -765,55 +771,70 @@ private long DeserializeMetadataLastAccessTimeUtc(byte[] data)
...
                     {
                         break;
                     }
 
-                    scannedEntries++;
-                    Counters[ContentLocationDatabaseCounters.GarbageCollectMetadataEntriesScanned].Increment();
-                    
-                    var lastAccessTimeUtc = DeserializeMetadataLastAccessTimeUtc(kvp.Value);
-                    if (lastAccessTimeUtc < cutoffTimeUtc)
+                    entries.Push((
```
ljcollins25: >entries [](start = 20, length = 7)

One optimization would be to compare to the top element first when the queue is over the limit. If its older, you don't need to add, you can just immediately remove.

PR #791, opened 2019-08-22
Stopwatch isn't threadsafe

kewilli:
Stopwatch isn't thread-safe, so it cannot be shared across multiple content hashers.

### DIFF
--- a/Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
+++ b/Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
@@ -183,8 +183,6 @@ namespace BuildXL.Cache.ContentStore.Hashing
             private static readonly byte[] EmptyByteArray = new byte[0];
             private static readonly Task<bool> TrueTask = Task.FromResult(true);
 
-            private static readonly Stopwatch _sw = Stopwatch.StartNew();
-
             private readonly ActionBlock<Pool<Buffer>.PoolHandle> _hashingBufferBlock;
 
             private readonly Stream _baseStream;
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
```diff
@@ -387,10 +385,9 @@ private Task<bool> PushHashBufferAsync(byte[] buffer, int offset, int count)
 
             private void TransformBlock(byte[] buffer, int offset, int count, byte[] outputBuffer, int outputOffset)
             {
-                var start = _sw.ElapsedTicks;
+                Stopwatch sw = Stopwatch.StartNew();
```
SergeyTeplyakov: I would suggest using StopwatchSlim instead, because Stopwatch.StartNew allocates in the heap on each call.

Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
```diff
@@ -387,10 +385,9 @@ private Task<bool> PushHashBufferAsync(byte[] buffer, int offset, int count)
 
             private void TransformBlock(byte[] buffer, int offset, int count, byte[] outputBuffer, int outputOffset)
             {
-                var start = _sw.ElapsedTicks;
+                Stopwatch sw = Stopwatch.StartNew();
                 _hashAlgorithm.TransformBlock(buffer, offset, count, outputBuffer, outputOffset);
-                var elapsedTicks = _sw.ElapsedTicks - start;
-                Interlocked.Add(ref _ticksSpentHashing, elapsedTicks);
+                Interlocked.Add(ref _ticksSpentHashing, sw.ElapsedTicks);
```
SergeyTeplyakov: And, btw, why do you think that Stopwatch is not threadsafe?

Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
kewilli: StopwatchSlim is defined in ContentStore.Library, a dependent of ContentStore.Hashing.

Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
kewilli: Discussed offline. `timeSpentHashing` is sometimes listed with the exact same elapsed time as neighboring copies. This is particularly strange when the overall copy time is less than the hashing time.

After our discussion, we agreed that we don't know what has happened here, but there's something fishy about `ElapsedTicks` (`1/Frequency` vs `Elapsed`'s count of 100 nanosecond periods) and we're willing to combine behind the more commonly used `Elapsed` field.

PR #792, opened 2019-08-23
More macOS Catalina EndpointSecurity adjustments

Kr1sso:
Take care of Catalina oddities (system data partition prefix) and add probe emitting piggybacking on the lookup event until we get KAuth callback equivalents for Endpoint Security.

### DIFF
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -784,6 +784,12 @@ namespace BuildXL
                             opt =>
                             {
                                 var parsedOption = CommandLineUtilities.ParseEnumOption<SandboxKind>(opt);
+#if PLATFORM_OSX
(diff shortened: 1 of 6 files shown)


Public/Src/App/Bxl/Args.cs
```diff
@@ -784,6 +784,12 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                             opt =>
                             {
                                 var parsedOption = CommandLineUtilities.ParseEnumOption<SandboxKind>(opt);
+#if PLATFORM_OSX
+                                if (parsedOption == SandboxKind.MacOsEndpointSecurity && !OperatingSystemHelper.IsMacOSCatalinaOrHigher)
```
dannyvv: >EndpointSecurity [](start = 69, length = 16)

No I think I know what ES stands for :) I always read Engineering System :P

Kr1sso: haha, yeah its EndpointSecurity - sorry 💃

PR #796, opened 2019-08-23
Move Bxp electron app to tools folter out from the buildxl released bits.

dannyvv:


olkononenko:
`Move Bxp electron app` <- typo? should this be `Bxl` ?

Public/Src/Deployment/buildXL.dsc
```diff
@@ -31,24 +31,16 @@ namespace BuildXL {
             ...addIfLazy(qualifier.targetRuntime !== "osx-x64", () => [{
                 subfolder: r`tools`,
                 contents: [
```
olkononenko: >contents: [ [](start = 16, length = 11)

I might be missing something here. we used to place two dirs - 'bxp' and 'bxp-server'. Now we are placing only 'bxp-server'

Public/Src/Deployment/tools.dsc
```diff
@@ -66,4 +66,31 @@ namespace Tools {
             targetLocation: r`${qualifier.configuration}/tools/Orchestrator/${qualifier.targetRuntime}`
         });
     }
+
+    namespace BuildExplorer {
```
olkononenko: >BuildExplorer [](start = 14, length = 13)

Is it used anywhere?

PR #813, opened 2019-08-29
Printing producer pip description when we have multiple consumers

rijulluman-msft:
[AB#1586413](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1586413)

### DIFF
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -5458,7 +5458,7 @@ namespace BuildXL.Scheduler
         }
 
         /// <summary>
-        /// Gets the first consumer description associated with a FileOrDirectory artifact.
+        /// Gets the consumer description if and only if a single consumer is associated with a FileOrDirectory artifact.
         /// </summary>
         /// <param name="artifact">The artifact for which to get the first consumer description.</param>
         /// <returns>The first consumer description or null if there is no consumer.</returns>
@@ -5466,14 +5466,18 @@ namespace BuildXL.Scheduler
         public string GetConsumerDescription(in FileOrDirectoryArtifact artifact)
         {
             var producerId = PipGraph.GetProducer(artifact);
+            string singleConsumer = null;
             foreach (var consumerEdge in DataflowGraph.GetOutgoingEdges(producerId.ToNodeId()))
             {
+                if (singleConsumer != null) {
+                    return null;    // More than 1 consumers
+                }
                 Pip consumer = PipGraph.GetPipFromPipId(consumerEdge.OtherNode.ToPipId());
-                return consumer.GetDescription(Context);
+                singleConsumer = consumer.GetDescription(Context);
             }
 
-            // No consumer
-            return null;
+            // Single consumer description or null for 0 consumers
+            return singleConsumer;
         }
 
         [SuppressMessage("Microsoft.Design", "CA1033:InterfaceMethodsShouldBeCallableByChildTypes")]


Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -2818,7 +2818,7 @@ private string GetAssociatedPipDescription(FileOrDirectoryArtifact declaredArtif
                 if (declaredArtifact.FileArtifact.IsSourceFile)
                 {
                     consumerDescription = m_host.GetConsumerDescription(declaredArtifact);
```
ljcollins25: >GetConsumerDescription [](start = 49, length = 22)

This is the only place where this method is called. Maybe change its behavior to return null if there is more than one consumer (and update comment on IFileContentManagerHost.GetConsumerDescription to represent new behavior of the API.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5476,6 +5476,29 @@ public string GetConsumerDescription(in FileOrDirectoryArtifact artifact)
             return null;
         }
 
+        /// <summary>
+        /// Returns true when associated with a FileOrDirectory artifact has more than 1 consumers.
+        /// </summary>
+        /// <param name="artifact">The artifact for which to find consumer count > 1.</param>
+        /// <returns>The true when consumer count = 0 or 1 or false if consumer count > 1.</returns>
+        [SuppressMessage("Microsoft.Design", "CA1033:InterfaceMethodsShouldBeCallableByChildTypes")]
+        public bool HasMultipleConsumers(in FileOrDirectoryArtifact artifact)
```
ljcollins25: >HasMultipleConsumers [](start = 20, length = 20)

Per my other suggestion, this method is no longer needed, but for future reference you can get the number of outgoing/incoming edges by calling DataflowGraph.GetOutgoingEdgesCount(nodeId)

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5458,22 +5458,26 @@ string IFileContentManagerHost.GetProducerDescription(in FileOrDirectoryArtifact
...
         /// <returns>The first consumer description or null if there is no consumer.</returns>
         [SuppressMessage("Microsoft.Design", "CA1033:InterfaceMethodsShouldBeCallableByChildTypes")]
         public string GetConsumerDescription(in FileOrDirectoryArtifact artifact)
         {
             var producerId = PipGraph.GetProducer(artifact);
+            string singleConsumer = null;
             foreach (var consumerEdge in DataflowGraph.GetOutgoingEdges(producerId.ToNodeId()))
             {
+                if (singleConsumer != null) {
+                    return null;    // More than 1 consumers
```
olkononenko: >// More than 1 consumers [](start = 36, length = 24)

I don't think this is a right approach. Now all the callers need to handle this 'null'. 
This method is called in `GetAssociatedPipDescription` - multiple consumers -> returning `null` -> the caller will have to search for description of a producer... How does it align with the error logging issue this PR claims to be addressing?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5458,22 +5458,26 @@ string IFileContentManagerHost.GetProducerDescription(in FileOrDirectoryArtifact
...
         /// <param name="artifact">The artifact for which to get the first consumer description.</param>
         /// <returns>The first consumer description or null if there is no consumer.</returns>
         [SuppressMessage("Microsoft.Design", "CA1033:InterfaceMethodsShouldBeCallableByChildTypes")]
         public string GetConsumerDescription(in FileOrDirectoryArtifact artifact)
         {
             var producerId = PipGraph.GetProducer(artifact);
+            string singleConsumer = null;
             foreach (var consumerEdge in DataflowGraph.GetOutgoingEdges(producerId.ToNodeId()))
             {
+                if (singleConsumer != null) {
```
almiliMSFT: nit:
```suggestion
                if (singleConsumer != null)
                {
```

Public/Src/Engine/Scheduler/Scheduler.cs
almiliMSFT: The callers had to handle `null` before too

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5458,22 +5458,26 @@ string IFileContentManagerHost.GetProducerDescription(in FileOrDirectoryArtifact
...
+                }
                 Pip consumer = PipGraph.GetPipFromPipId(consumerEdge.OtherNode.ToPipId());
-                return consumer.GetDescription(Context);
+                singleConsumer = consumer.GetDescription(Context);
             }
 
-            // No consumer
-            return null;
+            // Single consumer description or null for 0 consumers
+            return singleConsumer;
```
almiliMSFT: I don't know if this is the right change, but if it is, can you please write it in a less convoluted way, e.g.,:
```csharp
    var producerId = PipGraph.GetProducer(artifact);
    var numConsumers = DataflowGraph.GetOutgoingEdgesCount(producerId.ToNodeId());
    if (numConsumers == 1)
    {
        var consumerPipId = DataflowGraph.GetOutgoingEdges(producerId.ToNodeId()).First().OtherNode.ToPipId();
        var consumer = PipGraph.GetPipFromPipId(consumerPipId);
        return consumer.GetDescription(Context);
    }
    else
    {
        return null;
    }
```
No `for` loop is needed if we only care about the first consumer and/or whether there exist more than 1 consumer.

PR #816, opened 2019-08-29
Add parameter for environment variables to be set before analyzing dscript so that they can be fed from the output of another pip

erickulcyk:


Public/Src/Tools/UnitTests/BxlScriptAnalyzer/AnalyzerTest.cs
```diff
@@ -121,10 +122,10 @@ public class AnalyzerTest<TAnalyzer> : TemporaryStorageTestBase
             bool preserveTrivia = false)
         {
             var analyzer = new TAnalyzer();
-
+            var pathTable = new PathTable();
             var args = new Args(
-                config: @"b:\Fake.config.dsc",
-                filter: string.Empty,
+                commandLineConfig: new CommandLineConfiguration() { Startup = { ConfigFile = AbsolutePath.Create(pathTable, @"b:\Fake.config.dsc") } },
```
almiliMSFT:
```suggestion
                commandLineConfig: new CommandLineConfiguration() { Startup = { ConfigFile = AbsolutePath.Create(pathTable, X("b/Fake.config.dsc")) } },
```

almiliMSFT: this should fix the Mac test failures

erickulcyk:
azp run public builds pr validation

PR #817, opened 2019-08-29
Allow minimal graph enumerations to elide absent path probes

ljcollins25:
Allow minimal graph enumerations to elide absent path probes to increase path set stability and reduce path set size.

### DIFF
--- a/Public/Src/FrontEnd/UnitTests/Script.Interpretation/Test.BuildXL.FrontEnd.Script.Interpretation.dsc
+++ b/Public/Src/FrontEnd/UnitTests/Script.Interpretation/Test.BuildXL.FrontEnd.Script.Interpretation.dsc
@@ -25,6 +25,9 @@ namespace Script.Interpretation {
             importFrom("BuildXL.FrontEnd").Sdk.dll,
         ],
         //increase weight for frequent timeout pip
-        runTestArgs: { weight: 2 },
+        runTestArgs: { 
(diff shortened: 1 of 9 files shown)


olkononenko:
I'd put this behind a flag.

Public/Src/FrontEnd/UnitTests/Script.Interpretation/Test.BuildXL.FrontEnd.Script.Interpretation.dsc
```diff
@@ -25,6 +25,8 @@ namespace Script.Interpretation {
             importFrom("BuildXL.FrontEnd").Sdk.dll,
         ],
         //increase weight for frequent timeout pip
-        runTestArgs: { weight: 2 },
+        runTestArgs: { 
+            weight: 2,
```
mpysson: > weight: 2, [](start = 11, length = 11)

It's hard to say how much this actually helps but you might also try increasing this to something more dramatic like 8. Our test timeouts in our internal CloudBuild based runs have actually gotten quite a bit better recently but this assembly still had a few instances of bumping up to our timeout there in the last week

mpysson:
Thanks for adding the "uh oh" arg.

---
In reply to: [526347954](https://github.com/microsoft/BuildXL/pull/817#issuecomment-526347954) [](ancestors = 526347954)

PR #820, opened 2019-08-30
 Add invalid to all protobuf enums and correctly update them

sahilmgandhi:
[AB#1590638](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1590638)

Fix all protobuf enums to have the default 0 value to be invalid (since at some point we may want to use an enum in a key and protobuf does not serialize a value of 0 which could break our forwards/backwards compatibility)


Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto
```diff
@@ -338,176 +350,58 @@ enum PipType{
     PipType_SealDirectory = 9;
 }
 
-enum SealDirectoryType{
-    SealDirectoryType_SealDirectory = 0;
-    CompositeSharedOpaqueDirectory = 1;
-}
-
-enum PipQueryContext {
```
sahilmgandhi: I removed these enums since we don't use them anywhere

Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto
```diff
@@ -988,14 +882,6 @@ message LocationData{
     int32 Position = 3;
 }
 
-message PipDataEntry{
```
sahilmgandhi: Removed this enum since it is no longer needed.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/Xldb/XldbProtobufExtensions.cs
```diff
@@ -679,7 +703,7 @@ public static ProcessPip ToProcessPip(this Process pip, PathTable pathTable, Xld
             var xldbProcessPip = new ProcessPip
             {
                 GraphInfo = parentPip,
-                ProcessOptions = (Options)pip.ProcessOptions,
+                ProcessOptions = pip.ProcessOptions == Process.Options.None ? Options.None : (Options) ((int)pip.ProcessOptions << 1),
```
olkononenko: >pip.ProcessOptions == Process.Options.None ? Options.None : (Options) ((int)pip.ProcessOptions << 1), [](start = 33, length = 101)

I don't think this is correct - if it's zero, you are kipping it as zero, otherwise you shift it?

olkononenko:
>     // Deprecated in favor of ProcessFingerprintComputation (enum 10)

if it's deprecated, why is it still here?

---
Refers to: Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto:32 in b08a88a. [](commit_id = b08a88a35e24e9738210effdbf32ff073bba58b3, deletion_comment = False)

Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto
```diff
@@ -45,15 +44,17 @@ enum ExecutionEventId{
 }
 
 enum PipOutputOrigin{
-    Produced = 0;
-    UpToDate = 1;
-    DeployedFromCache = 2;
-    NotMaterialized = 3;
+    PipOutputOrigin_Invalid = 0;
```
olkononenko: >_Invalid [](start = 19, length = 8)

nit: I don't like '_invalid' suffix. Maybe call it 'uninitialized'...? Don't have a nice suggestion here

Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto
```diff
@@ -338,176 +350,58 @@ enum PipType{
     PipType_SealDirectory = 9;
 }
 
-enum SealDirectoryType{
-    SealDirectoryType_SealDirectory = 0;
-    CompositeSharedOpaqueDirectory = 1;
-}
-
```
olkononenko: we don't use this enum?

Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto
```diff
@@ -338,176 +350,58 @@ enum PipType{
...
-    PipGraphPostValidation = 71;
-    FingerprintStore = 72;
-    PipGraphGetSealDirectoryByKind = 73;
-    GetSealDirectoryFingerprint = 74;
-    Explorer = 75;
-    PreserveOutput = 76;
-    End = 77;
-}
-
 enum Options{
```
olkononenko: >Options [](start = 5, length = 7)

what about `IncrementalTool` value? how do you handle it?

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/Xldb/XldbProtobufExtensions.cs
sahilmgandhi: Options.None has a value of 1 (I did this since if you shift 0, you will get 0, so I force set it to 1, and everything else can be shifted)

Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto
sahilmgandhi: Nope, we only use it serialize/deserialize and dont store it in any member variable for SealDirectories.

---
In reply to: [320543153](https://github.com/microsoft/BuildXL/pull/820#discussion_r320543153) [](ancestors = 320543153)

Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto
sahilmgandhi: I like uninitialized, or even _defaultPlaceholder? or _defaultZero?

---
In reply to: [320542914](https://github.com/microsoft/BuildXL/pull/820#discussion_r320542914) [](ancestors = 320542914)

Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto
sahilmgandhi: Let me go ahead and add that.

---
In reply to: [320543502](https://github.com/microsoft/BuildXL/pull/820#discussion_r320543502) [](ancestors = 320543502)

sahilmgandhi:
>     // Deprecated in favor of ProcessFingerprintComputation (enum 10)

This mirrors the enum in our buildXL code. I left it in here for consistency reason, but if you want i can remove it (since we will never use this enum value as I skip it in the XlgToDb analyzer)

---
In reply to: [527703896](https://github.com/microsoft/BuildXL/pull/820#issuecomment-527703896) [](ancestors = 527703896)

---
Refers to: Public/Src/Tools/Xldb.Proto/EnumsAndHelperStructs.proto:32 in b08a88a. [](commit_id = b08a88a35e24e9738210effdbf32ff073bba58b3, deletion_comment = False)

PR #822, opened 2019-09-03
Add dynamic shim filtering

erikma:
Add capability to load and use a filter DLL for externally controlled approval of substitute shim injection.


erikma:
This is a WIP that will have several parts as I'm working with BuildXL packages and need to iterate. UTs are pending, probably 1-2 PRs from now as integration testing completes.

PR #825, opened 2019-09-03
Allow for error regex to match across multiple lines

almiliMSFT:
[AB#1570216](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1570216)

When multi-line regex matching for error messages is enabled, pip output is processes in chunks (of 10000 lines by default) against which the regex is applied and used to find all matches; those matches are returned verbatim (subject to truncating and and other existing logic)

Previously, pip output was always processed line by line, where the error regex is checked against each line and if any matches are found the whole line is returned/printed to the user.  This is in contrast to multi-line regex matching where only the exact match is printed. 

When multi-line regex matching is not enabled, the behavior remains the same.

### DIFF
--- a/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
@@ -33,7 +33,7 @@ namespace Transformer {
 
         /** Grant permissions permanently, i.e., until the service pip terminates. */
         permanent
-    } 
+    }
 
     @@public
     export interface ExecuteArgumentsCommon extends ExecuteArgumentsComposible {
@@ -83,9 +83,12 @@ namespace Transformer {
 
         /** Regex that would be used to extract warnings from the output. */
         warningRegex?: string;
-        
+
         /** Regex that would be used to extract errors from the output. */
         errorRegex?: string;
+
+        /** Options used to construct error regex (see the RegexOptions C# enum for valid values) */
+        errorRegexOptions?: number;
 
         /** Semaphores to acquire */
         acquireSemaphores?: SemaphoreInfo[];
--- a/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
@@ -87,8 +87,8 @@ namespace Transformer {
         /** Regex that would be used to extract errors from the output. */
         errorRegex?: string;
 
-        /** Options used to construct error regex (see the RegexOptions C# enum for valid values) */
-        errorRegexOptions?: number;
+        /** Whether 'errorRegex' should be matched across multiple lines. */
+        errorRegexEnableMultiLineMatches?: number;
 
         /** Semaphores to acquire */
         acquireSemaphores?: SemaphoreInfo[];
--- a/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
@@ -87,8 +87,23 @@ namespace Transformer {
         /** Regex that would be used to extract errors from the output. */
         errorRegex?: string;
 
-        /** Whether 'errorRegex' should be matched across multiple lines. */
-        errorRegexEnableMultiLineMatches?: boolean;
+        /** 
+         * When false (or not set): process output is scanned for error messages line by line;
+         * 'errorRegex' is applied to each line and if ANY match is found the ENTIRE line is reported.
+         *
+         * When true: process output is scanned in chunks of up to 1000 lines; 'errorRegex' is applied to
+         * each chunk and only the matches are reported.  Furthermore, if 'errorRegex' contains a capture 
+         * group named "ErrorMessage", the value of that group is reported; otherwise, the value of the
+         * entire match is reported.
+         * 
+         *   NOTE: because this scanning is done against chunks of text (instead of the entire process output),
+         *         false negatives are possible if an error message spans across multiple chunks.  The scanning
+         *         is done in chunks because attempting to construct a single string from the entire process
+         *         output can easily lead to OutOfMemory exception.
+         * 
+         * Default: false
+         */
+        enableMultiLineErrorScanning?: boolean;
 
         /** Semaphores to acquire */
         acquireSemaphores?: SemaphoreInfo[];
--- a/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
@@ -91,7 +91,7 @@ namespace Transformer {
          * When false (or not set): process output is scanned for error messages line by line;
          * 'errorRegex' is applied to each line and if ANY match is found the ENTIRE line is reported.
          *
-         * When true: process output is scanned in chunks of up to 1000 lines; 'errorRegex' is applied to
+         * When true: process output is scanned in chunks of up to 10000 lines; 'errorRegex' is applied to
          * each chunk and only the matches are reported.  Furthermore, if 'errorRegex' contains a capture 
          * group named "ErrorMessage", the value of that group is reported; otherwise, the value of the
          * entire match is reported.
--- a/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
+++ b/Public/Sdk/Public/Transformers/Transformer.Execute.dsc
@@ -88,7 +88,7 @@ namespace Transformer {
         errorRegex?: string;
 
         /** Whether 'errorRegex' should be matched across multiple lines. */
-        errorRegexEnableMultiLineMatches?: number;
+        errorRegexEnableMultiLineMatches?: boolean;
 
         /** Semaphores to acquire */
         acquireSemaphores?: SemaphoreInfo[];
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -517,26 +517,77 @@ namespace BuildXL.Processes
             return m_warningRegex.IsMatch(line);
         }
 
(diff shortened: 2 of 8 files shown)


Public/Sdk/Public/Transformers/Transformer.Execute.dsc
```diff
@@ -83,10 +83,13 @@ namespace Transformer {
 
         /** Regex that would be used to extract warnings from the output. */
         warningRegex?: string;
-        
+
         /** Regex that would be used to extract errors from the output. */
         errorRegex?: string;
 
+        /** Options used to construct error regex (see the RegexOptions C# enum for valid values) */
+        errorRegexOptions?: number;
```
mpysson: >errorRegexOptions?: number; [](start = 8, length = 27)

While extensible, it feels a bit funny to expose it this way. I think multiline is the only thing we want to control here. SHould we just expose a bool?

almiliMSFT: I can change it to bool.

almiliMSFT: Changed to bool

Public/Src/FrontEnd/Script/Ambients/Transformers/AmbientTransformer.Process.cs
```diff
@@ -545,7 +547,10 @@ private void ProcessExecuteArguments(Context context, ObjectLiteral obj, Process
             var errorRegex = Converter.ExtractString(obj, m_executeErrorRegex, allowUndefined: true);
             if (errorRegex != null)
             {
-                processBuilder.ErrorRegex = new RegexDescriptor(StringId.Create(context.StringTable, errorRegex), RegexOptions.None);
+                var enableMultiLine = Converter.ExtractOptionalBoolean(obj, m_executeErrorRegexEnableMultiLineMatches);
+                processBuilder.ErrorRegex = new RegexDescriptor(
+                    StringId.Create(context.StringTable, errorRegex),
+                    enableMultiLine == true ? RegexOptions.Singleline : RegexOptions.None);
```
mpysson: >enableMultiLine == true ? RegexOptions.Singleline : RegexOptions.None [](start = 20, length = 69)

This is splitting hairs a bit but our configuration option to process stdout/stedrr line by line isn't exactly the same as someone setting RegexOptions.Singleline. RegexOptions.Singleline is about letting regex matches span line breaks, essentially letting "." match all characters including \n instead of breaking on it. 
[https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options?view=netframework-4.8#Singleline](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options?view=netframework-4.8#Singleline)

Had someone set RegexOptions.Singleline today we wouldn't have honored it. I guess you're using that as a mechanism to thread this context through. I'm wondering if it would be more clear if we added a bool to RegexDescription to more specifically denote that the regex should be applied line by line vs. against the entire stream

mpysson: >RegexOptions.Singleline [](start = 46, length = 23)

Should we just instruct users to specify the inline processing options instead of setting the enum in the engine itself?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -517,26 +517,85 @@ private bool IsWarning(string line)
...
+                    }
+
+                    // An unusually long string causes pathologically slow Regex back-tracking.
+                    // To avoid that, only scan the first 400 characters. That's enough for
+                    // the longest possible prefix: MAX_PATH, plus a huge subcategory string, and an error location.
+                    // After the regex is done, we can append the overflow.
+                    if (line.Length > 400)
+                    {
+                        line = line.Substring(0, 400);
+                    }
```
mpysson: I wonder if we should just remove this. It makes sense for warnings where we process every single output line of every pip. On failures we're less concerned about perf

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ErrorRegexTests.cs
```diff
@@ -0,0 +1,110 @@
...
+* <error>err2</error> * <error>err3</error> *" };
+
+                yield return new object[] { useStdErr, Text, "(?s)<error>[^<]*</error>", MultiLineScanning, @"
+<error> *
+* err1 *
+* </error>
+<error>err2</error>
+<error>err3</error>" };
+
+                yield return new object[] { useStdErr, Text, "(?s)<error>[\\s*]*(?<ErrorMessage>.*?)[\\s*]*</error>", MultiLineScanning, @"
```
mpysson: >ErrorMessage [](start = 83, length = 12)

Thanks for adding this

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
```diff
@@ -83,10 +83,28 @@ namespace Transformer {
...
-        
+
         /** Regex that would be used to extract errors from the output. */
         errorRegex?: string;
 
+        /** 
+         * When false (or not set): process output is scanned for error messages line by line;
+         * 'errorRegex' is applied to each line and if ANY match is found the ENTIRE line is reported.
+         *
+         * When true: process output is scanned in chunks of up to 1000 lines; 'errorRegex' is applied to
```
Kr1sso: Nit: `OutputChunkInLines` is set to 10000

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -246,6 +246,30 @@ public sealed partial class Process : Pip
...
+        /// 
+        ///   NOTE: because this scanning is done against chunks of text (instead of the entire process output),
+        ///         false negatives are possible if an error message spans across multiple chunks.  The scanning
+        ///         is done in chunks because attempting to construct a single string from the entire process
+        ///         output can easily lead to OutOfMemory exception.
+        /// </summary>
+        /// <remarks>
+        /// Regarding fingerprinting: <see cref="ErrorRegex"/> and <see cref="WarningRegex"/> are
+        /// currently part of the process fingerprint, even though those two properties cannot affect
+        /// the outcome (success vs failure) of the process.  This is kind of strange and should probably
```
mpysson: The warning regex is special and should be part of the fingerprint since you could have /warnaserror on. Even without that, the outcome presented to the user could differ based on the warning regex. We evaluate the regex when the pip misses and store some state as to whether a warning should be replayed on a subsequent run so that needs to be in the fingerprint.

It should be safe to remove the error regex from the fingerprint though

almiliMSFT: I'll update the comment

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -252,6 +252,29 @@ public sealed partial class Process : Pip
         [PipCaching(FingerprintingRole = FingerprintingRole.Semantic)]
         public RegexDescriptor ErrorRegex { get; }
 
+        /// <summary>
+        /// When false (or not set): process output is scanned for error messages line by line;
+        /// 'errorRegex' is applied to each line and if ANY match is found the ENTIRE line is reported.
+        /// 
+        /// When true: process output is scanned in chunks of up to 1000 lines; 'errorRegex' is applied to
+        /// each chunk and only the matches are reported.Furthermore, if 'errorRegex' contains a capture
```
almiliMSFT:
```suggestion
        /// each chunk and only the matches are reported. Furthermore, if 'errorRegex' contains a capture
```

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -252,6 +252,29 @@ public sealed partial class Process : Pip
...
+        /// 
+        /// When true: process output is scanned in chunks of up to 1000 lines; 'errorRegex' is applied to
+        /// each chunk and only the matches are reported. Furthermore, if 'errorRegex' contains a capture
+        /// group named "ErrorMessage", the value of that group is reported; otherwise, the value of the
+        /// entire match is reported.
+        /// 
+        ///   NOTE: because this scanning is done against chunks of text (instead of the entire process output),
+        ///         false negatives are possible if an error message spans across multiple chunks.  The scanning
+        ///         is done in chunks because attempting to construct a single string from the entire process
+        ///         output can easily lead to OutOfMemory exception.
```
almiliMSFT:
```suggestion
        ///         output can easily lead to an "out of memory" exception.
```

Public/Sdk/Public/Transformers/Transformer.Execute.dsc
almiliMSFT:
```suggestion
         * When true: process output is scanned in chunks of up to 10000 lines; 'errorRegex' is applied to
```

almiliMSFT: thanks, will fix

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -252,6 +252,29 @@ public sealed partial class Process : Pip
         [PipCaching(FingerprintingRole = FingerprintingRole.Semantic)]
         public RegexDescriptor ErrorRegex { get; }
 
+        /// <summary>
+        /// When false (or not set): process output is scanned for error messages line by line;
+        /// 'errorRegex' is applied to each line and if ANY match is found the ENTIRE line is reported.
+        /// 
+        /// When true: process output is scanned in chunks of up to 1000 lines; 'errorRegex' is applied to
```
almiliMSFT:
```suggestion
        /// When true: process output is scanned in chunks of up to 10000 lines; 'errorRegex' is applied to
```

PR #827, opened 2019-09-03
Compute static fingerprint in the pipgraph fragments

erickulcyk:
Currently, computing pip graph fingerprint is done as the graph is being built in pipgraph.builder.  This is expensive, and so we move it to be done at the time of pip graph fragment creation.  A new field is added in pip, and if it exists, the value of it is used instead of recomputing.  The one caveat is if the pip graph fragments uses sealled directories from another fragment.  This is left as a todo because Office currently does not exercise this functionality.

### DIFF
--- a/Public/Src/Engine/Scheduler/Graph/GraphFragmentBuilderTopSort.cs
+++ b/Public/Src/Engine/Scheduler/Graph/GraphFragmentBuilderTopSort.cs
@@ -43,8 +43,8 @@ namespace BuildXL.Scheduler.Graph
             var extraFingerprintSalts = new ExtraFingerprintSalts(
                 configuration,
                 PipFingerprintingVersion.TwoPhaseV2,
-                string.Empty,
-                searchPathToolsHash: null);
+                configuration.Cache.CacheSalt,
+                new DirectoryMembershipFingerprinterRuleSet(configuration, pipExecutionContext.StringTable).ComputeSearchPathToolsHash());
 
             m_pipStaticFingerprinter = new PipStaticFingerprinter(
                 pipExecutionContext.PathTable,
@@ -82,7 +82,7 @@ namespace BuildXL.Scheduler.Graph
             var result = base.AddCopyFile(copyFile, valuePip);
             AddFileDependent(copyFile.Source, copyFile);
             m_fileProducers[copyFile.Destination] = copyFile.PipId;
-            copyFile.StaticFingerPrint = m_pipStaticFingerprinter.ComputeWeakFingerprint(copyFile).Hash;
+            copyFile.StaticFingerprint = m_pipStaticFingerprinter.ComputeWeakFingerprint(copyFile).Hash;
             return result;
         }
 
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/Scheduler/Graph/GraphFragmentBuilderTopSort.cs
```diff
@@ -29,12 +32,48 @@ public class GraphFragmentBuilderTopSort : GraphFragmentBuilder
...
         /// <summary>
         /// Creates an instance of <see cref="GraphFragmentBuilder"/>.
         /// </summary>
         public GraphFragmentBuilderTopSort(LoggingContext loggingContext, PipExecutionContext pipExecutionContext, IConfiguration configuration) : base(loggingContext, pipExecutionContext, configuration)
         {
+            var extraFingerprintSalts = new ExtraFingerprintSalts(
+                configuration,
+                PipFingerprintingVersion.TwoPhaseV2,
+                string.Empty,
+                searchPathToolsHash: null);
```
erickulcyk: todo: figure this out.

Public/Src/Pips/Dll/Operations/Pip.cs
```diff
@@ -195,6 +206,16 @@ internal void Serialize(PipWriter writer)
             writer.Write((byte)PipType);
             writer.Start(GetType());
             InternalSerialize(writer);
+            if (StaticFingerprint.Length >0)
```
ljcollins25: >>0 [](start = 41, length = 2)

Spacing.

ljcollins25:
>     public class GraphFragmentBuilderTopSort : GraphFragmentBuilder

Why isn't this logic in the base type? Also, it should be configurable.

---
Refers to: Public/Src/Engine/Scheduler/Graph/GraphFragmentBuilderTopSort.cs:30 in a73706d. [](commit_id = a73706d32781e3c05ad81d63f74c98e9dc87cf15, deletion_comment = False)

erickulcyk:
>     public class GraphFragmentBuilderTopSort : GraphFragmentBuilder

It needs the graph that is built up in this class, so it can't be in the base class.  Perhaps this class should be renamed so it implies more than top sort.

---
In reply to: [527708247](https://github.com/microsoft/BuildXL/pull/827#issuecomment-527708247) [](ancestors = 527708247)

---
Refers to: Public/Src/Engine/Scheduler/Graph/GraphFragmentBuilderTopSort.cs:30 in a73706d. [](commit_id = a73706d32781e3c05ad81d63f74c98e9dc87cf15, deletion_comment = False)

Public/Src/Engine/Scheduler/Graph/GraphFragmentBuilderTopSort.cs
```diff
@@ -29,12 +32,48 @@ public class GraphFragmentBuilderTopSort : GraphFragmentBuilder
...
+                PipFingerprintingVersion.TwoPhaseV2,
+                configuration.Cache.CacheSalt,
+                new DirectoryMembershipFingerprinterRuleSet(configuration, pipExecutionContext.StringTable).ComputeSearchPathToolsHash());
+
+            m_pipStaticFingerprinter = new PipStaticFingerprinter(
+                pipExecutionContext.PathTable,
+                GetSealDirectoryFingerprint,
+                GetDirectoryProducerFingerprint,
+                extraFingerprintSalts,
+                PathExpander.Default)
```
ljcollins25: >PathExpander.Default [](start = 16, length = 20)

I think you need a real mount path expander here. System mounts get tokenized with the real mount path expander.

PR #830, opened 2019-09-04
Add sdk-level nullability options of csharp projects

SergeyTeplyakov:
Add an option to csc and managed sdk to enable nullable reference types for the entire project.

### DIFF
--- a/Public/Sdk/Public/Managed/Tools/Csc/csc.dsc
+++ b/Public/Sdk/Public/Managed/Tools/Csc/csc.dsc
@@ -94,6 +94,8 @@ export function compile(inputArgs: Arguments) : Result {
         Cmd.sign("/debug",                 args.emitDebugInformation),
         Cmd.sign("/highentropyva",         args.highEntropyVa),
         Cmd.option("/subsystemversion:",   args.subSystemVersion),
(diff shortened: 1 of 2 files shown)


Public/Sdk/Public/Managed/Tools/Csc/csc.dsc
```diff
@@ -266,6 +268,11 @@ export interface Arguments extends Transformer.RunnerArguments{
     defines?: string[];
     /** This option instructs the compiler to only use ISO-1 C# language features, i.e which basically boils down to C# 1.0 language features. */
     languageVersion?: string;
+    
+    /** Specify nullable context option enable|disable. */
+    nullable?: boolean;
+    /** Specify nullable context option enable|disable|safeonly|warnings|safeonlywarnings.*/
+    nullabilityContext?: NullabilityContext;
```
almiliMSFT: Why do you need both `nullable` and `nullabilityContext`?  It looks like `nullabilityContext` subsumes `nullable`.

SergeyTeplyakov: I think 'nullable: bool' is just a shorthand for nullabilityContext: NullabilityContext.enabled.

I decided to expose the same options that csc exposes.

PR #831, opened 2019-09-04
Don't fingerprint GlobalUnsafePassthroughEnvironmentVariables

mpysson:
Having these environment variables accrue to the weak fingerprint prevents dev cache from being effective and adds unnecessary churn to pip fingerprints

Fixes [AB#1593520](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1593520)


Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/BaselineTests.cs
```diff
@@ -1097,6 +1099,32 @@ public void MoveDirectoryFailed()
             SetExpectedFailures(2, 0);
         }
 
+        /// <summary>
+        /// Test to validate that global passthrough environment variables are visible to processes
+        /// </summary>
+        [Fact]
+        public void GlobalPassthroughEnvironmentVariables()
+        {
+            string testEnvironmentVariable = "ENV" + Guid.NewGuid().ToString().Replace("-", string.Empty);
```
norahuang: Should we add one more step in this case in the beginning, which testEnvironmentVariable is not set at all?

mpysson: Good suggestion. To save on test execution time I'll just set and query for another environment variable that I'm not passing in the global untracked scopes to make sure it isn't visible.

PR #832, opened 2019-09-04
Redis memoization store

JuanCarlosGI:
Implement a memoization store backed by Redis. This implementation is independent of content, and is just concerned with metadata, even though it would now be possible to collaborate with content stores to make sure that everything is consistent.


Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
```diff
@@ -0,0 +1,192 @@
...
+using StackExchange.Redis;
+using BuildXL.Cache.ContentStore.UtilitiesCore.Internal;
+using System.Security.Cryptography;
+
+namespace BuildXL.Cache.MemoizationStore.Distributed.Stores
+{
+    /// <summary>
+    /// Implements a memoization database through Redis.
+    /// </summary>
+    internal class RedisMemoizationDatabase : MemoizationDatabase
```
jbayardo: Question: there seems not to be any garbage collection mechanism for Redis. Is it not needed?

jbayardo: sealed?

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        /// <inheritdoc />
+        public override async Task<Result<LevelSelectors>> GetLevelSelectorsAsync(OperationContext context, Fingerprint weakFingerprint, int level)
+        {
+            var weakFingerprintKey = GetKey(weakFingerprint);
+
+            var redisValues = await _redis.GetHashKeysAsync(context, weakFingerprintKey, context.Token);
+
+            var result = new List<Selector>(redisValues.Length / 2);
```
jbayardo: Question: Why is it sized to half the results?

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
```diff
@@ -0,0 +1,192 @@
...
+                    byte[] tokenFieldNameBytes = SerializeSelector(strongFingerprint.Selector, isHash: true);
+
+
+                    var result = await _redis.ExecuteBatchAsync(context, batch => batch.CompareExchangeAsync(key, selectorBytes, tokenFieldNameBytes, expectedReplacementToken, replacementBytes, newReplacementToken), RedisOperation.CompareExchange);
+                    return new Result<bool>(result);
+                });
+        }
+
+        /// <inheritdoc />
+        public override Task<IEnumerable<StructResult<StrongFingerprint>>> EnumerateStrongFingerprintsAsync(OperationContext context)
```
jbayardo: Question: Why isn't the operation supported? Is it OK not to support it?

Public/Src/Cache/ContentStore/Distributed/Redis/Scripts/CompareExchange.lua
```diff
@@ -0,0 +1,18 @@
+﻿-- bool CompareExchange(string: weakFingerprintKey, byte[]: selectorFieldName, byte[] tokenFieldName, string expectedToken, byte[] contentHashList, string newReplacementToken)
+
+local weakFingerprintKey = KEYS[1];
+local selectorFieldName = ARGV[1];
+local tokenFieldName = ARGV[2];
+local expectedToken = ARGV[3];
+local contentHashList = ARGV[4];
+local newReplacementToken = ARGV[5];
+
+local actualToken = redis.call("HGET", weakFingerprintKey, tokenFieldName);
```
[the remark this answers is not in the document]
JuanCarlosGI: From the redis docs: A Redis script is transactional by definition, so everything you can do with a Redis transaction, you can also do with a script, and usually the script will be both simpler and faster.

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
JuanCarlosGI: Everything in Redis has TTL and will be automatically garbage collected

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
JuanCarlosGI: Because of the schema that we chose. To avoid comparing byte[], we upload a hash of the contenthashlist, to be able to only compare strings.

You can imagine the schema to be: WeakFingerprint -> Dictionary<Selector, (ContentHashList:byte[], Hash:string)>, with the difference that, since Redis does not have the concept of value tuple, we duplicate each Selector, prefixed with either a 1 or a 0, depending if we're storing the ContentHasList or the hash

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
JuanCarlosGI: The operation is only used for testing, and nothing more. It's ok to not implement this, as long as we disable the corresponding tests that use this. The remaining tests are sufficient to test the functionality of the store.

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
ljcollins25: AFAIK, that is not true. We need to set the TTL in the batch which performs the CompareExchange.

---
In reply to: [320995030](https://github.com/microsoft/BuildXL/pull/832#discussion_r320995030) [](ancestors = 320995030)

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
ljcollins25: Rather than a comment in the PR, add this to the a comment on the method or class. Also, its a guid rather than a hash of the content hash list now.

---
In reply to: [320996591](https://github.com/microsoft/BuildXL/pull/832#discussion_r320996591) [](ancestors = 320996591)

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
jbayardo: I would probably explain this in a comment, given that the other stores do implement it, but feel free to ignore.

---
In reply to: [320996981](https://github.com/microsoft/BuildXL/pull/832#discussion_r320996981) [](ancestors = 320996981)

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisCacheFactory.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        private readonly TimeSpan _memoizationExpiryTime;
+
+        /// <nodoc />
+        public RedisCacheFactory(
+            IConnectionStringProvider contentConnectionStringProvider,
+            IConnectionStringProvider machineLocationConnectionStringProvider,
+            IClock clock,
+            TimeSpan contentHashBumpTime,
+            TimeSpan memoizationExpiryTime,
```
jbayardo: >memoization [](start = 21, length = 11)

metadata

Public/Src/Cache/DistributedCache.Host/Service/Internal/CacheServerFactory.cs
```diff
@@ -54,26 +54,26 @@ public CacheServerFactory(DistributedCacheServiceArguments arguments)
...
             }
         }
 
-        private (LocalContentServer, LocalCacheServer) CreateLocalServer(LocalServerConfiguration localServerConfiguration, bool enableMetadataStore)
+        private (LocalContentServer, LocalCacheServer) CreateLocalServer(LocalServerConfiguration localServerConfiguration, DistributedContentSettings distributedSettings)
         {
             Func<AbsolutePath, IContentStore> contentStoreFactory = path => ContentStoreFactory.CreateContentStore(_fileSystem, path, evictionAnnouncer: null, distributedEvictionSettings: default, contentStoreSettings: default, trimBulkAsync: null);
 
-            if (enableMetadataStore)
+            if (distributedSettings.EnableMetadataStore)
```
jbayardo: >distributedSettings. [](start = 16, length = 20)

distributedSettings may be null

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
ljcollins25: Assigning a TTL to keys via KeyExpire handle garbage collection.

---
In reply to: [320977839](https://github.com/microsoft/BuildXL/pull/832#discussion_r320977839) [](ancestors = 320977839)

Public/Src/Cache/MemoizationStore/Distributed/Stores/RedisMemoizationDatabase.cs
ljcollins25: This is an expensive operation in Redis. Yes, its OK not to support it. There can be unbounded amount of data returned by this call so it is not supported in some cases. Its mostly for some niche uses like externally managed garbage collection.

---
In reply to: [320978949](https://github.com/microsoft/BuildXL/pull/832#discussion_r320978949) [](ancestors = 320978949)

Public/Src/Cache/DistributedCache.Host/Configuration/DistributedContentSettings.cs
```diff
@@ -40,6 +40,11 @@ public class DistributedContentSettings
         [JsonConstructor]
         private DistributedContentSettings()
         {
+            if (MemoizationExpiryTimeMinutes == 0)
+            {
+                // Sync memoization with content unless specified.
+                MemoizationExpiryTimeMinutes = ContentHashBumpTimeMinutes;
+            }
```
ljcollins25: ContentHashBumpTimeMinutes is typically set after construction. This logic should be in the property getter with a backing field.

Public/Src/Cache/DistributedCache.Host/Configuration/DistributedContentSettings.cs
```diff
@@ -94,6 +99,12 @@ private DistributedContentSettings(IDictionary<string, string> connectionSecretN
         [DataMember]
         public int ContentHashBumpTimeMinutes { get; set; } = 2880;
 
+        /// <summary>
+        /// TTL to be set in Redis for memoization entries.
+        /// </summary>
+        [DataMember]
+        public int MemoizationExpiryTimeMinutes { get; set; }
```
ljcollins25: >MemoizationExpiryTimeMinutes [](start = 19, length = 28)

Prepend with Redis so its clear.

PR #841, opened 2019-09-06
Set a limit for reconnection in case of recoverable states

semihokur:


Public/Src/Engine/Dll/Distribution/Grpc/ClientConnectionManager.cs
```diff
@@ -69,20 +69,43 @@ public async Task MonitorConnectionAsync()
...
+        private async Task<bool> TryReconnect()
+        {
+            int numRetries = 0;
+            bool connectionSucceeded = false;
+
+            while (numRetries < GrpcSettings.MaxRetry)
+            {
+                numRetries++;
+
+                // Try connecting with timeout (default 5min)
```
olkononenko: >(default 5min) [](start = 47, length = 14)

nit: I wouldn't put this here; if we change the defaults later, we'll need it to update the comment as well...

Public/Src/Engine/Dll/Distribution/Grpc/ClientConnectionManager.cs
```diff
@@ -69,20 +69,43 @@ public async Task MonitorConnectionAsync()
...
+        {
+            int numRetries = 0;
+            bool connectionSucceeded = false;
+
+            while (numRetries < GrpcSettings.MaxRetry)
+            {
+                numRetries++;
+
+                // Try connecting with timeout (default 5min) 
+                connectionSucceeded = await TryConnectChannelAsync(GrpcSettings.CallTimeout, nameof(MonitorConnectionAsync));
```
olkononenko: >MonitorConnectionAsync [](start = 100, length = 22)

nit: `TryReconnect` ?

Public/Src/Engine/Dll/Distribution/Grpc/ClientConnectionManager.cs
```diff
@@ -69,20 +69,43 @@ public async Task MonitorConnectionAsync()
...
+                if (connectionSucceeded)
+                {
+                    return true;
+                }
+
+                if (!connectionSucceeded && IsNonRecoverableState(Channel.State))
```
olkononenko: >if [](start = 16, length = 2)

nit: `else if (IsNonRecoverableState(Channel.State)) `

PR #844, opened 2019-09-06
Add ProcativeCopyResult to aggregate children copies

JuanCarlosGI:
Add ProcativeCopyResult to aggregate children copies. This is for better tracing.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Sessions/ProactiveCopyResult.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Sessions/ProactiveCopyResult.cs
@@ -1,0 +1,66 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using BuildXL.Cache.ContentStore.Interfaces.Results;
+
+namespace BuildXL.Cache.ContentStore.Distributed.Sessions
+{
+    /// <nodoc />
+    internal class ProactiveCopyResult : ResultBase
+    {
+        /// <nodoc />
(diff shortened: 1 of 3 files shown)


Public/Src/Cache/ContentStore/Distributed/Sessions/ProactiveCopyResult.cs
```diff
@@ -0,0 +1,66 @@
...
+            _ringMachineResult = ringCopyResult;
+            _outsideRingMachineResult = outsideRingCopyResult;
+        }
+
+        private static string GetErrorMessage(BoolResult ringCopyResult, BoolResult outsideRingCopyResult)
+        {
+            if (!ringCopyResult.Succeeded || !outsideRingCopyResult.Succeeded)
+            {
+                return
+                    $"Success count: {(ringCopyResult.Succeeded ^ outsideRingCopyResult.Succeeded ? 1 : 0)} " +
```
SergeyTeplyakov: It should be possible to have 2 here as well. Right?

SergeyTeplyakov: Never mind, this is only for the error case.

---
In reply to: [321931539](https://github.com/microsoft/BuildXL/pull/844#discussion_r321931539) [](ancestors = 321931539)

SergeyTeplyakov: It means that at least one operation has failed. So we're good here.

---
In reply to: [321931621](https://github.com/microsoft/BuildXL/pull/844#discussion_r321931621) [](ancestors = 321931621,321931539)

PR #845, opened 2019-09-06
Share stores with Grpc

jbayardo:
Creating a LocalCacheServer would force creation of 2 DistributedContentStores, one of which wouldn't be initialized.

The uninitialized one was given to the GrpcContentServer, and the initialized one would be used by the LocalCacheServer. The GrpcContentServer only ever uses the ContentStore given when doing grpc copying, so all other operations would run just fine.

### DIFF
--- a/Public/Src/Cache/MemoizationStore/Library/Service/LocalCacheServer.cs
+++ b/Public/Src/Cache/MemoizationStore/Library/Service/LocalCacheServer.cs
@@ -42,15 +42,9 @@ namespace BuildXL.Cache.MemoizationStore.Service
         : base(logger, fileSystem, scenario, cacheFactory, localContentServerConfiguration)
         {
             var storesByName = new Dictionary<string, IContentStore>();
-            foreach (var kvp in localContentServerConfiguration.NamedCacheRoots)
+            foreach (var kvp in StoresByName)
             {
-                AbsolutePath cacheRootPath = kvp.Value;
-                fileSystem.CreateDirectory(cacheRootPath);
-
-                var cache = cacheFactory(cacheRootPath);
-                Contract.Assert(cache is IContentStore, $"Attempted to setup a cache named '{kvp.Key}' that is not an {nameof(IContentStore)} at path {cacheRootPath}, type used is {cache.GetType().Name}");
-
-                storesByName.Add(kvp.Key, (IContentStore)cache);
+                storesByName.Add(kvp.Key, (IContentStore)kvp.Value);
             }
 
             _grpcContentServer = new GrpcContentServer(logger, capabilities, this, storesByName);
--- a/Public/Src/Cache/MemoizationStore/Library/Service/LocalCacheServer.cs
+++ b/Public/Src/Cache/MemoizationStore/Library/Service/LocalCacheServer.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics.ContractsLight;
+using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/MemoizationStore/Library/Service/LocalCacheServer.cs
```diff
@@ -42,15 +42,9 @@ public class LocalCacheServer : LocalContentServerBase<ICache, ICacheSession>
         : base(logger, fileSystem, scenario, cacheFactory, localContentServerConfiguration)
         {
             var storesByName = new Dictionary<string, IContentStore>();
-            foreach (var kvp in localContentServerConfiguration.NamedCacheRoots)
```
dcwuser: After this change, any stores in localContentServerConfiguration.NamedCacheRoots will be ignored. What if they are different than StoresByName? What if someone tries to intialize differently?

dcwuser: And what if someone else has already added stores to that dictionary, which they can do because it is visible to other inheritors and indeed to everyone in assembly.

---
In reply to: [321938661](https://github.com/microsoft/BuildXL/pull/845#discussion_r321938661) [](ancestors = 321938661)

Public/Src/Cache/MemoizationStore/Library/Service/LocalCacheServer.cs
```diff
@@ -42,15 +42,9 @@ public class LocalCacheServer : LocalContentServerBase<ICache, ICacheSession>
...
+            foreach (var kvp in StoresByName)
             {
-                AbsolutePath cacheRootPath = kvp.Value;
-                fileSystem.CreateDirectory(cacheRootPath);
-
-                var cache = cacheFactory(cacheRootPath);
-                Contract.Assert(cache is IContentStore, $"Attempted to setup a cache named '{kvp.Key}' that is not an {nameof(IContentStore)} at path {cacheRootPath}, type used is {cache.GetType().Name}");
-
-                storesByName.Add(kvp.Key, (IContentStore)cache);
+                storesByName.Add(kvp.Key, (IContentStore)kvp.Value);
```
dcwuser: Very much dislike use of local variable with same name as property; caused a good 1-2 minutes of confusion.

dcwuser:
I am convince-able to fix in this way if be eliminate (or maybe heavily clarify via comments) initialization/access inconsistency problems. But the real fix I want to see is the eliminate of seperate Service implementations for GrpcCache and GrpcContent.

Public/Src/Cache/MemoizationStore/Library/Service/LocalCacheServer.cs
jbayardo: I have made StoresByName a protected only variable; I don't know why it was internal in the first place.

---
In reply to: [321938831](https://github.com/microsoft/BuildXL/pull/845#discussion_r321938831) [](ancestors = 321938831,321938661)

Public/Src/Cache/MemoizationStore/Library/Service/LocalCacheServer.cs
jbayardo: Renamed to a better suited name.

---
In reply to: [321938875](https://github.com/microsoft/BuildXL/pull/845#discussion_r321938875) [](ancestors = 321938875)

jbayardo:
We discussed offline and agreed to perform a refactoring of the Grpc logic next week, and check this in today so we can start seeing results as soon as possible.

---
In reply to: [529044482](https://github.com/microsoft/BuildXL/pull/845#issuecomment-529044482) [](ancestors = 529044482)

PR #849, opened 2019-09-09
Add logging for RocksDb critical failures

jbayardo:
If code running inside `store.Use()` generates an exception at any point, the RocksDb instance is subsequently "deprecated", which means it will fail to open after the disposal of the current instance.

This is NOT necessarily due to a RocksDb critical error, but only a design choice by `KeyValueStoreAccessor`. The issue is that the handling of that exception depends on the caller. This code ensures at least we will always know the reason why that happened.

We should look into a different failure handling methodology.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
@@ -195,7 +195,12 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                 Tracer.Info(context, $"Creating rocksdb store at '{storeLocation}'.");
 
                 var possibleStore = KeyValueStoreAccessor.Open(storeLocation,
-                    additionalColumns: new[] { nameof(Columns.ClusterState), nameof(Columns.Metadata) }, rotateLogs: true);
+                    additionalColumns: new[] { nameof(Columns.ClusterState), nameof(Columns.Metadata) },
+                    rotateLogs: true,
+                    failureHandler: failure => {
+                        Tracer.Error(context, $"RocksDb critical error caused store deprecation: {failure.DescribeIncludingInnerFailures()}");
+                    });
+
                 if (possibleStore.Succeeded)
                 {
                     var oldKeyValueStore = _keyValueStore;


Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -195,7 +195,12 @@ private BoolResult Load(OperationContext context, StoreSlot activeSlot, bool cle
                 Tracer.Info(context, $"Creating rocksdb store at '{storeLocation}'.");
 
                 var possibleStore = KeyValueStoreAccessor.Open(storeLocation,
-                    additionalColumns: new[] { nameof(Columns.ClusterState), nameof(Columns.Metadata) }, rotateLogs: true);
+                    additionalColumns: new[] { nameof(Columns.ClusterState), nameof(Columns.Metadata) },
+                    rotateLogs: true,
+                    failureHandler: failure => {
```
SergeyTeplyakov: Formatting:)

SergeyTeplyakov: BTW, based on this line of code, the accessor actually throws unknown exception but not swallowing it: [http://index/?leftProject=BuildXL.KeyValueStore&leftSymbol=opeuwx8lnhai&file=KeyValueStoreAccessor.cs&line=601](http://index/?leftProject=BuildXL.KeyValueStore&leftSymbol=opeuwx8lnhai&file=KeyValueStoreAccessor.cs&line=601)


---
In reply to: [322401909](https://github.com/microsoft/BuildXL/pull/849#discussion_r322401909) [](ancestors = 322401909)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -195,7 +195,12 @@ private BoolResult Load(OperationContext context, StoreSlot activeSlot, bool cle
                 Tracer.Info(context, $"Creating rocksdb store at '{storeLocation}'.");
 
                 var possibleStore = KeyValueStoreAccessor.Open(storeLocation,
-                    additionalColumns: new[] { nameof(Columns.ClusterState), nameof(Columns.Metadata) }, rotateLogs: true);
+                    additionalColumns: new[] { nameof(Columns.ClusterState), nameof(Columns.Metadata) },
+                    rotateLogs: true,
```
SergeyTeplyakov: BTW, an alternative fix for this is to actually change KeyValueStoreAccessor and always return a "Failure" with the right nested failure that caused the store to be inaccessible.

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: It doesn't swallow it, but we can swallow it sometimes. It is also not straightforward to search for that error, while this one is more explicit

---
In reply to: [322403543](https://github.com/microsoft/BuildXL/pull/849#discussion_r322403543) [](ancestors = 322403543,322401909)

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
jbayardo: Not necessarily, because we sometimes swallow the failures

---
In reply to: [322404029](https://github.com/microsoft/BuildXL/pull/849#discussion_r322404029) [](ancestors = 322404029)

SergeyTeplyakov: That was my point: to change KeyValueStoreAccessor to "remember" the failure. But this is not critical and the current fix works.

---
In reply to: [322404029](https://github.com/microsoft/BuildXL/pull/849#discussion_r322404029) [](ancestors = 322404029)

PR #850, opened 2019-09-09
Remove Microsoft.ContentStoreApp.exe

kewilli:
Removing Microsoft.ContentStoreApp.exe from nuget packages. This is the executable which was used prior to the product rename. None of our direct consumers use this anymore (AzDev/CB), so it seems safe to remove.

[AB#1365823](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1365823)

### DIFF
--- a/Public/Src/Cache/ContentStore/BuildXL.Cache.ContentStore.dsc
+++ b/Public/Src/Cache/ContentStore/BuildXL.Cache.ContentStore.dsc
@@ -22,16 +22,7 @@ namespace Default {
             {
                 subfolder: r`App`,
                 contents: [
-                    App.exe,
-                    // Back-Compat naming
-                    {
-                        file: App.exe.runtime.binary,
-                        targetFileName: "Microsoft.ContentStoreApp.exe",
-                    },
-                    {
-                        file: App.exe.runtime.pdb,
-                        targetFileName: "Microsoft.ContentStoreApp.pdb",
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/BuildXL.Cache.ContentStore.dsc
```diff
@@ -22,16 +22,7 @@ namespace Default {
...
-                    App.exe,
-                    // Back-Compat naming
-                    {
-                        file: App.exe.runtime.binary,
-                        targetFileName: "Microsoft.ContentStoreApp.exe",
-                    },
-                    {
-                        file: App.exe.runtime.pdb,
-                        targetFileName: "Microsoft.ContentStoreApp.pdb",
-                    }
```
mpysson: I'm not aware of anyone consuming these directly, though that doesn't mean it to be true. It would probably be good to double check with Office 4 Windows. Assuming they don't need any time to migrate anything it should be fine. Can you make sure it gets into the next release's release notes as breaking changes?

kewilli: Contacted OE, they have removed their final reference to `Microsoft.ContentStoreApp.exe` and believe that I'm clear to push this change.

PR #851, opened 2019-09-09
Add 3 finer control arguments for Azure DevOps console output

norahuang:


Public/Src/App/Bxl/Args.cs
```diff
@@ -684,6 +684,18 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                             "optimizeConsoleOutputForAzureDevOps",
                             "ado",
                             sign => loggingConfiguration.OptimizeConsoleOutputForAzureDevOps = sign),
+                        OptionHandlerFactory.CreateBoolOption2(
+                            "optimizeProgressUpdatingForAzureDevOps",
```
mpysson: >optimizeProgressUpdatingForAzureDevOps [](start = 29, length = 38)

nit - probably ok to omit the long forms for these temporary options

Public/Src/App/Bxl/Args.cs
```diff
@@ -684,6 +684,18 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                             "optimizeConsoleOutputForAzureDevOps",
                             "ado",
                             sign => loggingConfiguration.OptimizeConsoleOutputForAzureDevOps = sign),
+                        OptionHandlerFactory.CreateBoolOption2(
+                            "optimizeProgressUpdatingForAzureDevOps",
+                            "adoProgress",
+                            sign => loggingConfiguration.OptimizeProgressUpdatingForAzureDevOps = sign),
+                        OptionHandlerFactory.CreateBoolOption2(
+                            "optimizeVsoAnnotationsForAzureDevOps",
+                            "adoVso",
```
mpysson: >adoVso [](start = 29, length = 6)

how about "adoTask" since it is creating a "task" event in AzureDevOpsListener.cs?

mpysson: maybe that's confusing with "task" in the sense of a work item. Another thought is to suffix all of these with logging to disambiguate: adoTaskLoggin, adoProgressLogging, adoWarningErrorLogging. 

---
In reply to: [323013451](https://github.com/microsoft/BuildXL/pull/851#discussion_r323013451) [](ancestors = 323013451)

PR #852, opened 2019-09-09
Detailed cache materialization error log

semihokur:


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4179,6 +4178,31 @@ public int GetPipPriority(PipId pipId)
         /// <inheritdoc />
         public DirectoryTranslator DirectoryTranslator { get; }
 
+        [SuppressMessage("Microsoft.Design", "CA1033:InterfaceMethodsShouldBeCallableByChildTypes")]
+        string IPipExecutionEnvironment.GetProducerExecutionInfo(in FileOrDirectoryArtifact artifact)
+        {
+            var producerId = PipGraph.GetProducer(artifact);
+            if (!producerId.IsValid)
+            {
+                return "N/A";
```
mpysson: >return "N/A"; [](start = 16, length = 13)

When will this happen in practice?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4179,6 +4178,31 @@ public int GetPipPriority(PipId pipId)
...
+            PipExecutionStep step = PipExecutionStep.ExecuteProcess;
+            if (perfInfo.StepDurations[(int)step].TotalMilliseconds == 0)
+            {
+                step = PipExecutionStep.RunFromCache;
+            }
+
+            var workerId = perfInfo.Workers.Value[(int)step];
+            var worker = m_workers[(int)workerId];
+            bool isWorkerReleasedEarly = worker.WorkerEarlyReleasedTime != null;
+            return $"{producer.FormattedSemiStableHash} {step} on Worker#{workerId} ({m_workers[(int)workerId].Status} - IsReleasedEarly: {isWorkerReleasedEarly})";
```
mpysson: > {isWorkerReleasedEarly} [](start = 137, length = 24)

Can we lift this into a specific counter for the number of files or pips that failed due to failed materializations from a worker that has been released? That'd give us a very clear signal to quickly see if we're having materialization issues due to early release without having to churn through a lot of full text logs in kusto

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4179,6 +4178,31 @@ public int GetPipPriority(PipId pipId)
...
+            {
+                return "N/A";
+            }
+
+            var producer = PipGraph.GetPipFromPipId(producerId);
+
+            RunnablePipPerformanceInfo perfInfo = m_runnablePipPerformance[producerId];
+
+            PipExecutionStep step = PipExecutionStep.ExecuteProcess;
+            if (perfInfo.StepDurations[(int)step].TotalMilliseconds == 0)
```
mpysson: >(int)step].TotalMilliseconds == 0 [](start = 39, length = 33)

This feels like a fuzzy way to get this signal. Can you utilize some other state like whether CacheLookupPerfInfo was set?

PR #853, opened 2019-09-09
VS solution generator: support for multiple target frameworks

almiliMSFT:
This new VS solution generator supports projects targeting multiple frameworks.  Concretely, it generates .csproj files that:
  1. import `Sdk="Microsoft.NET.Sdk"`
  1. contain a `<TargetFrameworks>` property instead of `<TargetFramework>`
  1. have all properties/items conditionalized based on the framework they are targeting 

Th new generator is currently hidden behind an undocumented `/vsNew` switch.  The old one remains the default until the new one has been vetted more thoroughly.

The new generator also fixes a few problems present in the old one; namely:
  - adds support for aliased references
  - detects C# `/define` args in a more robust way

[AB#1537173](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1537173)


Public/Src/Engine/Dll/BuildXL.Engine.dsc
```diff
@@ -39,6 +39,7 @@ namespace Engine {
             importFrom("BuildXL.Cache.MemoizationStore").Interfaces.dll,
             importFrom("BuildXL.Pips").dll,
             importFrom("BuildXL.Ide").Generator.dll,
+            importFrom("BuildXL.Ide").Generator.Old.dll,
```
almiliMSFT: This `Generator.Old` project is a verbatim copy of the old generator project.  In git those files all appear as "new", but it's safe to ignore them for the purpose of this PR.

The reason why I copied the old project into a new folder instead of putting the new generator into a new folder is so that `git diff` shows the changes between the new and the old generator more clearly.

The old generator remains the default until the new one is vetted more thoroughly.

Public/Src/IDE/Generator.Old/BuildXL.Ide.Generator.Old.dsc
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft. All rights reserved.
```
almiliMSFT: safe to ignore everything under `Public/Src/IDE/Generator.Old` (since it is the verbatim copy of what used to be `Public/Src/IDE/Generator`

Public/Src/IDE/Generator/MsbuildFile.cs
```diff
@@ -86,7 +90,7 @@ internal abstract class MsbuildFile
             // Add '.g' suffix only if the projects files will be in the spec root
             if (Context.CanWriteToSrc)
             {
-                projectExtension = ".g" + projectExtension;
+                projectExtension = ".2.g" + projectExtension;
```
Kr1sso: Is this to allow for old/new file co-existence?

almiliMSFT: Indeed

almiliMSFT: I'll change it back to `.g` once the old generator is deprecated

Public/Src/App/Bxl/Args.cs
```diff
@@ -1129,6 +1129,9 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
                         OptionHandlerFactory.CreateBoolOption(
                             "vs",
                             sign => ideConfiguration.IsEnabled = sign),
+                        OptionHandlerFactory.CreateBoolOption(
+                            "vsNew", // temporary undocumented option for enabling new VS solution generation
```
mpysson: >New [](start = 31, length = 3)

Is there some label for the new strategy? There may come a day where this is years old and not justifying of "new" anymore .:smile: If we're unimaginative maybe just v2

almiliMSFT: I hope this is not going to survive for a whole year.  I thought I'd remove this in a month or so

Public/Src/IDE/Generator/CsprojFile.cs
```diff
@@ -364,10 +369,28 @@ private void IterateCscArguments(Project project, PipData arguments, bool isNest
...
+                            {
+                                project.SetProperty("OutputType", strValue.Substring(Target.Length));
+                                break;
+                            }
+
+                            if (strValue.StartsWith(Define, StringComparison.OrdinalIgnoreCase))
+                            {
+                                project.SetProperty("DefineConstants", strValue.Substring(Define.Length).Trim('"'));
+                                break;
+                            }
```
mpysson: why not include these as cases in the switch statement above?

Public/Src/IDE/Generator/CommonBuildFiles/CSharp.targets
```diff
@@ -7,6 +7,5 @@
     <LangVersion Condition="'$(LangVersion)' == 'preview' AND '$(VisualStudioVersion)' == '15.0'">latest</LangVersion>
   </PropertyGroup>
 
-  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
```
mpysson: >$(MSBuildToolsPath)\Microsoft.CSharp.targets [](start = 19, length = 44)

Probably a datapoint from your survey but I'm skeptical unit tests will work as expected without our funny deployment and test execution logic. Not that the VS test integration we had previously was very dialed in anyway.

Public/Src/IDE/Generator/CsprojFile.cs
almiliMSFT: The switch cases work when args are represented with 2 pip fragments: `/define:` and then another string fragment.  Because of escaping of whatever other reason, when we pass this argument to csc.exe, it is represented as a single StringLiteral fragment (`/define:"Const1;Const2"`) which is why a case for `/define:` would never get matched.

almiliMSFT: In fact, there is already 
```
                        case "/define:":
                            action = (obj) => project.SetProperty("DefineConstants", (string)obj);
                            break;
```
present in the code, but that doesn't do the trick

Public/Src/IDE/Generator/CommonBuildFiles/CSharp.targets
almiliMSFT: Agreed.  I didn't focus much on tests, but will do in a next iteration

PR #854, opened 2019-09-10
Add some stats re cache lookup

olkononenko:
[AB#1596610](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1596610)


Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1857,6 +1864,15 @@ private static void ReportFileAccesses(ExecutionResult processExecutionResult, F
                     environment.State.ExecutionLog?.ProcessFingerprintComputation(processFingerprintComputationResult);
                 }
 
+                Logger.Log.PipCacheLookupStats(
+                    operationContext,
+                    process.FormattedSemiStableHash,
+                    // if we can augment the FP, it means that we are NOT processing an augmented one
+                    !canAugmentWeakFingerprint,
```
ljcollins25: >canAugmentWeakFingerprint [](start = 21, length = 25)

I don't think this is true.

PR #855, opened 2019-09-10
Grpc server refactor

jbayardo:


Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
```diff
@@ -2,8 +2,8 @@
 
 package ContentStore.Grpc;
 
-// Defines an interface of a content server
-service ContentServer {
+// Defines an interface of a cache server
+service CacheServer {
```
SergeyTeplyakov: I think this is breaking change on the protocol level and it means that when the new grpc client will try to use an old casaas they will fail.

jbayardo: Yes, that is the case indeed. I actually just checked that the current implementation only fails when an old client connects to a new server (locally)

jbayardo: Fixed by still providing the old implementation as well as the current one. It's a temporary transition plan that should be in place until all stamps have the bits shipped

Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
```diff
@@ -78,11 +85,8 @@ public abstract class LocalContentServerBase<TStore, TSession> : StartupShutdown
         /// </summary>
         /// <remarks>
         /// This is only supposed to be used by this class and inheritors, do not make internal or public.
-        /// 
-        /// It is also expected to be immutable, because inheritors may copy it around. So do NOT add keys after
-        /// constructed.
         /// </remarks>
-        protected readonly Dictionary<string, TStore> StoresByName = new Dictionary<string, TStore>();
+        public readonly IReadOnlyDictionary<string, TStore> StoresByName;
```
dcwuser: What I think you're trying to do here is a good idea, but this is an incomplete implementation of that idea. Any arbitrary code can still read this field, cast to Dictionary, and modify entries. The complete way to do this is to create a private dictionary field, a public IReadOnlyDictionary property, and have the property return a ReadOnlyDictionary that wraps the private dictionary. Extra points for making sure the wrapper is only constructed once, not on every call.

dcwuser:
>     public class GrpcContentServer : StartupShutdownSlimBase

This only still exists because of transition reasons, right? In final version, GrpcContentServer should go away entirely.

---
Refers to: Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs:36 in bf080e6. [](commit_id = bf080e61ceee3bb47b5c4e9fcc74df237b9b29f9, deletion_comment = False)

jbayardo:
>     public class GrpcContentServer : StartupShutdownSlimBase

Well, removing this for now is problematic: in the distributed case, when we create the server, it has only an IContentStore (no Memoization). In the future -assuming we move all operations to the server- we will have a single server that does both.

Right now, doing this would complicate the logic: we'd need an ICache that throws for metadata operations to back the server, or do some "runtime duck typing", which is not very nice either.

This change still manages to remove the need for 2 different objects, as GrpcCacheServer now inherits from GrpcContentServer.

---
In reply to: [530162004](https://github.com/microsoft/BuildXL/pull/855#issuecomment-530162004) [](ancestors = 530162004)

---
Refers to: Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs:36 in bf080e6. [](commit_id = bf080e61ceee3bb47b5c4e9fcc74df237b9b29f9, deletion_comment = False)

jbayardo:
It needs to be noted that this PR breaks compatibility with clients using old bits of the gRPC client for Memoization. This is because of the naming conundrum; it can be easily fixed, but since those bits are only being used in DM_S1 right now, it seems better to disable them for deployment and then enable back again.

PR #858, opened 2019-09-10
Use /p to set the inputchanges path

norahuang:
The QTest SDK need to aware the existence of the input changes passed from GBR. Change /InputChanges command line option to a environment parameter /p:[Sdk.BuildXL]inputChanges

### DIFF
--- a/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
@@ -6,7 +6,11 @@ import {Artifact, Cmd, Transformer, Tool} from "Sdk.Transformers";
 const root = Environment.hasVariable("QTEST_DEPLOYMENT_PATH") ? d`${Environment.getFileValue("QTEST_DEPLOYMENT_PATH")}` : d`.`;
 const qCodeCoverageEnumType = Environment.hasVariable("[Sdk.BuildXL]qCodeCoverageEnumType")
     ? Environment.getStringValue("[Sdk.BuildXL]qCodeCoverageEnumType")
-    : "None";
+    : "None"; 
+
+const qInputChangesFilePath = Environment.hasVariable("[Sdk.BuildXL]inputChanges") 
+    ? f`${Environment.getFileValue("[Sdk.BuildXL]inputChanges")}`
+    : undefined;
 
 @@public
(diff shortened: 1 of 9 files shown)


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4915,11 +4920,18 @@ private void ProcessFileChanges(LoggingContext loggingContext, SchedulerState sc
         {
             InputChangeList inputChangeList = null;
 
-            if (m_configuration.Schedule.InputChanges.IsValid)
+            var inputChangesFilePath = Environment.GetEnvironmentVariable("[Sdk.BuildXL]inputChanges");
```
mpysson: >var inputChangesFilePath = Environment.GetEnvironmentVariable("[Sdk.BuildXL]inputChanges"); [](start = 12, length = 91)

The ability to consume an externally specified changes is something that's potentially useful to BuildXL beyond this code coverage feature. I don't really want to couple it to the QTest sdk. It's also a bit sneaky to toggle the BuildXL behavior just based on an environment variable. I think we should preserve the command line configuration.

Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -126,7 +130,7 @@ export function runQTest(args: QTestArguments): Result {
      
     let changeAffectedInputListWrittenFile = undefined;
     let changeAffectedInputListWrittenFileArg = {};
-    if (qCodeCoverageEnumType === "DynamicCodeCov"){
+    if (qCodeCoverageEnumType === "DynamicCodeCov" && File.exists(qInputChangesFilePath)){
```
mpysson: >&& File.exists(qInputChangesFilePath) [](start = 51, length = 37)

why is this necessary? Don't we want the ability to run code coverage without the inputChangesFile being set?

norahuang: I need to know if it is a change based code coverage or a full code coverage. I wanted to use the existence of the inputChanges to distinguish these two cases. Without knowing the full code coverage doesn't work. There is another approach which uses a separated /p: to indicated the change based one. This will make the code coverage feature separated from input changes. But the downside is introducing a extra option.

norahuang:
Close this PR with another which adds a extra /p: instead of changing /InputChanges to /p:

PR #859, opened 2019-09-10
Fix crash on worker if master fails so early

semihokur:
Fixes [AB#1596357](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1596357)

### DIFF
--- a/Public/Src/Engine/Dll/Distribution/WorkerService.cs
+++ b/Public/Src/Engine/Dll/Distribution/WorkerService.cs
@@ -75,7 +75,7 @@ namespace BuildXL.Engine.Distribution
         /// <summary>
         /// Returns a task representing the completion of the exit operation
         /// </summary>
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/Dll/Distribution/WorkerService.cs
```diff
@@ -339,8 +344,7 @@ public void Exit(string failure, bool isUnexpected = false)
 
             if (!reportSuccess)
             {
-                // Only log the error, if this thread set exit response
-                Logger.Log.DistributionWorkerExitFailure(m_appLoggingContext, failure);
+                m_masterFailureMessage = failure;
```
mpysson: >failure [](start = 41, length = 7)

It might be a good idea to log a warning at the time that the failure message from the master comes in. That could be helpful for debugging other potential unexpected issues, like our eventual error not actually getting logged due to a different race condition or Exit() being called multiple times somehow

PR #862, opened 2019-09-11
Reset post-initialization state when ProcessStateAsync succeeds.

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -466,27 +466,47 @@ internal async Task<BoolResult> RestoreCheckpointAsync(OperationContext context,
...
-                
-                if (!checkpointState)
-                {
-                    // The error is already logged.
-                    return checkpointState;
-                }
+                // A post initialization process may fail due to a transient issue, like a storage failure or an inconsistent checkpoint's state.
+                // The transient error can go away and the system may recover itself by calling this method again.
+
+                // In this case we need to reset _postInitializationTask and move it's state from "failure" to "success"
```
kewilli: Nit: `it's` -> `its`

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -466,27 +466,47 @@ internal async Task<BoolResult> RestoreCheckpointAsync(OperationContext context,
...
+                // and unblock all the public operations that will fail if post-initialization task is unsuccessful.
 
-                return await RestoreCheckpointAsync(context, checkpointState.Value, inline);
+                _postInitializationTask = Task.FromResult(result);
             }
-            finally
+
+            return result;
+
+            async Task<BoolResult> processStateCoreAsync()
```
dcwuser: Minor: Not a fan of these private-methods-in-private-methods. Harder to distinguish inner method code from outer method code. In any case, needs to start with a comment describing function.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -466,27 +466,47 @@ internal async Task<BoolResult> RestoreCheckpointAsync(OperationContext context,
 
         private async Task<BoolResult> ProcessStateAsync(OperationContext context, bool inline)
         {
-            try
+            var result = await processStateCoreAsync();
+
+            if (result.Succeeded &&
```
dcwuser: Technically this is a race, since multiple ProcessStateAsyncs could be running. Clear that's not a problem 99.99% of the time, but if it does ever happen that two are finishing at the same time, is that a problem?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
SergeyTeplyakov: I actually pretty like it. Because if you have two private methods: one is called ProcessStateAsync and another one is ProcessStateAndChangePostInitializationTaskAsync then it is possible to accidentally switch and use the wrong one.

In this case, the local method is intended to be local.

---
In reply to: [323499701](https://github.com/microsoft/BuildXL/pull/862#discussion_r323499701) [](ancestors = 323499701)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
SergeyTeplyakov: We have only one operation running at the same time.
And even if they can, I don't think we can ended up with an inconsistent state.

Field assignment is atomic and task transition is thread safe. One thread definitely can see that the task is not null, but after that it can't get into a null state. So we can't get NRE.

The worst thing that may happen here is double assignment when two operations are done at the same time and two threads will decide to update the field.

---
In reply to: [323499999](https://github.com/microsoft/BuildXL/pull/862#discussion_r323499999) [](ancestors = 323499999)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -466,27 +466,47 @@ internal async Task<BoolResult> RestoreCheckpointAsync(OperationContext context,
 
         private async Task<BoolResult> ProcessStateAsync(OperationContext context, bool inline)
         {
-            try
+            var result = await processStateCoreAsync();
```
ljcollins25: >processStateCoreAsync [](start = 31, length = 21)

I think it would be cleaner (with added benefit of logging) to wrap the method body in PerformOperation and then just set _postInitializationTask to BoolResult.SuccessTask in case where it succeeds (no checks on current task status). We might also consider tracking the time since the last successful result for the purposes of flagging repeated checkpoint restore or create checkpoint failures for too long of a period of time.

PR #865, opened 2019-09-12
Resource based canceling with shared opaque output producers

Kr1sso:
This PR introduced proper support for memory observation (VM pressure + process tree consumption) and resource based canceling for macOS. 

It **also adds logic** to deal with issues present if outputs are created in shared opaque directories and the producing pip gets canceled and rerun due to resource based cancelation mechanics.

### DIFF
--- a/Public/Src/Engine/Dll/Tracing/LogEventId.cs
+++ b/Public/Src/Engine/Dll/Tracing/LogEventId.cs
@@ -216,7 +216,7 @@ namespace BuildXL.Engine.Tracing
         FailedToAcquireDirectoryLock = 7116,
         UsingRedirectedUserProfile = 7117,
         FailedToRedirectUserProfile = 7118,
-        ResourceBasedCancellationIsEnabledWithSharedOpaquesPresent = 7119,
+        // ResourceBasedCancellationIsEnabledWithSharedOpaquesPresent = 7119,
         BusyOrUnavailableOutputDirectoriesException = 7120,
         GrpcSettings = 7121,
 
--- a/Public/Src/Engine/Dll/Tracing/LogEventId.cs
+++ b/Public/Src/Engine/Dll/Tracing/LogEventId.cs
@@ -216,7 +216,7 @@ namespace BuildXL.Engine.Tracing
         FailedToAcquireDirectoryLock = 7116,
         UsingRedirectedUserProfile = 7117,
         FailedToRedirectUserProfile = 7118,
-        // ResourceBasedCancellationIsEnabledWithSharedOpaquesPresent = 7119,
+        // was: ResourceBasedCancellationIsEnabledWithSharedOpaquesPresent = 7119,
         BusyOrUnavailableOutputDirectoriesException = 7120,
         GrpcSettings = 7121,
 
--- a/Public/Src/Engine/Processes/SandboxedProcessMac.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessMac.cs
@@ -35,8 +35,8 @@ namespace BuildXL.Processes
             internal PerformanceCollector.Aggregation JobKernelTimeMs { get; } = new PerformanceCollector.Aggregation();
             internal PerformanceCollector.Aggregation JobUserTimeMs { get; } = new PerformanceCollector.Aggregation();
             internal PerformanceCollector.Aggregation PeakMemoryBytes { get; } = new PerformanceCollector.Aggregation();
-            internal PerformanceCollector.Aggregation BytesRead { get; } = new PerformanceCollector.Aggregation();
-            internal PerformanceCollector.Aggregation BytesWritten { get; } = new PerformanceCollector.Aggregation();
+            internal PerformanceCollector.Aggregation DiskBytesRead { get; } = new PerformanceCollector.Aggregation();
+            internal PerformanceCollector.Aggregation DiskBytesWritten { get; } = new PerformanceCollector.Aggregation();
         }
 
         /// <summary>
@@ -50,6 +50,8 @@ namespace BuildXL.Processes
 
         private readonly Timer m_perfTimer;
         private readonly PerfAggregator m_perfAggregator;
+
+        private IEnumerable<ReportedProcess> m_survivingChildProcesses = null;
 
         private long m_processKilledFlag = 0;
 
@@ -166,8 +168,8 @@ namespace BuildXL.Processes
             // get memory usage for the process tree
             proc.m_perfAggregator.PeakMemoryBytes.RegisterSample(Dispatch.GetActivePeakMemoryUsage(default, proc.ProcessId));
 
-            proc.m_perfAggregator.BytesRead.RegisterSample(buffer.BytesRead);
-            proc.m_perfAggregator.BytesWritten.RegisterSample(buffer.BytesWritten);
+            proc.m_perfAggregator.DiskBytesRead.RegisterSample(buffer.DiskBytesRead);
+            proc.m_perfAggregator.DiskBytesWritten.RegisterSample(buffer.DiskBytesWritten);
 
             // reschedule the timer to fire again in PerfProbeInterval time (2 seconds)
             if (!proc.Killed)
@@ -238,6 +240,12 @@ namespace BuildXL.Processes
                 LogProcessState($"IOException caught while feeding the standard input: {e.Message}");
                 await KillAsync();
             }
+        }
+
+        /// <inheritdoc />
+        protected override IEnumerable<ReportedProcess> GetSurvivingChildProcesses()
+        {
+            return m_survivingChildProcesses;
         }
 
         /// <inheritdoc />
@@ -403,8 +411,8 @@ namespace BuildXL.Processes
                     {
                         ReadOperationCount = 1,
                         WriteOperationCount = 1,
-                        ReadTransferCount = Convert.ToUInt64(m_perfAggregator.BytesRead.Total),
-                        WriteTransferCount = Convert.ToUInt64(m_perfAggregator.BytesWritten.Total)
+                        ReadTransferCount = Convert.ToUInt64(m_perfAggregator.DiskBytesRead.Total),
+                        WriteTransferCount = Convert.ToUInt64(m_perfAggregator.DiskBytesWritten.Total)
                     }),
                     PeakMemoryUsage = (ulong)m_perfAggregator.PeakMemoryBytes.Maximum,
                     KernelTime = TimeSpan.FromMilliseconds(m_perfAggregator.JobKernelTimeMs.Latest),
@@ -467,7 +475,7 @@ namespace BuildXL.Processes
                         break;
                     }
 
-                    if (HasProcessExitBeenReceived)
+                    if (HasProcessExitBeenReceived && !Killed)
                     {
                         // Proceed if the event queue is beyond the point when ProcessExit was received
                         if (m_processExitTimeNs <= minEnqueueTime)
@@ -479,6 +487,7 @@ namespace BuildXL.Processes
                             }
 
                             LogProcessState($"Process timed out because nested process termination timeout limit was reached.");
+                            m_survivingChildProcesses = CoalesceProcesses(GetCurrentlyActiveProcesses());
                             processTreeTimeoutSource.SetResult(Unit.Void);
                             break;
                         }
@@ -584,6 +593,11 @@ namespace BuildXL.Processes
 
         private void ReportFileAccess(ref AccessReport report)
         {
+            if (ReportsCompleted())
+            {
+                return;
+            }
+
             if (IgnoreReportedAccesses &&
                 report.Operation != FileOperation.OpProcessStart &&
                 report.Operation != FileOperation.OpProcessExit)
--- a/Public/Src/Engine/Processes/SandboxedProcessMac.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessMac.cs
@@ -13,6 +13,7 @@ using System.Threading.Tasks.Dataflow;
 using BuildXL.Interop;
 using BuildXL.Interop.MacOS;
 using BuildXL.Native.IO;
+using BuildXL.Native.Processes;
 using BuildXL.Utilities;
 using BuildXL.Utilities.Tasks;
 using JetBrains.Annotations;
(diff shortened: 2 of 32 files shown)


Public/Src/Engine/Dll/Tracing/LogEventId.cs
```diff
@@ -216,7 +216,7 @@ public enum LogEventId : ushort
         FailedToAcquireDirectoryLock = 7116,
         UsingRedirectedUserProfile = 7117,
         FailedToRedirectUserProfile = 7118,
-        ResourceBasedCancellationIsEnabledWithSharedOpaquesPresent = 7119,
+        // ResourceBasedCancellationIsEnabledWithSharedOpaquesPresent = 7119,
```
ljcollins25: >ResourceBasedCancellationIsEnabledWithSharedOpaquesPresent [](start = 11, length = 58)

Remove?

Kr1sso: Whats our common practice here? A few lines above we have commented out ones and some that say // XXX was YYYY. Do we need to keep any traces around?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3562,6 +3577,13 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
                             // reschedule and choose worker again after adjusting expected RAM utilization
                             processRunnable.SetWorker(null);
 
+                            // Because the scheduler will re-run this pip, we have to nuke all outputs created under shared opaque directories
+                            var sharedOpaqueOutputs = FlagAndReturnSharedOpaqueOutputs(environment, processRunnable);
```
ljcollins25: >sharedOpaqueOutputs [](start = 32, length = 19)

This line should also be inside the worker.IsLocal condition.

Public/Src/Engine/Dll/Tracing/LogEventId.cs
mpysson:
```suggestion
        // was: ResourceBasedCancellationIsEnabledWithSharedOpaquesPresent = 7119,
```

mpysson: I usually put a "was" comment in there so if it comes up with people running old versions of BuildXL we can still see what it meant in the past without doing too much research. Then when the line is sufficiently old we can also recycle the ID

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2175,9 +2188,11 @@ private void UpdateResourceAvailability(PerformanceCollector.MachinePerfInfo per
...
                 if (!m_scheduleConfiguration.DisableProcessRetryOnResourceExhaustion)
                 {
-                    // Free down to the specified max RAM utilization percentage with 10% slack
-                    var desiredRamToFreePercentage =
-                        (perfInfo.RamUsagePercentage.Value - m_configuration.Schedule.MaximumRamUtilizationPercentage) + 10;
+                    // Free down to the specified max RAM utilization percentage with 10% slack, if VM memory pressure is critical
+                    // we instantly try to free up 15% of the overall system memory to improve stability
+                    var desiredRamToFreePercentage = !isMemoryPressureCritical
+                        ? (perfInfo.RamUsagePercentage.Value - m_configuration.Schedule.MaximumRamUtilizationPercentage) + 10
+                        : 15;
```
mpysson: 10% vs 15%: Do we really need different thresholds based on the OS? Can we just utilize the externally configurable MaximumRamUtilizationPercentage option instead? I feel like internally padding the number differently based on OS is a bit unexpected from end user perspective.

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
```diff
@@ -42,15 +42,58 @@ int GetRamUsageInfo(RamUsageInfo *buffer, long bufferSize)
...
+
+    pid_t pids[max_proc_count];
+    int child_count = proc_listchildpids(pid, pids, (int) max_proc_count);
+    *success &= child_count >= 0;
+
+    for (int i = 0; (i < child_count) && *success; i++)
+    {
+        int pid = pids[i];
+        rusage_info_current rusage;
+        *success &= proc_pid_rusage(pid, RUSAGE_INFO_CURRENT, (void **)&rusage) == 0;
```
mpysson: >RUSAGE_INFO_CURRENT, [](start = 41, length = 20)

This data is exposed by GetPeakWorkingSetSize. Is it really peak or is it current?

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
```diff
@@ -42,15 +42,58 @@ int GetRamUsageInfo(RamUsageInfo *buffer, long bufferSize)
...
+        {
+            children_usage += ProcessTreeWorkingSetSize(pid, max_proc_count, success);
+        }
+    }
+
+    return children_usage;
+
+}
+
 int GetPeakWorkingSetSize(pid_t pid, uint64_t *buffer)
```
mpysson: >GetPeakWorkingSetSize [](start = 4, length = 21)

getrusage's man page makes it look like you can call it twice for RUSAGE_SELF and RUSAGE_CHILDREN and then add the rs_maxrss field from both to get the overall peak. Can you do that to simplify all of the logic around querying the child processes individually?

Public/Src/Utilities/Configuration/ISandboxConfiguration.cs
```diff
@@ -171,13 +171,8 @@ public interface ISandboxConfiguration
...
         /// <summary>
-        /// Whether to measure CPU times (user/system) of sandboxed processes.  Default: false.
+        /// Whether to measure CPU times (user/system) of sandboxed processes.  Default: true.
         /// </summary>
-        /// <remarks>
-        /// In principle thre should be no reason not to measure CPU times.  But on macOS this amounts to wrapping
-        /// every process in '/usr/bin/time', which could lead to some unexpected behavior.  Hence this option to
-        /// explicitly turn in on or off.
-        /// </remarks>
         bool KextMeasureProcessCpuTimes { get; }
```
mpysson: >Kext [](start = 13, length = 4)

Rename to remove the "kext" part since it applies to the ES sandbox too

Public/Src/Utilities/Interop/MacOS/Process.cs
```diff
@@ -30,19 +30,29 @@ public struct ProcessTimesInfo
             /// User time of a given process in nanoseconds.
             /// </summary>
             public ulong UserTimeNs;
+
+            /// <summary>
+            /// User time of a given process in nanoseconds.
```
mpysson: > User time of a given process in nanoseconds. [](start = 15, length = 45)

Fix comments

Public/Src/Utilities/Interop/MacOS/Process.cs
```diff
@@ -30,19 +30,29 @@ public struct ProcessTimesInfo
...
             /// </summary>
             public ulong UserTimeNs;
+
+            /// <summary>
+            /// User time of a given process in nanoseconds.
+            /// </summary>
+            public ulong BytesRead;
+
+            /// <summary>
+            /// User time of a given process in nanoseconds.
```
mpysson: >ser time of a given process in nanoseconds. [](start = 17, length = 43)

fix

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
Kr1sso: `getrusage` does not allow for pid queries, it is always relative to the calling process / thread. there is no other way to query the whole tree for its memory consumption.

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
Kr1sso: It's current really - looking at the code in general it seemed to me like  "peak" is used by the job object as it is the only readable information. In the context this is called BuildXL is actually doing calculations to free RAM and get below a threshold. Do you think peak is what we should use?

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
mpysson: Ah. That's too bad. I guess to use the nice API we'd then have to first launch some special process to do reporting that would then spawn the pip. That'd be too much complexity and overhead. Might be worth adding a comment why getrusage isn't appropriate

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
mpysson: The windows code is using peak for historical runtime tracking. I'm not sure for cancellation. I guess for cancellation of the currently running pips it would be best to have max of historical peak and current. I'd mirror the windows implementation for our own sanity. But doing the peak is probably best if we had to be pick one.

Public/Src/Engine/Processes/SandboxedProcessMac.cs
```diff
@@ -34,6 +35,8 @@ private class PerfAggregator
             internal PerformanceCollector.Aggregation JobKernelTimeMs { get; } = new PerformanceCollector.Aggregation();
             internal PerformanceCollector.Aggregation JobUserTimeMs { get; } = new PerformanceCollector.Aggregation();
             internal PerformanceCollector.Aggregation PeakMemoryBytes { get; } = new PerformanceCollector.Aggregation();
+            internal PerformanceCollector.Aggregation BytesRead { get; } = new PerformanceCollector.Aggregation();
```
almiliMSFT: What bytes?  From disk, network, both?

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
```diff
@@ -42,15 +42,58 @@ int GetRamUsageInfo(RamUsageInfo *buffer, long bufferSize)
...
+{
+    uint64_t children_usage = 0;
+
+    pid_t pids[max_proc_count];
+    int child_count = proc_listchildpids(pid, pids, (int) max_proc_count);
+    *success &= child_count >= 0;
+
+    for (int i = 0; (i < child_count) && *success; i++)
+    {
+        int pid = pids[i];
```
almiliMSFT: Symbol `pid` is already defined as one of the function arguments (`pid_t pid`). How does this even compile?  Please rename

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
```diff
@@ -42,15 +42,58 @@ int GetRamUsageInfo(RamUsageInfo *buffer, long bufferSize)
     return KERN_SUCCESS;
 }
 
+static uint64_t ProcessTreeWorkingSetSize(pid_t pid, const rlim_t max_proc_count, bool *success)
+{
+    uint64_t children_usage = 0;
+
+    pid_t pids[max_proc_count];
```
almiliMSFT: rename `pids` to `child_pids`

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
```diff
@@ -42,15 +42,58 @@ int GetRamUsageInfo(RamUsageInfo *buffer, long bufferSize)
...
+
+    for (int i = 0; (i < child_count) && *success; i++)
+    {
+        int pid = pids[i];
+        rusage_info_current rusage;
+        *success &= proc_pid_rusage(pid, RUSAGE_INFO_CURRENT, (void **)&rusage) == 0;
+        children_usage += rusage.ri_resident_size;
+
+        pid_t childs_pids[max_proc_count];
+        int count = proc_listchildpids(pid, childs_pids, (int) max_proc_count);
```
almiliMSFT: why do you need to call `proc_listchildpids` here again?  When you recursively call `ProcessTreeWorkingSetSize(pid, max_proc_count, success)` it will get called then and then if the count is 0 it will immediately return.

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
```diff
@@ -42,15 +42,58 @@ int GetRamUsageInfo(RamUsageInfo *buffer, long bufferSize)
     return KERN_SUCCESS;
 }
 
+static uint64_t ProcessTreeWorkingSetSize(pid_t pid, const rlim_t max_proc_count, bool *success)
```
almiliMSFT: is `proc_pid_rusage` ever called for this `pid`?  It looks like it only queries its children and for each of them called `proc_pid_rusage`, but not for this root process itself

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
```diff
@@ -42,15 +42,58 @@ int GetRamUsageInfo(RamUsageInfo *buffer, long bufferSize)
...
+}
+
 int GetPeakWorkingSetSize(pid_t pid, uint64_t *buffer)
 {
+    struct rlimit rl;
+    bool success = getrlimit(RLIMIT_NPROC, &rl) == 0;
+
     rusage_info_current rusage;
-    if (proc_pid_rusage(pid, RUSAGE_INFO_CURRENT, (void **)&rusage) != 0)
+    success &= proc_pid_rusage(pid, RUSAGE_INFO_CURRENT, (void **)&rusage) == 0;
```
almiliMSFT: I think it would be better if `ProcessTreeWorkingSetSize` would include `proc_pid_rusage` result for the root pid and all of its children so that you don't have to call it here first.

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
```diff
@@ -42,15 +42,58 @@ int GetRamUsageInfo(RamUsageInfo *buffer, long bufferSize)
...
+
+    // We look at the resident size for the complete process tree because we care about "real" memory used,
```
almiliMSFT: add a comment about how this "real memory" is different from what `rusage.ri_lifetime_max_phys_footprint` returns

Public/Src/Sandbox/MacOs/Interop/Posix/process.h
```diff
@@ -11,9 +11,11 @@ typedef struct {
     double exitTime;
     unsigned long systemTime;
     unsigned long userTime;
-} ProcessTimesInfo;
+    unsigned long bytesRead;
+    unsigned long bytesWritten;
+} ProcessResourceUsage;
 
-int GetProcessTimes(pid_t pid, ProcessTimesInfo *buffer, long bufferSize, bool includeChildProcesses);
+int GetProcessResourceUsage(pid_t pid, ProcessResourceUsage *buffer, long bufferSize, bool includeChildProcesses);
```
almiliMSFT: don't forget to publish new Mac runtime and bump up the version

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
Kr1sso: Yeah, thats a C quirk that allows to overwrite definitions in local scopes. I renamed for clarity.

Public/Src/Sandbox/MacOs/Interop/Posix/memory.c
Kr1sso: We do this already, although the actual number is the current process tree memory size, the `AccountingInformation` for a sandboxed process returns `PeakMemoryUsage = (ulong)m_perfAggregator.PeakMemoryBytes.Maximum` so this works as expected because that value is used for the historic data.

PR #869, opened 2019-09-12
Rename DominoInvocation and ExtraEventDataReported Events

sahilmgandhi:
[AB#1597957](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1597957)


olkononenko:
there is domino invocation in `Public\Src\Utilities\Utilities\Tracing\EventId.cs`

sahilmgandhi:
> there is domino invocation in
> `Public\Src\Utilities\Utilities\Tracing\EventId.cs`
> `Public\Src\Utilities\Utilities\Tracing\FormattingEventListener.cs`

Yea those ones are used for some telemetry and should not be renamed

PR #872, opened 2019-09-12
Make proactive copy configurable inside/outside ring

JuanCarlosGI:
Make it possible to configure on/off for inside and outside ring independently


Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -283,42 +283,57 @@ private Task<ProactiveCopyResult> RequestProactiveCopyIfNeededAsync(OperationCon
                             return ProactiveCopyResult.CopyNotRequiredResult;
                         }
 
-                        IReadOnlyList<MachineLocation> buildRingMachines;
+                        IReadOnlyList<MachineLocation> buildRingMachines = null;
 
                         // Get random machine inside build ring
                         Task<BoolResult> insideRingCopyTask;
-                        if (_buildIdHash != null)
+                        if (Settings.EnableProactiveCopyInsideRing)
```
SergeyTeplyakov: Does it make sense to switch from having two bools to have "ProactiveCopyMode" with 3 values - Disable, InsideTheRingOnly, Full?

SergeyTeplyakov: If you'll stay with the current mode, we should assert that is impossible to "Enable" proactive copy but "disable" both - copy inside and outside the ring.

---
In reply to: [324778894](https://github.com/microsoft/BuildXL/pull/872#discussion_r324778894) [](ancestors = 324778894)

JuanCarlosGI: I slightly prefer having the two configurations be independent of each other, so I'll add the assertion to make sure that we're on a valid state at all points.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -283,42 +283,57 @@ private Task<ProactiveCopyResult> RequestProactiveCopyIfNeededAsync(OperationCon
                             return ProactiveCopyResult.CopyNotRequiredResult;
                         }
 
-                        IReadOnlyList<MachineLocation> buildRingMachines;
+                        IReadOnlyList<MachineLocation> buildRingMachines = null;
 
                         // Get random machine inside build ring
                         Task<BoolResult> insideRingCopyTask;
-                        if (_buildIdHash != null)
+                        if (Settings.ProactiveCopyMode.HasFlag(ProactiveCopyMode.InsideRing))
```
SergeyTeplyakov: Keep in mind that this operation causes 2 boxing allocations.

so an alternative (if you want) is if (Settings.ProactiveCopyMode & ProactiveCopyMode.InsideRing != 0).

Public/Src/Cache/DistributedCache.Host/Service/Internal/DistributedContentStoreFactory.cs
```diff
@@ -218,8 +218,8 @@ public IContentStore CreateContentStore(AbsolutePath localCacheRoot)
                         PinConfiguration = pinConfiguration,
                         EmptyFileHashShortcutEnabled = _distributedSettings.EmptyFileHashShortcutEnabled,
                         RetryIntervalForCopies = _distributedSettings.RetryIntervalForCopies,
-                        EnableProactiveCopy = _distributedSettings.EnableProactiveCopy,
                         TimeoutForProactiveCopies = TimeSpan.FromMinutes(_distributedSettings.TimeoutForProactiveCopiesMinutes),
+                        ProactiveCopyMode = (ProactiveCopyMode)Enum.Parse(typeof(ProactiveCopyMode), _distributedSettings.ProactiveCopyMode),
```
SergeyTeplyakov: nit: I would prefer to create an extension/helper method for parsing the input.

Right now if the settings are not correct the error message would be: "Requested value 'yourenumvaluewithtypo' was not found.

So an alternative may be a dedicated helper that will throw a bit more descriptive error message like:

"Failed to parse ProactiveCopyMode property 'yourenumvalueiwthtypeo'.'.

But this is minor, because the default error message contains an invalid value.

JuanCarlosGI: Leaving as is, which follows the normal pattern used in our code.

PR #875, opened 2019-09-13
Fix Drop Associate method (missing files mismatch)

olkononenko:
It appears that during drop associate call we were using a field that might not contain the full list of missing blobs.

[AB#1598050](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1598050)

### DIFF
--- a/Public/Src/Tools/DropDaemon/VsoClient.cs
+++ b/Public/Src/Tools/DropDaemon/VsoClient.cs
@@ -341,6 +341,12 @@ namespace Tool.DropDaemon
 
             var startTime = DateTime.UtcNow;
 
+            // m_dropClient.AssociateAsync does some internal batching. For each batch, it will create AssociationsStatus.
+            // The very first created AssociationsStatus is stored and later returned as Item2 in the tuple.
+            // Elements from all AssociationsStatus.Missing are added to the same IEnumerable<BlobIdentifier> and returned
+            // as Item2 in the tuple.
+            // If the method creates more than one batch (i.e., more than one AssociationsStatus is created), the returned
+            // associateResult.Item1 will not match associateResult.Item2.Missing.
             Tuple<IEnumerable<BlobIdentifier>, AssociationsStatus> associateResult = await m_dropClient.AssociateAsync(
                 DropName,
                 blobsForAssociate.ToList(),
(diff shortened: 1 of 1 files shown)


narasamdya:
Is this a fix? Or is this an attempt to fix? I mean, are the changes here meant for getting more information about some issue? If they are, perhaps, you should change the title of your PR/commit.

olkononenko:
yes, it's a fix. there are lots of logging added, but this line is what should fix the issue `result.Missing = associateResult.Item1;` (line #371)

---
In reply to: [531341503](https://github.com/microsoft/BuildXL/pull/875#issuecomment-531341503) [](ancestors = 531341503)

Public/Src/Tools/DropDaemon/VsoClient.cs
```diff
@@ -347,9 +347,31 @@ private async Task<AssociationsStatus> AssociateAsync(FileBlobDescriptor[] blobs
...
-            Interlocked.Add(ref Stats.NumFilesAssociated, blobsForAssociate.Length - associateResult.Item2.Missing.Count());
-            return associateResult.Item2;
+
+            var missingBlobIdsCount = associateResult.Item1.Count();
+            var associationsStatusMissingBlobsCount = associateResult.Item2.Missing.Count();
+            Interlocked.Add(ref Stats.NumFilesAssociated, blobsForAssociate.Length - missingBlobIdsCount);
+
+            if (missingBlobIdsCount != associationsStatusMissingBlobsCount)
+            {
+                m_logger.Warning("Mismatch in the number of missing files during Associate call -- missingBlobIdsCount={0}, associationsStatusMissingBlobsCount={1}",
```
mpysson: >Warning [](start = 25, length = 7)

Should this be a warning? Based on your theory this is just the way it works. Definitely log it for sake of debugging and to confirm our suspicion, but it probably deserves to be lower level like verbose

mpysson:
>             Tuple<IEnumerable<BlobIdentifier>, AssociationsStatus> associateResult = await m_dropClient.AssociateAsync(

Could you add some comments describing our understanding of the result of this API? That will help explain why we have the code below to check for agreement between Tuple1 and the Missing field on Tuple2

---
Refers to: Public/Src/Tools/DropDaemon/VsoClient.cs:344 in 6f068dd. [](commit_id = 6f068dde7e9fed1b24a482e9c1debf8c8d58aa8b, deletion_comment = False)

PR #879, opened 2019-09-16
Avoid locking down RocksDb for user-caused exceptions

jbayardo:
Currently, a piece of code like this:

```
accessor.Use((store) => {
    throw new Exception();
});
```

Will create a file that locks down RocksDb for any further use, even though the actual instance does _not_ have a problem. For non-critical stores, that is fine, and probably the "safest" option.

For use-cases like the _RocksDbMemoizationStore_ or the _RocksDbContentLocationDatabase_, which are important to the cache, this behavior is bad: we may fail builds simply because of a problem that we _can_ actually withstand. If the problem is widespread, then we can even bring down an entire stamp.

This PR introduces the notion of exception handling modes:

- In STRICT mode, the behavior is the default that we have been using.
- In RELAXED mode, the intent is to only lock down the database if the exception can be _proven_ to come from RocksDb, thereby avoiding cases like the one displayed above.

It also adds a callback for cases when the store gets disabled, so the user can recover the DB (this will be used in a future PR).

This also changes a bad behavior when doing snapshots: right now, if a store is disabled and a snapshot is created, then the snapshot won't be disabled. With these changes, the snapshot will be marked as disabled as well. _This is something to discuss_.

The way to distinguish the two kinds of exceptions is via examining a stack trace of the exception received, and checking if it is in the RocksDb assembly. The reason we do it this way instead of using a proxy is because a proxy would have difficulty dealing with `IEnumerable` and lambdas called inside the `RocksDbStore`.

WIP:
- More tests for relaxed mode.


Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs
```diff
@@ -575,30 +665,30 @@ public KeyValueStoreAccessor(KeyValueStoreAccessor accessor)
...
                 {
-                    return HandleException(ex);
+                    return use(m_store, state);
                 }
-                // Provide an opportunity for caller to handle any unknown exception type, including unmanaged ones, then rethrow
                 catch (Exception ex)
                 {
-                    HandleException(ex);
-                    throw;
+                    var result = HandleException(ex, out var rethrow);
```
SergeyTeplyakov: One way to deal with this pattern is to give an option to the caller (that provides an error handling function) to "handle" the error.

For instance, this pattern is used by TaskScheduler.UnobservedTaskException. The client subscribes to this event and in the callback they can set e.Handled = true; to avoid error propagation.

We can consider using the same pattern here, because it fits the needs, imo.

SergeyTeplyakov: Another benefit of the proposed approach is simplicity: I just quickly looked at stack walking logic inside HandleExcpetion function and it looked very scary to me.

---
In reply to: [324777386](https://github.com/microsoft/BuildXL/pull/879#discussion_r324777386) [](ancestors = 324777386)

Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs
```diff
@@ -22,6 +22,31 @@ namespace BuildXL.Engine.Cache.KeyValueStores
...
+            /// <summary>
+            /// Any exception produced within code that uses the key-value store will cause the store to shutdown,
+            /// call the invalidation handler, call the failure handler, and return a failure containing the exception.
+            /// 
+            /// If the exception is user-produced, it is thrown for the user to handle.
+            /// </summary>
+            /// <remarks>
+            /// This is the default due to backwards-compatibility concerns.
+            /// </remarks>
+            STRICT,
```
dcwuser: Convention in C# world is not to ALLCAP enum values, but use PascalCase.

Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs
```diff
@@ -638,17 +736,75 @@ public Possible<Unit> Use(Action<IBuildXLKeyValueStore> use)
...
+            {
+                // The SEHException class handles SEH (structured exception handling) errors that are thrown from 
+                // unmanaged code, but that have not been mapped to another .NET Framework exception. The SEHException
+                // class also corresponds to the HRESULT E_FAIL (0x80004005).
+                isRocksDbError = true;
+            }
+
+            if (isRocksDbError)
+            {
+                return (true, false);
```
dcwuser: In strict policy above, handler is called for RocksDbErrors, but in relaxed policy, it is not. Is that intentional/desirable?

jbayardo: That is intentional. In the relaxed policy, only the critical error handler is called. In this way, the failure handler doesn't have to have something like "if this is critical exception, then X"

Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs
```diff
@@ -89,10 +114,25 @@ public partial class KeyValueStoreAccessor : IDisposable
         public bool ReadOnly { get; }
 
         /// <summary>
-        /// Called when an exception is thrown, allows for custom exception handling.
+        /// Determines when the failure handler will be called, when the store will be locked down because of an 
+        /// error, and whether an exception should be rethrown.
+        /// </summary>
+        private readonly Func<Failure<Exception>, (bool lockdown, bool rethrow)> m_errorPolicy;
```
SergeyTeplyakov: Instead of using a tuple, I would suggest to wrap this thing into a separate type, similar to UnobservedTaskExceptionEventArgs ([https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.unobservedtaskexceptioneventargs?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.unobservedtaskexceptioneventargs?view=netframework-4.8)).

To be honest, I'm not a big fan of using tuples in the public surface.

Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs
```diff
@@ -638,17 +736,75 @@ public Possible<Unit> Use(Action<IBuildXLKeyValueStore> use)
...
+                // unmanaged code, but that have not been mapped to another .NET Framework exception. The SEHException
+                // class also corresponds to the HRESULT E_FAIL (0x80004005).
+                isUserError = false;
+            }
+
+            m_failureHandler?.Invoke(failure);
+            return (true, isUserError);
+        }
+
+        private (bool lockdown, bool rethrow) RelaxedErrorPolicy(Failure<Exception> failure)
```
SergeyTeplyakov: When we discuss this thing, I suggested (I think) to actually move the entire logic into the clients.

For instance, instead of having strict vs relaxed mode, I would delegate this decision to the client in a form of a call back.

It means that instead of taking two argumetns (the mode and the handler), the clients will be able to provide a handler similr to one from TaskUnbservedException handler when the callback can set the result whether the error is handled (non strict mode) or not (strict mode).

// the client code:

var accessor = new Accessor(MyHandler);

private void Handler(HandlerArgs args) {
  // the client is strict:
  args.Handled = true;
}

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -197,9 +196,14 @@ private BoolResult Load(OperationContext context, StoreSlot activeSlot, bool cle
                 var possibleStore = KeyValueStoreAccessor.Open(storeLocation,
                     additionalColumns: new[] { nameof(Columns.ClusterState), nameof(Columns.Metadata) },
                     rotateLogs: true,
-                    failureHandler: failure =>
+                    failureHandler: failureEvent =>
                     {
-                        Tracer.Error(context, $"RocksDb critical error caused store deprecation: {failure.DescribeIncludingInnerFailures()}");
+                        bool isUserError = failureEvent.Rethrow;
+                        failureEvent.Invalidate = !isUserError;
+                        failureEvent.Rethrow = isUserError;
```
SergeyTeplyakov: 2 nits:

1. Rethrow is set to itself:)
2. Does it make sense to have a helper method on the failureEvent instance, something: "InvalidateOnUserError" that will set the right flags?

jbayardo: That would be weird, because it would effectively have to operate on the flag's current state, which may or may not match what is expected if the user changed it. Hence,

```
failureEvent.InvalidateOnUserError()
failureEvent.InvalidateOnUserError()
```

Would be strictly wrong (because it wouldn't be idempotent). I'd rather leave it as-is and added a comment.

Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs
```diff
@@ -308,6 +342,9 @@ private struct VersionConstants
         /// <param name="openBulkLoad">
         /// Have RocksDb open for bulk loading.
         /// </param>
+        /// <param name="invalidationHandler">
+        /// <see cref="m_invalidationHandler"/>
```
JuanCarlosGI: nit: Since m_invalidationHandler is private, doesn't it make sense to just copy the description instead of referencing the private member?

jbayardo: Will address this in a separate PR, as I'm going to move these into a separate configuration object.

PR #880, opened 2019-09-16
Add mac assemblies in cache packages

JuanCarlosGI:
For now, netcoreapp3.0 assemblies will be put into windows and mac subfolders. This is the only short term solution that we have time for now, and this will be of use to Artifact Services, which is trying to move to netcoreapp.

The layout change of our NuGet packages should not be a breaking change since no one seems to be using our netcoreapp3 assemblies.


Public/Src/Deployment/privatePackages.dsc
```diff
@@ -94,6 +95,17 @@ namespace PrivatePackages {
                         importFrom("BuildXL.Utilities.Instrumentation").withQualifier(winx64Qualifier).Common.dll.runtime,
                     ],
                 },
+                {
+                    subfolder: r`lib/netcoreapp3.0/osx-x64`,
+                    contents: [
+                        importFrom("BuildXL.Utilities").withQualifier(osxx64Qualifier).dll.runtime,
```
Kr1sso: Putting the common subset into a function and calling it with the qualifier would reduce duplication / potential errors. The whole cache reusable dll business is malformed (we have to fix this later), as we should just publish the old assemblies .NET and one assembly for .NET Core, which is targeting .NETStandard2+.

PR #882, opened 2019-09-16
Restore checkpoint in parallel

SergeyTeplyakov:
I was considering adding another configuration option but decided to stay with the same.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
@@ -367,16 +367,38 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                     // Parse the checkpoint info for the checkpoint being restored
                     var newCheckpointInfo = ParseCheckpointInfo(checkpointFile);
 
-                    foreach (var (key, value) in newCheckpointInfo)
-                    {
-                        await RestoreFileAsync(context, checkpointTargetDirectory, key, value).ThrowIfFailure();
-                    }
+                    await RestoreFilesAsync(context, checkpointTargetDirectory, newCheckpointInfo);
 
                     // Finalize by adding the incremental checkpoint info file to the local incremental checkpoint directory
                     await HardlinkWithFallBackAsync(context, checkpointFile, _incrementalCheckpointInfoFile);
                     UpdateIncrementalCheckpointInfo(newCheckpointInfo);
                     return BoolResult.Success;
                 });
+        }
+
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
```diff
@@ -379,6 +376,31 @@ private Task<BoolResult> RestoreCheckpointIncrementalAsync(OperationContext cont
...
+                await ParallelAlgorithms.WhenDoneAsync(
+                    _configuration.IncrementalCheckpointDegreeOfParallelism,
+                    context.Token,
+                    action: async (addItem, kvp) =>
+                    {
+                        var key = kvp.Key;
+                        var value = kvp.Value;
+                        await RestoreFileAsync(context, checkpointTargetDirectory, key, value).ThrowIfFailure();
+                    },
+                    items: newCheckpointInfo.ToArray());
```
jbayardo: Question: what happens if one of the parallel restores fails here?

SergeyTeplyakov: In this case WhenDoneAsync will fail with aggregateexception.

---
In reply to: [325331241](https://github.com/microsoft/BuildXL/pull/882#discussion_r325331241) [](ancestors = 325331241)

ljcollins25:
>     internal sealed class CheckpointManager

Should we go ahead and add timeouts around Create/Restore checkpoint?

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs:29 in 7deb6a0. [](commit_id = 7deb6a03c23d257e1e54f7cd92cd2122aad9acfe, deletion_comment = False)

ljcollins25:
>     internal sealed class CheckpointManager

Actually, that might be non-trivial since we need to ensure that no files are locked which might caused failures in subsequent operations.

---
In reply to: [532680719](https://github.com/microsoft/BuildXL/pull/882#issuecomment-532680719) [](ancestors = 532680719)

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs:29 in 7deb6a0. [](commit_id = 7deb6a03c23d257e1e54f7cd92cd2122aad9acfe, deletion_comment = False)

jbayardo:
>     internal sealed class CheckpointManager

What do  you mean by "no files are locked"? which files?

---
In reply to: [532681435](https://github.com/microsoft/BuildXL/pull/882#issuecomment-532681435) [](ancestors = 532681435,532680719)

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs:29 in 7deb6a0. [](commit_id = 7deb6a03c23d257e1e54f7cd92cd2122aad9acfe, deletion_comment = False)

PR #884, opened 2019-09-16
More logging for retried processes

norahuang:
[AB#1554006](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1554006)

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutionResult.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutionResult.cs
@@ -138,14 +138,15 @@ namespace BuildXL.Processes
             long processStartTime,
             long maxDetoursHeapSize,
             ContainerConfiguration containerConfiguration,
-            string stdLog)
+            Tuple<AbsolutePath, Encoding> encodedStandardError,
+            Tuple<AbsolutePath, Encoding> encodedStandardOutput)
         {
             return new SandboxedProcessPipExecutionResult(
                 SandboxedProcessPipExecutionStatus.ShouldBeRetriedDueToUserSpecifiedExitCode,
                 observedFileAccesses: default(SortedReadOnlyArray<ObservedFileAccess, ObservedFileAccessExpandedPathComparer>),
                 sharedDynamicDirectoryWriteAccesses: default(Dictionary<AbsolutePath, IReadOnlyCollection<AbsolutePath>>),
-                encodedStandardError: null,
-                encodedStandardOutput: null,
+                encodedStandardError: encodedStandardError,
+                encodedStandardOutput: encodedStandardOutput,
                 numberOfWarnings: 0,
                 unexpectedFileAccesses: null,
                 primaryProcessTimes: primaryProcessTimes,
@@ -158,8 +159,7 @@ namespace BuildXL.Processes
                 allReportedFileAccesses: null,
                 detouringStatuses: detouringStatuses,
                 maxDetoursHeapSize: maxDetoursHeapSize,
-                containerConfiguration: containerConfiguration,
-                stdLog: stdLog);
+                containerConfiguration: containerConfiguration);
         }
 
         internal static SandboxedProcessPipExecutionResult RetryProcessDueToAzureWatsonExitCode(
@@ -311,8 +311,7 @@ namespace BuildXL.Processes
             IReadOnlyList<ReportedFileAccess> allReportedFileAccesses,
             IReadOnlyList<ProcessDetouringStatusData> detouringStatuses,
             long maxDetoursHeapSize,
-            ContainerConfiguration containerConfiguration,
-            string stdLog="")
+            ContainerConfiguration containerConfiguration)
         {
             Contract.Requires(
                 (status == SandboxedProcessPipExecutionStatus.PreparationFailed || status == SandboxedProcessPipExecutionStatus.ShouldBeRetriedDueToUserSpecifiedExitCode) ||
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -1294,6 +1294,20 @@ internal static string PipWriteFileFailedMessage(string pipDescription, string p
...
         public abstract void PipWillBeRetriedDueToExitCode(LoggingContext context, long pipSemiStableHash, string pipDescription, int exitCode, int remainingRetries);
 
+        [GeneratedEvent(
+            (int)EventId.PipRetry,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Verbose,
+            Keywords = (int)(Keywords.UserMessage | Keywords.UserError),
+            EventTask = (int)Tasks.PipExecutor,
+            Message = EventConstants.PipSpecPrefix + " - Retry failed process")]
+        public abstract void PipProcessRetry(
```
mpysson: >PipProcessRetry [](start = 29, length = 15)

It would be good to include the stdout/stderr along with the exit code to show why it is being retried

mpysson:
>                             // There are very rare cases where a child process is started not Detoured and we don't observe any file accesses from such process.

Hm, it seems this is the internal retry where somethign goes wrong with detouring the file. I think the logging we're missing is on the user configurable retries based on the processes exit code

---
Refers to: Public/Src/Engine/Scheduler/PipExecutor.cs:1347 in eb96191. [](commit_id = eb96191dc1dff8d49216cfc2ae8f26f42399a1c1, deletion_comment = False)

mpysson:
>                             // There are very rare cases where a child process is started not Detoured and we don't observe any file accesses from such process.

Here's a PR for some context:
[https://github.com/microsoft/BuildXL/pull/414](https://github.com/microsoft/BuildXL/pull/414)

Iman added some logging for sake of debugging when a process retried. Unfortunately that caused crashes since if the process succeeded on retry there would be an error logged but the build would pass. So i had to remove it. This task was about getting that logging of the stdout/stderr back. So I guess this one is just a matter of updated SamdboxedProcessPipExecutor to log in that case. And it looks like we already have a message for this scenario (PipWillBeRetriedDueToExitCode). So that event just needs to get pip stdout/stderr logging added to it.

---
In reply to: [532294730](https://github.com/microsoft/BuildXL/pull/884#issuecomment-532294730) [](ancestors = 532294730)

---
Refers to: Public/Src/Engine/Scheduler/PipExecutor.cs:1347 in eb96191. [](commit_id = eb96191dc1dff8d49216cfc2ae8f26f42399a1c1, deletion_comment = False)

mpysson:
>                                     Logger.Log.PipWillBeRetriedDueToExitCode(

Are you missing some changes to this log method to accommodate the new parameters?

---
Refers to: Public/Src/Engine/Scheduler/PipExecutor.cs:1487 in d8da891. [](commit_id = d8da891f1a33af452b317dec6493d09547a7ca12, deletion_comment = False)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -3954,6 +3957,58 @@ private async Task<bool> TryLogOutputAsync(SandboxedProcessResult result)
             }
         }
 
+        private async Task<string> GetStdLogStringAsync(SandboxedProcessResult result)
```
narasamdya: >GetStdLogStringAsync [](start = 35, length = 20)

This code is 99% similar to `TryLogOutputAsync`. Can you refactor it?

Public/Src/Engine/Processes/SandboxedProcessPipExecutionResult.cs
```diff
@@ -309,7 +311,8 @@ internal static SandboxedProcessPipExecutionResult MismatchedMessageCountFailure
             IReadOnlyList<ReportedFileAccess> allReportedFileAccesses,
             IReadOnlyList<ProcessDetouringStatusData> detouringStatuses,
             long maxDetoursHeapSize,
-            ContainerConfiguration containerConfiguration)
+            ContainerConfiguration containerConfiguration,
+            string stdLog="")
```
narasamdya: >stdLog="" [](start = 19, length = 9)

Perhaps `string stdLog = null`. Then, in the assignment below you have `StdLog = stdLog ?? string.Empty;`.

narasamdya:
>                 encodedStandardOutput: null,

Why don't we just assign this value when the process was retried. Then in `PipExecutor` we simply read that files. So you don't have to pass a gigantic `stdLog` string here.

---
Refers to: Public/Src/Engine/Processes/SandboxedProcessPipExecutionResult.cs:148 in 0db7714. [](commit_id = 0db77143a17764ea13af3786d4b00ebb5f656186, deletion_comment = False)

narasamdya:
Unit test, please.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1374,6 +1374,17 @@ private void LogCommandLineTooLong(SandboxedProcessInfo info)
 
                         if (m_pip.RetryExitCodes.Contains(result.ExitCode) && m_remainingUserRetryCount > 0)
                         {
+                            Tuple<AbsolutePath, Encoding> encodedStandardError = null;
+                            Tuple<AbsolutePath, Encoding> encodedStandardOutput = null;
+                            if (await TrySaveAndLogStandardOutputAsync(result))
```
narasamdya: >TrySaveAndLogStandardOutputAsync [](start = 38, length = 32)

`TrySaveAndLogStandard*` logs the content of the file. So, why do we need to log it again in `PipExecutor.cs` ?

PR #887, opened 2019-09-17
Change working directory for running VmCommandProxy

narasamdya:
Currently when running VmCommandProxy, the working directory is set to the working directory of the pip. However this can "conflict" with the operation that the pip will do to that working directory. 

VmCommandProxy writes log files to that working directory.  QTest pips set their working directory to the deployment directory. Later the pips robocopy this directory. The robocopy process will cause DFA when it finds the logs written by VmCommandProxy.

[AB#1590524](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1590524)

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -1022,9 +1022,10 @@ namespace BuildXL.Processes
                 {
                     Contract.Assert(m_sandboxConfig.AdminRequiredProcessExecutionMode == AdminRequiredProcessExecutionMode.ExternalVM);
 
-                    Tracing.Logger.Log.PipProcessStartExternalVm(m_loggingContext, m_pip.SemiStableHash, m_pip.GetDescription(m_context));
-
+                    // Initialize VM once.
                     await m_vmInitializer.LazyInitVmAsync.Value;
+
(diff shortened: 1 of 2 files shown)


Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1022,10 +1022,11 @@ private bool ShouldSandboxedProcessExecuteExternal
                 {
                     Contract.Assert(m_sandboxConfig.AdminRequiredProcessExecutionMode == AdminRequiredProcessExecutionMode.ExternalVM);
 
-                    Tracing.Logger.Log.PipProcessStartExternalVm(m_loggingContext, m_pip.SemiStableHash, m_pip.GetDescription(m_context));
-
+                    // Initialize VM once.
                     await m_vmInitializer.LazyInitVmAsync.Value;
 
+                    Tracing.Logger.Log.PipProcessStartExternalVm(m_loggingContext, m_pip.SemiStableHash, m_pip.GetDescription(m_context));
```
norahuang: Why you need to move the logging here? what is the difference?

narasamdya: The logging is a bit confusing before. It shows that the VM execution starts before initialization happens. So, this ordering should remove that confusion.

PR #888, opened 2019-09-17
Add netcoreapp3.0 bits to forCloudBuild package

kewilli:


Public/Src/Deployment/privatePackages.dsc
```diff
@@ -60,7 +61,30 @@ namespace PrivatePackages {
                         DetoursServices.Deployment.natives,
                         importFrom("BuildXL.Utilities").withQualifier(net472Qualifier).Branding.brandingManifest
                     ]
-                }
+                },
+                
+                // Net Core App 3.0
```
erikma: >// Net Core App 3.0 [](start = 16, length = 19)

Also need netstandard2.0 if available

Public/Src/Deployment/privatePackages.dsc
```diff
@@ -60,7 +61,30 @@ namespace PrivatePackages {
                         DetoursServices.Deployment.natives,
                         importFrom("BuildXL.Utilities").withQualifier(net472Qualifier).Branding.brandingManifest
                     ]
-                }
+                },
+                
+                // Net Core App 3.0
+                Nuget.createAssemblyLayout(importFrom("BuildXL.Engine").withQualifier(winx64Qualifier).Processes.dll),
```
Kr1sso: There is tons of duplication here - we'll need to change things around to support pure .netstandard2.0 frameworks only. Please extract the common assemblies into a function that returns them and takes a qualifier as parameter, so you can call it where appropriate.

Public/Src/Deployment/privatePackages.dsc
erikma: nm, netstandard not available, net472+necoreapp3.0 are fine

Public/Src/Deployment/privatePackages.dsc
erikma: no need for netstandard after all, but the duplication is a concern.

PR #889, opened 2019-09-17
Component Governance for NuGet Packages

rijulluman-msft:
[AB#1580270](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1580270)


Public/Src/FrontEnd/Nuget/NugetCgManifestGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+using BuildXL.Utilities.Collections;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace BuildXL.FrontEnd.Nuget
+{
+    /// <summary>
+    /// NugetCgManifestGenerator is used for creation and comparasion of the cgmanifest.json file.
+    /// cgmanifest.json contains all the Nuget Packages used in BuildXL with all their versions in use
+    /// cgmanifest.json is used by Component Governance in Cloud Build to determine security risks within components used by BuildXL
```
mpysson: >in Cloud Build [](start = 56, length = 14)

Not necessarily just in CB. Suggest remove

Public/Src/FrontEnd/Nuget/NugetCgManifestGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// Compares <paramref name="lhsManifest"/> and <paramref name="rhsManifest"/> for equality;
+        /// returns true if they are equal, and false otherwise. 
+        /// 
+        /// This equality check is case-insensitive and white space agnostic.
+        /// </summary>
+        public bool CompareForEquality(string lhsManifest, string rhsManifest)
+        {
+            try
+            {
+                return JToken.DeepEquals(JObject.Parse(lhsManifest), JObject.Parse(rhsManifest));
```
mpysson: >return JToken.DeepEquals(JObject.Parse(lhsManifest), JObject.Parse(rhsManifest)); [](start = 16, length = 81)

We always want this file to be generated and not human edited. Therefore it would be ok for you to just hash the contents of the file and use that for the comparison.

Public/Src/FrontEnd/Nuget/NugetResolver.cs
```diff
@@ -20,6 +25,7 @@ namespace BuildXL.FrontEnd.Script
     /// </summary>
     public sealed class NugetResolver : DScriptSourceResolver
     {
+        internal const string NugetResolverName = "ComponentGovernance";
```
mpysson: >ComponentGovernance [](start = 51, length = 19)

Nit - I'd call this manifest generator or something since we aren't actually performing the CG checks here. I wouldn't want to mislead anyone

Public/Src/FrontEnd/Nuget/NugetResolver.cs
```diff
@@ -62,12 +69,87 @@ public override async Task<bool> InitResolverAsync(IResolverSettings resolverSet
...
+                Configuration.FrontEnd.ValidateCgManifestForNugets.IsValid)
+            {
+                var cgManfiestGenerator = new NugetCgManifestGenerator(Context);
+                string generatedCgManifest = cgManfiestGenerator.GenerateCgManifestForPackages(maybePackages.Result);
+                string existingCgManifest = "{}";
+
+                if ( !Configuration.FrontEnd.GenerateCgManifestForNugets.IsValid &&
+                      Configuration.FrontEnd.ValidateCgManifestForNugets.IsValid )
+                {
+                    // Validation of existing cgmainfest.json results in failure due to mismatch. Should fail the build in this case.
```
mpysson: >cgmainfest.json [](start = 46, length = 15)

the filename is user configurable so it might be better to just refer to it as the manifest

Public/Src/FrontEnd/Nuget/NugetResolver.cs
```diff
@@ -62,12 +69,87 @@ public override async Task<bool> InitResolverAsync(IResolverSettings resolverSet
...
+                        throw new BuildXLException(@"Cannot read cgmanifest.json file from disk", e);
+                    }
+                    catch (FileNotFoundException e)
+                    {
+                        throw new BuildXLException(@"Cannot read cgmanifest.json file from disk", e);
+                    }
+                    if (!cgManfiestGenerator.CompareForEquality(generatedCgManifest, existingCgManifest))
+                    {
+                        throw new BuildXLException(@"Existing cgmanifest.json file is outdated, please generate a new one using /GenerateCgManifestForNugets:path");
+                    }
```
mpysson: Is the caller handling the error logging? Our typical pattern would be to log a unique error event here and then return false to indicate teh failure for the caller

Public/Src/FrontEnd/Nuget/NugetResolver.cs
```diff
@@ -62,12 +69,87 @@ public override async Task<bool> InitResolverAsync(IResolverSettings resolverSet
...
+                // CgManifest FileNotFound, continue to write the new file
+                // No operations required as the empty existingCgManifest will not match with the newly generated cgManifest
+                catch (DirectoryNotFoundException) { }
+                catch (FileNotFoundException) { }
+
+                if (!cgManfiestGenerator.CompareForEquality(generatedCgManifest, existingCgManifest))
+                {
+                    
+
+                    if (Configuration.FrontEnd.GenerateCgManifestForNugets.IsValid)
```
mpysson: extra whitespace

mpysson:
We need to think through how this will work if we update the manifest generator in the future. Our build automation will attempt to do runs with both the current version of bxl.exe and the new version where you would have changes. That makes changing the format tricky. How should we perform an update?

This is a bit theoretical since the format of the manifest is dictated by the consumer, but it could happen

mpysson:
> {

It's be great if we could generate some comment into this saying it is auto-generated and people shouldn't try to manually modify it.

---
Refers to: cg/nuget/cgmanifest.json:1 in 1b527da. [](commit_id = 1b527dab31fe9fedaaca0d3a277628329693ed09, deletion_comment = False)

Public/Src/FrontEnd/Nuget/NugetResolver.cs
mpysson: And the const member's name too

---
In reply to: [325449958](https://github.com/microsoft/BuildXL/pull/889#discussion_r325449958) [](ancestors = 325449958)

mpysson:
> {

Discussed offline. I guess that isn't possible

---
In reply to: [532482244](https://github.com/microsoft/BuildXL/pull/889#issuecomment-532482244) [](ancestors = 532482244)

---
Refers to: cg/nuget/cgmanifest.json:1 in 1b527da. [](commit_id = 1b527dab31fe9fedaaca0d3a277628329693ed09, deletion_comment = False)

cg/nuget/README.md
```diff
@@ -0,0 +1,4 @@
+cgmanifest.json is an auto-generated file using the /generateCgManifestForNugets:<path> agrument while using bxl.
+Any changes to this file will be over written whenever we run the bxl command.
+If you are adding additional NuGet Packages to the project, they will automatically be added in the cgmanifest.json file whenever you build the project using the bxl command.
+If you need to add custom entries into a cgmanifest.json file, please create a new file with the name cgmanifest.json in a different folder anywhere within the project, since any changes to the cgmanifest.json in this folder wil always be over-written.
```
mpysson: I think the idea is that this is never necessary. This file is always auto-generated so if you need somethign updated you either need to update the source of truth (config.dsc) or change the source code of the generator

mpysson:
Discussed offline

---
In reply to: [532482051](https://github.com/microsoft/BuildXL/pull/889#issuecomment-532482051) [](ancestors = 532482051)

Public/Src/FrontEnd/Nuget/NugetAnalyzedPackage.cs
```diff
@@ -70,6 +70,9 @@ public sealed class NugetAnalyzedPackage
         /// <nodoc />
         public string Id => PackageOnDisk.Package.Id;
 
+        /// <nodoc />
+        public string NugetName { get; set; }
```
almiliMSFT: Is this property every used?

Public/Src/FrontEnd/Nuget/NugetCgManifestGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
```
almiliMSFT: 👍 for using `DeepEquals`!

Public/Src/FrontEnd/Nuget/NugetCgManifestGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+                // The existing Manifest file was in invalid JSON format.
+                // Hence it does not match.
+                return false;
+            }
+            
+        }
+
+        private string ExtractNugetVersion(Package p)
+        {
+            return p.Path.GetParent(Context.PathTable).GetName(Context.PathTable).ToString(Context.StringTable);
```
almiliMSFT: Let's leave a comment that this relies on the folder structure created by the Nuget resolver.   Ideally, it would be great to find the code that creates that folder structure and cross reference this method in there to say that if that gets changed this method needs to be updated as well.

Public/Src/FrontEnd/Nuget/NugetResolver.cs
```diff
@@ -62,12 +69,89 @@ public override async Task<bool> InitResolverAsync(IResolverSettings resolverSet
...
+                    {
+                        Logger.ReportComponentGovernanceValidationError(Context.LoggingContext, "Cannot read Component Governance Manifest file from disk\n" + e.ToString());
+                        return false;
+                    }
+                    catch (FileNotFoundException e)
+                    {
+                        Logger.ReportComponentGovernanceValidationError(Context.LoggingContext, "Cannot read Component Governance Manifest file from disk\n" + e.ToString());
+                        return false;
+                    }
+                    if (!cgManfiestGenerator.CompareForEquality(generatedCgManifest, existingCgManifest))
```
almiliMSFT: nit: new line before `if`

Public/Src/FrontEnd/Nuget/NugetResolver.cs
```diff
@@ -62,12 +69,89 @@ public override async Task<bool> InitResolverAsync(IResolverSettings resolverSet
...
                 m_packages[package.Id] = package;
                 m_owningModules[package.ModuleId] = package;
             }
 
+            if (Configuration.FrontEnd.GenerateCgManifestForNugets.IsValid ||
+                Configuration.FrontEnd.ValidateCgManifestForNugets.IsValid)
+            {
+                var cgManfiestGenerator = new NugetCgManifestGenerator(Context);
+                string generatedCgManifest = cgManfiestGenerator.GenerateCgManifestForPackages(maybePackages.Result);
+                string existingCgManifest = "INVALID";
```
almiliMSFT: Why does this variable need to be declared here with a bogus value?   Please remove.

Public/Src/FrontEnd/Nuget/NugetResolver.cs
```diff
@@ -62,12 +69,89 @@ public override async Task<bool> InitResolverAsync(IResolverSettings resolverSet
...
+                    // We are calling FrontEndHost.Engine.RecordFrontEndFile towards the end of this function because we may update this file after the read below
+                    // Updating the file will cause a hash mismatch and the build to fail if this file is read again downstream
+                    existingCgManifest = File.ReadAllText(Configuration.FrontEnd.GenerateCgManifestForNugets.ToString(Context.PathTable));
+                }
+                // CgManifest FileNotFound, continue to write the new file
+                // No operations required as the empty existingCgManifest will not match with the newly generated cgManifest
+                catch (DirectoryNotFoundException) { }
+                catch (FileNotFoundException) { }
+
+                if (!cgManfiestGenerator.CompareForEquality(generatedCgManifest, existingCgManifest))
```
almiliMSFT: Why do you want to perform this check here?  Can't you just write the file to disk regardless if it already exists and whether its existing content is the same?

Public/Src/FrontEnd/Nuget/NugetAnalyzedPackage.cs
rijulluman-msft: Yes, it is used in GetOriginalNugetPackageName

Public/Src/FrontEnd/Nuget/NugetResolver.cs
```diff
@@ -62,12 +69,89 @@ public override async Task<bool> InitResolverAsync(IResolverSettings resolverSet
...
             }
 
+            if (Configuration.FrontEnd.GenerateCgManifestForNugets.IsValid ||
+                Configuration.FrontEnd.ValidateCgManifestForNugets.IsValid)
+            {
+                var cgManfiestGenerator = new NugetCgManifestGenerator(Context);
+                string generatedCgManifest = cgManfiestGenerator.GenerateCgManifestForPackages(maybePackages.Result);
+                string existingCgManifest = "INVALID";
+
+                if ( !Configuration.FrontEnd.GenerateCgManifestForNugets.IsValid &&
```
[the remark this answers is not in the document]
rijulluman-msft: The !Configuration.FrontEnd.GenerateCgManifestForNugets.IsValid was used to ignore the Validation flag when the generate flag is set, since the generated file will always be valid

Public/Src/FrontEnd/Nuget/NugetResolver.cs
rijulluman-msft: ok, I'll add a null check inside the compare function

Public/Src/FrontEnd/Nuget/NugetResolver.cs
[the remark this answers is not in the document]
almiliMSFT: Yeah, but maybe the intended semantics of `/validate /generate` should be "validate existing file then generate a new one".  Also, those two flags may point to entirely different files!  Therefore, I'd treat them completely separately, i.e.,: (1) first run validate if specified, then (2) run generate if specified.

PR #890, opened 2019-09-17
Reset RocksDb when the metadata store fails to open

jbayardo:


Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs
```diff
@@ -231,6 +231,9 @@ private struct VersionConstants
         /// <param name="openBulkLoad">
         /// Have RocksDb open for bulk loading.
         /// </param>
+        /// <param name="onStoreReset">
+        /// Callback for when the store gets reset due to `onFailureDeleteExistingStoreAndRetry`
```
kewilli: >onFailureDeleteExistingStoreAndRetry [](start = 59, length = 36)

Nit: `<paramref>`

Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs
```diff
@@ -304,6 +308,9 @@ private struct VersionConstants
...
```
kewilli: >onFailureDeleteExistingStoreAndRetry [](start = 59, length = 36)

Ditto

SergeyTeplyakov:
>             bool openReadOnly = false,

it seems that its time to create a configuration class:)

---
Refers to: Public/Src/Utilities/KeyValueStore/KeyValueStoreAccessor.cs:243 in 4140640. [](commit_id = 41406400096fd1c8998a558ec9051b32bc4d06f7, deletion_comment = False)

PR #895, opened 2019-09-18
Stop tracing operation starts in ContentSessionBase

JuanCarlosGI:
Note that PutResult already contains info about the hash, so we are able to not include the hash in the extraEndMessage.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
+++ b/Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
@@ -149,8 +149,8 @@ namespace BuildXL.Cache.ContentStore.Sessions
                 operationContext => operationContext.PerformOperationAsync(
                     Tracer,
                     () => PlaceFileCoreAsync(operationContext, contentHash, path, accessMode, replacementMode, realizationMode, urgencyHint, _counters[ContentSessionBaseCounters.PlaceFileRetries]),
-                    extraStartMessage: $"({contentHash.ToShortString()},{path},{accessMode},{replacementMode},{realizationMode})",
-                    extraEndMessage: (_) => $"input={contentHash.ToShortString()}",
+                    traceOperationStarted: false,
+                    extraEndMessage: (_) => $"input=({contentHash.ToShortString()},{path},{accessMode},{replacementMode},{realizationMode})",
                     counter: _counters[ContentSessionBaseCounters.PlaceFile]));
         }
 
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
```diff
@@ -253,8 +253,8 @@ protected ContentSessionBase(string name, CounterTracker counterTracker = null)
                 operationContext => operationContext.PerformOperationAsync(
                     Tracer,
                     () => PutFileCoreAsync(operationContext, contentHash, path, realizationMode, urgencyHint, _counters[ContentSessionBaseCounters.PutFileRetries]),
-                    extraStartMessage: $"({path},{realizationMode},{contentHash.ToShortString()}) trusted=false",
-                    extraEndMessage: _ => "trusted=false",
+                    traceOperationStarted: false,
+                    extraEndMessage: _ => $"({path},{realizationMode}) trusted=false",
```
dcwuser: Did we loose contentHash?

Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
```diff
@@ -211,10 +211,10 @@ protected ContentSessionBase(string name, CounterTracker counterTracker = null)
                 operationContext => operationContext.PerformOperationAsync(
                     Tracer,
                     () => PutFileCoreAsync(operationContext, hashType, path, realizationMode, urgencyHint, _counters[ContentSessionBaseCounters.PutFileRetries]),
-                    extraStartMessage: $"({path},{realizationMode},{hashType}) trusted=false",
```
kewilli: >hashType [](start = 68, length = 8)

Do we want to carry this along?

Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
```diff
@@ -253,8 +253,8 @@ protected ContentSessionBase(string name, CounterTracker counterTracker = null)
                 operationContext => operationContext.PerformOperationAsync(
                     Tracer,
                     () => PutFileCoreAsync(operationContext, contentHash, path, realizationMode, urgencyHint, _counters[ContentSessionBaseCounters.PutFileRetries]),
-                    extraStartMessage: $"({path},{realizationMode},{contentHash.ToShortString()}) trusted=false",
```
kewilli: >contentHash [](start = 68, length = 11)

Ditto?

Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
JuanCarlosGI: I think we should keep path. Realization mode I'm not so sure of, but the whole purpose of this is to keep all the current information, so I think we could remove unnecessary info on a later PR.

PR #896, opened 2019-09-18
Reduce tracing.

SergeyTeplyakov:
This change reduces the amount of tracing by about 20%


Public/Src/Cache/ContentStore/Distributed/NuCache/RedisGlobalStore.cs
```diff
@@ -703,11 +703,11 @@ public Task<BoolResult> InvalidateLocalMachineAsync(OperationContext context)
         }
 
         /// <inheritdoc />
-        public Task<BoolResult> PutBlobAsync(OperationContext context, ContentHash hash, byte[] blob)
```
JuanCarlosGI: Any particular reason to make this method async?

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStore.cs
```diff
@@ -1291,11 +1291,11 @@ public Result<IReadOnlyList<DateTime>> GetEffectiveLastAccessTimes(OperationCont
         }
 
         /// <inheritdoc />
-        public Task<BoolResult> PutBlobAsync(OperationContext context, ContentHash hash, byte[] blob)
+        public async Task<BoolResult> PutBlobAsync(OperationContext context, ContentHash hash, byte[] blob)
```
JuanCarlosGI: Why make this async?

Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
```diff
@@ -32,14 +32,17 @@ namespace BuildXL.Cache.ContentStore.Sessions
...
+        private readonly CounterCollection<ContentSessionBaseCounters> _counters;
 
         /// <inheritdoc />
         public string Name { get; }
 
         /// <nodoc />
         protected virtual bool TracePinFinished => true;
 
+        /// <nodoc />
+        protected virtual bool TraceErrorsOnly => false;
```
JuanCarlosGI: Will we not do a virtual property for TraceOperationStart, so that DistributedContentStore still logs starts?

Public/Src/Cache/ContentStore/Distributed/NuCache/RedisGlobalStore.cs
SergeyTeplyakov: yep: type conversion. This method calls PutBlobAsync that returns a type that derives from BoolResult.
There is no covarient conversion from 'Task<DerivedType>' to 'Task<BaseType>'. In order to get it you have to await the result.

Public/Src/Cache/ContentStore/Library/Sessions/ContentSessionBase.cs
SergeyTeplyakov: If TraceErrorsOnly is true, then the base session won't trace anything at all: no starts or stops.

JuanCarlosGI: Yeah, I meant an entirely different property; I only put the comment here because it seemed the logical place where the new property would live.

SergeyTeplyakov: Got it. Added another property.

Public/Src/Cache/ContentStore/Distributed/Redis/RedisBlobAdapter.cs
```diff
@@ -55,10 +55,64 @@ public RedisBlobAdapter(RedisDatabaseAdapter redis, TimeSpan blobExpiryTime, lon
             _tracer = tracer;
         }
 
+        /// <nodoc />
+        public class PutBlobResult : BoolResult
```
dcwuser: I really like this design for results with more data.

Public/Src/Cache/ContentStore/Distributed/Redis/RedisBlobAdapter.cs
```diff
@@ -55,10 +55,64 @@ public RedisBlobAdapter(RedisDatabaseAdapter redis, TimeSpan blobExpiryTime, lon
...
+            {
+                string baseResult = $"Hash=[{Hash.ToShortString()}], BlobSize=[{BlobSize}]";
+                if (Succeeded)
+                {
+                    if (AlreadyInRedis)
+                    {
+                        return $"{baseResult}. PutBlob skipped: the hash is already in Redis.";
+                    }
+
+                    return $"{baseResult}. Successfully reserved space in {RedisKey}. NewCapacity=[{NewCapacityInRedis}].";
```
dcwuser: Maybe less words and more formatting for this stuff. E.g. "Result=[{Success}] Key=[{RedisKey}] NewCapacity=[{NewCapacityInRedis}]" here and "AlreadyInRedis=[true]" above.

PR #897, opened 2019-09-18
Don't sort environment variables when creating pip graph

erickulcyk:
Don't sort environment variables when creating pip graph, instead add them order independent to the fingerprint

This saves about 33% of evaluation time from buildxlscriptanalyzer, about 10% of overall time in buildxlscriptanalyzer

### DIFF
--- a/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
+++ b/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
@@ -56,6 +56,7 @@ namespace BuildXL.Scheduler.Fingerprints
         private readonly ExpandedPathFileArtifactComparer m_expandedPathFileArtifactComparer;
         private readonly Comparer<DirectoryArtifact> m_directoryComparer;
         private readonly Comparer<FileArtifactWithAttributes> m_expandedPathFileArtifactWithAttributesComparer;
+        private readonly Comparer<EnvironmentVariable> m_environmentVariableComparer;
 
         /// <summary>
         /// The tokenizer used to handle path roots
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -293,7 +295,7 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process
                 AddPipData(fingerprinter, "ResponseFileData", process.ResponseFileData);
             }
 
-            fingerprinter.AddCollection<EnvironmentVariable, ReadOnlyArray<EnvironmentVariable>>(
+            fingerprinter.AddOrderIndependentCollection<EnvironmentVariable, ReadOnlyArray<EnvironmentVariable>>(
                 "EnvironmentVariables",
```
narasamdya: >"EnvironmentVariables", [](start = 16, length = 23)

Please add a unit test confirming that we still have a cache hit when environment variables are re-ordered.

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -103,6 +104,7 @@ public string FingerprintSalt
             PathExpander = pathExpander ?? PathExpander.Default;
             m_expandedPathFileArtifactComparer = new ExpandedPathFileArtifactComparer(m_pathTable.ExpandedPathComparer, pathOnly: false);
             m_directoryComparer = Comparer<DirectoryArtifact>.Create((d1, d2) => m_pathTable.ExpandedPathComparer.Compare(d1.Path, d2.Path));
+            m_environmentVariableComparer = Comparer<EnvironmentVariable>.Create((ev1, ev2) => { return ev1.Name.ToString(pathTable.StringTable).CompareTo(ev2.Name.ToString(pathTable.StringTable)); });
```
ljcollins25: >CompareTo [](start = 145, length = 9)

Create a OrdinalStringId comparer and use that to compare the names.

PR #898, opened 2019-09-18
Bandwidth check for gRPC copies

JuanCarlosGI:
The logic for the bandwidth check was taken from: http://index/?leftProject=CloudBuild.Services.Common&leftSymbol=acmnpfersqik&file=Communication%5CFsServer%5CFsServerClient.cs (BandwidthCheckedCopyAsync)

This will cause a breaking change in CloudBuild, since MinimumSpeedInMbPerSec is now double? instead of double. The actual fix for the breaking change should be trivial.

Note: as part of the PR, I moved the gRPC copy client to ContentStore.Distributed, since it is only used there.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
+++ b/Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
@@ -42,7 +42,7 @@ namespace BuildXL.Cache.ContentStore.Utils
             {
                 var startPosition = destinationStream.Position;
                 var timer = Stopwatch.StartNew();
-                await Impl();
+                await impl();
                 timer.Stop();
                 var endPosition = destinationStream.Position;
 
@@ -53,10 +53,10 @@ namespace BuildXL.Cache.ContentStore.Utils
             }
             else
             {
-                await Impl();
+                await impl();
             }
 
-            async Task Impl()
+            async Task impl()
             {
                 // This method should not fail with exceptions because the resulting task may be left unobserved causing an application to crash
                 // (given that the app is configured to fail on unobserved task exceptions).
--- a/Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
+++ b/Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
@@ -9,7 +9,7 @@ using System.Threading.Tasks;
 using BuildXL.Cache.ContentStore.Interfaces.Extensions;
 using BuildXL.Cache.ContentStore.Tracing.Internal;
 
-namespace BuildXL.Cache.ContentStore.Distributed.Utilities
+namespace BuildXL.Cache.ContentStore.Utils
 {
     /// <summary>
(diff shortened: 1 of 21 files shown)


ljcollins25:

Is there a reason we can't move this to ContentStore rather than moving Grpc stuff up to Distributed layer?

---
Refers to: Public/Src/Cache/ContentStore/Distributed/Utilities/BandwidthChecker.cs:1 in 16d976d. [](commit_id = 16d976de9fb63270102d491d9ea17127b3311552, deletion_comment = False)

JuanCarlosGI:

There is no reason, other than I think Distributed is the logical place for this to live in.

---
In reply to: [533670847](https://github.com/microsoft/BuildXL/pull/898#issuecomment-533670847) [](ancestors = 533670847)

---
Refers to: Public/Src/Cache/ContentStore/Distributed/Utilities/BandwidthChecker.cs:1 in 16d976d. [](commit_id = 16d976de9fb63270102d491d9ea17127b3311552, deletion_comment = False)

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -0,0 +1,132 @@
...
+                var bytesCopied = endPosition - startPosition;
+                var speed = bytesCopied / timer.Elapsed.TotalSeconds / (1024 * 1024);
+                _historicalBandwidthLimitSource.AddBandwidthRecord(speed);
+            }
+            else
+            {
+                await Impl();
+            }
+
+            async Task Impl()
```
SergeyTeplyakov: nit: local functions should be lowercased

Public/Src/Cache/ContentStore/Library/Utils/IBandwidthLimitSource.cs
```diff
@@ -0,0 +1,126 @@
...
+            _maxRecordsStored = maxRecordsStored;
+            _bandwidthRecords = new double[maxRecordsStored];
+        }
+
+        /// <summary>
+        /// Adds a record for future use.
+        /// </summary>
+        public void AddBandwidthRecord(double value)
+        {
+            // Do a quick check to protect ourselves against invalid data: NaN's, negative, and zero values.
```
SergeyTeplyakov: double.NaN <= 0.0 is false.

So you're not protected here against NaN's.

Public/Src/Cache/ContentStore/Library/Utils/IBandwidthLimitSource.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        /// <inheritdoc />
+        public double GetMinimumSpeedInMbPerSec()
+        {
+            var statistics = GetBandwidthRecordStatistics();
+            if (statistics.Count < _maxRecordsStored / 2)
+            {
+                // If not enough records, don't impose a limit.
+                return (0.0);
```
SergeyTeplyakov: nit: do you need braces here (and at line 78)

PR #903, opened 2019-09-19
Updating .NETCoreApp package to 3.0

Kr1sso:
This PR updates our internal .NETCore assemblies to ingest the RC1 release.


Public/Src/Cache/ContentStore/Library/FileSystem/TrackingFileStream.cs
```diff
@@ -32,28 +32,28 @@ private string Path
         }
 
         /// <inheritdoc />
-        public TrackingFileStream([NotNull] string path, FileMode mode)
+        public TrackingFileStream([JetBrains.Annotations.NotNull] string path, FileMode mode)
```
mpysson: >[JetBrains.Annotations. [](start = 34, length = 23)

wow, this is pretty annoying. What is this clashing with? I didn't see a NotNull in NetCore 3.0. Maybe I'm not looking hard enough

ljcollins25: If we have some contention on this type name, we need to do something less intrusive. Let's chat when you get a chance.

---
In reply to: [326408729](https://github.com/microsoft/BuildXL/pull/903#discussion_r326408729) [](ancestors = 326408729)

PR #907, opened 2019-09-20
Inline unsafe ambient calls when arguments are literals

narasamdya:
Inline unsafe ambient calls when arguments are literals. Inlining is done during the AST conversion phase.


erickulcyk:
Can you add an example of usage?

narasamdya:
You don't need to change anything. `Unsafe.outputFile` is translated into a proper AST

---
In reply to: [533633338](https://github.com/microsoft/BuildXL/pull/907#issuecomment-533633338) [](ancestors = 533633338)

PR #910, opened 2019-09-20
Make CacheServerFactory public again

kewilli:
`CacheServerFactory` was created from `ContentServerFactory`. The latter was public, but the former is internal. This type is used externally, so this PR returns it to its old access modifier.


Public/Src/Cache/DistributedCache.Host/Service/Internal/CacheServerFactory.cs
```diff
@@ -25,7 +25,7 @@ namespace BuildXL.Cache.Host.Service.Internal
     /// <summary>
     /// Creates and configures cache server instances.
     /// </summary>
-    internal class CacheServerFactory
+    public class CacheServerFactory
```
jbayardo: Nit: Add comment saying this is used externally.

It wasn't obvious nor expected when I renamed it

PR #911, opened 2019-09-20
Collision handling for symbols

semihokur:
Collision handling behavior can be specified by the developer: 
(1) DebugEntryCreateBehavior.ThrowIfExists:
     - We log a warning message on both SymbolDaemon and BuildXL log file.
     - We fail the IPC pip.
(2) DebugEntryCreateBehavior.SkipIfExists:
     - We log a warning message on both SymbolDaemon and BuildXL log file.
(3) DebugEntryCreateBehavior.OverwriteIfExists
     - We log a warning message on both SymbolDaemon and BuildXL log file.
     - We retry adding the symbol with OverwriteIfExists behavior.

### DIFF
--- a/Public/Src/Tools/SymbolDaemon/VsoSymbolClient.cs
+++ b/Public/Src/Tools/SymbolDaemon/VsoSymbolClient.cs
@@ -2,12 +2,15 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using BuildXL.Ipc.Common;
+using BuildXL.Ipc.ExternalApi;
 using BuildXL.Ipc.Interfaces;
+using BuildXL.Utilities;
 using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Services.Content.Common;
 using Microsoft.VisualStudio.Services.Content.Common.Authentication;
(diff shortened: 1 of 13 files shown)


Public/Src/Tools/SymbolDaemon/VsoSymbolClient.cs
```diff
@@ -124,11 +132,42 @@ public async Task<AddDebugEntryResult> AddFileAsync(SymbolFile symbolFile)
...
-                DefaultDebugEntryCreateBehavior,
-                CancellationToken);
+            List<DebugEntry> result;
+
+            try
+            {
+                result = await m_symbolClient.CreateRequestDebugEntriesAsync(
+                    RequestId,
+                    symbolFile.DebugEntries.Select(e => CreateDebugEntry(e)),
+                    DebugEntryCreateBehavior.ThrowIfExists,
```
olkononenko: >DebugEntryCreateBehavior.ThrowIfExists, [](start = 20, length = 39)

nit: add a comment explaining why we use throw if exists here

Public/Src/Tools/SymbolDaemon/VsoSymbolClient.cs
```diff
@@ -124,11 +132,42 @@ public async Task<AddDebugEntryResult> AddFileAsync(SymbolFile symbolFile)
...
+                        : $"SymbolDaemon will retry creating debug entry with {m_debugEntryCreateBehavior} behavior");
+
+                // Log a warning message in BuildXL log file
+                Analysis.IgnoreResult(await m_apiClient.LogMessage(message, isWarning: true));
+ 
+                if (m_debugEntryCreateBehavior == DebugEntryCreateBehavior.ThrowIfExists)
+                {
+                    // Log an error message in SymbolDaemon log file
+                    m_logger.Error(message);
+                    throw new DebugEntryExistsException(message);
```
olkononenko: >throw new DebugEntryExistsException(message); [](start = 20, length = 45)

just `throw;`

semihokur: I add a new message to the exception; as the existing message is too confusing. That's why, I did not want to use 'throw'.

PR #912, opened 2019-09-20
Optimize pin bulk tracing.

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Distributed/Sessions/ReadOnlyDistributedContentSession.cs
```diff
@@ -760,6 +760,22 @@ private async Task<IEnumerable<Task<Indexed<PinResult>>>> PinFromContentLocation
...
+            // Inform the content directory that we copied the files.
+            // Looking for distributed pin results that were successful by copying the content locally.
+            var localCopies = pinnings.Select((rp, index) => (result: rp, index)).Where(x => x.result.Result is DistributedPinResult dpr && dpr.CopyLocally).ToList();
+
+            BoolResult updated = await UpdateContentTrackerWithNewReplicaAsync(operationContext, localCopies.Select(lc => new ContentHashWithSize(lc.result.Record.ContentHash, lc.result.Record.Size)).ToList(), cancel, UrgencyHint.Nominal);
+            if (!updated.Succeeded)
+            {
+                // We failed to update the tracker. Need to update the results.
+                string hashesAsString = string.Join(", ", localCopies.Select(lc => lc.result.Record.ContentHash.ToShortString()));
+                Tracer.Info(operationContext, $"Pin failed for hashes {hashesAsString}: local copy succeeded, but could not inform content directory due to {updated.ErrorMessage}.");
```
JuanCarlosGI: >Info [](start = 23, length = 4)

Debug? I think we need to talk as a team to define what qualifies as Info, Warn, Debug, Diagnostic, etc

SergeyTeplyakov: I would say that this should Info or Warning. 
If we'll decide to stop tracing debug messages (which is reasonable) this trace should stay.

PR #915, opened 2019-09-23
Bug Fix: Pip source dependencies cannot be hashed logging

rijulluman-msft:
[AB#1594480](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1594480)


Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -1341,23 +1342,25 @@ private async Task<Possible<Unit>> TryHashFileArtifactsAsync(PipArtifactsState s
                     {
                         // Directory artifact contents are not hashed since they will be hashed dynamically
                         // if the pip accesses them, so the file is the declared artifact
-                        state.HashTasks.Add(TryQueryContentAsync(
+                        var artifactContentInfo = await TryQueryContentAsync(
                             file,
```
semihokur: Is it intentional to remove the parallelism here?

rijulluman-msft: Fixed in new commit

PR #918, opened 2019-09-23
Validate SST files during checkpointing

jbayardo:
In all error cases in CBTest, we obtain the file properly from remote storage, but RocksDb fails to start anyways due to corruption. The only explanation I can find is that we are not checking that its hash matches after copying. Another idea worth exploring is if we can simply hard link the files instead, so as to avoid that additional hash check. I _think_ it should be OK as long as we keep the files pinned for as long as the checkpoint is in use, but I'm not sure how to verify.

This also adds checks for Azure Blob downloads, as they don't check by default. One thing to keep in mind is that, **by default, the blob client will use a single thread; we may want to change this**. This latter change does not come from observations, but as a matter of principle.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
@@ -15,6 +15,7 @@ using BuildXL.Cache.ContentStore.Tracing;
 using BuildXL.Utilities.Tasks;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
+using Microsoft.WindowsAzure.Storage.RetryPolicies;
 using static BuildXL.Cache.ContentStore.Utils.DateTimeUtilities;
 using DateTimeUtilities = BuildXL.Cache.ContentStore.Utils.DateTimeUtilities;
 using OperationContext = BuildXL.Cache.ContentStore.Tracing.Internal.OperationContext;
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
```diff
@@ -39,6 +40,15 @@ public class BlobCentralStorage : CentralStorage
         /// <inheritdoc />
         protected override Tracer Tracer { get; } = new Tracer(nameof(BlobCentralStorage));
 
+        private static readonly BlobRequestOptions DefaultBlobStorageRequestOptions = new BlobRequestOptions()
```
SergeyTeplyakov: >DefaultBlobStorageRequestOptions [](start = 51, length = 32)

I'm curious, is it a breaking changing option or not? I mean, is it ok to upload the blob with StoreBlobContentMD5 == false and load it with this property as true?

SergeyTeplyakov: I think it is.

---
In reply to: [327371682](https://github.com/microsoft/BuildXL/pull/918#discussion_r327371682) [](ancestors = 327371682)

Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
```diff
@@ -96,15 +106,15 @@ protected override async Task<BoolResult> TryGetFileCoreAsync(OperationContext c
                 AttemptResult result = await TaskUtilities.WithTimeoutAsync(async token =>
                     {
                         var blob = container.GetBlockBlobReference(blobName);
-                        var exists = await blob.ExistsAsync(null, null, token);
+                        var exists = await blob.ExistsAsync(DefaultBlobStorageRequestOptions, null, token);
```
SergeyTeplyakov: I'm not sure that we should use these new default options here... What do you think?

Public/Src/Cache/ContentStore/Distributed/NuCache/BlobCentralStorage.cs
jbayardo: It is an optional feature, as per their documentation.

PR #920, opened 2019-09-23
Make some filesystemcontentstore traces optional.

SergeyTeplyakov:
This PR makes some of the messages from `FileSystemContentStoreInternal` optional (and disabled by default).

For instance, hashing, place file and hard link creation are responsible for about 8% of overall kusto traffic. And now these messages are still available but only when the configuration turns on.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Tracing/EvictResult.cs
+++ b/Public/Src/Cache/ContentStore/Library/Tracing/EvictResult.cs
@@ -116,11 +116,17 @@ namespace BuildXL.Cache.ContentStore.Tracing
             return new DeleteResult(DeleteResult.ResultCode.ContentNotDeleted, contentHash, EvictedSize, PinnedSize);
         }
 
+        /// <nodoc />
+        public TimeSpan Age => CreationTime - LastAccessTime;
+
+        /// <nodoc />
+        public TimeSpan? EffectiveAge => CreationTime - EffectiveLastAccessTime;
+
         /// <inheritdoc />
         public override string ToString()
         {
             return Succeeded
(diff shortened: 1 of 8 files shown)


Public/Src/Cache/ContentStore/Library/Tracing/EvictResult.cs
```diff
@@ -116,11 +116,17 @@ public DeleteResult ToDeleteResult(ContentHash contentHash)
...
+
+        /// <nodoc />
+        public TimeSpan? EffectiveAge => CreationTime - EffectiveLastAccessTime;
+
         /// <inheritdoc />
         public override string ToString()
         {
             return Succeeded
-                ? $"Success Size={EvictedSize} Files={EvictedFiles} Pinned={PinnedSize} LastAccessTime={LastAccessTime} Age={CreationTime - LastAccessTime} ReplicaCount={ReplicaCount} EffectiveLastAccessTime={EffectiveLastAccessTime} EffectiveAge={CreationTime - EffectiveLastAccessTime}"
+                ? $"Success Size={EvictedSize} Files={EvictedFiles} Pinned={PinnedSize} LastAccessTime={LastAccessTime} Age={Age} ReplicaCount={ReplicaCount} EffectiveLastAccessTime={EffectiveLastAccessTime} EffectiveAge={EffectiveAge}"
```
JuanCarlosGI: >LastAccessTime={LastAccessTime} [](start = 88, length = 31)

Weren't we going to remove LastAccessTime and EffectiveLastAccessTime, in favor of just having Age and EffectiveAge?

SergeyTeplyakov: We trace the eviction manually instead of using `Result.ToString`.

PR #921, opened 2019-09-24
Caching for symbol indexing pip

olkononenko:
[AB#1596291](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1596291)

### DIFF
--- a/Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
+++ b/Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
@@ -5,9 +5,11 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.ContractsLight;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
+using BuildXL.Cache.ContentStore.Hashing;
 using BuildXL.Ipc.Common;
 using BuildXL.Ipc.ExternalApi;
 using BuildXL.Ipc.Interfaces;
@@ -15,6 +17,7 @@ using BuildXL.Storage;
 using BuildXL.Utilities.CLI;
 using BuildXL.Utilities.Tasks;
 using Microsoft.VisualStudio.Services.Symbol.App.Core.Tracing;
+using Microsoft.VisualStudio.Services.Symbol.Common;
 using Microsoft.VisualStudio.Services.Symbol.WebApi;
 using Newtonsoft.Json;
 using Tool.ServicePipDaemon;
@@ -97,6 +100,14 @@ namespace Tool.SymbolDaemon
             IsRequired = false,
             DefaultValue = SymbolConfig.DefaultEnableTelemetry,
         });
+
+        internal static readonly StrOption SymbolMetadataFile = new StrOption("symbolMetadata")
+        {
+            ShortName = "sm",
+            HelpText = "Path to file with symbols metadata.",
+            IsRequired = false,
+            IsMultiValue = false,
+        };
 
         internal static SymbolConfig CreateSymbolConfig(ConfiguredCommand conf)
         {
@@ -207,7 +218,7 @@ namespace Tool.SymbolDaemon
         internal static readonly Command AddSymbolFilesCmd = RegisterCommand(
             name: "addsymbolfiles",
             description: "[RPC] invokes the 'addsymbolfiles' operation.",
-            options: new Option[] { IpcServerMonikerRequired, File, FileId, HashOptional },
+            options: new Option[] { IpcServerMonikerRequired, File, FileId, HashOptional, SymbolMetadataFile },
             clientAction: SyncRPCSend,
             serverAction: async (conf, daemon) =>
             {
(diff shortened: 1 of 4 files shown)


Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
```diff
@@ -165,15 +168,55 @@ function addFiles(createResult: SymbolCreateResult, args: OperationArguments, fi
...
+function indexSymbolFiles(files: File[]) : DerivedFile {
+    const symbolDataFileName = "symbol_data.txt";
+
+    Contract.requires(
+        files.length !== 0,
+        "The list of files cannot be empty"
+    );   
+
+    // get a directory where we are going to store the result of indexing
+    const outDir = Context.getNewOutputDirectory("symbol_indexing");
```
mpysson: >getNewOutputDirectory [](start = 27, length = 21)

Cool. I guess this already does the right thing since it takes the stack frame of the requesting evaluated value into account for uniquness

Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
```diff
@@ -165,15 +168,55 @@ function addFiles(createResult: SymbolCreateResult, args: OperationArguments, fi
...
+        arguments: [
+            Cmd.argument("indexFiles"),
+            Cmd.options("--file ", files.map(f => Artifact.none(f))),
+            Cmd.options("--hash ", files.map(f => Artifact.vsoHash(f))),
+            Cmd.option("--symbolMetadata ", Artifact.output(outputPath)),
+        ],
+        unsafe: {
+            // SymStoreUtil opens some files (namely .pdb) with ReadWrite access,
+            // this causes DFAs and other issues.
+            untrackedPaths: files
```
mpysson: You're untracking the file you're indexing? That doesn't seem legit. Also, all should be correctly specified as inputs otherwise if this runs on a different machine in distribution they won't be materialized

Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
```diff
@@ -165,15 +168,55 @@ function addFiles(createResult: SymbolCreateResult, args: OperationArguments, fi
...
+        "The list of files cannot be empty"
+    );   
+
+    // get a directory where we are going to store the result of indexing
+    const outDir = Context.getNewOutputDirectory("symbol_indexing");
+
+    // the result of indexing
+    const outputPath = p`${outDir.path}/${symbolDataFileName}`;
+
+    const executeArgumetns = <Transformer.ExecuteArguments>{
```
mpysson: >tns [](start = 23, length = 3)

typo

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
```diff
@@ -218,11 +229,117 @@ private static Client CreateClient(string serverMoniker, IClientConfig config)
...
+            Contract.Assert(files.Length > 0);
+            Contract.Assert(!string.IsNullOrEmpty(outputFile));
+            Contract.Assert(!hashes.Any(hash => hash.HashType != HashType.Vso0), "Unsupported hash type");
+
+            var indexer = new SymbolIndexer(SymbolAppTraceSource.SingleInstance);
+            var symbolsMetadata = new Dictionary<ContentHash, HashSet<DebugEntryData>>();
+
+            for (int i = 0; i < files.Length; i++)
+            {
+                HashSet<DebugEntryData> symbols;
```
mpysson: >HashSet<DebugEntryData> symbols; [](start = 16, length = 32)

instantiate this outside of the loop and clear it within to save allocations

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
```diff
@@ -218,11 +229,117 @@ private static Client CreateClient(string serverMoniker, IClientConfig config)
...
+
+        private static void IndexFilesAndStoreMetadataToFile(string[] files, ContentHash[] hashes, string outputFile)
+        {
+            Contract.Assert(files.Length == hashes.Length);
+            Contract.Assert(files.Length > 0);
+            Contract.Assert(!string.IsNullOrEmpty(outputFile));
+            Contract.Assert(!hashes.Any(hash => hash.HashType != HashType.Vso0), "Unsupported hash type");
+
+            var indexer = new SymbolIndexer(SymbolAppTraceSource.SingleInstance);
+            var symbolsMetadata = new Dictionary<ContentHash, HashSet<DebugEntryData>>();
```
mpysson: >() [](start = 86, length = 2)

You can instantiate this with a set size to avoid some growth allocations

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
```diff
@@ -231,13 +348,38 @@ private static async Task<IIpcResult> AddSymbolFilesInternalAsync(ConfiguredComm
...
             }
 
+            if (string.IsNullOrEmpty(symbolMetadataFile))
+            {
+                return new IpcResult(
+                    IpcResultStatus.GenericError,
+                    "Invalid path to symbol metadata file.");
+            }
+
+            Dictionary < ContentHash, HashSet < DebugEntryData >> symbolMetadata;
```
mpysson: >Dictionary < ContentHash, HashSet < DebugEntryData > [](start = 12, length = 52)

nit - spaces

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
```diff
@@ -218,11 +229,117 @@ private static Client CreateClient(string serverMoniker, IClientConfig config)
...
+                writer.WriteLine(symbolsMetadata.Count);
+
+                foreach (var kvp in symbolsMetadata)
+                {
+                    writer.WriteLine(kvp.Key.Serialize());
+                    writer.WriteLine(kvp.Value.Count);
+
+                    foreach (var debugEntry in kvp.Value)
+                    {
+                        writer.WriteLine(JsonConvert.SerializeObject(debugEntry));
```
mpysson: >JsonConvert.SerializeObject(debugEntry) [](start = 41, length = 39)

I'm not sure how complicated the DebugEntryData object is but I'd expect it to be more performant to couple this serialization to its current implementation rather than using a generic json serializer.

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
```diff
@@ -218,11 +229,117 @@ private static Client CreateClient(string serverMoniker, IClientConfig config)
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Deserializes DebugEntryData's from a file.
+        /// </summary>
+        public static Dictionary<ContentHash, HashSet<DebugEntryData>> DeserializeSymbolsMetadata(string fileName)
+        {
+            var result = new Dictionary<ContentHash, HashSet<DebugEntryData>>();
```
mpysson: >(); [](start = 77, length = 3)

instantiate once you have hashCount below

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
```diff
@@ -218,11 +229,117 @@ private static Client CreateClient(string serverMoniker, IClientConfig config)
...
+        internal static readonly Command IndexFilesCmd = RegisterCommand(
+            name: "indexFiles",
+            description: "Indexes the specified files and saves SymbolData into a file.",
+            options: new Option[] { File, HashOptional, SymbolMetadataFile },
+            needsIpcClient: false,
+            clientAction: (ConfiguredCommand conf, IClient rpc) =>
+            {
+                var files = File.GetValues(conf.Config).ToArray();
+
+                // hashes are sent from BXL by serializing FileContentInfo's
```
norahuang: nit: FileContentInfo's  -> FileContentInfo?

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
```diff
@@ -218,11 +229,117 @@ private static Client CreateClient(string serverMoniker, IClientConfig config)
...
+            using (var writer = new StreamWriter(outputFile))
+            {                
+                writer.WriteLine(symbolsMetadata.Count);
+
+                foreach (var kvp in symbolsMetadata)
+                {
+                    writer.WriteLine(kvp.Key.Serialize());
+                    writer.WriteLine(kvp.Value.Count);
+
+                    foreach (var debugEntry in kvp.Value)
```
norahuang: Can kvp.Value be written as Json object? Instead of each entry be written as a Json object?

Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
```diff
@@ -144,13 +144,16 @@ function addFiles(createResult: SymbolCreateResult, args: OperationArguments, fi
...
     if (files.length === 0) {
         return undefined;
-    }               
+    }
+
+    const symbolMetadataFile = indexSymbolFiles(files);
 
     const fileMessageBody = files.length !== 0
         ? [
             Cmd.options("--file ", files.map(fi => Artifact.input(fi))),
```
mpysson: >Artifact.input [](start = 51, length = 14)

Not totally sure how this works, just calling out for you to double check: For a normal Process pip this will specify the file as an input to ProcessBuilder which will cause the file to be materialized when the process is a cache miss. IPC pips are always a "miss". So just double checking that it doesn't have the side effect of having the files unconditionally materialized

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
```diff
@@ -218,11 +229,117 @@ private static Client CreateClient(string serverMoniker, IClientConfig config)
...
+                int hashCount = int.Parse(reader.ReadLine());
+
+                for (int i = 0; i < hashCount; i++)
+                {
+                    ContentHash.TryParse(reader.ReadLine(), out var hash);
+
+                    var symbols = new HashSet<DebugEntryData>(DebugEntryDataComparer.Instance);
+                    result.Add(hash, symbols);
+
+                    int debugEntryCount = int.Parse(reader.ReadLine());
```
norahuang: If kvp.Value is serialized as an object, you can deserialized the written Json Object into an array and get the count and entries easily.

Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
olkononenko: these files are inputs into: 1) indexingPip and 2) ipcPip

For (1) - these file must be materialized if we are going to index them, i.e., no cache hit
For (2) - these files are listed as inputs so pip can take dependency on them. In addition to that, these files are listed in lazilyMaterializedDependencies block. Adding elements to that block does not add dependency, it just populates a list of elements that should not be materialized.
For (2) cont'd - symbolMetadataFile is not on lazily materialized because Daemo is reading that file to get the metadata.

---
In reply to: [327711924](https://github.com/microsoft/BuildXL/pull/921#discussion_r327711924) [](ancestors = 327711924)

Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
olkononenko: Yeah, I had sorting/hashing of inputs, and used that hash as an input into getNewOutputDirectory. However, Lance pointed out that this is an unnecessary step that can be avoided, and bxl will handle any collisions

---
In reply to: [327694554](https://github.com/microsoft/BuildXL/pull/921#discussion_r327694554) [](ancestors = 327694554)

Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
```diff
@@ -165,15 +168,55 @@ function addFiles(createResult: SymbolCreateResult, args: OperationArguments, fi
...
+
+    // the result of indexing
+    const outputPath = p`${outDir.path}/${symbolDataFileName}`;
+
+    const executeArgumetns = <Transformer.ExecuteArguments>{
+        tool: tool,
+        workingDirectory: outDir,
+        arguments: [
+            Cmd.argument("indexFiles"),
+            Cmd.options("--file ", files.map(f => Artifact.none(f))),
```
mpysson: >none [](start = 59, length = 4)

fix as per our conversation

Public/Src/Tools/SymbolDaemon/VsoSymbolClient.cs
```diff
@@ -128,7 +128,12 @@ public async Task<Request> CreateAsync(CancellationToken token)
         public async Task<AddDebugEntryResult> AddFileAsync(SymbolFile symbolFile)
         {
             Contract.Requires(symbolFile.IsIndexed, "File has not been indexed.");
-            Contract.Requires(symbolFile.DebugEntries.Count > 0, "File contains no symbol data.");
+            
+            if (symbolFile.DebugEntries.Count == 0)
+            {
+                // early return if there are no debug entries.
+                return AddDebugEntryResult.NoSymbolData;
+            }
```
semihokur: If there are no symbols, we will still have the success result. That's fine, i guess. What about logging a warning on the BuildXL side, similarly to what I've been doing for collision? 

You can use "logmessagecommand" for that.

Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
olkononenko: discussed offline

---
In reply to: [327696192](https://github.com/microsoft/BuildXL/pull/921#discussion_r327696192) [](ancestors = 327696192)

Public/Src/Tools/SymbolDaemon/SymbolDaemon.cs
olkononenko: removing JsonConvert

---
In reply to: [327711199](https://github.com/microsoft/BuildXL/pull/921#discussion_r327711199) [](ancestors = 327711199)

semihokur:
>                     DebugEntryCreateBehavior.ThrowIfExists,

(this was related to my changes regarding collision handling)

add a comment saying: 
"First, we create debug entries with ThrowIfExists behavior not to silence the collision errors. "

---
Refers to: Public/Src/Tools/SymbolDaemon/VsoSymbolClient.cs:147 in 68ff987. [](commit_id = 68ff987492f05a10a462c14426b8ff299821be09, deletion_comment = False)

PR #923, opened 2019-09-24
Force restore from last checkpoint when content location database is invalidated

jbayardo:
When an event (such as database corruption) triggers a database invalidation, we currently have to wait until the next checkpoint is made available and restored. This PR changes the mechanic so that if this happens, a restore is triggered.

Notice that even though the restore is triggered, it will only happen if there is no other ongoing restore.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
@@ -88,6 +88,41 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 
         private readonly object _cacheFlushTimerLock = new object();
 
+        /// <summary>
+        /// Event callback that's triggered when the database is permanently invalidated. 
+        /// </summary>
+        public event EventHandler<DatabaseInvalidatedEventArgs> DatabaseInvalidated;
+
+        /// <nodoc />
+        public class DatabaseInvalidatedEventArgs : EventArgs
+        {
+            /// <nodoc />
+            public OperationContext Context { get; }
+
+            /// <nodoc />
+            public Failure<Exception> Failure { get; }
+
+            /// <nodoc />
+            public DatabaseInvalidatedEventArgs(OperationContext context, Failure<Exception> failure)
+            {
+                Contract.Requires(failure != null);
+                Context = context;
+                Failure = failure;
+            }
+        }   
+
+        /// <nodoc />
+        protected void OnDatabaseInvalidated(OperationContext context, Failure<Exception> failure)
+        {
+            Contract.Requires(failure != null);
+
+            // Notice that no update to the internal state is required when invalidation happens. By definition,
+            // nothing can be done to this instance after invalidation: all incoming and ongoing operations should fail
+            // (because it is triggered by RocksDb). The only way to resume operation is to reload from a checkpoint,
+            // which resets the internal state correctly.
+            DatabaseInvalidated?.Invoke(this, new DatabaseInvalidatedEventArgs(context, failure));
+        }
+
         /// <nodoc />
         protected ContentLocationDatabase(IClock clock, ContentLocationDatabaseConfiguration configuration, Func<IReadOnlyList<MachineId>> getInactiveMachines)
         {
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
@@ -91,25 +91,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// <summary>
         /// Event callback that's triggered when the database is permanently invalidated. 
         /// </summary>
-        public event EventHandler<DatabaseInvalidatedEventArgs> DatabaseInvalidated;
-
-        /// <nodoc />
-        public class DatabaseInvalidatedEventArgs : EventArgs
-        {
-            /// <nodoc />
-            public OperationContext Context { get; }
-
-            /// <nodoc />
-            public Failure<Exception> Failure { get; }
-
-            /// <nodoc />
-            public DatabaseInvalidatedEventArgs(OperationContext context, Failure<Exception> failure)
-            {
-                Contract.Requires(failure != null);
-                Context = context;
-                Failure = failure;
-            }
-        }   
+        public Action<OperationContext, Failure<Exception>> DatabaseInvalidated;
 
         /// <nodoc />
         protected void OnDatabaseInvalidated(OperationContext context, Failure<Exception> failure)
@@ -120,7 +102,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             // nothing can be done to this instance after invalidation: all incoming and ongoing operations should fail
             // (because it is triggered by RocksDb). The only way to resume operation is to reload from a checkpoint,
             // which resets the internal state correctly.
-            DatabaseInvalidated?.Invoke(this, new DatabaseInvalidatedEventArgs(context, failure));
+            DatabaseInvalidated?.Invoke(context, failure);
         }
 
         /// <nodoc />
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
@@ -32,6 +32,7 @@ using DateTimeUtilities = BuildXL.Cache.ContentStore.Utils.DateTimeUtilities;
 using static BuildXL.Cache.ContentStore.Utils.DateTimeUtilities;
 using BuildXL.Utilities.Tasks;
 using BuildXL.Utilities;
+using BuildXL.Cache.ContentStore.Interfaces.Synchronization.Internal;
 
 namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 {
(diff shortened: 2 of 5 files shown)


Public/Src/Utilities/Utilities/Tasks/TaskUtilities.cs
```diff
@@ -208,7 +208,8 @@ public static async Task<SemaphoreReleaser> AcquireAsync(this SemaphoreSlim sema
         /// Synchronously acquire a semaphore
         /// </summary>
         /// <param name="semaphore">The semaphore to acquire</param>
-        public static SemaphoreReleaser AcquireSemaphore(this SemaphoreSlim semaphore)
+        /// <param name="millisecondsTimeout">Time to wait to acquire the semaphore</param>
+        public static SemaphoreReleaser AcquireSemaphore(this SemaphoreSlim semaphore, int millisecondsTimeout = Timeout.Infinite)
```
SergeyTeplyakov: It seems that the timeout is not used.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1257,6 +1262,20 @@ public Task<Result<byte[]>> GetBlobAsync(OperationContext context, ContentHash h
             return GlobalStore.GetBlobAsync(context, hash);
         }
 
+        private async void OnContentLocationDatabaseInvalidationAsync(object sender, ContentLocationDatabase.DatabaseInvalidatedEventArgs e)
+        {
+            // If multiple threads fail at the same time (i.e. corruption error), this cheaply deduplicates the
+            // restores, avoids redundant logging.
+            using (_databaseInvalidationGate.AcquireSemaphore(0))
```
SergeyTeplyakov: I'm curious, have you seen more then one logging at the same time? Because ProcessStateAsync will take care of concurrency any ways... And I Can't see how you actually reducing the amount of logging by "serializing" the body of this using block.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1257,6 +1262,20 @@ public Task<Result<byte[]>> GetBlobAsync(OperationContext context, ContentHash h
             return GlobalStore.GetBlobAsync(context, hash);
         }
 
+        private async void OnContentLocationDatabaseInvalidationAsync(object sender, ContentLocationDatabase.DatabaseInvalidatedEventArgs e)
```
SergeyTeplyakov: Are you sure that you want to kill the process if this method will fail (due to bugs, for instance?)

One option to avoid this is to have a void (non-async) handler that will call async operation and will call fire and forget on it.

SergeyTeplyakov: I mean, bugs are possible. Like, null refs, so I don't want to kill the app because of relatively innocent bug in the code.

---
In reply to: [327867211](https://github.com/microsoft/BuildXL/pull/923#discussion_r327867211) [](ancestors = 327867211)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -88,6 +88,41 @@ public abstract class ContentLocationDatabase : StartupShutdownSlimBase
 
         private readonly object _cacheFlushTimerLock = new object();
 
+        /// <summary>
+        /// Event callback that's triggered when the database is permanently invalidated. 
+        /// </summary>
+        public event EventHandler<DatabaseInvalidatedEventArgs> DatabaseInvalidated;
```
ljcollins25: >EventHandler [](start = 21, length = 12)

Just use Action. No need for EventHandler since sender is never used.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
ljcollins25: >AcquireSemaphore [](start = 45, length = 16)

Maybe Monitor.TryEnter instead of SemaphoreSlim here. Do the other threads need to block or is it sufficient to just prevent reentrancy?

SergeyTeplyakov: But do we actually need it at all? ProcessStateAsync prevents re-entrancy.

---
In reply to: [327874599](https://github.com/microsoft/BuildXL/pull/923#discussion_r327874599) [](ancestors = 327874599)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1257,6 +1262,20 @@ public Task<Result<byte[]>> GetBlobAsync(OperationContext context, ContentHash h
...
+        {
+            // If multiple threads fail at the same time (i.e. corruption error), this cheaply deduplicates the
+            // restores, avoids redundant logging.
+            using (_databaseInvalidationGate.AcquireSemaphore(0))
+            {
+                Tracer.Error(e.Context, $"Content location database has been invalidated. Forcing a restore from the last checkpoint. Error: {e.Failure.DescribeIncludingInnerFailures()}");
+
+                // There is nothing we can do from here if the checkpoint fails to restore. We will just wait until the
+                // next heartbeat.
+                await ProcessStateAsync(e.Context, inline: false, forceRestore: true).IgnoreFailure();
```
ljcollins25: >ProcessStateAsync [](start = 22, length = 17)

Somewhere we need to ensure that this isn't run concurrently with Heartbeat.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
ljcollins25: Looking at RunOutOfBandAsync it doesn't seem to be threadsafe.

---
In reply to: [327874969](https://github.com/microsoft/BuildXL/pull/923#discussion_r327874969) [](ancestors = 327874969,327874599)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
SergeyTeplyakov: Ok. I confused ProcessStateAsync with RestoreCheckpointAsync. RestoreCheckpointAsync makes sure that only one thread runs the operation, but there is no protection inside ProcessStateAsync.

Maybe we should just restart the timer here?

---
In reply to: [327875373](https://github.com/microsoft/BuildXL/pull/923#discussion_r327875373) [](ancestors = 327875373)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
jbayardo: This prevents re-entrancy by forcing the other threads to do a non-blocking return. I forgot to add the 0 in the Wait inside, but that is the idea at least

---
In reply to: [327876278](https://github.com/microsoft/BuildXL/pull/923#discussion_r327876278) [](ancestors = 327876278,327874969,327874599)

jbayardo: Yes, I have seen this. An example situation is as follows:

- Imagine K threads running in parallel with .Use()
- DB gets corrupted, threads perform some DB operation after the corruption, cause a throw
- All threads simultaneously disable the store (there is no semaphore in that code)
- Hence, the invalidation handler is called K times, and possibly spaced out

This semaphore is there to ensure that we only log an error once per machine. Let me know if you want a query that logs this example.

---
In reply to: [327866401](https://github.com/microsoft/BuildXL/pull/923#discussion_r327866401) [](ancestors = 327866401)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
jbayardo: End of the day we need to make sure that there is only one caller doing the Heartbeat, I have added a non-blocking semaphore in the heartbeat so as to avoid the undesired situations, and call directly the HeartbeatAsync function

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -358,11 +366,27 @@ protected override async Task<BoolResult> ShutdownCoreAsync(OperationContext con
...
-        internal Task<BoolResult> HeartbeatAsync(OperationContext context)
+        internal async Task<BoolResult> HeartbeatAsync(OperationContext context)
         {
-            return context.PerformOperationAsync(
-                Tracer,
-                () => ProcessStateAsync(context, inline: false));
+            // This makes sure that the heartbeat is only run once for each call to this function. It is non-blocking
+            // wait, so no need to use the async version.
+#pragma warning disable AsyncFixer02
+            if (!_heartbeatGate.Wait(0))
```
SergeyTeplyakov: One way to remove these pragmas in two places and simplify the code a bit is to change WaitToken extension method from SemaphoreSlimExtension to take the wait time or to add an extension method called TryWait() that will return SemaphoreSlimToken (a disposable instance).

In this case you'll remove two pragmas and will be able to use this:

bool acquired
using _gate.TryWait(out var acquierd);
if (!acquired)
{
  return BoolResult.Success;
}

return await context.PerformOperationAsync(

What do you think?

Public/Src/Cache/ContentStore/Interfaces/Synchronization/SemaphoreSlimToken.cs
```diff
@@ -42,6 +42,20 @@ public static SemaphoreSlimToken Wait(SemaphoreSlim semaphore)
...
 
+        /// <summary>
+        ///     Wait on a SemaphoreSlim and return a token that, when disposed, calls Release() on the SemaphoreSlim.
+        ///     If the deadline is reached, the token is not acquired and a fake disposable is returned.
+        /// </summary>
+        /// <param name="semaphore">The semaphore to wait on</param>
+        /// <param name="millisecondsTimeout">Maximum time to wait for the semaphore, in milliseconds</param>
+        /// <param name="acquired">Whether the semaphore was obtained or not</param>
+        /// <returns>A token that, when disposed, calls Release() on the SemaphoreSlim</returns>
+        public static SemaphoreSlimToken TryWait(SemaphoreSlim semaphore, int millisecondsTimeout, out bool acquired)
```
SergeyTeplyakov: 2 very nit comments (feel free to ignore):

1. Move this to SemaphoreSlimExtensions to simplify usage
2. Add an overload without timeout (will pass 0 by default)
3. use new 'using var' in ConentLocationDatabase (where this extension is used) to remove extra block:

using var _gate.TryWait(out var acquired);

PR #925, opened 2019-09-24
Place without first calling pin.

dcwuser:
Current logic calls pin for any hash before placing it. This causes more cache activity, more cache logging, and makes it hard to distinguish okay-to-fail pins testing for content from error-to-fail pins that Domino regards as already guaranteed. 

It was not trivial to remove pin before place, because the structure of the code that does it is linked to the code that attempts to recover files not found in the cache.

I also added counters to find out how often file recovery is attempted and how often it is successful.

### DIFF
--- a/Public/Src/Utilities/Configuration/Mutable/ScheduleConfiguration.cs
+++ b/Public/Src/Utilities/Configuration/Mutable/ScheduleConfiguration.cs
@@ -72,6 +72,7 @@ namespace BuildXL.Utilities.Configuration.Mutable
 
             UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing = false;
             InputChanges = AbsolutePath.Invalid;
+            // PinBeforePlace = false;
         }
 
         /// <nodoc />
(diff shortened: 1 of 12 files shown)


Public/Src/Utilities/Configuration/Mutable/ScheduleConfiguration.cs
```diff
@@ -72,6 +72,7 @@ public ScheduleConfiguration()
 
             UnsafeDisableSharedOpaqueEmptyDirectoryScrubbing = false;
             InputChanges = AbsolutePath.Invalid;
+            // PinBeforePlace = false;
```
SergeyTeplyakov: Remove or uncomment?

SergeyTeplyakov: Added a comment on the first iteration.

---
In reply to: [327867667](https://github.com/microsoft/BuildXL/pull/925#discussion_r327867667) [](ancestors = 327867667)

PR #926, opened 2019-09-24
Remove some unnecessary DominoMessage cache logs

JuanCarlosGI:
Removed messages:

BuildCacheCache.GetContentnHashList start
ServiceClientCacheSession starts


Public/Src/Cache/MemoizationStore/Library/Service/ServiceClientCacheSession.cs
```diff
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
ljcollins25: >using System.Diagnostics [](start = 0, length = 24)

Is this still needed?

PR #929, opened 2019-09-24
Introduce /abTesting feature for perf testing

semihokur:
Example usage: 
If we would like to evaluate the perf characteristics of different IO concurrency limits, we can pass three different /abTesting arguments: 

bxl.exe ... /abTesting:lowIO="/maxMaterialize:2 /maxIO:1" /abTesting:mediumIO="/maxMaterialize:10 /maxIO:5" /abTesting:highIO="/maxMaterialize:20 /maxIO:10"

Then, BuildXL will randomly choose one of the given options, apply those arguments to the current build, and add the key (e.g., mediumIO) and hash of the arguments (e.g., "/maxMaterialize:20 /maxIO:10".GetHashCode()) to the TraceInfo so that we can query the perf for lowIO, mediumIO, and highIO builds. 

If RelatedActivityId is passed to the build, master and workers will choose the same set of arguments as we use the RelatedActivityId as a random seed. It means that we will use the same set of abTesting arguments for Enlist, Meta, and Product builds in master and workers.

### DIFF
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -171,11 +171,6 @@ namespace BuildXL
         [SuppressMessage("Microsoft.Performance", "CA1809")]
         public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfiguration arguments)
         {
-            foreach (var a in args)
-            {
-                Console.WriteLine("***    " + a);
-            }
-
             try
             {
                 var cl = new CommandLineUtilities(args);
@@ -216,7 +211,7 @@ namespace BuildXL
                     {
                         OptionHandlerFactory.CreateOption(
                             "abTesting",
-                            opt => CommandLineUtilities.ParsePropertyOption(opt, loggingConfiguration.ABTestingArgs)),
+                            opt => CommandLineUtilities.ParsePropertyOption(opt, startupConfiguration.ABTestingArgs)),
                         OptionHandlerFactory.CreateOption2(
                             "additionalConfigFile",
                             "ac",
@@ -1215,7 +1210,7 @@ namespace BuildXL
                 // Iterate through each argument, looking each argument up in the table.
                 IterateArgs(cl, configuration, specialCaseUnsafeOptions);
 
-                int numABTestingOptions = loggingConfiguration.ABTestingArgs.Count;
+                int numABTestingOptions = startupConfiguration.ABTestingArgs.Count;
                 if (numABTestingOptions > 0)
                 {
                     // If RelatedActivityId is populated, use it as a seed for random number generation
@@ -1225,8 +1220,8 @@ namespace BuildXL
                         new Random(loggingConfiguration.RelatedActivityId.GetHashCode());
 
                     int randomNum = randomGen.Next(numABTestingOptions);
-                    var randomOption = loggingConfiguration.ABTestingArgs.ToList()[randomNum];
-                    string abTestingId = randomOption.Key;
+                    var randomOption = startupConfiguration.ABTestingArgs.ToList()[randomNum];
+                    string abTestingKey = randomOption.Key;
                     string abTestingArgs = randomOption.Value;
 
                     // Add key and the hash code of the arguments to traceInfo for telemetry purposes. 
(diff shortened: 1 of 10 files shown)


Public/Src/App/Bxl/Args.cs
```diff
@@ -171,6 +171,11 @@ public static bool TryParseArguments(string[] args, PathTable pathTable, IConsol
         [SuppressMessage("Microsoft.Performance", "CA1809")]
         public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfiguration arguments)
         {
+            foreach (var a in args)
```
almiliMSFT: remove

Public/Src/App/UnitTests/Bxl/ArgsTests.cs
```diff
@@ -146,5 +147,22 @@ public void DisplayHelpText()
...
+        [Fact]
+        public void ABTestingOption()
+        {
+            ICommandLineConfiguration config;
+            PathTable pt = new PathTable();
+            var argsParser = new Args();
+
+            string abTestingArg = "/incrementalScheduling+ /maxIO:1";
+
+            System.Diagnostics.Debugger.Launch();
```
almiliMSFT: remove

Public/Src/App/Bxl/Args.cs
```diff
@@ -1205,47 +1213,32 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
...
+
+                    // Add key and the hash code of the arguments to traceInfo for telemetry purposes. 
+                    // As the ID for the AB testing arguments is specified by the user, there is a chance to
+                    // give the same ID to the different set of arguments. That's why, we also add the hash code of 
+                    // the arguments to traceInfo.
+                    loggingConfiguration.TraceInfo.Add(
+                        TraceInfoExtensions.ABTesting, 
+                        $"{abTestingId};{abTestingArgs.GetHashCode().ToString()}");
+
+                    var cl2 = new CommandLineUtilities(abTestingArgs.Split(' '));
```
almiliMSFT: Is `abTestingArgs.Split(' ')` always going to be correct?  I don't think it will, since each BuildXL arg inside of `abTestingArgs` may contain spaces (e.g., `/abTesting:AB1="/f:spec='*/a b/spec.dsc'"` or something)

almiliMSFT: This should do the right thing:
```csharp
BuildXL.Utilities.CLI.WinParser().SplitArgs(abTestingArgs);
```

Public/Src/App/UnitTests/Bxl/ArgsTests.cs
```diff
@@ -146,5 +147,22 @@ public void DisplayHelpText()
             HelpText.DisplayHelp(global::BuildXL.ToolSupport.HelpLevel.Standard);
             HelpText.DisplayHelp(global::BuildXL.ToolSupport.HelpLevel.Verbose);
         }
+
+        [Fact]
+        public void ABTestingOption()
```
mpysson: >ABTestingOption [](start = 20, length = 15)

Also add a test for the RelatedActivityId seed magic

Public/Src/App/Bxl/Args.cs
```diff
@@ -1205,47 +1213,32 @@ public bool TryParse(string[] args, PathTable pathTable, out ICommandLineConfigu
...
+                    int randomNum = randomGen.Next(numABTestingOptions);
+                    var randomOption = loggingConfiguration.ABTestingArgs.ToList()[randomNum];
+                    string abTestingId = randomOption.Key;
+                    string abTestingArgs = randomOption.Value;
+
+                    // Add key and the hash code of the arguments to traceInfo for telemetry purposes. 
+                    // As the ID for the AB testing arguments is specified by the user, there is a chance to
+                    // give the same ID to the different set of arguments. That's why, we also add the hash code of 
+                    // the arguments to traceInfo.
+                    loggingConfiguration.TraceInfo.Add(
```
mpysson: >TraceInfo [](start = 41, length = 9)

This will only make it into telemetry so it may not be obvious what options a build actually used based on the text log. I suggest you add something to the text log to make it self-contained.

Public/Src/App/Bxl/Args.cs
semihokur: thanks, replaced with WinParser.

Public/Src/App/Bxl/Args.cs
semihokur: good point, added a verbose event.

PR #934, opened 2019-09-25
Refactor BandwidthChecker and add more config options

JuanCarlosGI:
Refactored so config was exclusive to BandiwdthChecker, instead of also involving GrpcCopier. 

Bandiwidth check, instead of being implemented by each copier, can now be done via a BandwidthCheckedCopier, which can wrap an inner copier.

Added more configurations: MaxBandwidthLimit, BandwidthLimitMultiplier and HistoricalBandwidthRecordsStored.

Additionally, stop catching unnecessary exceptions, and always time out copies if an interval detects that current bandwidth is 0.

Finally, make sure to register bandwidths for failed copies in historical bandwidth check, to better reflect the current status of the system.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Utilities/BandwidthCheckedCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Utilities/BandwidthCheckedCopier.cs
@@ -1,0 +1,55 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.ContractsLight;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
+using BuildXL.Cache.ContentStore.Interfaces.Logging;
+using BuildXL.Cache.ContentStore.Interfaces.Results;
+using BuildXL.Cache.ContentStore.Interfaces.Tracing;
+using BuildXL.Cache.ContentStore.Tracing.Internal;
+using BuildXL.Cache.ContentStore.Utils;
+
+namespace BuildXL.Cache.ContentStore.Distributed.Utilities
+{
+    /// <summary>
+    /// File copier that times out copies based on their bandwidths.
+    /// </summary>
+    public class BandwidthCheckedCopier<T> : IFileCopier<T> where T : PathBase
+    {
+        private readonly IFileCopier<T> _inner;
+        private readonly BandwidthChecker _checker;
+        private readonly ILogger _logger;
+
+        /// <nodoc />
+        public BandwidthCheckedCopier(IFileCopier<T> inner, BandwidthChecker.Configuration config, ILogger logger)
+        {
+            _inner = inner;
+            _checker = new BandwidthChecker(config);
+            _logger = logger;
+        }
+
+        /// <inheritdoc />
+        public async Task<CopyFileResult> CopyToAsync(T sourcePath, Stream destinationStream, long expectedContentSize, CancellationToken cancellationToken)
+        {
+            try
+            {
+                var context = new OperationContext(new Context(_logger), cancellationToken); 
+                CopyFileResult result = null;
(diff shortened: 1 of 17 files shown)


Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -60,7 +71,8 @@ async Task impl()
             {
                 // This method should not fail with exceptions because the resulting task may be left unobserved causing an application to crash
                 // (given that the app is configured to fail on unobserved task exceptions).
-                var minimumSpeedInMbPerSec = _bandwidthLimitSource.GetMinimumSpeedInMbPerSec();
+                var minimumSpeedInMbPerSec = _bandwidthLimitSource.GetMinimumSpeedInMbPerSec() * _config.BandwidthLimitMultiplier;
+                minimumSpeedInMbPerSec = Math.Min(minimumSpeedInMbPerSec, _config.MaxBandwidthLimit);
```
SergeyTeplyakov: I'm not sure what is the reason for having MaxBandwidthLimit. Does it make sense to set it in any possible scenarios?

JuanCarlosGI: It is so that we don't get an over-optimistic estimate of what the bandwidth should be. If for some reason we have a bunch of copies that are faster than normal, we could end up in a state where the expectation would be unrealistic.

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -128,5 +132,60 @@ async Task impl()
                 }
             }
         }
+
+        /// <nodoc />
+        public struct Configuration
```
ljcollins25: >struct [](start = 15, length = 6)

Can we make this a class? That would make Default value a little clearer since default(Configuration) currently does not equal Configuration.Default but is still a "valid" instance.

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -94,8 +106,8 @@ async Task impl()
                             var position = destinationStream.Position;
 
                             var receivedMiB = (position - previousPosition) / BytesInMb;
-                            var currentSpeed = receivedMiB / _checkInterval.TotalSeconds;
-                            if (currentSpeed < minimumSpeedInMbPerSec)
+                            var currentSpeed = receivedMiB / _config.BandwidthCheckInterval.TotalSeconds;
```
SergeyTeplyakov: I hate copy pasting stuff, but I've make another change in this logic here: [https://dev.azure.com/mseng/Domino/_git/CloudBuild/pullrequest/507021?_a=files](https://dev.azure.com/mseng/Domino/_git/CloudBuild/pullrequest/507021?_a=files)

I think it make sense to adopt it.

And I would suggest to use the structure from the CB with                 a few helpers like: TraceCopyTaskFailures, TryGetPosition and CheckSufficientBandwidth.

Public/Src/Cache/ContentStore/Distributed/Utilities/BandwidthCheckedCopier.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <inheritdoc />
+        public async Task<CopyFileResult> CopyToAsync(T sourcePath, Stream destinationStream, long expectedContentSize, CancellationToken cancellationToken)
+        {
+            try
+            {
+                var context = new OperationContext(new Context(_logger), cancellationToken); 
+                CopyFileResult result = null;
+                await _checker.CheckBandwidthAtIntervalAsync(context, async token =>
+                {
+                    result = await _inner.CopyToAsync(sourcePath, destinationStream, expectedContentSize, token);
```
SergeyTeplyakov: Does it make sense to change CheckBandwidthAtIntervalAsync method signature and instead of providing Func<Task> to accept Func<Task<CopyFileResult>> ?

PR #936, opened 2019-09-25
[WIP] Properly build VSCode DScript extension with Npm CG

rijulluman-msft:
[AB#1478432](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1478432)


Public/Sdk/Experimental/NodeJs/npm.dsc
```diff
@@ -41,6 +41,39 @@ namespace Npm {
...
+        const arguments: Argument[] = [
+            Cmd.argument(Artifact.input(Node.npmCli)),
+            Cmd.argument("install")
+        ];
+
+        const result = Node.run({
+            arguments: arguments,
+            workingDirectory: workingDirectory,
+            outputs: [
+                nodeModulesPath
```
almiliMSFT: Specifying this folder here doesn't tell `Node` where to place outputs; instead this only declares where outputs are expected to be.  That means that this pip will simply run `npm install` from the working directory (which in your case will be the `client` folder); when this pip is executed, it will try to produce `client/node_modules` and not the folder specified in `nodeModulesPath`.

Public/Sdk/Experimental/NodeJs/npm.dsc
```diff
@@ -41,6 +41,39 @@ namespace Npm {
...
+            Cmd.argument(Artifact.input(Node.npmCli)),
+            Cmd.argument("run"),
+            Cmd.argument("compile")
+        ];
+
+        const result = Node.run({
+            arguments: arguments,
+            workingDirectory: folder,
+            outputs: [
+                outPath
```
almiliMSFT: same problem here

Public/Src/IDE/VsCode/BuildXL.IDE.VsCode.dsc
```diff
@@ -96,11 +101,9 @@ namespace LanguageService.Server {
                         Branding.pngFile,
                         json,
 
-                        // This contains the actual extension source as well as the
-                        // node_modules that it depends on.
                         Transformer.sealDirectory({
-                            root: d`pluginTemplate/extension`, 
-                            files: globR(d`pluginTemplate/extension`)
+                            root: extensionSourceFolder, 
+                            files: globR(extensionSourceFolder)
```
almiliMSFT: this `globR` executes at dscript evaluation time, so it will always glob and empty folder and return 0 files

Public/Src/IDE/VsCode/BuildXL.IDE.VsCode.dsc
```diff
@@ -55,6 +56,10 @@ namespace LanguageService.Server {
         let json = IDE.VersionUtilities.updateVersion(version, f`client/package.json`);
         let readme = IDE.VersionUtilities.updateVersion(Branding.version, f`client/README.md`);
 
+        const extensionSourceFolder = Context.getNewOutputDirectory(`VsixTemp`);
+        Npm.installFromPackageJson(d`client`, d`${extensionSourceFolder}/node_modules`);
```
almiliMSFT: To establish a dependency between pips, these functions should return some output artifacts and then you should use those artifacts down below to specify vsix deployment.  The way it is now there is no dependency between the deployment pip and these `npm install`/`npm run compile` pips; consequently, when you run a build these `npm` pips are not even executed

PR #945, opened 2019-09-26
Improve minimum eviction times

jbayardo:
This changes the pooling logic to fetch many times the size of the pool, and evict only a fraction of it that depends on the total content being stored.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/NuCacheCollectionUtilities.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/NuCacheCollectionUtilities.cs
@@ -237,29 +237,41 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         /// process, adding them to the same priority queue. This process repeats until the original enumerable has no
         /// more elements.
         /// </summary>
-        public static IEnumerable<T> QueryAndOrderInPages<T>(this IEnumerable<T> original, int pageSize, Comparer<T> comparer, Func<List<T>, IEnumerable<T>> query)
+        public static IEnumerable<T> QueryAndOrderInPages<T>(this IEnumerable<T> original, int pageSize, Comparer<T> comparer, Func<List<T>, IEnumerable<T>> query, float? takeoutFraction = null, int poolMultiplier = 2)
         {
+            Contract.Assert(pageSize > 0);
+            Contract.Assert(takeoutFraction == null || (takeoutFraction > 0 && takeoutFraction <= 1));
+            Contract.Assert(poolMultiplier > 0);
+
+            var poolSize = Math.Max(pageSize * poolMultiplier, pageSize);
+
+            // We either take the fraction given, or a single page at a time
+            var removalFraction = takeoutFraction ?? (1 / poolMultiplier);
+
             var source = original.GetEnumerator();
-            var queue = new PriorityQueue<T>(pageSize * 2, comparer);
+            var pool = new PriorityQueue<T>(poolSize, comparer);
             var sourceHasItems = true;
 
             while (true)
             {
-                if (sourceHasItems && queue.Count < pageSize)
+                if (sourceHasItems && pool.Count < poolSize)
                 {
-                    var itemsToQuery = new List<T>(pageSize);
-                    for (var i = 0; i < pageSize && source.MoveNext(); i++)
+                    // In this branch, we fill the queue up to maximum capacity by querying in batches of size at most
+                    // `pageSize`. Notice that this may be run up to `poolMultiplier` times in a row before producing
+                    // any results.
+                    var batchSize = Math.Min(poolSize - pool.Count, pageSize);
+
+                    var batch = new List<T>(batchSize);
+                    while (batch.Count < batch.Capacity && source.MoveNext())
                     {
-                        itemsToQuery.Add(source.Current);
+                        batch.Add(source.Current);
                     }
 
-                    if (itemsToQuery.Count > 0)
+                    if (batch.Count > 0)
                     {
-                        var results = query(itemsToQuery);
-
-                        foreach (var result in results)
+                        foreach (var candidate in query(batch))
                         {
-                            queue.Push(result);
+                            pool.Push(candidate);
                         }
                     }
                     else
@@ -268,14 +280,37 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                         sourceHasItems = false;
                     }
                 }
-
-                if (queue.Count == 0)
+                else if (pool.Count == 0)
                 {
+                    Contract.Assert(!sourceHasItems);
                     yield break;
                 }
+                else
+                {
+                    Contract.Assert(pool.Count > 0);
 
-                yield return queue.Top;
-                queue.Pop();
+                    int minimumYieldSize;
+                    if (!sourceHasItems)
+                    {
+                        // If the enumerator has no elements left, we know that the queue has the best order possible,
+                        // so we yield everything.
+                        minimumYieldSize = pool.Count;
+                    }
+                    else
+                    {
+                        minimumYieldSize = (int)Math.Floor(removalFraction * pool.Count);
+                        // To guarantee termination, we always yield at least one element
+                        minimumYieldSize = Math.Max(minimumYieldSize, 1);
+                        // Never yield more than one page of results. This is about maintaining baseline accuracy for very large fractions.
+                        minimumYieldSize = Math.Min(minimumYieldSize, pageSize);
+                    }
+
+                    Contract.Assert(minimumYieldSize <= pool.Count);
+                    for (var i = 0; i < minimumYieldSize; ++i) {
+                        yield return pool.Top;
+                        pool.Pop();
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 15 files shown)


Public/Src/Cache/ContentStore/Hashing/ContentHashWithLastAccessTime.cs
```diff
@@ -13,14 +13,14 @@ namespace BuildXL.Cache.ContentStore.Hashing
...
-            Hash = contentHash;
+            ContentHash = contentHash;
             LastAccessTime = lastAccessTime;
         }
 
         /// <summary>
         ///     Gets the content hash member.
         /// </summary>
-        public ContentHash Hash { get; }
+        public ContentHash ContentHash { get; }
```
SergeyTeplyakov: This may be a breaking change.

Is there any specific reason for this rename?

Public/Src/Cache/ContentStore/Distributed/NuCache/NuCacheCollectionUtilities.cs
```diff
@@ -237,29 +237,41 @@ public static IEnumerable<T> SkipOptimized<T>(this IReadOnlyList<T> original, in
         /// process, adding them to the same priority queue. This process repeats until the original enumerable has no
         /// more elements.
         /// </summary>
-        public static IEnumerable<T> QueryAndOrderInPages<T>(this IEnumerable<T> original, int pageSize, Comparer<T> comparer, Func<List<T>, IEnumerable<T>> query)
+        public static IEnumerable<T> QueryAndOrderInPages<T>(this IEnumerable<T> original, int pageSize, Comparer<T> comparer, Func<List<T>, IEnumerable<T>> query, float? takeoutFraction = null, int poolMultiplier = 2)
         {
+            Contract.Assert(pageSize > 0);
```
SergeyTeplyakov: Use requires here.

Public/Src/Cache/ContentStore/Hashing/ContentHashWithLastAccessTime.cs
jbayardo: Seemed like an unnecessary inconsistency wrt to the other ContentHashWith*. Didn't know it could be breaking. Just rolled it back

Public/Src/Cache/ContentStore/Distributed/NuCache/NuCacheCollectionUtilities.cs
```diff
@@ -237,29 +237,41 @@ public static IEnumerable<T> SkipOptimized<T>(this IReadOnlyList<T> original, in
         /// process, adding them to the same priority queue. This process repeats until the original enumerable has no
         /// more elements.
         /// </summary>
-        public static IEnumerable<T> QueryAndOrderInPages<T>(this IEnumerable<T> original, int pageSize, Comparer<T> comparer, Func<List<T>, IEnumerable<T>> query)
+        public static IEnumerable<T> QueryAndOrderInPages<T>(this IEnumerable<T> original, int pageSize, Comparer<T> comparer, Func<List<T>, IEnumerable<T>> query, float? takeoutFraction = null, int poolMultiplier = 2)
```
SergeyTeplyakov: I'm curious: is it possible to actually write the test for this that will detect the change in behavior?

Public/Src/Cache/ContentStore/Distributed/NuCache/NuCacheCollectionUtilities.cs
```diff
@@ -237,29 +237,41 @@ public static IEnumerable<T> SkipOptimized<T>(this IReadOnlyList<T> original, in
...
+
             var source = original.GetEnumerator();
-            var queue = new PriorityQueue<T>(pageSize * 2, comparer);
+            var pool = new PriorityQueue<T>(poolSize, comparer);
             var sourceHasItems = true;
 
             while (true)
             {
-                if (sourceHasItems && queue.Count < pageSize)
+                if (sourceHasItems && pool.Count < poolSize)
```
dcwuser: I am a confused about this if/else structure. As currently written, seems like
  while (true) {
    if (pool not full) {
      get another page
    } else {
      deliver results
    }
  }
Shouldn't it be
  while (true) {
    if (pool not full) {
      get another page
    }
    deliver results
  }
I don't want to fill the pool before we start delivering results.

Public/Src/Cache/ContentStore/Distributed/NuCache/NuCacheCollectionUtilities.cs
dcwuser: I expect pool multiplier to be more like 8-16. It's great that it's configurable, but can our defaults be more in line with expectations? Or is that a bad default?

Public/Src/Cache/ContentStore/Distributed/NuCache/TransitioningContentLocationStore.cs
```diff
@@ -271,21 +277,34 @@ public Task<BoolResult> UnregisterAsync(Context context, IReadOnlyList<ContentHa
...
+            // information, so we instead have an evictability metric. Here we obtain and sort by that evictability
+            // metric.
             var pageSize = _configuration.EvictionWindowSize;
+            var poolMultiplier = _configuration.MaximumEvictionPoolMultiplier;
+
+            float? takeoutFraction = null;
+            if (contentHashesWithInfo.Count > 0)
+            {
+                // We want to remove roughly the size of the pool divided by the amount of content at each iteration
+                takeoutFraction = 1 / contentHashesWithInfo.Count;
```
dcwuser: This looks like integer division, in which case takeoutFraction = 0 always. To get around this problem, use 1.0 instead of 1.

dcwuser: Also, I don't think this is what we want takeout fraction to be, even without the integer division issue. Every time we pull in a page, we want to return a fraction (pool count) / (cache count) of the pool, i.e. (pool count)^2 / (cache count) items. First off, this fraction is not fixed for the call duration of the call into QueryAndOrderInPages, but rather increases as the pool count increases from (page size) to (pool maximum). [This may disagreement may be related to the current code that fills the pool before returning anything.] Second, even if the fraction were fixed, it would be (pool size) / (cache count), not 1.0 / (cache count).

Public/Src/Cache/ContentStore/Distributed/NuCache/NuCacheCollectionUtilities.cs
```diff
@@ -268,14 +280,37 @@ public static IEnumerable<T> QueryAndOrderInPages<T>(this IEnumerable<T> origina
...
+                    {
+                        minimumYieldSize = (int)Math.Floor(removalFraction * pool.Count);
+                        // To guarantee termination, we always yield at least one element
+                        minimumYieldSize = Math.Max(minimumYieldSize, 1);
+                        // Never yield more than one page of results. This is about maintaining baseline accuracy for very large fractions.
+                        minimumYieldSize = Math.Min(minimumYieldSize, pageSize);
+                    }
+
+                    Contract.Assert(minimumYieldSize <= pool.Count);
+                    for (var i = 0; i < minimumYieldSize; ++i) {
```
dcwuser: Given my disagreement (and/or wrongness) about what pattern of yield sizes is under this code, it would be great if you could run in debug and tell me what series of minimumYieldSizes you see.

dcwuser:
Just to be clear about how I thought this would work, let me give an example with pageSize = 2K, cacheCount = 1M, and poolMultiplier = 4. I would expect:
   pull 2K candidates => pool size 2K => return 4 
   pull 2K candidates => pool size 4K => return 16
   pull 2K candidates => pool size 6K => return 36
   pull 2K candidates => pool size 8K => return 64
   pool is now full, so pull 2K and return 2K from now on
Transition would become less abrupt for larger poolMultiplier. I'm not particular about exact transition from not full to full behavior, but basic idea would be that we return a fraction until pool full, then return pages.

dcwuser:
Here's a version that keeps behavior of starting small, but makes transition to pool full behavior less abrupt: for each page pulled, increase the number returned by pageSize /poolMultiplier, until after poolMultiplier pulls, we are returning as many items as we pulled. In our example, that would look like:
  pull 2K candidates => pool size 2K => return 500
  pull 2K candidates => pool size 3.5K => return 1000
  pull 2K candidates => pool size 5.5K => return 1500
  pull 2K candidates => pool size 6.5K => return 2000
  pull 2K candidates => pool size 6.5K => return 2000
This doesn't return the theoretically justified tiny fraction at first, but it transitions more cleanly to the full pool case, and as long as poolMultiplier >> 2, it's still much better than our current behavior.

---
In reply to: [535702219](https://github.com/microsoft/BuildXL/pull/945#issuecomment-535702219) [](ancestors = 535702219)

dcwuser:
Actually, that same math with the increase being the slightly smaller value pageSize / (poolMultiplier + 1) lets the pool get full in a way that it didn't without the +1.
  pull 2K candidates => pool size 2K => return 400
  pull 2K candidates => pool size 3.6K => return 800
  pull 2K candidates => pool size 4.8K => return 1200
  pull 2K candidates => pool size 5.6K => return 1600
  pull 2K candidates => pool size 6.0K => return 2000
  pull 2K candidates => pool size 8.0K => return 2000
So that's my latest suggestion.

---
In reply to: [535704990](https://github.com/microsoft/BuildXL/pull/945#issuecomment-535704990) [](ancestors = 535704990,535702219)

PR #947, opened 2019-09-26
Fix SandboxedProcessMac.UpdatePerfCounters

almiliMSFT:
It shouldn't reschedule the timer if the process has exited

[AB#1604999](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1604999)

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessMac.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessMac.cs
@@ -172,7 +172,7 @@ namespace BuildXL.Processes
             proc.m_perfAggregator.DiskBytesWritten.RegisterSample(buffer.DiskBytesWritten);
 
             // reschedule the timer to fire again in PerfProbeInterval time (2 seconds)
-            if (!proc.Process.HasExited)
+            if (!proc.Process.HasExited) 
             {
                 proc.m_perfTimer.Change(dueTime: PerfProbeInternal, period: Timeout.InfiniteTimeSpan);
             }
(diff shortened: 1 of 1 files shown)


mpysson:
>             m_perfTimer.Dispose();

Any outstanding occurrences of the timer will keep firing even after Dispose is called. You may want to call .Change() to prevent it from firing and then Dispose(WaitHandle) to synchronously dispose so you can ensure it doesn't mess with other state that you might clean up lower down here.

[https://docs.microsoft.com/en-us/dotnet/api/system.threading.timer.dispose?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.threading.timer.dispose?view=netframework-4.8)

---
Refers to: Public/Src/Engine/Processes/SandboxedProcessMac.cs:303 in a1a5665. [](commit_id = a1a566511c67512d98bbd8adcdea26f8a3241612, deletion_comment = False)

Public/Src/Engine/Processes/SandboxedProcessMac.cs
```diff
@@ -172,7 +172,7 @@ private static void UpdatePerfCounters(object state)
             proc.m_perfAggregator.DiskBytesWritten.RegisterSample(buffer.DiskBytesWritten);
 
             // reschedule the timer to fire again in PerfProbeInterval time (2 seconds)
-            if (!proc.Killed)
+            if (!proc.Process.HasExited)
```
almiliMSFT:
```suggestion
            if (!proc.Process.HasExited) 
```

Public/Src/Engine/Processes/SandboxedProcessMac.cs
```diff
@@ -172,7 +172,7 @@ private static void UpdatePerfCounters(object state)
...
```
almiliMSFT:
```suggestion
            if (!proc.Process.HasExited)
```

PR #950, opened 2019-09-27
Make PipHistoricPerfData merge crashes more visible

Kr1sso:
Remove implicit casts and add some debugging capabilities for historic perf data crashes we are seeing rarely.

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessMac.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessMac.cs
@@ -414,7 +414,7 @@ namespace BuildXL.Processes
                         ReadTransferCount = Convert.ToUInt64(m_perfAggregator.DiskBytesRead.Total),
                         WriteTransferCount = Convert.ToUInt64(m_perfAggregator.DiskBytesWritten.Total)
                     }),
-                    PeakMemoryUsage = (ulong)m_perfAggregator.PeakMemoryBytes.Maximum,
+                    PeakMemoryUsage = Convert.ToUInt64(m_perfAggregator.PeakMemoryBytes.Maximum),
                     KernelTime = TimeSpan.FromMilliseconds(m_perfAggregator.JobKernelTimeMs.Latest),
                     UserTime = TimeSpan.FromMilliseconds(m_perfAggregator.JobUserTimeMs.Latest),
                 };
--- a/Public/Src/Engine/Scheduler/PipHistoricPerfData.cs
+++ b/Public/Src/Engine/Scheduler/PipHistoricPerfData.cs
@@ -6,6 +6,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.ContractsLight;
 using BuildXL.Pips;
 using BuildXL.Utilities;
+using static BuildXL.Utilities.FormattableStringEx;
 
 namespace BuildXL.Scheduler
 {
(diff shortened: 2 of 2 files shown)


Public/Src/Engine/Processes/SandboxedProcessMac.cs
```diff
@@ -414,7 +414,7 @@ internal override JobObject.AccountingInformation GetJobAccountingInfo()
                         ReadTransferCount = Convert.ToUInt64(m_perfAggregator.DiskBytesRead.Total),
                         WriteTransferCount = Convert.ToUInt64(m_perfAggregator.DiskBytesWritten.Total)
                     }),
-                    PeakMemoryUsage = (ulong)m_perfAggregator.PeakMemoryBytes.Maximum,
+                    PeakMemoryUsage = Convert.ToUInt64(m_perfAggregator.PeakMemoryBytes.Maximum),
```
Kr1sso: Potential culprit.

mpysson: >Convert.ToUInt64 [](start = 38, length = 16)

Convert.ToUInt64() still throws an OverflowException if PeakMemoryBytes is negative or too large. What does this change?

Public/Src/Engine/Scheduler/PipHistoricPerfData.cs
```diff
@@ -176,23 +177,30 @@ public PipHistoricPerfData MakeFresh()
...
-                GetMergeResult(DiskIOInKB, other.DiskIOInKB));
+            var durationResult = GetMergeResult(DurationInMs, other.DurationInMs);
+            var peakMemoryResult = GetMergeResult(PeakMemoryInKB, other.PeakMemoryInKB);
+            var processorInPercentResult = GetMergeResult(ProcessorsInPercents, other.ProcessorsInPercents);
+            var diskIOResult = GetMergeResult(DiskIOInKB, other.DiskIOInKB);
+
+            return new PipHistoricPerfData(DefaultTimeToLive, durationResult, peakMemoryResult, (ushort)processorInPercentResult, diskIOResult);
         }
 
         private static uint GetMergeResult(uint oldData, uint newData)
```
mpysson: >GetMergeResult [](start = 28, length = 14)

How about making this internal and hitting it with some unit tests to bounds test the parameters?

Public/Src/Engine/Scheduler/PipHistoricPerfData.cs
```diff
@@ -176,23 +177,30 @@ public PipHistoricPerfData MakeFresh()
...
-                GetMergeResult(DurationInMs, other.DurationInMs),
-                GetMergeResult(PeakMemoryInKB, other.PeakMemoryInKB),
-                (ushort)GetMergeResult(ProcessorsInPercents, other.ProcessorsInPercents),
-                GetMergeResult(DiskIOInKB, other.DiskIOInKB));
+            var durationResult = GetMergeResult(DurationInMs, other.DurationInMs);
+            var peakMemoryResult = GetMergeResult(PeakMemoryInKB, other.PeakMemoryInKB);
+            var processorInPercentResult = GetMergeResult(ProcessorsInPercents, other.ProcessorsInPercents);
+            var diskIOResult = GetMergeResult(DiskIOInKB, other.DiskIOInKB);
+
+            return new PipHistoricPerfData(DefaultTimeToLive, durationResult, peakMemoryResult, (ushort)processorInPercentResult, diskIOResult);
```
almiliMSFT: What's the change here?  Only reformatting?

Public/Src/Engine/Scheduler/PipHistoricPerfData.cs
```diff
@@ -176,23 +177,30 @@ public PipHistoricPerfData MakeFresh()
...
-            // An asymmetric merge that goes up fast but decreases slowly.
-            if (newData > oldData)
+            try
             {
-                return (uint)(newData + (ulong) oldData) / 2;
-            }
+                // An asymmetric merge that goes up fast but decreases slowly.
+                if (newData > oldData)
+                {
+                    return (uint)(newData + (ulong) oldData) / 2;
```
almiliMSFT: how about changing to `newData/2 + oldData/2`?

Public/Src/Engine/Scheduler/PipHistoricPerfData.cs
```diff
@@ -176,23 +177,30 @@ public PipHistoricPerfData MakeFresh()
...
+                }
 
-            return (uint)((((ulong)oldData * 9) + ((ulong)newData * 1)) / 10);
+                return (uint)((((ulong)oldData * 9) + ((ulong)newData * 1)) / 10);
+            }
+            catch(System.OverflowException ex)
```
almiliMSFT:
```suggestion
            catch (System.OverflowException ex)
```

Public/Src/Engine/Scheduler/PipHistoricPerfData.cs
Kr1sso: Sort of, before the stack trace could not indicate any line number hence its was not easy to say which counter caused a potential overflow.

Public/Src/Engine/Scheduler/PipHistoricPerfData.cs
Kr1sso: Indeed.

Public/Src/Engine/Processes/SandboxedProcessMac.cs
Kr1sso: Thats the point exactly, merely casting does not throw an exception. This way we can at least get a better idea of where bogus data is coming from.

PR #951, opened 2019-09-27
Fix files missing from drop when ipc pips run on workers

olkononenko:
Fix for [AB#1598050](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1598050). 

This one is a different issue though it's tracked by the same bug.

### DIFF
--- a/Public/Src/Engine/Scheduler/ApiServer.cs
+++ b/Public/Src/Engine/Scheduler/ApiServer.cs
@@ -129,7 +129,7 @@ namespace BuildXL.Scheduler
             }
 
             var getSealedDirectoryFilesCmd = cmd as GetSealedDirectoryContentCommand;
-            if(getSealedDirectoryFilesCmd != null)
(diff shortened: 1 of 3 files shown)


Public/Src/Engine/Scheduler/ApiServer.cs
```diff
@@ -239,9 +239,14 @@ private async Task<IIpcResult> ExecuteGetSealedDirectoryContent(GetSealedDirecto
 
             if (failedResults.Count > 0)
             {
-                new IpcResult(
+                return new IpcResult(
```
mpysson: >return  [](start = 16, length = 7)

Oh boy. I guess this is why we didn't see the error from drop daemon? :cry:

olkononenko: yeah

PR #954, opened 2019-09-27
Enable CustomPipDescription in event logs in ADO

norahuang:
Enable CustomPipDescription in event logs in ADO.

### DIFF
--- a/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
+++ b/Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
@@ -215,7 +215,6 @@ export function runQTest(args: QTestArguments): Result {
             tempDirectory: qtestRunTempDirectory,
             weight: args.weight,
             environmentVariables: [
-                { name: "[Sdk.BuildXL]qCodeCoverageEnumType", value: qCodeCoverageEnumType },
                 ...(args.qTestEnvironmentVariables || [])
             ],
             disableCacheLookup: Environment.getFlag("[Sdk.BuildXL]qTestForceTest"),
--- a/Public/Src/App/Bxl/AzureDevOpsListener.cs
+++ b/Public/Src/App/Bxl/AzureDevOpsListener.cs
@@ -44,8 +44,9 @@ namespace BuildXL
             Events eventSource,
             IConsole console,
             DateTime baseTime,
-            BuildViewModel buildViewModel)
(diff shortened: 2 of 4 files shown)


Public/Sdk/SelfHost/BuildXL/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -215,7 +215,6 @@ export function runQTest(args: QTestArguments): Result {
             tempDirectory: qtestRunTempDirectory,
             weight: args.weight,
             environmentVariables: [
-                { name: "[Sdk.BuildXL]qCodeCoverageEnumType", value: qCodeCoverageEnumType },
```
olkononenko: > { name: "[Sdk.BuildXL]qCodeCoverageEnumType", value: qCodeCoverageEnumType }, [](start = 15, length = 78)

is this relevant to this PR?

norahuang: This is a wrong commit, I will fix it.

Public/Src/App/Bxl/AzureDevOpsListener.cs
```diff
@@ -202,6 +203,8 @@ private void LogAzureDevOpsIssue(EventWrittenEventArgs eventData, string eventTy
             // report the entire message since Azure DevOps does not yet provide actionalbe information from the metadata.
             body = string.Format(CultureInfo.CurrentCulture, message, args);
```
olkononenko: I understand that this code was already there, but  why are we not calling FormattingEventListener.CreateFullMessageString method here? The big if block above appears to be copy pasted from that method (though some other bits were left there)...

Public/Src/Utilities/Utilities/Tracing/FormattingEventListener.cs
```diff
@@ -342,26 +342,34 @@ public static string CreateFullMessageString(EventWrittenEventArgs eventData, st
...
+
+            return full;
+        }
+
+        /// <summary>
+        /// Removes BuildXL generated pip description if custom privides its own one
+        /// </summary>
+        protected static void ProcessCustomPipDescription(ref string message, int numberOfArgs, bool useCustomPipDescription)
+        {
             // if args is empty, there is no pip description in the final string
```
olkononenko: > if args is empty [](start = 14, length = 17)

nit: I'd change this slightly cause it's not clear what are those "args" the comment is referring to. maybe 'If there are no arguments, ...' ? Cannot say that I like this one either :)

PR #955, opened 2019-09-27
Enable CustomPipDescription in event logs in ADO

norahuang:


Public/Src/App/Bxl/AzureDevOpsListener.cs
```diff
@@ -202,6 +203,8 @@ private void LogAzureDevOpsIssue(EventWrittenEventArgs eventData, string eventTy
             // report the entire message since Azure DevOps does not yet provide actionalbe information from the metadata.
             body = string.Format(CultureInfo.CurrentCulture, message, args);
```
olkononenko: >body = string.Format(CultureInfo.CurrentCulture, message, args); [](start = 12, length = 64)

I understand that this code was already there, but  why are we not calling FormattingEventListener.CreateFullMessageString method here? The big if block above appears to be copy pasted from that method (though some other bits were left there)...

Public/Src/Utilities/Utilities/Tracing/FormattingEventListener.cs
```diff
@@ -342,26 +342,34 @@ public static string CreateFullMessageString(EventWrittenEventArgs eventData, st
...
+
+            return full;
+        }
+
+        /// <summary>
+        /// Removes BuildXL generated pip description if custom privides its own one
+        /// </summary>
+        protected static void ProcessCustomPipDescription(ref string message, int numberOfArgs, bool useCustomPipDescription)
+        {
             // if args is empty, there is no pip description in the final string
```
olkononenko: >if args is empty [](start = 15, length = 16)

nit: I'd change this slightly cause it's not clear what are those "args" the comment is referring to. maybe 'If there are no arguments, ...' ? Cannot say that I like this one either :)

Public/Src/App/Bxl/AzureDevOpsListener.cs
norahuang: I will try to refactor FormattingEventListener, AzureDevOpsListener, and ConsleEventListener in another PR. In this one, let's make the function work first.

PR #956, opened 2019-09-30
Fix crash in ReportStatsForBuildSummary when execute phase doesn't exit cleanly

mpysson:


cg/nuget/cgmanifest.json
```diff
@@ -2670,7 +2670,7 @@
         "Type": "NuGet",
         "NuGet": {
           "Name": "runtime.osx-x64.BuildXL",
-          "Version": "1.99.99"
+          "Version": "2.0.1"
```
mpysson: @rijulluman-msft - this is unrelated to my change. Is it just due to staging of when we enabled the enforcement side of things?

Kr1sso: I haven't been in the loop about this - what does this indicate, cause if its not updated?

mpysson: It's automatically generated in dev builds. Soon, PR and rolling builds will fail if the cgmanifest file is out of sync with the packages that are fetched. The purpose of the file is to pump package information into the component governance system.

almiliMSFT: yes, this is exactly because the cgmanifest file checked in with Rijul's change was a bit stale

Kr1sso: Thanks!

rijulluman-msft: Yes, looks like the file I generated was a bit stale

PR #958, opened 2019-09-30
Make CG manifest generation uniform across platforms

almiliMSFT:


cg/nuget/cgmanifest.json
```diff
@@ -149,17 +149,17 @@
...
         }
       }
     },
     {
       "Component": {
         "Type": "NuGet",
         "NuGet": {
-          "Name": "Dotnet-Runtime",
-          "Version": "5.0.2"
+          "Name": "DotNet.Glob",
```
mpysson: These are going to flip flop based on whether people have synced to master. Also the LKG has one sort and dev bits have another sort. How should we handle the oscillation?
- Update LKG after this? (release candidate complications)
- Disable the manifest generation for a week?
- Send people a heads up email and live with it?

rijulluman-msft: The validation step uses deepEquals on json objects, so the sorting order will not cause validation failures
About syncing with master, I think there will be a merge conflict if both the master and new branch have changes to Nugets used. So a new merge will be required in the PR and new cgmanifest needs to be generated to keep the PR valid.

almiliMSFT: It is going to flip flop until we sort out the issues.

I've reverted the cgmanifest.json file  so that whatever is checked in is whatever is generated with the current LKG

PR #959, opened 2019-09-30
Proactive copy fixes

JuanCarlosGI:
Make sure that we register our location before requesting the proactive copy


SergeyTeplyakov:
>                     var putBlobResult = await ContentLocationStore.PutBlobAsync(context, result.ContentHash, recorder.RecordedBytes).FireAndForgetAndReturnTask(context);

I think this make no sense. Because we don't fire and we don't forget stuff:)


---
Refers to: Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs:241 in d413959. [](commit_id = d41395914d8c6d33480a67e0b9f63ec218b79e56, deletion_comment = False)

PR #961, opened 2019-09-30
Avoid duplicated error/warning message on Azure DevOps

norahuang:
OnError and OnWarning event handles write very similar messages both in ConsoleEventListener.cs  and AzureDevOpsListener.cs. This leads to duplicated messages on Azure DevOps console and the coloring of the logs from both messages is confusing. This PR removes writing the error and warning messages in ConsoleEventListener.cs. So there is only single message per error or warning event in Azure devops console.

### DIFF
--- a/Public/Src/App/Bxl/BuildXLApp.cs
+++ b/Public/Src/App/Bxl/BuildXLApp.cs
@@ -1426,7 +1426,7 @@ namespace BuildXL
                     m_noLogMask,
                     onDisabledDueToDiskWriteFailure: OnListenerDisabledDueToDiskWriteFailure,
                     maxStatusPips: m_configuration.FancyConsoleMaxStatusPips,
-                    optimizeForAzureDevOps: m_configuration.OptimizeConsoleOutputForAzureDevOps || m_configuration.OptimizeWarningOrErrorAnnotationsForAzureDevOps);
+                    optimizeForAzureDevOps: m_configuration.OptimizeConsoleOutputForAzureDevOps || m_configuration.OptimizeVsoAnnotationsForAzureDevOps);
 
                 listener.SetBuildViewModel(buildViewModel);
 
@@ -2100,8 +2100,7 @@ namespace BuildXL
 
             if (loggingConfiguration.OptimizeConsoleOutputForAzureDevOps
                 || loggingConfiguration.OptimizeProgressUpdatingForAzureDevOps
-                || loggingConfiguration.OptimizeVsoAnnotationsForAzureDevOps
-                || loggingConfiguration.OptimizeWarningOrErrorAnnotationsForAzureDevOps)
+                || loggingConfiguration.OptimizeVsoAnnotationsForAzureDevOps)
             {
                 // Use a very simple logger for azure devops
                 return new StandardConsole(colorize: false, animateTaskbar: false, supportsOverwriting: false, pathTranslator: translator);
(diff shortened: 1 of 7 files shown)


Public/Src/App/Bxl/ConsoleEventListener.cs
```diff
@@ -440,6 +446,12 @@ protected override void OnError(EventWrittenEventArgs eventData)
         /// <inheritdoc />
         protected override void OnWarning(EventWrittenEventArgs eventData)
         {
+            // AzureDevOpsListener has alreday write the event to console, avoid duplication
```
mpysson: >alreday write [](start = 39, length = 13)

already written

norahuang: fixed

Public/Src/App/Bxl/BuildXLApp.cs
```diff
@@ -1426,7 +1426,7 @@ private void ConfigureConsoleLogging(bool notWorker, BuildViewModel buildViewMod
                     m_noLogMask,
                     onDisabledDueToDiskWriteFailure: OnListenerDisabledDueToDiskWriteFailure,
                     maxStatusPips: m_configuration.FancyConsoleMaxStatusPips,
-                    optimizeForAzureDevOps: m_configuration.OptimizeConsoleOutputForAzureDevOps || m_configuration.OptimizeWarningOrErrorAnnotationsForAzureDevOps);
+                    optimizeForAzureDevOps: m_configuration.OptimizeConsoleOutputForAzureDevOps || m_configuration.OptimizeVsoAnnotationsForAzureDevOps);
```
mpysson: >m_configuration.OptimizeConsoleOutputForAzureDevOps || m_configuration.OptimizeVsoAnnotationsForAzureDevOps [](start = 44, length = 107)

How are these 2 flags different now?

mpysson:
>             bool optimizeForAzureDevOps = false)

I guess this now means something more like: hide errors & warnings because they'll be handled by the AzureDevOpsListener. Suggest rename/comment to make this clear.

Also, can we structure to make sure the two are kept in agreement and no user configuration can make errors disappear? Having builds fail without anything on the console could prove quite confusing

---
Refers to: Public/Src/App/Bxl/ConsoleEventListener.cs:190 in 6fde741. [](commit_id = 6fde7410bb9a261f7e60a9f27bbcd49f8fef0d68, deletion_comment = False)

Public/Src/App/Bxl/BuildXLApp.cs
norahuang: OptimizeWarningOrErrorAnnotationsForAzureDevOps was used to control outputting error/warning message in ConsoleEventListener. But since I rolled out, we don't need it any more. 

Instead, if OptimizeConsoleOutputForAzureDevOps  or OptimizeVsoAnnotationsForAzureDevOps is set, that mean AzureDevOpsListener is in place and ConsoleEventListener should be aware of it and don't output error and warning which are taken care by AzureDevOpsListener.

norahuang:
> bool optimizeForAzureDevOps = false)
> 
> 
> I guess this now means something more like: hide errors & warnings because they'll be handled by the AzureDevOpsListener. Suggest rename/comment to make this clear.
> Also, can we structure to make sure the two are kept in agreement and no user configuration can make errors disappear? Having builds fail without anything on the console could prove quite confusing
> 
> Refers to: Public/Src/App/Bxl/ConsoleEventListener.cs:190 in 6fde741. [](commit_id = 6fde741, deletion_comment = False)

I don't see any possible user configuration will lead to errors disappear. But we do still have a situation that when OptimizeProgressUpdatingForAzureDevOps is set, we still get duplicated errors. However, I think we will roll out the separated flags for azure dev opt optimization soon. So I don't want add this one in the check.

PR #963, opened 2019-09-30
Make sure that build machines in proactive copy are active

JuanCarlosGI:
... also reduce logging.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/TransitioningContentLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/TransitioningContentLocationStore.cs
@@ -243,6 +243,19 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             return _localLocationStore.ClusterState.GetRandomMachineLocation(except);
         }
 
+        /// <inheritdoc />
+        public bool IsMachineActive(MachineLocation machine)
+        {
+            if (_configuration.HasReadMode(ContentLocationMode.Redis))
+            {
+                return _redisContentLocationStore.IsMachineActive(machine);
+            }
+
+            return _localLocationStore.ClusterState.TryResolveMachineId(machine, out var machineId)
+                ? !_localLocationStore.ClusterState.InactiveMachines.Contains(machineId)
+                : false;
+        }
+
         #region IDistributedLocationStore Members
 
         /// <inheritdoc />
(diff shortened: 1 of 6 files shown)


Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStore.cs
```diff
@@ -1326,6 +1326,10 @@ public Result<MachineLocation> GetRandomMachineLocation(IReadOnlyList<MachineLoc
             return new Result<MachineLocation>("Could not select a machine location.");
         }
 
+
+        /// <inheritdoc />
```
SergeyTeplyakov: nit: extra new line before this method.

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStore.cs
```diff
@@ -1326,6 +1326,10 @@ public Result<MachineLocation> GetRandomMachineLocation(IReadOnlyList<MachineLoc
             return new Result<MachineLocation>("Could not select a machine location.");
         }
 
+
+        /// <inheritdoc />
+        public bool IsMachineActive(MachineLocation machine) => _idsByLocation.Keys.Contains(machine);
```
SergeyTeplyakov: >_idsByLocation [](start = 64, length = 14)

I think it should be "_idsByLocation.ContainsKey()", because the "Keys" property will acquire a global lock and copy all the keys.

Public/Src/Cache/ContentStore/Distributed/NuCache/TransitioningContentLocationStore.cs
```diff
@@ -243,6 +243,19 @@ public Result<MachineLocation> GetRandomMachineLocation(IReadOnlyList<MachineLoc
...
 
+        /// <inheritdoc />
+        public bool IsMachineActive(MachineLocation machine)
+        {
+            if (_configuration.HasReadMode(ContentLocationMode.Redis))
+            {
+                return _redisContentLocationStore.IsMachineActive(machine);
+            }
+
+            return _localLocationStore.ClusterState.TryResolveMachineId(machine, out var machineId)
```
SergeyTeplyakov: Does it make sense to add a helper IsMachineActive to cluster state? (directly, or using extension method?)

SergeyTeplyakov: My usual way of thinking is: whether the change will make code more readable or not, but not whether we'll use the method in another place.

For instance, the following code is easier (IMO) to read because it deals with the same level of abstraction:

return _configuration.HashReadMode(Redis) 
  ?   _redisContentLocationStore.IsMachineActive(machine)
  : _localLocationStore.IsMachineActive(machine);

But the current version digs deeper into the cluster's state and is lower level.

---
In reply to: [329783222](https://github.com/microsoft/BuildXL/pull/963#discussion_r329783222) [](ancestors = 329783222)

PR #966, opened 2019-09-30
Properly build 'buildxl.dscript' VSCode extension with Npm

rijulluman-msft:
[AB#1478432](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1478432)


Public/Sdk/Experimental/NodeJs/node.dsc
```diff
@@ -103,4 +103,27 @@ namespace Node {
 
         return pkgContents.getFile(executable);
     }
+
+    @@public
+    export function tscCompileOut(workingDirectory: Directory, ...dependencies: StaticDirectory[]) : OpaqueDirectory {
```
almiliMSFT: rename to `tscCompile`

Public/Sdk/Experimental/NodeJs/npm.dsc
```diff
@@ -41,14 +41,43 @@ namespace Npm {
         };
     }
 
+    @@public
+    export function npmInstallRoot(workingDirectory: StaticDirectory): OpaqueDirectory {
```
almiliMSFT: Rename to `npmInstall`

Public/Sdk/Public/Deployment/definition.dsc
```diff
@@ -53,7 +53,7 @@ export interface DeployedFileWithProvenance {
 @@public
 export interface FlattenedResult {
     flattenedFiles: Map<RelativePath, DeployedFileWithProvenance>,
-    flattenedOpaques: Map<RelativePath, OpaqueDirectory>,
+    flattenedOpaques: Map<RelativePath, [OpaqueDirectory, RelativePath]>,
```
almiliMSFT: maybe add a comment explaining that [OpaqueDirectory, RelativePath] is a tuple of (1) an opaque directory, and (2) a path relative to that opaque directory designating a subdirectory to be added to the deployment

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -16,6 +16,16 @@ export interface OnDiskDeployment {
 
     /** Optional primary file, i.e. an executable or test dll */
     primaryFile?: File;
+
+    /** Optional opaque directories robocopied/rsynced into this deployment */
+    targetOpaques?: OpaqueDirectory[];
+}
+
+@@public
+export interface OpaqueSubDirectory extends Deployable {
```
almiliMSFT: some comments on this interface as well as on its properties would be useful

almiliMSFT: also for the new functions below

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -39,6 +49,147 @@ export interface DeployToDiskArguments {
...
+                Cmd.argument(Artifact.none(source)),
+                Cmd.argument(Artifact.output(target))
+            ],
+            dependencies: [
+                sourceOpaqueDir
+            ]
+        };
+
+    const result = Transformer.execute(args);
+
```
almiliMSFT:
```suggestion
```

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -39,6 +49,147 @@ export interface DeployToDiskArguments {
...
+    return result.getOutputFile(target);
+}
+
+/**
+ * Based on the current platform schedules either a robocopy.exe or rsync pip to copy 'sourceDir' to 'targetDir'
```
almiliMSFT:
```suggestion
 * Based on the current platform schedules either a robocopy.exe or rsync pip to copy 'sourceDir' to 'targetDir'.
 * That pip takes a dependency on `sourceDirDep`.  If 'sourceDir' is not within `sourceDirDep.root`, disallowed
 * file accesses are almost certain to happen.
```

Public/Src/IDE/VsCode/BuildXL.IDE.VsCode.dsc
```diff
@@ -1,22 +1,44 @@
...
+    @public
+    export const installRootDir: OpaqueDirectory = Npm.npmInstallRoot(clientCopy);
 
+    @@public
+    export const compileOutDir: OpaqueDirectory = Node.tscCompileOut(clientCopy.root, clientCopy, installRootDir);
+
+    @@public
+    export const deployedNpmPackageLockFile = Deployment.copyFileFromOpaqueDirectory(
+        // Here we want to copy a file from inside an opaque output directory. 
+        // If done using Transformer.copyFile we would have no way of specifying a dependency of that copy pip to the pip producing this opaque directory, so we wouldn't be able to ensure that the copy operation runs after the opaque directory is produced.
```
almiliMSFT:
```suggestion
        // If done using Transformer.copyFile we would have no way of specifying a dependency 
        // of that copy pip to the pip producing this opaque directory, so we wouldn't be able to
        // ensure that the copy operation runs after the opaque directory is produced.
```

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -16,6 +16,17 @@ export interface OnDiskDeployment {
...
     /** Optional primary file, i.e. an executable or test dll */
     primaryFile?: File;
+
+    /** Optional opaque directories robocopied/rsynced into this deployment */
+    targetOpaques?: OpaqueDirectory[];
+}
+
+@@public
+export interface OpaqueSubDirectory extends Deployable {
+    // Used to represent deployment of a subdirectory of an opaque directory
```
almiliMSFT:
```suggestion
    /** Parent opaque directory */
```

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -16,6 +16,17 @@ export interface OnDiskDeployment {
...
+    opaque: OpaqueDirectory,
+    subDirectory: RelativePath,     // a path relative to that opaque directory designating a subdirectory to be added to the deployment
```
almiliMSFT:
```suggestion
    /** a path relative to that opaque directory designating a subdirectory to be added to the deployment */
    subDirectory: RelativePath,
```

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -16,6 +16,17 @@ export interface OnDiskDeployment {
 
     /** Optional primary file, i.e. an executable or test dll */
     primaryFile?: File;
+
+    /** Optional opaque directories robocopied/rsynced into this deployment */
+    targetOpaques?: OpaqueDirectory[];
+}
+
+@@public
```
almiliMSFT:
```suggestion
/**
 * Used to represent deployment of a subdirectory of an opaque directory.
 *
 * Call the 'createDeployableOpaqueSubDirectory' function below to create an instance of this type.
 */
@@public
```

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -39,6 +50,148 @@ export interface DeployToDiskArguments {
...
+                    flattenedFiles: result.flattenedFiles,
+                    flattenedOpaques: result.flattenedOpaques.add(targetFolder, [<OpaqueDirectory>opaque, sub]),
+                    visitedItems: result.visitedItems.add(d`{opaque.root}/${sub}`),
+                };
+            }
+        }
+    };
+}
+
+@@public
```
almiliMSFT:
```suggestion
/**
 * Schedules a platform-specific process to copy file from 'source' to 'target'.  This process takes a dependency 
 * on 'sourceOpaqueDir`, which should be the parent opaque directory containing the file 'source'.
 */
@@public
```

Public/Sdk/Experimental/NodeJs/node.dsc
```diff
@@ -103,4 +103,27 @@ namespace Node {
...
+            Cmd.argument("-p"),
+            Cmd.argument("."),
+        ];
+
+        const result = Node.run({
+            arguments: arguments,
+            workingDirectory: workingDirectory,
+            dependencies : [
+                ...dependencies
+            ],
```
smera: nit: simplify to 'dependencies'

Public/Sdk/Experimental/NodeJs/npm.dsc
```diff
@@ -41,14 +41,43 @@ namespace Npm {
...
+        const arguments: Argument[] = [
+            Cmd.argument(Artifact.input(Node.npmCli)),
+            Cmd.argument("install"),
+            Cmd.option("--cache ", Artifact.none(npmCachePath)), // Forces the npm cache to use this output folder for this object so that it doesn't write to user folder
+        ];
+
+        const result = Node.run({
+            arguments: arguments,
+            workingDirectory: wd,
+            dependencies: [ workingDirectory ].filter(x => x !== undefined),
```
smera: >[ workingDirectory ].filter(x => x !== undefined), [](start = 26, length = 50)

nit: maybe an 'addIf' will look simpler

smera: >workingDirectory [](start = 28, length = 16)

in what case this will be undefined?

Public/Sdk/Experimental/NodeJs/npm.dsc
```diff
@@ -41,14 +41,43 @@ namespace Npm {
         };
     }
 
+    @@public
+    export function npmInstall(workingDirectory: StaticDirectory): OpaqueDirectory {
```
smera: >workingDirectory [](start = 31, length = 16)

the name looks slightly misleading to me. Given that its type is a static directory, it is already representing a group of artifacts, and not just a working directory. What's the semantics of this collection of files? The output of a compilation?

smera:
>  * Helper type alias for handeling duplicate file deployments. It can decide to return one or the other, but commonly it should just fail.

handling

---
Refers to: Public/Sdk/Public/Deployment/definition.dsc:61 in cc9089b. [](commit_id = cc9089bb09150160e362e4673468c257370c27ef, deletion_comment = False)

Public/Sdk/Experimental/NodeJs/npm.dsc
almiliMSFT: The semantics is:
  - use that folder as the working directory
  - take a dependency on that folder (so that this pip is scheduled after this static directory is produced and that it can read everything in it)

Maybe rename to `rootDir`.

smera:
>      * @param item - The item that is deployable. Think of this as the 'this' pointer which is not accessable from interface implementations.

accessible 

---
Refers to: Public/Sdk/Public/Deployment/definition.dsc:87 in cc9089b. [](commit_id = cc9089bb09150160e362e4673468c257370c27ef, deletion_comment = False)

Public/Sdk/Public/Deployment/definition.dsc
```diff
@@ -53,7 +53,7 @@ export interface DeployedFileWithProvenance {
 @@public
 export interface FlattenedResult {
     flattenedFiles: Map<RelativePath, DeployedFileWithProvenance>,
-    flattenedOpaques: Map<RelativePath, OpaqueDirectory>,
+    flattenedOpaques: Map<RelativePath, [OpaqueDirectory, RelativePath]>,  // Tuple of (1) an opaque directory, and (2) a path relative to that opaque directory designating a subdirectory to be added to the deployment
```
smera: >RelativePath [](start = 58, length = 12)

this change seems to imply that now all flattened opaques will have a designated subdirectory to be added to the deployment? what if that's not the case?

almiliMSFT: when that's not the case, the `RelativePath` part is 
```
r`.`
```
so everything works exactly as it did before

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -39,6 +57,152 @@ export interface DeployToDiskArguments {
...
+            item: Object, 
+            targetFolder: RelativePath,
+            handleDuplicateFile: HandleDuplicateFileDeployment, 
+            result: FlattenedResult,
+            deploymentOptions?: Object,
+            provenance?: Diagnostics.Provenance) => 
+        {
+            const existingOpaque = result.flattenedOpaques.get(targetFolder);
+
+            if (existingOpaque !== undefined) {
```
smera: >existingOpaque [](start = 16, length = 14)

can this be undefined?

Public/Sdk/Public/Deployment/definition.dsc
smera: >[OpaqueDirectory, RelativePath] [](start = 40, length = 31)

I'd rather see a proper named interface for this. Using a tuple for local usage is fine IMO, but this definition makes expressions like [0][1], [0][0] and similar pop up all over the place, which are somewhat cryptic to follow

Public/Sdk/Public/Deployment/filesystem.dsc
almiliMSFT: This follows the same pattern as [here](https://github.com/microsoft/BuildXL/blob/42cf246cd46a20fce8e5b626732b73db8adb7c42/Public/Sdk/Public/Deployment/operations.dsc#L229)

That said, I'm not sure if this can be undefined.  I don't think it can.

Public/Sdk/Experimental/NodeJs/npm.dsc
almiliMSFT: it's never undefined and should be changed to 
```suggestion
            dependencies: [ workingDirectory ],
```

Public/Sdk/Public/Deployment/operations.dsc
```diff
@@ -240,7 +240,7 @@ function flattenStaticDirectory(staticDirectory: StaticDirectory, targetFolder:
                 // TODO: Validate if there is a flattenedFile already under this OpaqueDirectory. To implement this we'll need IsWithin on RelativePath
                 return {
                     flattenedFiles: result.flattenedFiles,
-                    flattenedOpaques: result.flattenedOpaques.add(targetFolder, <OpaqueDirectory>staticDirectory),
+                    flattenedOpaques: result.flattenedOpaques.add(targetFolder, [<OpaqueDirectory>staticDirectory, r`.`]),
```
smera: > r`.` [](start = 114, length = 5)

Ok, so this is an example of the comment I made on definition.dsc. In this case we don't really need a designated subdir.
My suggestion is to define the flattened opaque member as something like

Map<RelativePath, OpaqueDirectory | OpaqueDirectoryWithDesignatedSubdir]>

smera: Or maybe OpaqueSubDirectory, which you already defined, is good enough instead of this 'OpaqueDirectoryWithDesignatedSubdir' that I was suggesting?

---
In reply to: [330193191](https://github.com/microsoft/BuildXL/pull/966#discussion_r330193191) [](ancestors = 330193191)

almiliMSFT: How can you then easily distinguish between `OpaqueDirectory` and `OpaqueDirectoryWithDesignatedSubdir`?   I'm not sure if `typeof` returns something other than `object` when invoked against an OpaqueDirectory; if it does, then I like this proposal

Public/Sdk/Public/Deployment/definition.dsc
smera: Yeah, saw that in a use of this. But I don't love the change, it looks like the type is being special cased for this particular problem, and now all flattened opaque will have a designated subdirectory to deploy and r`.` has to be used is all cases where this subdir thing does not apply.
I'd rather see a new type defined that represents an opaque subdirectory and make this map contain a union type of opaques or opaque subdirs

---
In reply to: [330191168](https://github.com/microsoft/BuildXL/pull/966#discussion_r330191168) [](ancestors = 330191168)

Public/Sdk/Public/Deployment/operations.dsc
almiliMSFT: Now I think the type should just be `Map<RelativePath, OpaqueSubDirectory>`. 

I kind of like using a single type (like `OpaqueSubDirectory`) with both "parent opaque directory" and "relative path" properties than using a union type (mostly because it avoids any kind of checking/pattern matching to determine if the object is of type OpaqueDirectory or OpaqueSubDirectory)

smera: yes, I think typeof won't work. But you can use a known field to distinguish them (in a function where you would use a type predicate to narrow down the result as well)

---
In reply to: [330194376](https://github.com/microsoft/BuildXL/pull/966#discussion_r330194376) [](ancestors = 330194376)

almiliMSFT: Alright, let's give union type a try here

Public/Src/IDE/VsCode/BuildXL.IDE.VsCode.dsc
```diff
@@ -1,22 +1,46 @@
...
 import * as Deployment from "Sdk.Deployment";
 import * as Transformers from "Sdk.Transformers";
 import * as DetoursServices from "BuildXL.Sandbox.Windows";
 import * as Branding from "BuildXL.Branding";
 import * as VSIntegration from "BuildXL.Ide.VsIntegration";
+import {Node, Npm} from "Sdk.NodeJs";
 
-namespace LanguageService.Server {
+namespace VsCode.Client {
+    //A new namespace with empty qualifier space to ensure the values inside are evaluated only once
```
smera: >/A [](start = 5, length = 2)

// A new

Public/Src/IDE/VsCode/updateVsixTemplate.cmd
```diff
@@ -1,23 +0,0 @@
-@echo off
```
smera: can't we now remove everything under pluginTemplate as well?

Public/Sdk/Public/Deployment/operations.dsc
almiliMSFT: What do you think about using only a non-union type (like `OpaqueSubDirectory`) where the `subDirectory` property is optional?

In that case, one would here write:
```typescript
                    flattenedOpaques: result.flattenedOpaques.add(targetFolder, {opaque: <OpaqueDirectory>staticDirectory}),
 
```

Public/Src/IDE/VsCode/BuildXL.IDE.VsCode.dsc
```diff
@@ -1,22 +1,46 @@
...
+    const clientCopy: OpaqueDirectory = Deployment.copyDirectory(clientSealDir.root, Context.getNewOutputDirectory("client-copy"), clientSealDir);
+
+    @public
+    export const installRootDir: OpaqueDirectory = Npm.npmInstall(clientCopy);
 
+    @@public
+    export const compileOutDir: OpaqueDirectory = Node.tscCompile(clientCopy.root, clientCopy, installRootDir);
+
+    @@public
+    export const deployedNpmPackageLockFile = Deployment.copyFileFromOpaqueDirectory(
```
smera: >deployedNpmPackageLockFile [](start = 17, length = 26)

Don't completely get what this is doing, but do we actually need a special mount for it? Looking at the list of mounts we have defined for our selfhost, this one now stands out as a rare snowflake.

Public/Sdk/Public/Deployment/operations.dsc
smera: yes, that also works for me

---
In reply to: [330202985](https://github.com/microsoft/BuildXL/pull/966#discussion_r330202985) [](ancestors = 330202985)

Public/Src/IDE/VsCode/updateVsixTemplate.cmd
almiliMSFT: yes, we can now delete `pluginTemplate/extension`.  Doing that would make this PR have over a 1000 modified files though.  So after this PR gets approved, we can delete that folder.

Public/Src/IDE/VsCode/BuildXL.IDE.VsCode.dsc
almiliMSFT: We need to check this file into our repo. 

Where do you suggest we place this file?  All other component governance manifest files are under the `cg` folder.  In order to be able to write to that folder from a pip, I thought we had to create a mount for it.

smera: Maybe we can copy this package-lock file into an output location, and then use the copied version of it? The fact that we have a writeable place, that is not under Out, and where we also have source files doesn't look super clean to me. At least for now we have always kept outputs and sources well separated.

---
In reply to: [330205562](https://github.com/microsoft/BuildXL/pull/966#discussion_r330205562) [](ancestors = 330205562)

almiliMSFT: This writable mount is defined for `cg/npm` under which there are no source files.

There is nothing we need to use this `package-lock.json` file for; all we need to do is to check it into our repo.  So the question is where it should live.  

Even though the `Out` folder is git ignored, this lock file could live somewhere under `Out` and we could force add it to git.  That, however, seems like a worse solution to me: I don't think any of us would intuitively expect to find any files under `Out` checked into our repo. 

Having a small writable folder with no source files in it (like `cg/npm`) under a source folder (like `cg`) does not seem too bad to me.

smera: My proposal was to check in the file where it is in this PR (under cg/npm), but copy the file during the build under some out location, and use the copy of it. In that way you won't need to define any extra mount, and all output files will still be under Out/

---
In reply to: [330240761](https://github.com/microsoft/BuildXL/pull/966#discussion_r330240761) [](ancestors = 330240761)

Public/Sdk/Experimental/NodeJs/node.dsc
```diff
@@ -103,4 +103,25 @@ namespace Node {
...
+        const arguments: Argument[] = [
+            Cmd.argument(Artifact.none(f`${workingDirectory}/node_modules/typescript/lib/tsc.js`)),
+            Cmd.argument("-p"),
+            Cmd.argument("."),
+        ];
+
+        const result = Node.run({
+            arguments: arguments,
+            workingDirectory: workingDirectory,
+            dependencies : dependencies,
```
smera: >s  [](start = 23, length = 2)

nit: extra space

smera:
small request, could we please start resolving the comments that were already addressed? It is hard to give a glance at the general 'state' of the PR with so many iterations and active comments.

Public/Src/IDE/VsCode/updateVsixTemplate.cmd
rijulluman-msft: We will delete this folder in a separate pull request

Public/Sdk/Experimental/NodeJs/node.dsc
rijulluman-msft:
```suggestion
            dependencies: dependencies,
```

Public/Sdk/Public/Deployment/filesystem.dsc
```diff
@@ -16,6 +16,29 @@ export interface OnDiskDeployment {
...
+    /** Optional opaque directories robocopied/rsynced into this deployment */
+    targetOpaques?: OpaqueDirectory[];
+}
+
+@@public
+export interface OpaqueSubDirectory {
+    /** Parent opaque directory */
+    opaque: OpaqueDirectory,
+
+    /** A path relative to that opaque directory designating a subdirectory to be added to the deployment */
```
smera: >A path relative to that op [](start = 8, length = 26)

I'd say 'An optional path to...' and 'If not specified, the whole opaque directory will be added to the deployment'

PR #968, opened 2019-10-01
Tracing improvements

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Library/Utils/ResourcePool.cs
```diff
@@ -181,10 +183,10 @@ public async Task<ResourceWrapper<TObject>> CreateAsync(TKey key)
...
                     await EnsureCapacityAsync();
 
-                    returnWrapper = _resourceDict.GetOrAdd(key, k =>
-                    {
-                        return new ResourceWrapper<TObject>(() => _resourceFactory(k));
-                    });
+                    returnWrapper = _resourceDict.GetOrAdd(
+                        key,
+                        (k, resourceFactory) => { return new ResourceWrapper<TObject>(() => resourceFactory(k)); },
+                        _resourceFactory);
```
JuanCarlosGI: Why pass in the factory instead of just calling it? To avoid the capture?

SergeyTeplyakov: Yep. To avoid closure allocation on every call.

---
In reply to: [330168861](https://github.com/microsoft/BuildXL/pull/968#discussion_r330168861) [](ancestors = 330168861)

PR #972, opened 2019-10-01
More comments and rename the option to disable reconciliation

SergeyTeplyakov:


Public/Src/Cache/DistributedCache.Host/Configuration/DistributedContentSettings.cs
```diff
@@ -368,8 +368,12 @@ public int[] RetryIntervalForCopiesMs
         [DataMember]
         public bool IsRedisGarbageCollectionEnabled { get; set; } = false;
 
+        /// <summary>
+        /// Disabling reconciliation is an unsafe option that can cause builds to fail because the machine's state can be off compared to the LLS's state.
+        /// Please do not set this property for long period of time. 
+        /// </summary>
         [DataMember]
-        public bool? IsReconciliationEnabled { get; set; }
+        public bool? Unsafe_DisableReconciliation { get; set; }
```
ljcollins25: >bool? [](start = 15, length = 5)

I think we can just make this a normal bool instead of nullable.

PR #973, opened 2019-10-01
Wrong Pip Description fix for Hashing source failure 

rijulluman-msft:
[AB#1586413](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1586413)

### DIFF
--- a/Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
+++ b/Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
@@ -352,7 +352,7 @@ namespace BuildXL.Scheduler.Artifacts
                 // Get inputs
                 PopulateDependencies(pip, state.PipArtifacts, includeLazyInputs: true, onlySourceFiles: true);
 
-                var maybeInputsHashed = await TryHashFileArtifactsAsync(state, operationContext, pip.ProcessAllowsUndeclaredSourceReads);
+                var maybeInputsHashed = await TryHashFileArtifactsAsync(state, operationContext, pip.ProcessAllowsUndeclaredSourceReads, pip.GetDescription(Context));
 
                 if (!maybeInputsHashed.Succeeded)
                 {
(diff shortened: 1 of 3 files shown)


cg/nuget/cgmanifest.json
```diff
@@ -50,7 +50,7 @@
       "Component": {
         "Type": "NuGet",
         "NuGet": {
-          "Name": "Bond.Core.CSharp",
+          "Name": "Bond.CSharp",
```
almiliMSFT: Presumably you used your dev build to generate this file, which is why it's different from what's checked in.

To avoid "Primary BuildXL PR Validation" failing, use the LKG version to build (which should generate the same content as the one currently checked in) or simply revert this file.

Public/Src/Engine/Scheduler/Artifacts/FileContentManager.cs
```diff
@@ -1327,7 +1327,11 @@ private FileOrDirectoryArtifact GetDeclaredArtifact(FileArtifact file)
             return file;
         }
 
-        private async Task<Possible<Unit>> TryHashFileArtifactsAsync(PipArtifactsState state, OperationContext operationContext, bool allowUndeclaredSourceReads)
+        private async Task<Possible<Unit>> TryHashFileArtifactsAsync(
+            PipArtifactsState state, 
+            OperationContext operationContext, 
+            bool allowUndeclaredSourceReads, 
+            string pipDescription = null)
```
almiliMSFT: Do you need to pass `pipDescription` explicitly here?  There is `state.PipInfo.Description` which seems to be exactly what you want.

almiliMSFT: Now I see that `state.PipInfo` might not even be set at this point (mutable fields...)

rijulluman-msft: Yes, I tried it out first, but it was null in most cases

PR #976, opened 2019-10-01
Override ReportQueueProcessTimeout for some tests to make them faster

almiliMSFT:


Public/Src/Engine/UnitTests/Processes/SandboxedProcessMacTest.cs
```diff
@@ -72,6 +72,7 @@ private class ReportInstruction
         public async Task CheckProcessTreeTimoutOnReportQueueStarvationAsync()
         {
             var processInfo = CreateProcessInfoWithSandboxConnection(Operation.Echo("hi"));
+            processInfo.ReportQueueProcessTimeoutForTests = TimeSpan.FromSeconds(1);
```
almiliMSFT: this test would previously run more than 100s

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -400,6 +400,11 @@ public byte[] GetUnicodeEnvironmentBlock()
         /// </summary>
         public string Provenance => $"[Pip{PipSemiStableHash:X16} -- {PipDescription}] ";
 
+        /// <summary>
+        /// Overrides <see cref="SandboxedProcessMac.ReportQueueProcessTimeout"/> when running tests.
```
almiliMSFT:
```suggestion
        /// Overrides <see cref="SandboxedProcessMac.ReportQueueProcessTimeout"/> when running tests
```

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -64,7 +64,7 @@ public sealed class SandboxedProcessInfo
         /// <summary>
         /// A macOS kernel extension connection.
         /// </summary>
-        public ISandboxConnection SandboxConnection;
+        public ISandboxConnection SandboxConnection { get; private set; }
```
almiliMSFT:
```suggestion
        public ISandboxConnection SandboxConnection;
```

PR #977, opened 2019-10-01
Support source seal directory for affected input files

narasamdya:
Besides supporting source seal directory, this change also cleans up previous implementation and tests.

### DIFF
--- a/Public/Src/Engine/Scheduler/ChangeAffectedOutput/SourceChangeAffectedContents.cs
+++ b/Public/Src/Engine/Scheduler/ChangeAffectedOutput/SourceChangeAffectedContents.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using BuildXL.Pips.Operations;
@@ -16,7 +17,8 @@ namespace BuildXL.Scheduler.ChangeAffectedOutput
     /// </summary>
     public sealed class SourceChangeAffectedContents
     {
-        private readonly PathTable m_pathTable;
+        private PathTable PathTable => m_fileContentManager.Host.Context.PathTable;
+
         private readonly FileContentManager m_fileContentManager;
 
         /// <summary>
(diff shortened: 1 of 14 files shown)


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -5482,6 +5484,20 @@ SealDirectoryKind IFileContentManagerHost.GetSealDirectoryKind(DirectoryArtifact
             return GetSealDirectoryKind(directory);
         }
 
+        [SuppressMessage("Microsoft.Design", "CA1033:InterfaceMethodsShouldBeCallableByChildTypes")]
```
norahuang: Why implement this function here but not in FileConternManager.cs?

narasamdya: Because `FileContentManager` doesn't know if a directory is a source sealed directory; only scheduler knows, and the functionality has been implemented as `IsSourceSealedDirectory` of `IPipExecutionEnvironment`.

---
In reply to: [330658466](https://github.com/microsoft/BuildXL/pull/977#discussion_r330658466) [](ancestors = 330658466)

Public/Src/Engine/Scheduler/ChangeAffectedOutput/SourceChangeAffectedContents.cs
```diff
@@ -77,6 +90,14 @@ public IReadOnlyCollection<AbsolutePath> GetChangeAffectedInputs(Process process
                         changeAffectedInputs.Add(file.Path);
                     }
                 }
+
+                if (m_fileContentManager.Host.TryGetSourceSealDirectory(directory, out var sourceSealWithPatterns))
```
norahuang: I totally understand that this makes the algorithm looks more completed and would benefit other possible features in the future. 
But for QTest itself, it won't take the source code as input and instrument the source files. This part is not necessary for the change list code coverage feature.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ChangeAffectedInputTests.cs
```diff
@@ -55,19 +60,16 @@ public void DirectAffectedFileInputTest()
...
-            pipBuilderB.SetEnvironmentVariable(StringId.Create(Context.StringTable, "[Sdk.BuildXL]qCodeCoverageEnumType"), "DynamicCodeCov");
-            pipBuilderB.SetChangeAffectedInputListWrittenFilePath(changeAffectedWrittenFile);
+            pipBuilderB.SetChangeAffectedInputListWrittenFile(changeAffectedWrittenFile);
             SchedulePipBuilder(pipBuilderB);
 
-
             var inputChangesFile = CreateOutputFileArtifact();
             File.WriteAllText(ArtifactToString(inputChangesFile), ArtifactToString(aInput));
             Configuration.Schedule.InputChanges = inputChangesFile.Path;
-            Environment.SetEnvironmentVariable("[Sdk.BuildXL]qCodeCoverageEnumType", "DynamicCodeCov");
```
norahuang: Why removing all the enviroment reset of [Sdk.BuildXL]qCodeCoverageEnumType? They are here to test the cache behavior.

Public/Src/Engine/Scheduler/ChangeAffectedOutput/SourceChangeAffectedContents.cs
narasamdya: This feature is not for QTest only. It happens that only QTest benefits from this feature. Thus, this is for feature completeness. 

---
In reply to: [330716599](https://github.com/microsoft/BuildXL/pull/977#discussion_r330716599) [](ancestors = 330716599)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ChangeAffectedInputTests.cs
narasamdya: They are not needed. You removed the env var from the SDK. 

---
In reply to: [330721429](https://github.com/microsoft/BuildXL/pull/977#discussion_r330721429) [](ancestors = 330721429)

PR #982, opened 2019-10-02
Remove legacy quota keeper.

SergeyTeplyakov:
This PR removes the old quota keeper implementation and the flags associated with it.


JuanCarlosGI:
Glad to see all of this go! :)

PR #983, opened 2019-10-02
Fix BandwidthChecker

JuanCarlosGI:
BandwidthChecker was sometimes failing because it was trying to access the position of the disposed destination stream for calculating the bandwidth.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
+++ b/Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
@@ -83,7 +83,7 @@ namespace BuildXL.Cache.ContentStore.Utils
                     if (copyCompleted)
                     {
                         var result = await copyTask;
-                        var bytesCopied = result.Size ?? previousPosition - startPosition;
+                        var bytesCopied = result.Size ?? (previousPosition - startPosition);
 
                         return (result, bytesCopied);
                     }
--- a/Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
+++ b/Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
@@ -45,14 +45,12 @@ namespace BuildXL.Cache.ContentStore.Utils
         {
             if (_historicalBandwidthLimitSource != null)
             {
-                var startPosition = tryGetPosition(destinationStream, out var pos) ? pos : 0;
                 var timer = Stopwatch.StartNew();
-                var (result, lastPosition) = await impl();
+                var (result, bytesCopied) = await impl();
                 timer.Stop();
 
                 // Bandwidth checker expects speed in MiB/s, so convert it.
-                var bytesCopied = lastPosition - startPosition;
-                var speed = bytesCopied / timer.Elapsed.TotalSeconds / (1024 * 1024);
+                var speed = bytesCopied / timer.Elapsed.TotalSeconds / BytesInMb;
                 _historicalBandwidthLimitSource.AddBandwidthRecord(speed);
 
                 return result;
@@ -62,15 +60,15 @@ namespace BuildXL.Cache.ContentStore.Utils
                 return (await impl()).result;
             }
 
-            async Task<(CopyFileResult result, long lastPosition)> impl()
+            async Task<(CopyFileResult result, long bytesCopied)> impl()
             {
                 // This method should not fail with exceptions because the resulting task may be left unobserved causing an application to crash
                 // (given that the app is configured to fail on unobserved task exceptions).
                 var minimumSpeedInMbPerSec = _bandwidthLimitSource.GetMinimumSpeedInMbPerSec() * _config.BandwidthLimitMultiplier;
                 minimumSpeedInMbPerSec = Math.Min(minimumSpeedInMbPerSec, _config.MaxBandwidthLimit);
 
-                long firstPosition = tryGetPosition(destinationStream, out var pos) ? pos : 0;
-                long previousPosition = firstPosition;
+                var startPosition = tryGetPosition(destinationStream, out var pos) ? pos : 0;
+                long previousPosition = startPosition;
                 var copyCompleted = false;
                 using var copyCancellation = CancellationTokenSource.CreateLinkedTokenSource(context.Token);
                 var copyTask = copyTaskFactory(copyCancellation.Token);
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -45,93 +45,78 @@ public async Task<CopyFileResult> CheckBandwidthAtIntervalAsync(OperationContext
...
+
+                    copyCompleted = firstCompletedTask == copyTask;
+                    if (copyCompleted)
                     {
-                        // Wait some time for bytes to be copied
-                        var firstCompletedTask = await Task.WhenAny(copyTask,
-                            Task.Delay(_config.BandwidthCheckInterval, context.Token));
+                        var result = await copyTask;
+                        var lastPosition = result.Size.HasValue
+                            ? firstPosition + result.Size.Value
```
SergeyTeplyakov: should it be 'previousPosition+result.Size.Value'?

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -45,93 +45,78 @@ public async Task<CopyFileResult> CheckBandwidthAtIntervalAsync(OperationContext
...
-            async Task<CopyFileResult> impl()
+            async Task<(CopyFileResult result, long lastPosition)> impl()
             {
                 // This method should not fail with exceptions because the resulting task may be left unobserved causing an application to crash
                 // (given that the app is configured to fail on unobserved task exceptions).
                 var minimumSpeedInMbPerSec = _bandwidthLimitSource.GetMinimumSpeedInMbPerSec() * _config.BandwidthLimitMultiplier;
                 minimumSpeedInMbPerSec = Math.Min(minimumSpeedInMbPerSec, _config.MaxBandwidthLimit);
 
-                long previousPosition = 0;
+                long firstPosition = tryGetPosition(destinationStream, out var pos) ? pos : 0;
```
SergeyTeplyakov: nit: initialPosition, originalPosition?

SergeyTeplyakov: Does it make sense to obtain the start position in the enclosing method not inside the if, but on top of it and the impl can just reference it?

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
SergeyTeplyakov: Ignore this.

---
In reply to: [330672652](https://github.com/microsoft/BuildXL/pull/983#discussion_r330672652) [](ancestors = 330672652)

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -45,93 +45,78 @@ public async Task<CopyFileResult> CheckBandwidthAtIntervalAsync(OperationContext
...
+                return result;
             }
             else
             {
-                return await impl();
+                return (await impl()).result;
             }
 
-            async Task<CopyFileResult> impl()
+            async Task<(CopyFileResult result, long lastPosition)> impl()
```
SergeyTeplyakov: just an idea: does it make more sense maybe to return result and bytescopied? Because it seems that this is what the impl's caller actually needs...

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -45,93 +45,78 @@ public async Task<CopyFileResult> CheckBandwidthAtIntervalAsync(OperationContext
...
-                {
-                    return await impl();
-                }
-                finally
-                {
-                    timer.Stop();
-                    var endPosition = destinationStream.Position;
+                // Bandwidth checker expects speed in MiB/s, so convert it.
+                var bytesCopied = lastPosition - startPosition;
+                var speed = bytesCopied / timer.Elapsed.TotalSeconds / (1024 * 1024);
```
SergeyTeplyakov: A minor thing, but what do you think about this approach:

1. We can change the result of impl to return bytesCopied and
2. Add an extension method to compute mbps:

in this case the code will looks like this:

var (result, bytesCopied) = await impl;
var spped = bytesCopied.ToMbps(timer.Elapsed);
_historicalBandwidthLimitSource.AddBandwidthRecord(speed);

JuanCarlosGI: The extension method seems a little overkill to me. Otherwise, agree.

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -45,93 +45,76 @@ public async Task<CopyFileResult> CheckBandwidthAtIntervalAsync(OperationContext
...
+                        Task.Delay(_config.BandwidthCheckInterval, context.Token));
+
+                    copyCompleted = firstCompletedTask == copyTask;
+                    if (copyCompleted)
                     {
-                        // Wait some time for bytes to be copied
-                        var firstCompletedTask = await Task.WhenAny(copyTask,
-                            Task.Delay(_config.BandwidthCheckInterval, context.Token));
+                        var result = await copyTask;
+                        var bytesCopied = result.Size ?? previousPosition - startPosition;
```
SergeyTeplyakov: one very minor nit: could you please add braces?
because I have no idea if this code is correct.

Public/Src/Cache/ContentStore/Library/Utils/BandwidthChecker.cs
```diff
@@ -45,93 +45,76 @@ public async Task<CopyFileResult> CheckBandwidthAtIntervalAsync(OperationContext
...
-                                return new CopyFileResult(CopyFileResult.ResultCode.CopyBandwidthTimeoutError, $"Average speed was {currentSpeed}MiB/s - under {minimumSpeedInMbPerSec}MiB/s requirement. Aborting copy with {position} copied");
-                            }
-
-                            previousPosition = position;
-                        }
-                        catch (ObjectDisposedException)
-                        {
-                            // If the check task races with the copy completing, it might attempt to check the position of a disposed stream.
-                            // Don't bother logging because the copy completed successfully.
+                            var result = new CopyFileResult(CopyFileResult.ResultCode.CopyBandwidthTimeoutError, $"Average speed was {currentSpeed}MiB/s - under {minimumSpeedInMbPerSec}MiB/s requirement. Aborting copy with {position} copied");
```
SergeyTeplyakov: and very nit again: I think we should change the message and actually use 'bytesCopied' there. Because position is the position and it may be not equal to 'bytes copied'.

PR #987, opened 2019-10-02
Fix macOS cache package

JuanCarlosGI:
This change will make it so that we include all the runtime requirements for BuildXL.Native in the private package that we send to the AzureDevOps repo.


Public/Src/Deployment/privatePackages.dsc
```diff
@@ -127,7 +127,7 @@ namespace PrivatePackages {
                         importFrom("BuildXL.Utilities").withQualifier(osxx64Qualifier).dll.runtime,
                         importFrom("BuildXL.Utilities").withQualifier(osxx64Qualifier).Collections.dll.runtime,
                         importFrom("BuildXL.Utilities").withQualifier(osxx64Qualifier).Configuration.dll.runtime,
-                        importFrom("BuildXL.Utilities").withQualifier(osxx64Qualifier).Native.dll.runtime,
+                        importFrom("BuildXL.Utilities").withQualifier(osxx64Qualifier).Native.dll,
```
almiliMSFT: you should probably remove `.runtime` in other lines too

PR #991, opened 2019-10-02
Configure GrpcPoolSize

kewilli:
Allow GrpcPoolSize to be configured.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
+++ b/Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
@@ -268,7 +268,7 @@ namespace BuildXL.Cache.ContentStore.Service
 
                     await LoadHibernatedSessionsAsync(context);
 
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
```diff
@@ -290,10 +290,10 @@ async Task<BoolResult> tryStartupCoreAsync()
             }
         }
 
-        private void InitializeAndStartGrpcServer(int grpcPort, ServerServiceDefinition[] definitions, int requestCallTokensPerCompletionQueue)
+        private void InitializeAndStartGrpcServer(int grpcPort, ServerServiceDefinition[] definitions, int requestCallTokensPerCompletionQueue, int grpcPoolSize = 70)
```
SergeyTeplyakov: nit: the method is called only once, so it is kind of makes no sense to specify the default here.

PR #992, opened 2019-10-03
Improve time to first pip on workers by having workers download graph while scheduler is initializing/filtering.

ljcollins25:
Improve time to first pip on workers by having workers download graph while scheduler is initializing/filtering. Also improves, time to first pip on master by making historic metadata and historic perf data retrieval lazy and async. For Office Mularchy builds, this results in workers being ready to execute about the same time as the master.  Specifically for master (13.5 min to 11 min) and for worker (17 min to 11 min).

### DIFF
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -308,9 +308,13 @@ namespace BuildXL.Scheduler
                 worker.SyncResourceMappings(LocalWorker);
             }
 
-            m_workersStatusOperation = OperationTracker.StartOperation(Worker.WorkerStatusParentOperationKind, m_executePhaseLoggingContext);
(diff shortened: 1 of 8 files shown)


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -4922,6 +4926,11 @@ private void InitSchedulerRuntimeState(LoggingContext loggingContext, SchedulerS
             {
                 Contract.Requires(loggingContext != null);
 
+                // Start loading data for pip two phase cache and running time table. No need to wait since any operation against the components
+                // will block until the required component is ready.
+                m_pipTwoPhaseCache.StartLoading(waitForCompletion: false);
+                m_runningTimeTableTask.Start();
+
```
semihokur: In the past, these two were getting loaded earlier. I did not understand the reason why we postpone them.

PR #997, opened 2019-10-03
Removed Untracked Executables from Weak Fingerprint

rijulluman-msft:
[AB#1584112](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1584112)


Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -244,7 +246,15 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, SealDire
         /// </summary>
         protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process process)
         {
-            AddFileDependency(fingerprinter, "Executable", process.Executable);
+            if (process.UntrackedPaths.Contains(process.Executable.Path))       // Check if executable is untracked
+            {
+                // Donot record executable in weak fingerprint if executable is untracked
```
almiliMSFT:
```suggestion
                // Don't record executable in weak fingerprint if executable is untracked
```

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -244,7 +246,15 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, SealDire
         /// </summary>
         protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process process)
         {
-            AddFileDependency(fingerprinter, "Executable", process.Executable);
+            if (process.UntrackedPaths.Contains(process.Executable.Path))       // Check if executable is untracked
```
almiliMSFT:
```suggestion
            if (process.UntrackedPaths.Contains(process.Executable.Path) ||
                process.UntrackedScopes.Any(scope => process.Executable.Path.IsWithin(m_pathTable, scope)))
```

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -244,7 +246,15 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, SealDire
         /// </summary>
         protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process process)
         {
-            AddFileDependency(fingerprinter, "Executable", process.Executable);
+            if (process.UntrackedPaths.Contains(process.Executable.Path))       // Check if executable is untracked
+            {
+                // Donot record executable in weak fingerprint if executable is untracked
+                fingerprinter.Add("Executable", s_untrackedExecutable);
+            }
+            else {
```
almiliMSFT:
```suggestion
            else
            {
```

almiliMSFT:
You might need to increment the fingerprint version, see [`PipFingerprintingVersion.TwoPhaseV2`](https://github.com/microsoft/BuildXL/blob/3afebc155112f155db00815ca30df639748fec0f/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprintingVersion.cs#L40)

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -244,7 +246,17 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, SealDire
         /// </summary>
         protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process process)
         {
-            AddFileDependency(fingerprinter, "Executable", process.Executable);
+            if (process.UntrackedPaths.Contains(process.Executable.Path) ||
```
almiliMSFT: Add some unit tests as well. See `PipFingerprinterTests`

almiliMSFT: It would also be good to add an integration test.  See [`ChangeAffectedInputTests`](https://github.com/microsoft/BuildXL/blob/master/Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/ChangeAffectedInputTests.cs); add a new test method at the end of that class which: 
  - sets up a simple process pip (e.g., one that has a single `WriteFile` operation)
  - runs the scheduler and asserts the pip was executed (i.e., was a cache miss) and succeeded
  - changes the executable of that process pip
  - runs the scheduler again and asserts that the pip was was a cache hit
  - reverts the change it made to the executable of that process pip

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -380,6 +380,100 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            var executable = FileArtifact.CreateSourceFile(AbsolutePath.Create(pathTable, X(untrackedPathText)));
+
+            PipDataBuilder pdb = new PipDataBuilder(m_context.PathTable.StringTable);
+            string appDataSubdirName = "appdataSubDir";
+            pdb.Add(AbsolutePath.Create(pathTable, Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), appDataSubdirName)));
+            EnvironmentVariable envVar = new EnvironmentVariable(m_context.StringTable.AddString("testVar"), pdb.ToPipData(" ", PipDataFragmentEscaping.NoEscaping));
+
+            var dependencies = new HashSet<FileArtifact>() { executable };
+            var process = GetDefaultProcessBuilder(pathTable, executable)
+                .WithEnvironmentVariables(new EnvironmentVariable[] { envVar })
```
almiliMSFT: Adding env variables is non-essential for this test, so I suggest removing this.

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -380,6 +380,100 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+        [Fact]
+        public void TestUntrackedPathStaysOutOfFingerprints()
+        {
+            string untrackedPathText = "/x/pkgs/tool.exe";
+            var pathTable = m_context.PathTable;
+            var executable = FileArtifact.CreateSourceFile(AbsolutePath.Create(pathTable, X(untrackedPathText)));
+
+            PipDataBuilder pdb = new PipDataBuilder(m_context.PathTable.StringTable);
+            string appDataSubdirName = "appdataSubDir";
+            pdb.Add(AbsolutePath.Create(pathTable, Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), appDataSubdirName)));
```
almiliMSFT: unused --> remove

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -380,6 +380,100 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            EnvironmentVariable envVar = new EnvironmentVariable(m_context.StringTable.AddString("testVar"), pdb.ToPipData(" ", PipDataFragmentEscaping.NoEscaping));
+
+            var dependencies = new HashSet<FileArtifact>() { executable };
+            var process = GetDefaultProcessBuilder(pathTable, executable)
+                .WithEnvironmentVariables(new EnvironmentVariable[] { envVar })
+                .WithDependencies(dependencies)
+                .WithUntrackedPaths(new AbsolutePath[] { AbsolutePath.Create(pathTable, X(untrackedPathText)) })
+                //.WithUntrackedScopes(new AbsolutePath[] { AbsolutePath.Create(pathTable, OperatingSystemHelper.IsUnixOS ? "/tmp" : Environment.GetFolderPath(Environment.SpecialFolder.History)) })
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -380,6 +380,100 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            var fingerprinter = new PipContentFingerprinter(
+                m_context.PathTable,
+                GetContentHashLookup(executable),
+                ExtraFingerprintSalts.Default(),
+                pathExpander: expander)
+            {
+                FingerprintTextEnabled = true
+            };
+
+            var baselineFingerprint = fingerprinter.ComputeWeakFingerprint(process, out string fingerprintText);
```
almiliMSFT:
```suggestion
            fingerprinter.ComputeWeakFingerprint(process, out string fingerprintText);
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -380,6 +380,100 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+
+            // Validate the JSON fingerprinter does the same
+            var json = JsonFingerprinter.CreateJsonString(writer =>
+            {
+                fingerprinter.AddWeakFingerprint(writer, process);
+            },
+            pathTable: m_context.PathTable).ToUpperInvariant();
+
+            var fence = OperatingSystemHelper.IsUnixOS ? "\"" : "";
+            XAssert.IsFalse(json.Contains($"{fence}{executable.Path.ToString().ToUpperInvariant()}{fence}"));
```
almiliMSFT: why add `fence` around the path?

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -380,6 +380,100 @@ public void TestUserProfileStaysOutOfFingerprints()
             XAssert.IsTrue(json.Contains(Path.GetFileName(Environment.GetFolderPath(Environment.SpecialFolder.InternetCache)).ToUpperInvariant()));
         }
 
+        [Fact]
+        public void TestUntrackedPathStaysOutOfFingerprints()
```
almiliMSFT: This name is misleading, because untracked paths still remain part of the fingerprint. 
```suggestion
        public void TestUntrackedExecutablePathStaysOutOfFingerprints()
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -380,6 +380,100 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+                fingerprinter.AddWeakFingerprint(writer, process);
+            },
+            pathTable: m_context.PathTable).ToUpperInvariant();
+
+            var fence = OperatingSystemHelper.IsUnixOS ? "\"" : "";
+            XAssert.IsFalse(json.Contains($"{fence}{executable.Path.ToString().ToUpperInvariant()}{fence}"));
+        }
+
+        [Fact]
+        public void TestUntrackedScopeStaysOutOfFingerprints()
```
almiliMSFT:
```suggestion
        public void TestExecutableInAnUntrackedScopeStaysOutOfFingerprints()
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -380,6 +380,100 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+                pathExpander: expander)
+            {
+                FingerprintTextEnabled = true
+            };
+
+            var baselineFingerprint = fingerprinter.ComputeWeakFingerprint(process, out string fingerprintText);
+            fingerprintText = fingerprintText.ToUpperInvariant();
+
+            // Make sure the executable path doesn't show up in the fingerprint
+            XAssert.IsFalse(fingerprintText.Contains(executable.Path.ToString().ToUpperInvariant()));
```
almiliMSFT: I don't know what this `fingerprintText` is supposed to contain in the first place; this text could be any kind of gibberish, so just checking that it doesn't contain executable.Path is not very convincing to me.  In fact, I just ran your test with `.WithUntrackedPaths(...)` commented out and the test still passed, meaning that indeed this is a bogus thing to check. 

In general, please change this Fact into a Theory and add a bool argument like `untrackExePath` then based on the value of that arg either (1) untrack the exe path and assert that the exe path IS NOT part of the fingerprint, or (2) don't untrack the exe path and assert that the exe path IS part of the fingerprint.

almiliMSFT: I see now that the reason why this is is because of `executable.Path.ToString()` which always returns some bogus string; instead, use `executable.Path.ToString(pathTable)`.

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
almiliMSFT: and this `json.Contains` check never succeeds because all the backslashes are escaped in `json`, i.e., `json` looks something like `..."X:\\PKGS\\TOOL.EXE"...` and `executable.Path.ToString(pathTable)` is just `"X:\PKGS\TOOL.EXE"`

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+
+            return contentFingerprint;
+        }
+
+        [Fact]
+        public void TestExecutableWhenTrackedStaysInsideFingerprints()
+        {
+            var fingerprint = GetWeakFingerPrintForExecutable("/x/pkgs/tool.exe", true);
+            var fingerprint2 = GetWeakFingerPrintForExecutable("/x/pkgs/tool2.exe", true);
+            XAssert.IsFalse(fingerprint.Equals(fingerprint2));
```
almiliMSFT: Prefer `AreNotEqual` to `IsFalse` because the former will produce better message when the assertion doesn't hold
```suggestion
            XAssert.AreNotEqual(fingerprint, fingerprint2);
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            var fingerprint2 = GetWeakFingerPrintForExecutable("/x/pkgs/tool2.exe", true);
+            XAssert.IsFalse(fingerprint.Equals(fingerprint2));    
+        }
+
+        [Fact]
+        public void TestExecutableWhenUntrackedStaysOutOfFingerprints()
+        {
+            var fingerprint = GetWeakFingerPrintForExecutable("/x/pkgs/tool.exe", false, true);
+            var fingerprint2 = GetWeakFingerPrintForExecutable("/x/pkgs/tool2.exe", false, true);
+            //XAssert.IsTrue(fingerprint.Equals(fingerprint2)); // Will uncomment once development is done
```
almiliMSFT:
```suggestion
            //XAssert.AreEqual(fingerprint, fingerprint2); // Will uncomment once development is done
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+        }
+
+        [Fact]
+        public void TestExecutableWhenInsideUntrackedScopeStaysOutOfFingerprints()
+        {
+            var fingerprint = GetWeakFingerPrintForExecutable("/x/pkgs/tool.exe", false, false, "/x/pkgs");
+            var fingerprint2 = GetWeakFingerPrintForExecutable("/x/pkgs/tool2.exe", false, false, "/x/pkgs");
+            //XAssert.IsTrue(fingerprint.Equals(fingerprint2)); // Will uncomment once development is done
```
almiliMSFT: ditto

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+        }
+
+        [Fact]
+        public void TestExecutableWhenInsideUntrackedParentScopeStaysOutOfFingerprints()
+        {
+            var fingerprint = GetWeakFingerPrintForExecutable("/x/pkgs/sub1/tool.exe", false, false, "/x/pkgs");
+            var fingerprint2 = GetWeakFingerPrintForExecutable("/x/pkgs/sub2/tool2.exe", false, false, "/x/pkgs");
+            //XAssert.IsTrue(fingerprint.Equals(fingerprint2)); // Will uncomment once development is done
```
almiliMSFT: ditto

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
             XAssert.IsTrue(json.Contains(Path.GetFileName(Environment.GetFolderPath(Environment.SpecialFolder.InternetCache)).ToUpperInvariant()));
         }
 
+        private ContentFingerprint GetWeakFingerPrintForExecutable(
+            string exectuablePath, 
+            bool assertValue,  // true if fingerprint should contain executable path, false if output should be untracked
+            bool untracked = false,
```
almiliMSFT: I suggest changing this argument to `string[] untrackedPaths = null`

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
             XAssert.IsTrue(json.Contains(Path.GetFileName(Environment.GetFolderPath(Environment.SpecialFolder.InternetCache)).ToUpperInvariant()));
         }
 
+        private ContentFingerprint GetWeakFingerPrintForExecutable(
+            string exectuablePath, 
+            bool assertValue,  // true if fingerprint should contain executable path, false if output should be untracked
+            bool untracked = false, 
+            string untrackedScopePath = null)
```
almiliMSFT:
```suggestion
            string[] untrackedScopes = null)
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
             XAssert.IsTrue(json.Contains(Path.GetFileName(Environment.GetFolderPath(Environment.SpecialFolder.InternetCache)).ToUpperInvariant()));
         }
 
+        private ContentFingerprint GetWeakFingerPrintForExecutable(
+            string exectuablePath, 
+            bool assertValue,  // true if fingerprint should contain executable path, false if output should be untracked
```
almiliMSFT: I suggest removing this argument and removing all assertions (i.e., calls to `XAssert` methods) from this method.

The name `GetWeakFingerPrintForExecutable` suggests that this method is pure, i.e., that it will just compute the fingerprint, return it, and not have any other side effects.  

That said, I don't think it's necessary to check if the fingerprint text contains the executable path or not.  The real functional test specification is what you have below, i.e., asserting that (1) when the executable changes and it is not untracked => the fingerprints change, and (2) when the executable changes and it is untracked => the fingerprints don't change.

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
             XAssert.IsTrue(json.Contains(Path.GetFileName(Environment.GetFolderPath(Environment.SpecialFolder.InternetCache)).ToUpperInvariant()));
         }
 
+        private ContentFingerprint GetWeakFingerPrintForExecutable(
```
almiliMSFT:
```suggestion
        private ContentFingerprint GetWeakFingerprint(
```
This method returns a fingerprint for the pip, not for the executable, hence the suggested change.

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,94 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+                .WithDependencies(dependencies);
+
+            if (untracked)
+            {
+                processBuilder = processBuilder.WithUntrackedPaths(new AbsolutePath[] { AbsolutePath.Create(pathTable, X(exectuablePath)) });
+            }
+
+            if (untrackedScopePath != null)
+            {
+                processBuilder = processBuilder.WithUntrackedScopes(new AbsolutePath[] { AbsolutePath.Create(pathTable, OperatingSystemHelper.IsUnixOS ? "/tmp" : Environment.GetFolderPath(Environment.SpecialFolder.History)) });
```
almiliMSFT: In addition to untracking `/tmp` you should untrack `untrackedScopePath` as well

narasamdya:
I'm not sure about the approach here.

Even with these changes, the executable is still added to the dependencies by [http://index/?leftProject=BuildXL.FrontEnd.Script&leftSymbol=oy2g2qk3lm7t&file=Ambients%5CTransformers%5CAmbientTransformer.Process.cs&line=750](http://index/?leftProject=BuildXL.FrontEnd.Script&leftSymbol=oy2g2qk3lm7t&file=Ambients%5CTransformers%5CAmbientTransformer.Process.cs&line=750)
The addition will make the executable tracked.

Here's the change that you have to make 
- Ensure that the executable is not added to the `InputFiles` of `cachedTool` on `ProcessCachedTool`.
- Ensure in `ProcessBuilder` (not in `PipFingerprinter`) that executable is not part of dependencies when it is under an untracked scope or it is an untracked path.
- Change `FingerprintingRole` from `Content` to `Semantic` (see [http://index/?query=Process&rightProject=BuildXL.Pips&file=Operations%5CProcess.cs&line=133](http://index/?query=Process&rightProject=BuildXL.Pips&file=Operations%5CProcess.cs&line=133))
- In `AddWeakFingerprint` instead of `AddFileDependency(fingerprinter, "Executable", process.Executable)`, do `fingerprinter.Add("Executable", process.Executable)`. This ensure that we have the executable path (not content) in the fingerprint.

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -370,6 +389,7 @@ protected virtual void AddFileDependency(ICollectionFingerprinter fingerprinter,
                 }
                 else
                 {
+                    // HASHPATH changes here
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -248,7 +250,19 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, SealDire
         /// </summary>
         protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process process)
         {
-            AddFileDependency(fingerprinter, "Executable", process.Executable);
+            bool untrackedExecutable = false;
```
almiliMSFT: Please don't reassign value to variable whenever possible
```suggestion
            bool untrackedExecutable = 
                process.UntrackedPaths.Contains(process.Executable.Path) ||
                process.UntrackedScopes.Any(scope => process.Executable.Path.IsWithin(m_pathTable, scope));
```

almiliMSFT: Please don't reassign values to variables unless really necessary
```suggestion
            bool untrackedExecutable =
                process.UntrackedPaths.Contains(process.Executable.Path) ||
                process.UntrackedScopes.Any(scope => process.Executable.Path.IsWithin(m_pathTable, scope));
```

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -248,7 +250,19 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, SealDire
         /// </summary>
         protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process process)
         {
-            AddFileDependency(fingerprinter, "Executable", process.Executable);
+            bool untrackedExecutable = false;
+            if (process.UntrackedPaths.Contains(process.Executable.Path) ||
+                process.UntrackedScopes.Any(scope => process.Executable.Path.IsWithin(m_pathTable, scope)))
+            {
+                // Don't record executable in weak fingerprint if executable is untracked
+                fingerprinter.Add("Executable", s_untrackedExecutable);
```
almiliMSFT: I like Iman's proposal to unconditionally do `fingerprinter.Add("Executable", process.Executable)` here.

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -260,7 +274,12 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process
             AddFileOutput(fingerprinter, "StandardError", process.StandardError);
             AddFileOutput(fingerprinter, "StandardOutput", process.StandardOutput);
 
-            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f), m_expandedPathFileArtifactComparer);            
+            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => {
+                if (!(untrackedExecutable && process.Executable.Path.Equals(f.Path)))    // Don't record executable in weak fingerprint if executable is untracked
+                {
+                    AddFileDependency(fp, f);
+                }
+            }, m_expandedPathFileArtifactComparer);
```
almiliMSFT: I think this is the right way to do it and not mess with ProcessBuilder.  @narasamdya, do you agree?

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
```diff
@@ -260,7 +274,12 @@ protected virtual void AddWeakFingerprint(IFingerprinter fingerprinter, Process
             AddFileOutput(fingerprinter, "StandardError", process.StandardError);
             AddFileOutput(fingerprinter, "StandardOutput", process.StandardOutput);
 
-            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => AddFileDependency(fp, f), m_expandedPathFileArtifactComparer);            
+            fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>("Dependencies", process.Dependencies, (fp, f) => {
```
almiliMSFT: I think it would be cleaner to write this as
```csharp
var relevantDependencies = untrackedExecutable
    ? process.Dependencies.Except(new[] { process.Executable })
    : process.Dependencies;
fingerprinter.AddOrderIndependentCollection<FileArtifact, ReadOnlyArray<FileArtifact>>(
    "Dependencies", 
    relevantDependencies, 
    (fp, f) => AddFileDependency(fp, f), m_expandedPathFileArtifactComparer); 
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -28,6 +28,7 @@
 using Xunit;
 using Xunit.Abstractions;
 using static BuildXL.Utilities.FormattableStringEx;
+using Newtonsoft.Json;
```
almiliMSFT: order using

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,104 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            var fingerprint2 = GenerateAndAssertWeakFingerPrint("/x/pkgs/tool2.exe", false, new string[] { "/x/pkgs/tool2.exe" });
+            XAssert.AreNotEqual(fingerprint, fingerprint2);
+        }
+
+        [Fact]
+        public void TestExecutableWhenInsideUntrackedScopeStaysOutOfFingerprints()
+        {
+            var fingerprint = GenerateAndAssertWeakFingerPrint("/x/pkgs/tool.exe", false, null, new string[] { "/x/pkgs" });
+            var fingerprint2 = GenerateAndAssertWeakFingerPrint("/x/pkgs/tool2.exe", false, null, new string[] { "/x/pkgs" });
+            XAssert.AreEqual(fingerprint, fingerprint2);
```
almiliMSFT: I don't know what's expected behavior here.  Since both tool.exe and tool2.exe are untracked, one could argue that the fingerprints should be the same; but on the other hand, since the paths are different, it's also a fair argument that the fingerprints shouldn't be the same. 

In any case, there needs to be a test that checks that the fingerprint stays the same when only the content of the executable changes (and not its path)

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
             XAssert.IsTrue(json.Contains(Path.GetFileName(Environment.GetFolderPath(Environment.SpecialFolder.InternetCache)).ToUpperInvariant()));
         }
 
+        [Fact]
+        public void TestModifiedExecutableWhenUntrackedStaysOutOfFingerprints()
+        {
+            var exectuablePath = "/x/pkgs/tool.exe";
```
almiliMSFT: 1. typo: `exectuable` -> `executable`
2. call `X()` here once instead of calling it everywhere else where this path is used.
```suggestion
            var executablePath = X("/x/pkgs/tool.exe");
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+
+            var contentFingerprint2 = fingerprinter2.ComputeWeakFingerprint(process, out string fingerprintText2);
+            fingerprintText2 = fingerprintText2.ToUpperInvariant();
+
+            XAssert.AreEqual(contentFingerprint, contentFingerprint2);
+            XAssert.AreEqual(fingerprintText, fingerprintText2);
+        }
+
+        [Fact]
+        public void TestModifiedExecutableInsideUnrackedScopeStaysOutOfFingerprints()
```
almiliMSFT: as suggested earlier, let's not copy-paste test methods and instead use a single `Theory` method

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
             XAssert.IsTrue(json.Contains(Path.GetFileName(Environment.GetFolderPath(Environment.SpecialFolder.InternetCache)).ToUpperInvariant()));
         }
 
+        [Fact]
+        public void TestModifiedExecutableWhenUntrackedStaysOutOfFingerprints()
```
almiliMSFT:
```suggestion
        public void TestContentHashesOfUntrackedPathsStayOutOfWeakFingerprint(bool untrackPath, bool untrackScope, int contentHash1, int contentHash2, bool expectedEqualityOutcome)
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
             XAssert.IsTrue(json.Contains(Path.GetFileName(Environment.GetFolderPath(Environment.SpecialFolder.InternetCache)).ToUpperInvariant()));
         }
 
+        [Fact]
```
almiliMSFT:
```suggestion
        [Theory]
        [MemberData(false, false, 1, 1, true)] // not untracked in any way + same content hashes => same fingerprint
        [MemberData(false, false, 1, 2, false)] // not untracked in any way + same different content hashes => different fingerprints
        [MemberData(false, false, 1, 2, false)] // not untracked in any way + different content hashes => different fingerprints
        [MemberData(true, false, 1, 2, true)] // untracked path + different content hashes => same fingerprint
        [MemberData(false, true, 1, 2, true)] // untracked scope + different content hashes => same fingerprint
        [MemberData(true, true, 1, 2, true)] // untracked as path and as scope + different content hashes => same fingerprint
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+        public void TestModifiedExecutableWhenUntrackedStaysOutOfFingerprints()
+        {
+            var exectuablePath = "/x/pkgs/tool.exe";
+            var pathTable = m_context.PathTable;
+            var executable = FileArtifact.CreateSourceFile(AbsolutePath.Create(pathTable, X(exectuablePath)));
+
+            var dependencies = new HashSet<FileArtifact>() { executable };
+            var process = GetDefaultProcessBuilder(pathTable, executable)
+                .WithDependencies(dependencies)
```
almiliMSFT:
```suggestion
                .WithDependencies(dependencies);
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+                .WithUntrackedPaths(new AbsolutePath[] { AbsolutePath.Create(pathTable, X(exectuablePath)) })
```
almiliMSFT:
```suggestion
            if (untrackPath)
            {
                pb = pb.WithUntrackedPaths(new[] { AbsolutePath.Create(pathTable, exectuablePath) });
            }
            if (untrackScope)
            {
                pb = pb.WithUntrackedScopes(new[] { AbsolutePath.Create(pathTable, Path.GetDirectoryName(executablePath)) });
            }
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+                .Build();
```
almiliMSFT:
```suggestion
            var process = pb.Build();
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+
+            MountPathExpander expander = new MountPathExpander(pathTable);
+
+            var fingerprinter1 = new PipContentFingerprinter(
+                m_context.PathTable,
+                GetDummyContentHash(5),
```
almiliMSFT:
```suggestion
                GetDummyContentHash(contentHash1),
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            {
+                FingerprintTextEnabled = true
+            };
+
+            var contentFingerprint = fingerprinter1.ComputeWeakFingerprint(process, out string fingerprintText);
+            fingerprintText = fingerprintText.ToUpperInvariant();
+
+            var fingerprinter2 = new PipContentFingerprinter(
+                m_context.PathTable,
+                GetDummyContentHash(6),
```
almiliMSFT:
```suggestion
                GetDummyContentHash(contentHash2),
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+                ExtraFingerprintSalts.Default(),
+                pathExpander: expander)
+            {
+                FingerprintTextEnabled = true
+            };
+
+            var contentFingerprint2 = fingerprinter2.ComputeWeakFingerprint(process, out string fingerprintText2);
+            fingerprintText2 = fingerprintText2.ToUpperInvariant();
+
+            XAssert.AreEqual(contentFingerprint, contentFingerprint2);
```
almiliMSFT:
```suggestion
            XAssert.AreEqual(expectedEqualityOutcome, contentFingerprint.Equals(contentFingerprint2));
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            XAssert.AreEqual(fingerprintText, fingerprintText2);
```
almiliMSFT:
```suggestion
            XAssert.AreEqual(expectedEqualityOutcome, fingerprintText.Equals(fingerprintText2), $"fp1: '{fingerprintText}'; fp2: '{fingerprintText2}'");
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -66,6 +67,16 @@ internal static PipFragmentRenderer.ContentHashLookup GetContentHashLookup(FileA
             };
         }
 
+        internal static PipFragmentRenderer.ContentHashLookup GetDummyContentHash(int randomNumber)
```
almiliMSFT: Don't copy-paste.  Extent the existing `GetContentHashLookup` function to accept this `randomNumber` argument (there you can set its default value to 1 so that you don't have to update any other callers of `GetContentHashLookup`

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +411,91 @@ public void TestUserProfileStaysOutOfFingerprints()
...
 
+        [Fact]
+        public void TestModifiedExecutableWhenUntrackedStaysOutOfFingerprints()
+        {
+            var exectuablePath = "/x/pkgs/tool.exe";
+            var pathTable = m_context.PathTable;
+            var executable = FileArtifact.CreateSourceFile(AbsolutePath.Create(pathTable, X(exectuablePath)));
+
+            var dependencies = new HashSet<FileArtifact>() { executable };
+            var process = GetDefaultProcessBuilder(pathTable, executable)
```
almiliMSFT:
```suggestion
            var pb = GetDefaultProcessBuilder(pathTable, executable)
```

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/FileAccessPolicyTests.cs
```diff
@@ -70,20 +70,12 @@ public void ValidateCachingUntrackedPathInput(bool declareDependency, bool input
 
             // Modify untracked input file /dir/nestedFile
             File.WriteAllText(ArtifactToString(nestedFile), "nestedFile");
-            if (declareDependency)
-            {
-                // Changes in declared dependencies always cause cache misses
-                RunScheduler().AssertCacheMiss(pip.PipId);
-            }
+
```
almiliMSFT: With this PR, the semantics tested here clearly changed.  I think that's ok because I don't think anyone actually depends on this semantics.  I would just update the doc comment on this test method to say that content hashes of untracked paths are not added to the fingerprint.

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/FileAccessPolicyTests.cs
```diff
@@ -70,20 +70,12 @@ public void ValidateCachingUntrackedPathInput(bool declareDependency, bool input
 
             // Modify untracked input file /dir/nestedFile
             File.WriteAllText(ArtifactToString(nestedFile), "nestedFile");
-            if (declareDependency)
-            {
-                // Changes in declared dependencies always cause cache misses
```
almiliMSFT: With this PR, the semantics tested here clearly changes.

I don't know how important this test is.  It is a bit scary to see that if an explicitly declared dependency changes the pip is not invalidated.  After all, maybe it's better to exclude only the executable path from the fingerprint when it's untracked, instead of every single dependency.

@narasamdya, what do you think?  I'm leaning towards not breaking this test.

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,66 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            var executablePath = X("/x/pkgs/tool.exe");
+            var pathTable = m_context.PathTable;
+            var executable = FileArtifact.CreateSourceFile(AbsolutePath.Create(pathTable, executablePath));
+
+            var dependencies = new HashSet<FileArtifact>() { executable };
+            var pb = GetDefaultProcessBuilder(pathTable, executable)
+                .WithDependencies(dependencies);
+
+            if (untrackPath)
+
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/UnitTests/Scheduler/PipFingerprinterTests.cs
```diff
@@ -400,6 +401,66 @@ public void TestUserProfileStaysOutOfFingerprints()
...
+            {
+                pb = pb.WithUntrackedPaths(new[] { AbsolutePath.Create(pathTable, executablePath) });
+            }
+
+            if (untrackScope)
+
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Scheduler/Fingerprints/PipFingerprinter.cs
narasamdya: I agree. The only concerning part is complexity, which is quadratic. But how bad it is, I don't know; depend on the size of untrack paths/scopes.

---
In reply to: [332731236](https://github.com/microsoft/BuildXL/pull/997#discussion_r332731236) [](ancestors = 332731236)

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/FileAccessPolicyTests.cs
narasamdya: I'm leaning towards consistency. I think this PR changes the semantics, where untracked stuff overrides dependencies. 

I don't mind we're breaking this test.

However, I'm still thinking that the changes should go to the process builder.

---
In reply to: [333088339](https://github.com/microsoft/BuildXL/pull/997#discussion_r333088339) [](ancestors = 333088339)

almiliMSFT: Ok, let's just update the doc comment for this test method to reflect the new semantics

PR #1000, opened 2019-10-04
Bubble up additional retry telemetry along with the ability to add custom pip property telemetry

breidikl:
This work adds information needed to investigate and bucket retried pips.  The design is to add a couple events capturing the telemetry as follows:

ProcessRetries event – This is a single event for the build session which will fire only if retries took place. It has the following columns
    o	PipsSucceedingAfterUserRetry – csv of semistable hashes that succeeded after a user retry
    o	PipsFailingAfterLastUserRetry – csv of semistable hashes of pips failed after a user retry
    o	ProcessUserRetriesImpactedPips – count of distinct pips impacted by user retries
    o	ProcessUserRetries – count of user retries performed
    o	RetriedUserExecutionDuration – sum retried process wall clock time of all user retried pips
    o	RetriedInternalExecutionDuration – sum retried process wall clock time of all internal retried pips

ProcessPattern event – Single session level event that will give counters for various interesting things that happen during the build. If the regex has a match for that capture group a counter gets incremented. At the end of the build session a single event is sent with a summary of the hits for those user defined patterns. Columns would be:
    o	[PatternName]Count – Count of all pips that had a result for user defined capture group ProcessPattern_[ExtactedInformation]_EndProperty is determined by looking for patterns in tool output streams. When a pip fails (including prior to a retry), BuildXL would apply the regular expression to the output to look for the matching capture groups.   
    o	[PatternName]Pips – SemiStable hashes corresponding to the counter. This is separate from count because it may need to be truncated
    o	All additional patterns


breidikl:
Based on the current draft and the ultimate goals, there are still 3 things that will need to be figured out (beyond what's in that pull request) before I'll be able to be finished:

1. Add a way to save PipIds globally to report up at the end.  This is something like adding to a hashset the the schedulers environment class, though there's not a concurrent version of that (I'm open to suggestions).  
2. Add a way to create custom counters on the fly based on the extracted pip properties.  This is likely a concurrent dictionary in the schedulers environment class with the key being the property name, and the value being a hashset of impacted pips.  
3. Actually create the Log event to upload the data. Which would involve saving the count of impacted pips for each of the new items and truncating into a csv as appropriate.

Public/Src/Engine/Processes/SandboxedProcessPipExecutionResult.cs
```diff
@@ -292,6 +301,11 @@ internal static SandboxedProcessPipExecutionResult MismatchedMessageCountFailure
         /// </summary>
         public ContainerConfiguration ContainerConfiguration { get; }
 
+        /// <summary>
+        /// Extract if a pip property and it's count if a value matching the PipProperty regex was defined in the process output
+        /// </summary>
+        public Dictionary<string,int> PipProperties { get; set; }
```
narasamdya: nit: add space -> `<string, int>`

Public/Src/Engine/Processes/SandboxedProcessPipExecutionResult.cs
```diff
@@ -292,6 +301,11 @@ internal static SandboxedProcessPipExecutionResult MismatchedMessageCountFailure
         /// </summary>
         public ContainerConfiguration ContainerConfiguration { get; }
 
+        /// <summary>
+        /// Extract if a pip property and it's count if a value matching the PipProperty regex was defined in the process output
```
narasamdya: >it's [](start = 42, length = 4)

typo: its?

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1549,12 +1553,30 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
...
+
                             return result;
                         }
                     });
 
             processExecutionResult.ReportSandboxedExecutionResult(executionResult);
 
+            if (executionResult.PipProperties != null && executionResult.PipProperties.Count > 0)
+            {
+                //TODO create and/or update global pip Properties counters
```
narasamdya: >//TODO create and/or update global pip Properties counters [](start = 16, length = 58)

Do you want to collect/aggregate the pip properties for each retry or only the final one? In this current code, it seems that you're only interested in the final run.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1549,12 +1553,30 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
...
+                                {
+                                    //TODO update PipsFailingAfterLastUserRetry
+                                }
+                            }
+
                             return result;
                         }
                     });
 
             processExecutionResult.ReportSandboxedExecutionResult(executionResult);
```
narasamdya: >ReportSandboxedExecutionResult [](start = 35, length = 30)

`processExecutionResult` is the one that the `Scheduler` gets after executing the process. The method `processExecutionResult.ReportSandboxedExecutionResult` basically include the sandboxed result (i.e., `executionResult`) into `processExecutionResult`. So we don't need to add any method to `IPipExecutionEnvironment`.

narasamdya: In the Scheduler.cs, try find the following excerpt:
```
case PipExecutionStep.ExecuteProcess:
                    {
                        MarkPipStartExecuting();

                        if (processRunnable.Weight > 1)
                        {
                            // Only log for pips with non-standard process weights
                            Logger.Log.ProcessPipProcessWeight(loggingContext, processRunnable.Description, processRunnable.Weight);
                        }

                        processRunnable.Executed = true;

                        var executionResult = await worker.ExecuteProcessAsync(processRunnable);

                        // Don't count service pips in process pip counters
                        if (!processRunnable.Process.IsStartOrShutdownKind && executionResult.PerformanceInformation != null)
                        {
                            var perfInfo = executionResult.PerformanceInformation;

                            if (perfInfo != null)
                            {
                                m_groupedPipCounters.AddToCounters(
                                    processRunnable.Process,
                                    new[]
                                    {
                                        (PipCountersByGroup.IOReadBytes, (long)perfInfo.IO.ReadCounters.TransferCount),
                                        (PipCountersByGroup.IOWriteBytes, (long)perfInfo.IO.WriteCounters.TransferCount)
                                    },
                                    new[] { (PipCountersByGroup.ExecuteProcessDuration, perfInfo.ProcessExecutionTime) }
                                );
                            }
                        }
```

In that excerpt we update `m_groupedPipCounters`. We can follow this pattern by adding code there to update the pip properties that you're tracking. I think this is how we propagate values from `PipExecutor` to `Scheduler`.

---
In reply to: [332259811](https://github.com/microsoft/BuildXL/pull/1000#discussion_r332259811) [](ancestors = 332259811)

breidikl: I would then still need a way to pass up if there had been retries in that case.  The main concern is what type of structure should be used to globally collect the information.  The existing counters are either long or int, and have helpers for aggregating correctly across all the machines.  I basically need to do that for PipIds.

Public/Src/Engine/Scheduler/PipExecutor.cs
breidikl: I want the count of pips with a given property.  As you said above, the result is already passed through to the scheduler (so I should be able to do the update there).  I do pass through the earlier dictionary of pip properties when re-running the pips, though I'll need to check if those are returning early prior to parsing out the existing properties.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1687,7 +1695,46 @@ private void LogCommandLineTooLong(SandboxedProcessInfo info)
...
+                {
+                    pipProperties = matchedProperties.ToDictionary(val => val, val => 1);
+                }
+                else
+                {
+                    foreach (string property in matchedProperties)
+                    {
+                        if (pipProperties.ContainsKey(property))
+                        {
+                            pipProperties[property] = ++pipProperties[property];
```
narasamdya: >pipProperties[property] = ++pipProperties[property]; [](start = 28, length = 52)

Is `++pipProperties[property]` alone not sufficient? I looks funny to have the assignment and the `++` operator.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1687,7 +1695,46 @@ private void LogCommandLineTooLong(SandboxedProcessInfo info)
                 allReportedFileAccesses: allFileAccesses,
                 detouringStatuses: result.DetouringStatuses,
                 maxDetoursHeapSize: maxDetoursHeapSize,
-                containerConfiguration: m_containerConfiguration);
+                containerConfiguration: m_containerConfiguration,
+                pipProperties: pipProperties);
+        }
+
+        private async Task<Dictionary<string, int>> UpdatePipPropertiesCountsAsync(SandboxedProcessResult result, Dictionary<string, int> pipProperties)
```
narasamdya: >private async Task<Dictionary<string, int>> [](start = 8, length = 43)

I think we can have just `private async Task Update...`; no need to return a dictionary. Or this method will create a new instance of Dictionary, and thus you don't have to pass it as an argument.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1687,7 +1695,46 @@ private void LogCommandLineTooLong(SandboxedProcessInfo info)
...
-                containerConfiguration: m_containerConfiguration);
+                containerConfiguration: m_containerConfiguration,
+                pipProperties: pipProperties);
+        }
+
+        private async Task<Dictionary<string, int>> UpdatePipPropertiesCountsAsync(SandboxedProcessResult result, Dictionary<string, int> pipProperties)
+        {
+            const string PipPropertyPrefix = "PipProperty_";
+
+            Regex pipPropertyRegex = new Regex(PipPropertyPrefix + @"(?<" + PipPropertyPrefix + @">.+?)_EndProperty", RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.Multiline);
```
narasamdya: Please add a brief comment what this Regex is trying to achieve. A comment accompanied with an example would be better -- adding unit tests would be great.

Public/Src/Engine/Scheduler/PipExecutor.cs
breidikl: I've moved UpdatePipPropertiesCountsAsync earlier in the process, so that it counts the values before starting the retry as well.

Public/Src/Engine/Scheduler/PipExecutor.cs
breidikl: ExecutionResult can be created from the SandboxedProcessPipExecutionResult, though access to that is restricted (part of the private InnerUnsealedState) and we'll still need to pass through any explicit data we want available in the scheduler.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -538,24 +538,26 @@ private bool IsWarning(string line)
         {
             internal readonly Predicate<string> LinePredicate;
             internal readonly Regex Regex;
+            internal readonly string GroupName;
```
narasamdya: >internal readonly string GroupName; [](start = 12, length = 35)

Add comment that this `GroupName` is only associated with the `Regex`, and not `LinePredicate`. This association needs to be reflected in the contract as well.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -583,9 +585,9 @@ internal string ExtractMatches(string source, string outputSeparator = null)
                 }
             }
 
-            private static string ExtractTextFromMatch(Match match)
+            private string ExtractTextFromMatch(Match match)
```
narasamdya: >ExtractTextFromMatch [](start = 27, length = 20)

I'd rather make this `OutputFilter` component more general.  I think we should simply return `match.Value` if `GroupName` is `null` or `string.Empty`.

The caller to `OutputFilter`'s constructor is responsible for passing `ErrorMessageGroupName`.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -583,9 +585,9 @@ internal string ExtractMatches(string source, string outputSeparator = null)
                 }
             }
 
-            private static string ExtractTextFromMatch(Match match)
+            private string ExtractTextFromMatch(Match match)
             {
-                var errorMessageGroup = match.Groups[ErrorMessageGroupName];
+                Group errorMessageGroup = match.Groups[GroupName];
```
narasamdya: >errorMessageGroup [](start = 22, length = 17)

The variable `errorMessageGroup` may be misnomer here because we no longer use this method for matching against `ErrorMessageGroupName`.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -538,24 +538,26 @@ private bool IsWarning(string line)
...
                 Contract.Requires(regex != null);
             }
 
-            private OutputFilter(Predicate<string> linePredicate, Regex regex)
+            private OutputFilter(Predicate<string> linePredicate, Regex regex, string groupName)
             {
                 Contract.Requires(linePredicate != null || regex != null);
                 LinePredicate = linePredicate;
                 Regex = regex;
+                GroupName = groupName ?? ErrorMessageGroupName;
```
narasamdya: >ErrorMessageGroupName [](start = 41, length = 21)

I think the caller of `OutputFilter`'s constructor is responsible for passing this group name. I see only one caller, so it's should be easy to update.

narasamdya:
>         private readonly struct OutputFilter

This struct now deserves a separate file. This will also make us easy to test this component.

---
Refers to: Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs:537 in cb49c48. [](commit_id = cb49c48ab2888650a9765f36f076bf449fd54533, deletion_comment = False)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1687,7 +1694,39 @@ private void LogCommandLineTooLong(SandboxedProcessInfo info)
...
-                containerConfiguration: m_containerConfiguration);
+                containerConfiguration: m_containerConfiguration,
+                pipProperties: pipProperties);
+        }
+
+        private async Task UpdatePipPropertiesCountsAsync(SandboxedProcessResult result, Dictionary<string, int> pipProperties)
+        {
+            const string PipPropertyPrefix = "PipProperty_";
+
+            // Match everything between PipProperty_ and _EndProperty, assigned to a PipProperty_ group
```
narasamdya: >PipProperty_ and _EndPropert [](start = 40, length = 28)

For testing this, here's what I think

- Create a separate struct/class and file for `PipPropertyFilter`. We can add description about `PipProperty_` and `_EndProperty` in that class.
```
class PipPropertyFilter
{
    private const string ErrorMessageGroup = "..";
    private const string NamedPropertyPrefix = "...";
    private const string NamedPropertySuffix = "...";
    private static readonly Regex s_pipNamedPropertyRegex = new Regex(NamedPropertyPrefix + @"(?<" + NamedPropertyPrefix + @">.+?)_NamedPropertySuffix ", RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.Multiline);

    public readonly Regex Regex;
    public readonly string GroupName;
    private PipPropertyFilter(Regex regex, string groupName)
    {
       // ...
    }
   
    // Create a specialized error filter.
    public static PipPropertyFilter CreateErrorProperty(Regex regex) => new PipPropertyFilter(regex, ErrorMessageGroup);

    // Create a named filter.
    public static PipPropertyFilter CreateNamedProperty() => new PipPropertyFilter(s_pipNamedPropertyRegex, NamedPropertyPrefix);
}
```

- Create a separate file for `OutputFilter` struct. The `OutputFilter` struct now will take `PipPropertyFilter` as an argument.

- Create unit tests against the two classes.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
breidikl: GroupName is currently never null or string.Empty.  It's optionally passed via one of the constructors to override the default ErrorMessageGroupName value.  That allows all the existing callers to remain unchanged.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
breidikl: Is the ask to make GroupName an explicitly required parameter (and then update all the callers to pass that)?

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1687,7 +1694,39 @@ private void LogCommandLineTooLong(SandboxedProcessInfo info)
                 allReportedFileAccesses: allFileAccesses,
                 detouringStatuses: result.DetouringStatuses,
                 maxDetoursHeapSize: maxDetoursHeapSize,
-                containerConfiguration: m_containerConfiguration);
+                containerConfiguration: m_containerConfiguration,
+                pipProperties: pipProperties);
+        }
+
+        private async Task UpdatePipPropertiesCountsAsync(SandboxedProcessResult result, Dictionary<string, int> pipProperties)
```
narasamdya: >UpdatePipPropertiesCountsAsync [](start = 27, length = 30)

This is only called once per call to `ProcessSandboxedProcessResultAsync`. So I don't understand when the `pipProperties` gets "updated". In this case perhaps the code should be 
```
private async Task<Dictionary<string, int>> GetPipPropertiesCountAsync(...);
```

breidikl: It updates whatever value was set on the result, which could originally be set prior to a retry and passed through to the current execution.  You're correct that it won't be set prior to the first instance.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
breidikl: GroupName can currently only be added via a constructor that sets a regex, which is checked by the contract.  If a value is not passed, it already defaults to the previous hardcoded value that's already assumed by the existing callers.  Comment added about GroupName.

Public/Src/Engine/Processes/OutputFilter.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        /// <summary>
+        /// Group name from the ErrorRegex to use to extract error message.
+        /// When no such group exists, the entire match is used.
+        /// </summary>
+        private const string ErrorMessageGroupName = "ErrorMessage";
+
+        internal readonly Predicate<string> LinePredicate;
+        internal readonly Regex Regex;
+        // GroupName can be used with the Regex and defaults to ErrorMessageGroupName if nothing
```
narasamdya: >// [](start = 8, length = 2)

We typically use /// <summary and /// <remarks

Public/Src/Engine/Processes/OutputFilter.cs
```diff
@@ -0,0 +1,125 @@
...
+        private const string ErrorMessageGroupName = "ErrorMessage";
+
+        internal readonly Predicate<string> LinePredicate;
+        internal readonly Regex Regex;
+
+        /// <summary>
+        /// The GroupName to be used by the Regex
+        /// </summary>
+        /// <remarks>
+        /// Defaults to ErrorMessageGroupName if nothing is provided when constructing the OutputFiler with a Regex
```
narasamdya: >ErrorMessageGroupName [](start = 24, length = 21)

nit: `<see cref ...`

Public/Src/Engine/Scheduler/Distribution/ExecutionResultSerializer.cs
```diff
@@ -613,5 +621,51 @@ private void WriteFileArtifact(BuildXLWriter writer, FileArtifact file)
...
+            }
+            else
+            {
+                writer.Write(true);
+
+                writer.WriteCompact(pipProperties.Count);
+
+                foreach (string key in pipProperties.Keys)
+                {
+                    writer.WriteNullableString(key);
```
narasamdya: Could the key be `null`?

Public/Src/Engine/Scheduler/Distribution/ExecutionResultSerializer.cs
```diff
@@ -613,5 +621,51 @@ private void WriteFileArtifact(BuildXLWriter writer, FileArtifact file)
...
+
+            if (!hasPipProperties)
+            {
+                pipProperties = null;
+            }
+            else
+            {
+                int count = reader.ReadInt32Compact();
+
+                var tmpPipProperties = new Dictionary<string, int>();
```
narasamdya: Why do we need a temporary variable? Can the method just return the dictionary instead of using `out`?

Public/Src/Engine/Scheduler/Distribution/ExecutionResultSerializer.cs
```diff
@@ -613,5 +621,51 @@ private void WriteFileArtifact(BuildXLWriter writer, FileArtifact file)
             WriteAbsolutePath(writer, file.Path);
             writer.WriteCompact(file.RewriteCount);
         }
+
+        private static void ReadPipProperties(BuildXLReader reader, out IReadOnlyDictionary<string, int> pipProperties)
```
narasamdya: >ReadPipProperties [](start = 28, length = 17)

I'm thinking that we should use [http://index/?leftProject=BuildXL.Utilities&leftSymbol=wdncwk257pvr&file=BuildXLWriter.cs&line=338](http://index/?leftProject=BuildXL.Utilities&leftSymbol=wdncwk257pvr&file=BuildXLWriter.cs&line=338)
for reading and writing.
The pipProperties can be turned into a list of key-value pairs first.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1404,6 +1406,25 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
 
                                 ++retryCount;
 
+                                if (result.PipProperties != null)
```
narasamdya: >result.PipProperties [](start = 36, length = 20)

Who made result.PipProperties non-null the first time?

narasamdya: >if (result.PipProperties != null) [](start = 32, length = 33)

What if in the current run `result.PipProperties` turns out to be `null`? In that case, you may lose the pip properties from previous runs.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1404,6 +1406,25 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
...
                                 ++retryCount;
 
+                                if (result.PipProperties != null)
+                                {
+                                    if (pipProperties != null)
+                                    {
+                                        foreach (string pipPropertyKey in pipProperties.Keys)
+                                        {
+                                            result.PipProperties.TryGetValue(pipPropertyKey, out int value);
+                                            result.PipProperties[pipPropertyKey] = value + pipProperties[pipPropertyKey];
```
narasamdya: It took me some time to understand what you're doing here.

I'd rather override `result.PipProperties` with `pipProperties` at the end of the loop (just like `hasRetry`). 

I.e., `pipProperties` is the dictionary used to accumulate all obtained `result.PipProperties`. Then it updates the pip properties of returned `result`.

Public/Src/Engine/Scheduler/Distribution/ExecutionResultSerializer.cs
breidikl: Dictionaries can't have null keys.  The if condition ensures we only hit this block if we have dictionary items.

Public/Src/Engine/Scheduler/Distribution/ExecutionResultSerializer.cs
breidikl: What is the benefit of that over the methods added in the review, which does not require converting the dictionary to list first?

Public/Src/Engine/Scheduler/Distribution/ExecutionResultSerializer.cs
breidikl: I was trying to match the other methods in this file used by the serializer, though I agree just returning is clearer.  Updated.

Public/Src/Engine/Scheduler/PipExecutor.cs
breidikl: Good catch, updated.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1404,6 +1406,28 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
 
                                 ++retryCount;
 
+                                if (result.PipProperties != null)
+                                {
+                                    if (pipProperties != null)
+                                    {
+                                        foreach (string pipPropertyKey in pipProperties.Keys)
+                                        {
+                                            result.PipProperties.TryGetValue(pipPropertyKey, out int value);
```
narasamdya: What is `value` if `result.PipProperties` doesn't have the `pipPropertyKey`?

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1404,6 +1406,28 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
 
                                 ++retryCount;
 
+                                if (result.PipProperties != null)
```
narasamdya: >if [](start = 32, length = 2)

This is my suggestion to make the code less confusing (at least for me):
```
// Declare aggregate pip properties before the while loop.
var aggregatePipProperties = new Dictionary<string, int>(); // If memory is a concern, we can use object pool -- we have object pool.

while (true) 
{
...
    if (result.PipProperties != null) {
        foreach (var kvp in result.PipProperties) {
             if (aggregatePipProperties.TryGetValue(kvp.Key, out var value)) {
                 aggregatePipProperties[kvp.Key] = value + kvp.Value;
             } else {
                 aggregatePipProperties.Add(kvp.Key, kvp.Value);
             }
        }
    }
}

if (aggregatePipProperties.Count > 0) {
    // The aggregate result should override the last result.
    result.PipProperties = aggregatePipProperties;
}

```

What do you think?

Public/Src/Engine/Scheduler/Distribution/ExecutionResultSerializer.cs
narasamdya: Then, should we simply use `writer.Write(key)` ?

---
In reply to: [334751796](https://github.com/microsoft/BuildXL/pull/1000#discussion_r334751796) [](ancestors = 334751796)

Public/Src/Engine/Scheduler/Distribution/ExecutionResultSerializer.cs
narasamdya: Code reuse.

---
In reply to: [334752748](https://github.com/microsoft/BuildXL/pull/1000#discussion_r334752748) [](ancestors = 334752748)

Public/Src/Engine/Scheduler/PipExecutor.cs
breidikl: Sounds good to me, updated.

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -3565,6 +3565,24 @@ internal void FailedToHashInputFile(LoggingContext loggingContext, string pipDes
             Message = "USN journal statistics for scheduler")]
         public abstract void JournalProcessingStatisticsForSchedulerTelemetry(LoggingContext context, string scanningJournalStatus, IDictionary<string, long> stats);
 
+        [GeneratedEvent(
+            (int)EventId.ProcessRetries,
+            EventGenerators = EventGenerators.TelemetryOnly,
+            EventLevel = Level.Verbose,
+            EventTask = (ushort)Tasks.Scheduler,
+            Keywords = (int)Keywords.UserMessage,
+            Message = "ProcessRetries PipsSucceedingAfterUserRetry:{0} and PipsFailingAfterUserRetry:{1}")]
```
narasamdya: >{ [](start = 67, length = 1)

nit: add space between ':' and '{'. 

Also use named variable, i.e., instead of `{0}`, write `{pipsSucceedingAfterUserRetry}`. 

Similarly for `{1}`.

Public/Src/Engine/Scheduler/Tracing/Log.cs
```diff
@@ -3565,6 +3565,24 @@ internal void FailedToHashInputFile(LoggingContext loggingContext, string pipDes
...
+        public abstract void ProcessRetries(LoggingContext context, string pipsSucceedingAfterUserRetry, string pipsFailingAfterLastUserRetry);
+
+        [GeneratedEvent(
+            (int)EventId.ProcessPattern,
+            EventGenerators = EventGenerators.TelemetryOnly | Generators.Statistics,
+            EventLevel = Level.Verbose,
+            EventTask = (ushort)Tasks.Scheduler,
+            Keywords = (int)Keywords.UserMessage,
+            Message = "ProcessPattern {0}")]
```
narasamdya: >0 [](start = 39, length = 1)

Use named variable for reference.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -297,6 +302,11 @@ public partial class Scheduler : IPipScheduler, IPipExecutionEnvironment, IFileC
         private int m_maxUnresponsivenessFactor = 0;
         private DateTime m_statusLastCollected = DateTime.MaxValue;
 
+        private ConcurrentDictionary<string, long> m_aggregatedPipPropertiesCount = new ConcurrentDictionary<string, long>();
+        private ConcurrentDictionary<string, HashSet<string>> m_addgregatepipsPerPipProperty = new ConcurrentDictionary<string, HashSet<string>>();
+        private ConcurrentBigSet<string> m_pipsSucceedingAfterUserRetry = new ConcurrentBigSet<string>();
+        private ConcurrentBigSet<string> m_pipsFailingAfterLastUserRetry = new ConcurrentBigSet<string>();
```
narasamdya: Can you please create a separate class containing all of these `ConcurrentDictionary` and `ConcurrentBugSet`?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -1652,6 +1662,30 @@ public SchedulerPerformanceInfo LogStats(LoggingContext loggingContext, [CanBeNu
...
+                {
+                    pipPropertiesPips = item.Key + "Pips: " + string.Join(",", item.Value);
+                    countOfLoggedProperties++;
+                    if (countOfLoggedProperties == MaxListOfPipIdsForTelemetry)
+                    {
+                        break;
+                    }
+                }
+                Logger.Log.ProcessPattern(loggingContext, pipPropertiesPips, m_aggregatedPipPropertiesCount);
+            }
```
narasamdya: These if-logic can be put inside the classes that I mentioned before.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3645,6 +3679,43 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
...
+                            }
+                            else if (executionResult.Result == PipResultStatus.Failed)
+                            {
+                                PipExecutionCounters.IncrementCounter(PipExecutorCounter.ProcessUserRetriesFailedPipsCount);
+                                if (m_pipsFailingAfterLastUserRetry.Count < MaxListOfPipIdsForTelemetry)
+                                {
+                                    m_pipsFailingAfterLastUserRetry.Add(processRunnable.Process.FormattedSemiStableHash);
+                                }
+                            }
+                        }
```
narasamdya: These if-logic can be put inside the classes that I suggested before.

PR #1003, opened 2019-10-04
Add delay for the self check process.

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternalChecker.cs
```diff
@@ -15,11 +14,69 @@
...
+using BuildXL.Native.IO;
 using AbsolutePath = BuildXL.Cache.ContentStore.Interfaces.FileSystem.AbsolutePath;
 using FileInfo = BuildXL.Cache.ContentStore.Interfaces.FileSystem.FileInfo;
 
+#nullable enable
+
 namespace BuildXL.Cache.ContentStore.Stores
 {
+    /// <nodoc />
+    public sealed class SelfCheckSettings
```
dcwuser: I think this is big enough to trigger the one class per file rule.

SergeyTeplyakov: Will move in a separate PR, because the current pipeline passes.

PR #1006, opened 2019-10-04
Trace number of replicas and attempts for successful copies.

ljcollins25:


dcwuser:
I'm okay with this, but I would prefer to keep it a bit more structured and extensible by adding properties to CopyResult and logging via CopyResutl.ToString()

ljcollins25:
I was initially going to do something like that but its strange since these aren't really aspects of the result but rather context about the copy operation. Also, some of the other fields are even stranger since they are settings.

---
In reply to: [538585796](https://github.com/microsoft/BuildXL/pull/1006#issuecomment-538585796) [](ancestors = 538585796)

PR #1007, opened 2019-10-04
Use xattrs to mark shared opaque outputs on Mac

almiliMSFT:
The implementation for Windows remains the same and continues to use magic timestamps.

On Mac, instead of timestamps extended attributes are used.  Concretely, a file is a shared opaque output IFF it has a `com.microsoft.buildxl:shared_opaque_output` attribute and it's value is equal to `42` (at some point in the future we could insert a more meaningful value here, e.g., PipId)

[AB#1607996](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1607996)

### DIFF
--- a/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
+++ b/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
@@ -83,7 +83,9 @@ namespace BuildXL.Processes
         private static unsafe class Unix
         {
             private const string MY_XATTR_NAME = "com.microsoft.buildxl:shared_opaque_output";
-            private const long MY_XATTR_VALUE = 42;
+
+            // arbitrary value; in the future, we could store something more useful here (e.g., the producer PipId or something)
+            private const long MY_XATTR_VALUE = 42; 
 
             private const int XATTR_NOFOLLOW = 1;
 
@@ -100,7 +102,7 @@ namespace BuildXL.Processes
             private static extern long GetXattr(
                 [MarshalAs(UnmanagedType.LPStr)] string path,
                 [MarshalAs(UnmanagedType.LPStr)] string name,
-                void* value,
+                void *value,
                 ulong size,
                 uint position,
                 int options);
--- a/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
+++ b/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
@@ -166,7 +166,7 @@ namespace BuildXL.Processes
             //   - if 'expandedPath' is a symlink pointing to a non-existent file, 'expandedPath' should hence
             //     still be treated as an existent file
             //   - similarly, if 'expandedPath' is a symlink pointing to a directory, it should still be treated as a file
-            bool followSymlink = true;
+            bool followSymlink = !OperatingSystemHelper.IsUnixOS;
 
             // If the file is not there (the check may be happening against a reported write that later got deleted)
             // then there is nothing to do
--- a/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
+++ b/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
@@ -14,6 +14,14 @@ namespace BuildXL.Processes
     /// Utility class for identifying a file as being an output of a shared opaque.
     /// This helps the scrubber to be more precautious when deleting files under shared opaques.
     /// </summary>
+    /// <remarks>
+    /// Currently, we use two different strategies when running on Windows and non-Windows platforms
+    ///   - on Windows, we set a magic timestamp as file's creation (birth) date
+    ///   - on Mac, we set an extended attribute with a special name.
+    /// On Mac, some tools tend to change the timestamps (even the birth date) which is the reason for this.
+    ///
+    /// TODO: eventually we should unify this and use the same mechanism for all platforms, if possible.
+    /// </remarks>
     public static class SharedOpaqueOutputHelper
     {
         private static class Win
--- a/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
+++ b/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Runtime.InteropServices;
 using System.Security.AccessControl;
 using BuildXL.Native.IO;
 using BuildXL.Utilities;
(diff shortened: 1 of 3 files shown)


Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -56,7 +166,7 @@ public static bool IsSharedOpaqueOutput(string expandedPath)
             //   - if 'expandedPath' is a symlink pointing to a non-existent file, 'expandedPath' should hence
             //     still be treated as an existent file
             //   - similarly, if 'expandedPath' is a symlink pointing to a directory, it should still be treated as a file
-            bool followSymlink = !OperatingSystemHelper.IsUnixOS;
+            bool followSymlink = true;
```
almiliMSFT:
```suggestion
            bool followSymlink = !OperatingSystemHelper.IsUnixOS;
```

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -10,31 +11,140 @@
...
+                void *value,
+                ulong size,
+                uint position,
+                int options);
+
+            [DllImport("libc", EntryPoint = "getxattr")]
+            private static extern long GetXattr(
+                [MarshalAs(UnmanagedType.LPStr)] string path,
+                [MarshalAs(UnmanagedType.LPStr)] string name,
+                void* value,
```
almiliMSFT:
```suggestion
                void *value,
```

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -10,31 +11,140 @@
...
+                }
+                catch (BuildXLException ex)
+                {
+                    throw new BuildXLException(I($"Failed to open output file '{expandedPath}' for reading."), ex);
+                }
             }
-            catch (BuildXLException ex)
+        }
+
+        private static unsafe class Unix
```
smera: >Unix [](start = 36, length = 4)

If we are doing this change, why not doing it for both Win and Unix? It will make the code simpler, and considering this is a breaking change, better do it now that later.

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -10,31 +11,140 @@
...
+        {
+            private const string MY_XATTR_NAME = "com.microsoft.buildxl:shared_opaque_output";
+            private const long MY_XATTR_VALUE = 42;
```
smera: > 42 [](start = 47, length = 3)

I assume this is arbitrary, right? Maybe adding a comment stating that?

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -10,31 +11,140 @@
...
+
+            private const int XATTR_NOFOLLOW = 1;
+
+            [DllImport("libc", EntryPoint = "setxattr")]
+            private static extern int SetXattr(
+                [MarshalAs(UnmanagedType.LPStr)] string path,
+                [MarshalAs(UnmanagedType.LPStr)] string name,
+                void *value,
```
smera: >void *value [](start = 16, length = 11)

can't we declare this in a more specific way? e.g. long* ?

smera:
You probably already have thought about this, but just in case, this is a breaking change for mac, meaning that people have to be instructed to manually scrub old shared opaque outputs before ingesting this change. Same would apply for Win if we decide to include it as well in the change (which I would vote for)

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
almiliMSFT: That's exactly why: I don't really want this breaking change, I only want to validate if this fixes the problem for Mac builds.  And indeed, there doesn't seem to be anything wrong with using timestamps on Windows (because of timestamp faking) so we don't need to make this a breaking change on Windows.

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
almiliMSFT: Yes, completely arbitrary; will add a comment.

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
almiliMSFT: It doesn't have to be long*, it can be any pointer.  Here I'm simply mimicking the setxattr function signature:
```c
     int
     setxattr(const char *path, const char *name, void *value, size_t size, u_int32_t position, int options);
```
#WontFix

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
smera: That's fine, and I won't push hard for it. But I'd rather have a uniform marker. Less cognitive load and simpler code.
I don't expect this breaking change to be a problematic one: there are no real MSBuild customers yet, and WDG always scrubs everything themselves before building. So should be almost a noop.

---
In reply to: [332139469](https://github.com/microsoft/BuildXL/pull/1007#discussion_r332139469) [](ancestors = 332139469)

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
almiliMSFT:
```suggestion

           // arbitrary value; in the future, we could store something more useful here (e.g., the producer PipId or something)
            private const long MY_XATTR_VALUE = 42; 
```

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
almiliMSFT: Thanks for all those pointers.

I'll leave a comment saying that we should unify this eventually.

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -10,31 +11,142 @@
 namespace BuildXL.Processes
 {
     /// <summary>
-    /// Utility class for identifying a file as being an output of a shared opaque. This helps the scrubber to be more precautious when deleting files under shared opaques.
+    /// Utility class for identifying a file as being an output of a shared opaque.
+    /// This helps the scrubber to be more precautious when deleting files under shared opaques.
     /// </summary>
```
almiliMSFT:
```suggestion
    /// </summary>
    /// <remarks>
    /// Currently, we use two different strategies when running on Windows and non-Windows platforms
    ///   - on Windows, we set a magic timestamp as file's creation (birth) date
    ///   - on Mac, we set an extended attribute with a special name.
    /// On Mac, some tools tend to change the timestamps (even the birth date) which is the reason for this.
    ///
    /// TODO: eventually we should unify this and use the same mechanism for all platforms, if possible.
    /// </remarks>
```

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -10,31 +11,150 @@
...
+        private static unsafe class Unix
+        {
+            private const string MY_XATTR_NAME = "com.microsoft.buildxl:shared_opaque_output";
+
+            // arbitrary value; in the future, we could store something more useful here (e.g., the producer PipId or something)
+            private const long MY_XATTR_VALUE = 42; 
 
-                // Since these files should be just created outputs, this shouldn't happen and we bail out hard.
-                throw new BuildXLException(I($"Failed to open output file '{expandedPath}' for writing."), ex);
+            private const int XATTR_NOFOLLOW = 1;
```
almiliMSFT:
```suggestion
            // from xattr.h:
            // #define XATTR_NOFOLLOW   0x0001     /* Don't follow symbolic links */
            private const int XATTR_NOFOLLOW = 1;
```

PR #1014, opened 2019-10-07
A special symbol daemon for CB

olkononenko:
A special symbol daemon runner for CB builds (includes magic CB env vars)

### DIFF
--- a/Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
+++ b/Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
@@ -63,6 +63,24 @@ export const runner: SymbolRunner = {
     },
 };
 
+/**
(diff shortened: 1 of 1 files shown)


Public/Src/Tools/SymbolDaemon/Tool.SymbolDaemonRunner.dsc
```diff
@@ -350,3 +368,31 @@ function debugEntryCreateBehaviorToString(arg: DebugEntryCreateBehavior) {
             return "Unspecified";
     };
 }
+
+const cbEnvironmentVariables: string[] = [
```
almiliMSFT: these vars are already defined [here](https://github.com/microsoft/BuildXL/blob/006f3fc733a011c92417f0065c77ac04a0803df4/Public/Src/Tools/DropDaemon/Tool.DropDaemonRunner.dsc#L424-L433)

PR #1016, opened 2019-10-08
Fix "bxl -cache -vs" to do the right thing

JuanCarlosGI:
Previously we ran "bxl -cache /vs /q:DebugNet472 -LongRunningTest" to generate the cache solution

Now we can run "bxl -cache -vs" to achieve the same thing.

### DIFF
--- a/Shared/Scripts/bxl.ps1
+++ b/Shared/Scripts/bxl.ps1
@@ -557,7 +557,7 @@ if (!$skipFilter){
 
     if ($Cache) {
         # Only build Cache code.
-        if ($LongRunningTest) {
+        if ($LongRunningTest -or $Vs) {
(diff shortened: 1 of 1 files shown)


Shared/Scripts/bxl.ps1
```diff
@@ -557,7 +557,7 @@ if (!$skipFilter){
 
     if ($Cache) {
         # Only build Cache code.
-        if ($LongRunningTest) {
+        if ($LongRunningTest -or $Vs) {
```
almiliMSFT: I thought @SergeyTeplyakov fixed this a while ago

JuanCarlosGI: Right now it doesn't include the LongRunningTest flag by default

PR #1017, opened 2019-10-08
Don't shutdown a session if shutdown has already begun

kewilli:


Public/Src/Cache/ContentStore/Library/Service/LocalContentServerBase.cs
```diff
@@ -816,7 +816,11 @@ private async Task ReleaseSessionInternalAsync(Context context, int sessionId)
 
             TryUnsetBuildId(sessionHandle.SessionName);
 
-            await sessionHandle.Session.ShutdownAsync(context).ThrowIfFailure();
+            if (!sessionHandle.Session.ShutdownStarted)
```
JuanCarlosGI: Should we lock until we verify that ShutdownFinished?

PR #1021, opened 2019-10-09
Fix flushing behavior for scalability

jbayardo:


Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -605,7 +613,10 @@ protected void Delete(OperationContext context, ShortHash hash)
             Store(context, hash, entry: null);
         }
 
-        private Task ForceCacheFlushAsync(OperationContext context)
+        /// <summary>
+        /// Should only be called from <see cref="ForceCacheFlush(ContentStore.Tracing.Internal.OperationContext, ContentLocationDatabaseCounters?, bool)"/>
```
SergeyTeplyakov: Maybe instead of making this comment, this method should be made internal to avoid any other options?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -616,15 +627,26 @@ private Task ForceCacheFlushAsync(OperationContext context)
...
                     }
                     finally
                     {
-                        Interlocked.Exchange(ref _cacheUpdatesSinceLastFlush, 0);
+                        Interlocked.Add(ref _cacheUpdatesSinceLastFlush, -flushedEntries);
                         ResetFlushTimer();
+
+                        if (counter != null)
+                        {
+                            Counters[counter.Value].Increment();
```
SergeyTeplyakov: should it be 'counters[counter.Value]' instead?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -616,15 +627,26 @@ private Task ForceCacheFlushAsync(OperationContext context)
                 Tracer,
                 async () =>
                 {
+                    long flushedEntries = 0;
                     try
                     {
+                        // Make it likely that this runs in a separate thread other than the one that triggered the
+                        // flush
                         await Task.Yield();
```
SergeyTeplyakov: I believe this Task.Yield should be inside '_inMemoryCache.FlushAsync' not here.

Right now it is hard to tell, why the Yield is here. THis method just do its job and it is hard to tell why we follow this pattern here (calling Yield) but not in other places.

Moving this into InMemoryCache.FlushAsync make sense, because InMemoryCache.FlushAsync is not "very" async because it only acquires mutex asynchronously.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -308,8 +311,8 @@ protected IEnumerable<ShortHash> EnumerateSortedKeys(OperationContext context)
             if (IsInMemoryCacheEnabled)
             {
                 // If we don't check, then GC will happen anyways and we will -incorrectly- increment the counter
-                Counters[ContentLocationDatabaseCounters.NumberOfCacheFlushesTriggeredByGarbageCollection].Increment();
-                ForceCacheFlush(context);
+                ForceCacheFlush(context,
+                    counter: ContentLocationDatabaseCounters.NumberOfCacheFlushesTriggeredByGarbageCollection);
```
SergeyTeplyakov: Based on the client of this methods, it seems that the counter is not correct.

This method is not called from the GC and only called from the worker machine.

Could you please double check that?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
jbayardo: No, Counters is for the ContentLocationDatabase, counters is for the flush. I renamed the flush ones to make it less ambiguous

---
In reply to: [333121903](https://github.com/microsoft/BuildXL/pull/1021#discussion_r333121903) [](ancestors = 333121903)

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -638,9 +660,29 @@ private Task ForceCacheFlushAsync(OperationContext context)
...
         }
 
-        internal void ForceCacheFlush(OperationContext context)
+        internal bool ForceCacheFlush(OperationContext context, ContentLocationDatabaseCounters? counter = null, bool blocking = true)
         {
-            ForceCacheFlushAsync(context).GetAwaiter().GetResult();
+            // Ensure only one thread can start the flush
+            if (Interlocked.CompareExchange(ref _isFlushInProgress, 1, 0) == 1)
+            {
+                return false;
```
ljcollins25: >return false; [](start = 16, length = 13)

Is it ok to early out in the case where a blocking flush is requested?

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -638,9 +660,29 @@ private Task ForceCacheFlushAsync(OperationContext context)
                 }).ThrowIfFailure();
         }
 
-        internal void ForceCacheFlush(OperationContext context)
+        internal bool ForceCacheFlush(OperationContext context, ContentLocationDatabaseCounters? counter = null, bool blocking = true)
         {
-            ForceCacheFlushAsync(context).GetAwaiter().GetResult();
+            // Ensure only one thread can start the flush
+            if (Interlocked.CompareExchange(ref _isFlushInProgress, 1, 0) == 1)
```
dcwuser: After our discussion yesterday, I was going to suggest using SpinLock instead of lock for the inner comparison, but CompareExchange is even better. Well done improvement.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
jbayardo: No it's not, I added a wait handle for this case. Please re-review

jbayardo: No it's not, I added a wait handle for this case. Please re-review

---
In reply to: [333181654](https://github.com/microsoft/BuildXL/pull/1021#discussion_r333181654) [](ancestors = 333181654)

ljcollins25: At the very least, I think you should use ManualResetEventSlim instead of using a system event wait handle.

---
In reply to: [333220081](https://github.com/microsoft/BuildXL/pull/1021#discussion_r333220081) [](ancestors = 333220081,333181654)

PR #1026, opened 2019-10-10
Expose GrpcPoolSize

kewilli:


Public/Src/Cache/DistributedCache.Host/Configuration/LocalCasSettings.cs
```diff
@@ -51,7 +52,8 @@ public LocalCasSettings()
                 grpcPort: grpcPort,
                 grpcPortFileName: grpcPortFileName,
                 bufferSizeForGrpcCopies: bufferSizeForGrpcCopies,
-                gzipBarrierSizeForGrpcCopies: gzipBarrierSizeForGrpcCopies);
+                gzipBarrierSizeForGrpcCopies: gzipBarrierSizeForGrpcCopies,
+                grpcPoolSize: grpcPoolSize);
```
SergeyTeplyakov: Should it be "grpcThreadPoolSize", otherwise it is hard to tell what this thing actually configures...

Public/Src/Cache/DistributedCache.Host/Configuration/LocalCasServiceSettings.cs
```diff
@@ -107,5 +109,11 @@ public LocalCasServiceSettings()
         /// </summary>
         [DataMember]
         public int? GzipBarrierSizeForGrpcCopies { get; set; } = null;
+
+        /// <summary>
+        /// Thread pool size used by GRPC.
+        /// </summary>
+        [DataMember]
+        public int? GrpcPoolSize { get; set; } = 70;
```
SergeyTeplyakov: I believe the pattern here is to keep it null for nullable types and not to duplicate the default values in many places.

Public/Src/Cache/DistributedCache.Host/Configuration/LocalCasSettings.cs
SergeyTeplyakov: I see that this is the name that was used everywhere before, but I think in the global config it make sense to use a more descriptive name (I would actually vote for changing the name everywhere, but this is up to you).

---
In reply to: [333632905](https://github.com/microsoft/BuildXL/pull/1026#discussion_r333632905) [](ancestors = 333632905)

Public/Src/Cache/ContentStore/Library/Service/LocalServerConfiguration.cs
```diff
@@ -108,7 +108,7 @@ public LocalServerConfiguration OverrideServiceConfiguration(ServiceConfiguratio
         public string GrpcPortFileName { get; set; } = DefaultFileName;
 
         /// <nodoc />
-        public int GrpcPoolSize { get; private set; } = 70;
+        public int? GrpcPoolSize { get; set; } = 70;
```
SergeyTeplyakov: I think we should remove all the '70' hardcoded in other places and just rely on the default constant.

PR #1027, opened 2019-10-10
Better error handling for setxattr

almiliMSFT:
- check if file has write permissions before attempting to set xattrs
  - log more information with setting/getting xattrs fails

[AB#1607996](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1607996)

### DIFF
--- a/Public/Src/Utilities/Native/IO/Unix/FileUtilities.Unix.cs
+++ b/Public/Src/Utilities/Native/IO/Unix/FileUtilities.Unix.cs
@@ -668,7 +668,7 @@ namespace BuildXL.Native.IO.Unix
                 currentPermissions &= ~permissions;
             }
 
-            result = SetFilePermissionsForFilePath(path, currentPermissions);
+            result = SetFilePermissionsForFilePath(path, currentPermissions, followSymlink: false);
 
             if (result < 0)
             {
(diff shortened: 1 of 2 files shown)


Public/Src/Utilities/Native/IO/Unix/FileUtilities.Unix.cs
```diff
@@ -668,7 +668,7 @@ public void SetFileAccessControl(string path, FileSystemRights fileSystemRights,
                 currentPermissions &= ~permissions;
             }
 
-            result = SetFilePermissionsForFilePath(path, currentPermissions);
+            result = SetFilePermissionsForFilePath(path, currentPermissions, followSymlink: false);
```
smera: >followSymlink: false) [](start = 77, length = 21)

maybe a comment for why we need to do this?

almiliMSFT: will add

almiliMSFT:
```suggestion
            // don't follow symlinks because if we do that we may end up modifying source files via symlink outputs
            result = SetFilePermissionsForFilePath(path, currentPermissions, followSymlink: false);
```

PR #1029, opened 2019-10-10
Journal for shared opaque outputs

almiliMSFT:
During build: all observed writes into shared opaque directories are recorded in sideband, per-pip, log files.  They are recorded as soon as they are observed.

During scrubbing (before build): all such log files are discovered, scanned, and all recorded files are deleted.  This is consistent with the current behavior of unconditionally scrubbing all shared opaque outputs.  In the future, this process can become smarter and, for example, delay scrubbing until pip execution and then selectively scrubbing outputs of individual pips before they are executed.


Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs
```diff
@@ -59,8 +60,14 @@ public void SharedOpaqueDirectoryConsumptionCachingBehavior()
             var pipB = SchedulePipBuilder(builderB);
 
             // B should be able to consume the file in the opaque directory. Second build should have both cached
-            RunScheduler().AssertCacheMiss(pipA.Process.PipId, pipB.Process.PipId);
+            var result = RunScheduler().AssertCacheMiss(pipA.Process.PipId, pipB.Process.PipId);
+            AssertWritesJournaled(result, pipA, outputInSharedOpaque);
```
almiliMSFT: For now I only added this single integration test.

I'll try to hijack some low-level test function and there uniformly check that all outputs into shared opaque directories are properly journaled.

Public/Src/Engine/Dll/DirectoryScrubber.cs
```diff
@@ -111,23 +75,67 @@ public static IEnumerable<string> CollapsePaths(IEnumerable<string> paths)
...
+            HashSet<string> blockedPaths,
+            HashSet<string> nonDeletableRootDirectories,
+            MountPathExpander mountPathExpander = null)
+        { 
             int directoriesEncountered = 0;
             int filesEncountered = 0;
             int filesRemoved = 0;
             int directoriesRemovedRecursively = 0;
 
+
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -585,6 +582,8 @@ private bool FileAccessReportLineReceived<T>(ref T data, FileAccessReportProvide
                 return true;
             }
 
+
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -585,6 +582,8 @@ private bool FileAccessReportLineReceived<T>(ref T data, FileAccessReportProvide
                 return true;
             }
 
+
+
```
almiliMSFT:
```suggestion
```

Public/Src/Tools/SandboxExec/Program.cs
```diff
@@ -374,7 +374,8 @@ public List<string> DedupeAccessReports(params ISet<ReportedFileAccess>[] access
         /// <returns>SandboxedProcessInfo object that is configured to explicitly report all observed file accesses</returns>
         public static SandboxedProcessInfo CreateSandboxedProcessInfo(string processFileName, SandboxExecRunner instance)
         {
-            var sandboxProcessInfo = new SandboxedProcessInfo(fileStorage: instance, fileName: processFileName, disableConHostSharing: true, sandboxConnection: instance.m_sandboxConnection);
+            var sandboxProcessInfo = new SandboxedProcessInfo(
+                new PathTable(), fileStorage: instance, fileName: processFileName, disableConHostSharing: true, sandboxConnection: instance.m_sandboxConnection);
```
smera: >new PathTable() [](start = 16, length = 15)

not getting the whole context yet, but the fact that you are creating a new path table sounds fishy. Can't you use the single path table that always floats around?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs
smera: I would put this in its own integration test. Now it is mixed up with a test that checks caching behavior

---
In reply to: [333684941](https://github.com/microsoft/BuildXL/pull/1029#discussion_r333684941) [](ancestors = 333684941)

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -771,21 +771,46 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
...
+
+            var journalFiles = SharedOpaqueJournal
+                .FindAllJournalFiles(configuration.Layout.SharedOpaqueJournalDirectory.ToString(scheduler.Context.PathTable))
+                .ToArray();
+            var distinctRecordedWrites = journalFiles
+                .AsParallel()
+                .WithDegreeOfParallelism(Environment.ProcessorCount)
+                .WithCancellation(scheduler.Context.CancellationToken)
+                .SelectMany(SharedOpaqueJournal.ReadRecordedWritesFromJournalWrapExceptions)
+                .Distinct(StringComparer.OrdinalIgnoreCase)
```
smera: >Distinct(StringComparer.OrdinalIgnoreCase) [](start = 17, length = 42)

Do we actually need Distinct() here?
My understanding is that duplicates can only occur when a double write happened (and the corresponding build failed). So it doesn't look like the most common case. An actual race trying to delete the same file is not a problem AFAIK. On the other hand, a Distinct() among lots of files could be time consuming

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -771,21 +771,46 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
...
+                .Where(FileUtilities.FileExistsNoFollow)
```
smera: >.Where(FileUtilities.FileExistsNoFollow) [](start = 16, length = 40)

And a similar comment here. Do you actually need to probe every single file? You could try to unconditionally delete it, and if the file is not there, we are good as well

Public/Src/Engine/Dll/DirectoryScrubber.cs
```diff
@@ -375,5 +372,50 @@ public bool RemoveExtraneousFilesAndDirectories(CancellationToken cancellationTo
...
+                dueTime: BuildXLEngine.GetTimerUpdatePeriodInMs(m_loggingConfiguration),
+                period: BuildXLEngine.GetTimerUpdatePeriodInMs(m_loggingConfiguration)))
+            {
+                filePaths
+                    .AsParallel()
+                    .WithDegreeOfParallelism(m_maxDegreeParallelism)
+                    .WithCancellation(m_cancellationToken)
+                    .ForAll(path =>
+                    {
+                        if (FileUtilities.FileExistsNoFollow(path) && TryScrubFile(m_loggingContext, path))
```
smera: >FileUtilities.FileExistsNoFollow(path) [](start = 28, length = 38)

Do you need to probe beforehand?

Public/Src/Engine/Dll/DirectoryScrubber.cs
```diff
@@ -375,5 +372,50 @@ public bool RemoveExtraneousFilesAndDirectories(CancellationToken cancellationTo
...
+                        {
+                            Interlocked.Increment(ref numRemoved);
+                        }
+                    });
+                Tracing.Logger.Log.ScrubbingFinished(m_loggingContext, 0, filePaths.Length, numRemoved, 0);
+                return numRemoved;
+            }
+        }
+
+        private bool TryScrubFile(LoggingContext loggingContext, string path)
```
smera: >TryScrubFile [](start = 21, length = 12)

nit: TryDeleteFile maybe?

Public/Src/Engine/Dll/DirectoryScrubber.cs
```diff
@@ -111,18 +75,61 @@ public static IEnumerable<string> CollapsePaths(IEnumerable<string> paths)
...
+            IEnumerable<string> blockedPaths,
+            IEnumerable<string> nonDeletableRootDirectories,
+            MountPathExpander mountPathExpander = null)
         {
+            var finalPathsToScrub = CollapsePaths(pathsToScrub).ToList();
+            var finalBlockedPaths = new HashSet<string>(blockedPaths, StringComparer.OrdinalIgnoreCase);
+            var finalNnonDeletableRootDirectories = new HashSet<string>(nonDeletableRootDirectories, StringComparer.OrdinalIgnoreCase);
+            if (mountPathExpander != null)
+            {
+                finalNnonDeletableRootDirectories.UnionWith(mountPathExpander.GetAllRoots().Select(p => p.ToString(mountPathExpander.PathTable)));
```
smera: >Nnon [](start = 21, length = 4)

Non

Public/Src/Engine/Dll/Tracing/Log.cs
```diff
@@ -1868,6 +1868,26 @@ public void BusyOrUnavailableOutputDirectories(LoggingContext context, string ob
...
         public abstract void ConfigUnsafeDisableSharedOpaqueEmptyDirectoryScrubbing(LoggingContext context);
 
+        [GeneratedEvent(
+            (int)EventId.ScrubbingOutputsFromJournalStarted,
+            EventGenerators = EventGenerators.LocalOnly,
+            EventLevel = Level.Informational,
+            Keywords = (int)(Keywords.UserMessage | Keywords.Overwritable),
+            EventTask = (int)Tasks.Engine,
+            EventOpcode = (byte)EventOpcode.Start,
+            Message = EventConstants.PhasePrefix + "Scrubbing shared opaque outputs explicitly recorded in the journal.")]
```
smera: >Scrubbing [](start = 52, length = 9)

I'd love that we eventually switch from 'scrub' to 'delete' as the term we expose to the end user when deleting shared opaque outputs.
I think it is just an implementation detail that we use the same 'scrubber' class for both the /scrub option and for deleting old shared opaque outputs. Bxl uses 'scrub' to mean 'remove extraneous files', but what we actually do for shared opaque outputs is just a deletion.
(and I know I'm personally to blame for leaking that term to being with :( )

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -67,24 +67,14 @@ public sealed class SandboxedProcessInfo
         public ISandboxConnection SandboxConnection;
 
         /// <summary>
-        /// Holds the path remapping information for a process that needs to run in a container
+        /// On optional journal where to record file writes under shared opaque directories as soon as they happen.
```
smera: >On [](start = 12, length = 2)

An

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -814,6 +816,14 @@ public async Task<SandboxedProcessPipExecutionResult> RunAsync(CancellationToken
             }
         }
 
+        private SharedOpaqueJournal CreateSharedOpaqueJournalIfConfigured()
+        {
+            // don't use this journal if pip's semistable hash is 0 or if journal root directory is not set up
```
smera: >if pip's semistable hash is 0 [](start = 38, length = 29)

What is the meaning of a process pip with semistable hash 0?

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -97,7 +98,8 @@ public int GetLastMessageCount()
             long pipSemiStableHash,
             string pipDescription,
             LoggingContext loggingContext,
-            IDetoursEventListener detoursEventListener = null)
+            IDetoursEventListener detoursEventListener,
+            SharedOpaqueJournal writeJournal)
```
smera: >SharedOpaqueJournal [](start = 12, length = 19)

maybe a [CanBeNull] here?

Public/Src/Tools/SandboxExec/Program.cs
smera: Ok, ignore my comment above, just saw that you just moved existing code around

---
In reply to: [334109096](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334109096) [](ancestors = 334109096)

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
```diff
@@ -0,0 +1,183 @@
...
+using System.IO;
+using System.Linq;
+using BuildXL.Pips.Operations;
+using BuildXL.Utilities;
+using BuildXL.Utilities.Collections;
+
+namespace BuildXL.Processes
+{
+    /// <summary>
+    /// TODO
```
smera: >/// TODO  [](start = 4, length = 9)

just, in case, TODO :)

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
```diff
@@ -0,0 +1,183 @@
...
+        /// <param name="pathTable">Path table.</param>
+        /// <param name="rootDirectories">Only paths under one of the root directories are recorded in <see cref="RecordFileWrite(AbsolutePath)"/>.</param>
+        /// <param name="journalPath">File to which to write recorded accesses.</param>
+        public SharedOpaqueJournal(PathTable pathTable, IEnumerable<AbsolutePath> rootDirectories, AbsolutePath journalPath)
+        {
+            Contract.Requires(pathTable.IsValid);
+            Contract.Requires(rootDirectories != null);
+
+            m_pathTable = pathTable;
+            m_rootDirectories = rootDirectories.ToArray();
```
smera: >rootDirectories.ToArray(); [](start = 32, length = 26)

Maybe require an IReadOnlyCollection to begin with? So you avoid copies if the user of this constructor already has one at hand

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        /// <summary>
+        /// Finds and returns all journal files that exist in directory denoted by <paramref name="directory"/>
+        /// </summary>
+        public static IEnumerable<string> FindAllJournalFiles(string directory)
+        {
+            return Directory.Exists(directory)
+                ? Directory.EnumerateFiles(directory, "*", SearchOption.TopDirectoryOnly)
+                : CollectionUtilities.EmptyArray<string>();
+        }
```
smera: This looks slightly weird here, since there is in fact no logic to distinguish a journal file from one that is not. If I'm getting this right, the assumption is that all journal files are placed flat under a well-known directory.
So it doesn't look like the responsibility of this class to know that. I'd move this logic to whoever decides to put all journal files under the same directory (I guess it is EngineSchedule)

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
```diff
@@ -0,0 +1,183 @@
...
+        public static AbsolutePath GetJournalFileForProcess(PathTable pathTable, AbsolutePath journalDirectory, Process process)
+        {
+            Contract.Requires(journalDirectory.IsValid);
+            return journalDirectory.Combine(pathTable, process.FormattedSemiStableHash);
+        }
+
+        /// <summary>
+        /// Finds and returns all journal files that exist in directory denoted by <paramref name="directory"/>
+        /// </summary>
+        public static IEnumerable<string> FindAllJournalFiles(string directory)
```
smera: >FindAllJournalFiles [](start = 42, length = 19)

Another consideration is that the number of files in the same journal dir folder can grow to be a very big number, given that there is a 1:1 correspondence to the number of pips. In similar cases in the past we 'clustered' files in different subdirs, based on some grouping factor.
This is usually to avoid enumerations and other file operations that don't deal well with a huge amount of files in the same dir

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
```diff
@@ -0,0 +1,183 @@
...
+        /// <returns>
+        /// <code>true</code> if <paramref name="path"/> was recorded, <code>false</code> 
+        /// if the path was filtered out because of <see cref="m_rootDirectories"/>.
+        /// </returns>
+        /// <remarks>
+        /// NOT THREAD-SAFE.
+        /// </remarks>
+        public bool RecordFileWrite(AbsolutePath path)
+        {
+            bool isUnderSharedOpaqueDirectory = m_rootDirectories.Any(dir => path.IsWithin(m_pathTable, dir));
```
smera: >isUnderSharedOpaqueDirectory [](start = 17, length = 28)

This will record outputs that happen to fall under shared opaques but are not shared opaque outputs. Is there any logic to prevent bxl from deleting those?

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs
almiliMSFT: agreed; this was a temp solution just to have at least one check in an integration test

Public/Src/Tools/SandboxExec/Program.cs
almiliMSFT: exactly

Public/Src/Engine/Dll/EngineSchedule.cs
almiliMSFT: A path is written to a journal as soon as a Write operation is reported by the sandbox.  The same process can write multiple times to the same file, so yes, we end up having a ton of duplicates here.

almiliMSFT: I'm not sure about the perf implications and the behavior under race conditions.  Judging by a few examples I've seen, the number of entries in a journal can easily be 5x the number of distinct paths; calling File.Delete for each entry doesn't sound beneficial to me.  Also conceptually, if we don't deduplicate here, we cannot report how many files we actually wanted to delete and how many we actually deleted.

Public/Src/Engine/Dll/EngineSchedule.cs
almiliMSFT: will try

Public/Src/Engine/Dll/DirectoryScrubber.cs
almiliMSFT: I don't.  I did it only to get a precise count of deleted files.  If you call our FileUtilities.Delete method for a file that doesn't exists, it will return true.

Public/Src/Engine/Dll/DirectoryScrubber.cs
almiliMSFT: will change

Public/Src/Engine/Dll/Tracing/Log.cs
almiliMSFT: I'll change it here to delete

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
almiliMSFT: Some tests use that and I wanted to avoid running this on those tests. 

What ends up happening, in a single test there might be multiple pips with semi stable hash 0, and then we end up creating multiple journal objects concurrently writing to the same journal file (which is not good)

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
almiliMSFT: thanks :)

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
almiliMSFT: indeed

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
almiliMSFT: True.  And there is no logic to prevent BuidXL from deleting those.  But that's how it already is.  Such files are already getting deleted by the scrubbing logic because even today we set magic timestamps on *every* that lives inside a shared opaque directory (even if it is a statically declared one)

smera: I don't think that's true. We only flag shared opaques that are identified as such, not all output files

---
In reply to: [334136868](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334136868) [](ancestors = 334136868)

almiliMSFT: #96

Public/Src/Engine/Dll/EngineSchedule.cs
smera: That 5X surprises me... shared opaque outputs are enforced to be uniquely produced. So a duplicate file means a build break.

---
In reply to: [334131960](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334131960) [](ancestors = 334131960)

smera: And following the assumption that duplicate files are not common, I don't think reporting more files in that case is a big problem

---
In reply to: [334138336](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334138336) [](ancestors = 334138336,334131960)

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
smera: Just checked. We are using ExecutionResult.SharedDynamicDirectoryWriteAccesses, which only contains shared opaque outputs

---
In reply to: [334137271](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334137271) [](ancestors = 334137271,334136868)

Public/Src/Engine/Dll/EngineSchedule.cs
almiliMSFT: Each shared opaque output is indeed uniquely produced by a single pip; but that one pip can write to it as many times as it wants, and every time it writes to it we get a report about it from the sandbox.  The journal records reports from the sandbox as soon as they are received, hence duplication.

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
smera: So I don't think there is a big problem today with deleting *all* the produced outputs beforehand, because that will only happen to pips that have at least one shared opaque (right?).
But maybe a slightly more complicated solution, but that will accrue to the general 'delayed' deletion that we want to eventually go to could be:
1) after flagging all shared opaques as such, after the pip is done, add a marker to the journal file indicating the pip finished successfully
2) On deletion, if the marker is there, only remove the files that you see in the journal AND have the shared opaque marker
3) On deletion, if you don't see the marker for a given journal file, delete all. The pip failed the last time anyway.

---
In reply to: [334139103](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334139103) [](ancestors = 334139103,334137271,334136868)

Public/Src/Engine/Dll/EngineSchedule.cs
smera: Ah, you are correct, yes. You could consider keeping a set of already written paths and check before writing again. We are already keeping similar sets/maps as such, from what I could see.

---
In reply to: [334130813](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334130813) [](ancestors = 334130813)

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
almiliMSFT: Here is a [unit test](https://github.com/microsoft/BuildXL/blob/f3be55c8bfa386fb03db4e29a991250755e75c7d/Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/SharedOpaqueDirectoryTests.cs#L1399) that checks that all outputs that fall under a shared opaque directory are marked as shared opaque outputs.

smera: Ah, you are right, forgot that we made that change. Indeed both static and dynamic outputs are flagged. That might not be the case for exclusive opaques though, but I don't think that's very impactful

---
In reply to: [334139103](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334139103) [](ancestors = 334139103,334137271,334136868)

Public/Src/Engine/Dll/EngineSchedule.cs
smera: That would also reduce the number of writes/size of those files (which from what you are seeing seems to be 5X what they could be)

---
In reply to: [334141163](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334141163) [](ancestors = 334141163,334130813)

almiliMSFT: made SharedOpaqueJournal dedupe paths on the fly, before writing them to disk

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
almiliMSFT: added a comment about it

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -99,7 +89,8 @@ public sealed class SandboxedProcessInfo
             bool testRetries = false,
             LoggingContext loggingContext = null,
             IDetoursEventListener detoursEventListener = null,
-            ISandboxConnection sandboxConnection = null)
+            ISandboxConnection sandboxConnection = null,
+            SharedOpaqueJournal sharedOpaqueOutputsJournal = null)
```
narasamdya: >SharedOpaqueJournal sharedOpaqueOutputsJournal  [](start = 12, length = 47)

This has to be made serializable to be able to run in VM.

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
almiliMSFT: In the end I left it here because it should be consistent with `GetJournalFileForProcess`.

Here's what I did to make it better:
  - `GetJournalFileForProcess` returns a path like "Pip{SemiStableHash}.journal"
  - this method searches for all "Pip*.journal" files
  - added a CODESYNC comment

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
almiliMSFT: added subduers

narasamdya:
I'm thinking of renaming Journal to TransactionLog to not confuse it with USN change journal.

Public/Src/Engine/Dll/EngineSchedule.cs
almiliMSFT: removed probing from here

Public/Src/Engine/Dll/DirectoryScrubber.cs
almiliMSFT: The problem is that `FileUtilities.DeleteFile` is not to be trusted. :) In fact, if you give it a path to a folder, it will delete that folder (maybe only if that folder is empty, but it doesn't matter).

I think probing for file existence here is a small price to pay.

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
almiliMSFT: exactly, outputs in exclusive opaques are not flagged, and that doesn't change with this PR either

almiliMSFT:
> I'm thinking of renaming Journal to TransactionLog to not confuse it with USN change journal.

I'm happy to rename, but `TransactionLog` is not kind of misleading.  How about `WriteLog`?

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
almiliMSFT: I saw that but I was kind of hoping this wouldn't be necessary.

I'll make it serializable

smera:
What about OutputLog, or PipOutputLog?

---
In reply to: [541232957](https://github.com/microsoft/BuildXL/pull/1029#issuecomment-541232957) [](ancestors = 541232957)

Public/Src/Engine/Processes/SandboxedProcessInfo.cs
almiliMSFT: implemented serialization

almiliMSFT:
> What about OutputLog, or PipOutputLog?

fine with me

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -110,6 +117,7 @@ public int GetLastMessageCount()
             FileUnexpectedAccesses = new HashSet<ReportedFileAccess>();
             m_manifest = manifest;
             m_detoursEventListener = detoursEventListener;
+            m_writeJournal = writeJournal;
```
almiliMSFT: rename to `outputLog`

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
```diff
@@ -0,0 +1,227 @@
...
+        /// When <code>null</code>, all paths are recorded.
+        /// </summary>
+        [CanBeNull]
+        public IReadOnlyList<string> RootDirectories { get; }
+
+        /// <summary>
+        /// Creates a journal for a given process.
+        /// 
+        /// Shared opaque directory outputs of <paramref name="process"/> are used as root directories and
+        /// <see cref="Pip.FormattedSemiStableHash"/> is used as journal base name.
```
almiliMSFT: fix `<see />`

Public/Src/Engine/Processes/SharedOpaqueJournal.cs
```diff
@@ -0,0 +1,227 @@
...
+                return new BuildXLWriter(
+                    stream: new FileStream(JournalPath, FileMode.Create, FileAccess.Write, FileShare.Read | FileShare.Delete),
+                    debug: false,
+                    logStats: false,
+                    leaveOpen: false);
+            });
+        }
+
+        /// <summary>
+        /// Given a root directory (<paramref name="journalDirectory"/>), returns the full path to the journal corresponding to process <paramref name="process"/>.
```
almiliMSFT:
```suggestion
        /// Given a root directory (<paramref name="journalDirectory"/>), returns the full path to the journal file corresponding to process <paramref name="process"/>.
```

Public/Src/Engine/UnitTests/Engine/SharedOpaqueOutputLoggerTests.cs
```diff
@@ -0,0 +1,240 @@
...
+            var path = Path.Combine(TemporaryDirectory, "path1");
+            var logFile = Path.Combine(myDir, "Pip0");
+            
+            CreateOutputLoggerAndRecordPaths(logFile, new[] { path, path, path });
+
+            XAssert.ArrayEqual(
+                new[] { path },
+                ReadRecordedPathsFromSharedOpaqueOutputLog(logFile).ToArray());
+        }
+
```
Kr1sso: Nit: remove empty lines

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Those paths are expected to be absolute paths of files/directories that were written to by the previous build.
+        ///
+        /// NOTE: this method does not validate the recorded paths in any way.  That means that each returned string may be
+        ///   - a path pointing to an absent file
+        ///   - a path pointing to a file
+        ///   - a path pointing to a directory.
+        /// 
+        /// NOTE: unless the log file was corrupted (i.e., it was produced by an instance of this class):
```
Kr1sso: Should this say `wasn't` because only other impls could corrupt the file?

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+        public bool RecordFileWrite(PathTable pathTable, string absolutePath)
+            => RecordFileWrite(pathTable, AbsolutePath.Create(pathTable, absolutePath));
+
+        /// <summary>
+        /// Records that the file at location <paramref name="path"/> was written to.
+        /// 
+        /// Returns whether the path was recorded or skipped.
+        /// </summary>
+        /// <returns>
+        /// <code>true</code> if <paramref name="path"/> was recorded, <code>false</code>
```
Kr1sso: This is confusing, could you add a remark saying it should be recorded if the path is within any given root directory?

Public/Src/Engine/Dll/DirectoryScrubber.cs
```diff
@@ -111,18 +75,61 @@ public static IEnumerable<string> CollapsePaths(IEnumerable<string> paths)
         /// <remarks>
         /// A directory path is valid to scrub if it is under a scrubbable mount.
         /// </remarks>
-        private bool ValidateDirectory(string directory, out SemanticPathInfo foundSemanticPathInfo)
+        private bool ValidateDirectory(MountPathExpander mpe, string directory, out SemanticPathInfo foundSemanticPathInfo)
```
narasamdya: >mpe [](start = 57, length = 3)

nit: Given other argument is in understandable name, I think, instead of `mpe`, you should name it `expander` or `mountPathExpander`.

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: Rephrased it.
```suggestion
        /// NOTE: if the log file was produced by an instance of this class (and wasn't corrupted in any way)
```

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -771,21 +772,42 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
...
             }
 
+            var scrubber = new DirectoryScrubber(
+                cancellationToken: scheduler.Context.CancellationToken,
+                loggingContext: loggingContext,
+                loggingConfiguration: configuration.Logging,
+                maxDegreeParallelism: Environment.ProcessorCount,
+                tempDirectoryCleaner: tempCleaner);
+
+            var journalFiles = SharedOpaqueOutputLogger.FindAllProcessPipOutputLogFiles(configuration.Layout.SharedOpaqueSidebandDirectory.ToString(scheduler.Context.PathTable));
```
narasamdya: nit: `journalFiles` -> `sharedOpaqueLogFiles`.

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        /// <summary>
+        /// <see cref="RecordFileWrite(PathTable, AbsolutePath)"/>
+        /// </summary>
+        public bool RecordFileWrite(PathTable pathTable, string absolutePath)
+            => RecordFileWrite(pathTable, AbsolutePath.Create(pathTable, absolutePath));
+
+        /// <summary>
+        /// Records that the file at location <paramref name="path"/> was written to.
+        ///
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// <summary>
+        /// <see cref="RecordFileWrite(PathTable, AbsolutePath)"/>
+        /// </summary>
+        public bool RecordFileWrite(PathTable pathTable, string absolutePath)
+            => RecordFileWrite(pathTable, AbsolutePath.Create(pathTable, absolutePath));
+
+        /// <summary>
+        /// Records that the file at location <paramref name="path"/> was written to.
+        /// 
+        /// Returns whether the path was recorded or skipped.
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT:
```suggestion
        /// <code>true</code> if <paramref name="path"/> is within any given root directory (<see cref="RootDirectories"/>) and hence was recorded; <code>false</code> otherwise.
```

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// </summary>
+        /// <returns>
+        /// <code>true</code> if <paramref name="path"/> was recorded, <code>false</code> 
+        /// if the path was filtered out because of <see cref="RootDirectories"/>.
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/UnitTests/Engine/SharedOpaqueOutputLoggerTests.cs
almiliMSFT:
```suggestion
```

Public/Src/Engine/UnitTests/Engine/SharedOpaqueOutputLoggerTests.cs
```diff
@@ -0,0 +1,240 @@
...
+
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+                {
+                    yield return nextString;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Same as <see cref="ReadRecordedPathsFromSharedOpaqueOutputLog(string)"/> except that all exceptions are wrapped in <see cref="BuildXLException"/>
+        /// </summary>
+        public static string[] ReadRecordedPathsFromSharedOpaqueOutputLogWrapExceptions(string filePath)
```
narasamdya: Why do we need two public user facing API? Why can't we just have `ReadRecordedPathsFromSharedOpaqueOutputLogWrapExceptions` ?

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// <remarks>
+        /// NOT THREAD-SAFE.
+        /// </remarks>
+        public bool RecordFileWrite(PathTable pathTable, AbsolutePath path)
+        {
+            if (RootDirectories == null || RootDirectories.Any(dir => path.IsWithin(pathTable, AbsolutePath.Create(pathTable, dir))))
+            {
+                if (m_recordedPathsCache.Add(path))
+                {
+                    m_lazyBxlWriter.Value.Write(path.ToString(pathTable));
```
narasamdya: >m_lazyBxlWriter.Value [](start = 20, length = 21)

Any reason why the writer is made lazy?

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+                logPath: reader.ReadString(),
+                rootDirectories: reader.ReadNullable(r => r.ReadReadOnlyList(r2 => r2.ReadString())));
+        }
+        #endregion
+
+        private static string ReadStringOrNull(BuildXLReader bxlReader)
+        {
+            try
+            {
+                return bxlReader.ReadString();
```
narasamdya: >ReadString [](start = 33, length = 10)

Should this be `ReadNullable<string>` ?

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
narasamdya: >corrupted [](start = 42, length = 9)

We typically use `FileEnvelope` to determine if the file is corrupted or not. See: [http://index/?leftProject=BuildXL.Utilities&leftSymbol=fyoqpbmtsdt6&file=FileEnvelopeId.cs](http://index/?leftProject=BuildXL.Utilities&leftSymbol=fyoqpbmtsdt6&file=FileEnvelopeId.cs)

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: no, this is `ReadString` or return `null` if failed to read string for any reason

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: Yes, the reason is that the tests can use non-existing file names if they are not going to write anything to this log.

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: will change

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// <param name="rootDirectories">Only paths under one of the root directories are recorded in <see cref="RecordFileWrite(PathTable, AbsolutePath)"/>.</param>
+        public SharedOpaqueOutputLogger(string logPath, [CanBeNull] IReadOnlyList<string> rootDirectories)
+        {
+            LogPath = logPath;
+            RootDirectories = rootDirectories;
+            m_recordedPathsCache = new HashSet<AbsolutePath>();
+
+            m_lazyBxlWriter = Lazy.Create(() =>
+            {
+                Directory.CreateDirectory(Path.GetDirectoryName(LogPath));
```
narasamdya: >D [](start = 16, length = 1)

Ah, I see why you made the writer lazy. For robustness, some IO exception should be caught and reported here.

almiliMSFT: Exactly

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        /// <summary>
+        /// Creates a new output logger for a given process.
+        /// 
+        /// Shared opaque directory outputs of <paramref name="process"/> are used as root directories and
+        /// <see cref="GetOutputLogFileForProcess"/> is used as log base name.
+        ///
+        /// <seealso cref="SharedOpaqueOutputLogger(string, IReadOnlyList{string})"/>
+        /// </summary>
+        public SharedOpaqueOutputLogger(PipExecutionContext context, Process process, AbsolutePath logDirectory)
```
narasamdya: I believe this class can be made free from `PipExecutionContext`, `Process`, and `PathTable` (as well as `AbsolutePath`). 

The absolute path is only used when you want to check if a path is inside a directory. But in that case, can you not use check for substring?

Public/Src/Engine/Dll/DirectoryScrubber.cs
almiliMSFT: changed

Public/Src/Engine/Dll/EngineSchedule.cs
almiliMSFT: changed

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,233 @@
...
+    /// <summary>
+    /// Responsible for keeping a log of all paths written to by a given process pip.
+    /// 
+    /// Paths are flushed to an underlying file as soon as they're reported via <see cref="RecordFileWrite(PathTable, AbsolutePath)"/>.
+    /// 
+    /// A number of root directories may be set in which case the log will record only 
+    /// those paths that fall under one of those directories.  The typical use case is to
+    /// use all shared opaque directory outputs of a process as root directories.
+    /// </summary>
+    /// <remarks>
```
narasamdya: ><remarks> [](start = 8, length = 9)

Can you make some remarks here stating that no  path table can be used here because the class has to be made serializable for VM execution.

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: I removed the WrapExceptions method.

The reason why left the other one is because the other one can be a true iterator (i.e., it's not possible to yield a value from a try/catch block)

narasamdya:
Can you add a simple test here to make sure that this feature is compatible with VM execution?
[http://index/?leftProject=IntegrationTest.BuildXL.Scheduler&file=ExternalToolExecutionTests.cs](http://index/?leftProject=IntegrationTest.BuildXL.Scheduler&file=ExternalToolExecutionTests.cs)

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: This class is already free of `PipExecutionContext`, `Process`, and `PathTable` (in the sense that it has no fields of any of those types); it's only this constructor that accepts those arguments, and the reason why this constructor exists is just convenience (since this logger will be created per process pip).

If you really don't want to see this constructor here, I can turn it into a static factory method and move it to PipExecutor (or wherever it's currently being called from)

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: will do

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,238 @@
...
+    /// NOTE: not thread-safe
+    /// 
+    /// NOTE: must be serializable in order to be compatible with VM execution; for this 
+    ///       reason, this class must not have a field of type <see cref="PathTable"/>.
+    /// </remarks>
+    public sealed class SharedOpaqueOutputLogger : IDisposable
```
smera: >SharedOpaqueOutputLogger [](start = 24, length = 24)

is there any logic in place to prevent us from file corruption? Thinking in the case where the writer dies half way through writing to the sideband file. And a next build trying to read from it and getting some unexpected/malformed paths

almiliMSFT: There isn't at the moment.  I'll look into using FileEnvelope, as Iman suggested.

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
```diff
@@ -0,0 +1,333 @@
...
+        public static string[] ReadRecordedPathsFromSidebandFile(string filePath)
+        {
+            if (!File.Exists(filePath))
+            {
+                return CollectionUtilities.EmptyArray<string>();
+            }
+
+            using (var reader = CreateSidebandFileReader(filePath))
+            {
+                reader.ReadHeader(ignoreChecksum: true);
```
smera: >ReadHeader [](start = 23, length = 10)

Considering this, I'd add an <exception > tag in the comments explaining that an exception will be thrown if the sideband file is corrupted

Public/Src/Engine/Dll/EngineSchedule.cs
```diff
@@ -771,21 +772,43 @@ private bool ShouldSerializeOptimizationDataStructurePostExecution()
...
+                maxDegreeParallelism: Environment.ProcessorCount,
+                tempDirectoryCleaner: tempCleaner);
+
+            var sharedOpaqueSidebandDirectory = configuration.Layout.SharedOpaqueSidebandDirectory.ToString(scheduler.Context.PathTable);
+            var sharedOpaqueSidebandFiles = SharedOpaqueOutputLogger.FindAllProcessPipSidebandFiles(sharedOpaqueSidebandDirectory);
+            var distinctRecordedWrites = sharedOpaqueSidebandFiles
+                .AsParallel()
+                .WithDegreeOfParallelism(Environment.ProcessorCount)
+                .WithCancellation(scheduler.Context.CancellationToken)
+                .SelectMany(SharedOpaqueOutputLogger.ReadRecordedPathsFromSidebandFile)
```
smera: >.SelectMany(SharedOpaqueOutputLogger.ReadRecordedPathsFromSidebandFile) [](start = 16, length = 71)

Don't you need a catch block here for the corrupted file case? (and I think you could limit it to a particular file being corrupted, and let the other potential sideband files to be honored)

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: Added FileEnvelope for checking file integrity.  However, there is an option to read whatever we can even if the file wasn't closed cleanly, because that was one of the indented use cases for this to begin with (e.g., bxl crashes, doesn't mark shared opaque, and doesn't cleanly close sideband files either)

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: Added FileEnvelope for checking file integrity.  However, there is an option to read whatever we can even if the file wasn't closed cleanly, because that was one of the indented use cases for this to begin with (e.g., bxl crashes, doesn't mark shared opaque, and doesn't cleanly close sideband files either)

Public/Src/Engine/Processes/SharedOpaqueOutputLogger.cs
almiliMSFT: This doesn't throw (because of `ignoreChecksum: true`).

Public/Src/Engine/Dll/EngineSchedule.cs
almiliMSFT: `SharedOpaqueOutputLogger.ReadRecordedPathsFromSidebandFile` shouldn't throw unless something really really bad happens (in which case we should probably know about it and not swallow it)

smera: ok, I guess I was missing what happens when the file is corrupted then (which I thought it would be about ignoring the corrupted file)

---
In reply to: [334722307](https://github.com/microsoft/BuildXL/pull/1029#discussion_r334722307) [](ancestors = 334722307)

PR #1033, opened 2019-10-11
Explicitly set u+w before calling setxattr

almiliMSFT:


Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -134,34 +174,13 @@ public static bool IsSharedOpaqueOutput(string expandedPath)
         /// <exception cref="BuildXLException">When unsuccessful</exception>
         public static void SetPathAsSharedOpaqueOutput(string expandedPath)
         {
-            // In the case of a no replay, this case can happen if the file got into the cache as a static output,
```
almiliMSFT: There was a reason why before any of my changes this was only being called when running on Windows: the implementation for Unix simply does something else; there are some cache components that depend on that behavior so I couldn't just changed the implementation for Unix, so instead I implemented it in `Unix.SetPathAsSharedOpaqueOutput`

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -33,6 +33,16 @@ private static class Win
             /// <exception cref="BuildXLException">When the timestamp cannot be set</exception>
             public static void SetPathAsSharedOpaqueOutput(string expandedPath)
             {
+                // In the case of a no replay, this case can happen if the file got into the cache as a static output,
```
almiliMSFT: just reverting this to how it originally was: call it only here for Windows instead of calling it uniformly regardless of the current platform.

PR #1034, opened 2019-10-11
Feature complete unit tests for binary graph fragments

narasamdya:
[AB#1604485](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1604485)
[AB#1604486](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1604486)


Public/Src/Engine/UnitTests/Scheduler/ArgumentsBuilder.cs
```diff
@@ -10,52 +10,221 @@
 
 namespace Test.BuildXL.Scheduler
 {
+    /// <summary>
+    /// Interface for building argument data.
+    /// </summary>
+    internal interface IArgumentsDataBuilder
```
erickulcyk: >IArgumentsDataBuilder [](start = 23, length = 21)

should this be in it's own file?

Public/Src/Engine/UnitTests/Scheduler/TestPipGraphFragment.cs
```diff
@@ -81,42 +88,56 @@ public TestPipGraphFragment(LoggingContext loggingContext, string sourceRoot, st
...
+                .Serialize(AbsolutePath.Create(Context.PathTable, path), PipGraph, m_moduleId.Value.ToString(Context.StringTable), useTopSortSerialization: m_useTopSort);
+
+        ///// <summary>
+        ///// Serializes this instance of pip graph fragment with top-sort to a file.
+        ///// </summary>
+        //public void SerializeTopSort(string path) =>
+        //    new PipGraphFragmentSerializer(
+        //        Context,
+        //        new PipGraphFragmentContext())
+        //        .SerializeTopSort(AbsolutePath.Create(Context.PathTable, path), m_pipGraph.RetrieveScheduledPips().ToList(), m_moduleId.Value.ToString(Context.StringTable));
```
erickulcyk: is this needed

erickulcyk:
It looks like nothing of substance was changed in the non-test code, right?  As in, we don't need an fixes before releasing.

Public/Src/Engine/UnitTests/Scheduler/ArgumentsBuilder.cs
narasamdya: When things become complex, I'll move it to a separate file. But this one is only used by unit tests. I don't feel like introducing too many new files if not necessary (e.g., for refactoring).

---
In reply to: [334140655](https://github.com/microsoft/BuildXL/pull/1034#discussion_r334140655) [](ancestors = 334140655)

narasamdya:
The unit tests confirm that no logic changes for binary graphs.

---
In reply to: [541194011](https://github.com/microsoft/BuildXL/pull/1034#issuecomment-541194011) [](ancestors = 541194011)

PR #1042, opened 2019-10-14
Untrack generated nuget props files for MSBuild-scheduled pips

smera:
Nuget restore generates nuget.g.props files with absolute paths embedded. This blocks shared cache from working on machines were those paths are not uniform. This PR makes the MSBuild pip construction process to untrack those.

### DIFF
--- a/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
+++ b/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
@@ -330,6 +330,16 @@ namespace BuildXL.FrontEnd.MsBuild
                     continue;
                 }
 
+                // Nuget restore generates nuget.g.props files that have absolute paths embedded. This blocks shared cache from working when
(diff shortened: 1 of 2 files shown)


Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -330,6 +330,16 @@ public bool TrySchedulePipForFile(ProjectWithPredictions project, QualifierId qu
                     continue;
                 }
 
+                // Nuget restore generates nuget.g.props files that have absolute paths embedded. This blocks shared cache from working when
+                // the project file (and therefore the corresponding .pkgrefgen folder) is placed on machine-dependent folders
+                // Even though untracking this file is not completely safe from a caching perspective, in practice this should be harmless
```
cdmihai: Harmless because any cache invalidation changes on the `nuget.g.props` would also imply changes to `project.assets.json`?

smera: Exactly

PR #1043, opened 2019-10-14
Retry when pipe is busy on inheritable pipe creations

narasamdya:
The root cause of the bug is still unknown.

This change doesn't solve the bug, but try to address it by retry. If, after retry, the bug still occurs, then we will reach out to the Windows team.

[AB#1614763](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1614763)

### DIFF
--- a/Public/Src/Engine/Processes/Internal/Pipes.cs
+++ b/Public/Src/Engine/Processes/Internal/Pipes.cs
@@ -108,17 +108,48 @@ namespace BuildXL.Processes.Internal
                 readSideFlags |= FileFlagsAndAttributes.FileFlagOverlapped;
             }
 
-            OpenFileResult openReadSideResult = FileUtilities.TryCreateOrOpenFile(
-                pipeName,
-                FileDesiredAccess.GenericRead,
-                FileShare.None,
-                FileMode.Open,
-                readSideFlags,
-                out readHandle);
+            int maxRetry = 3;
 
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/Processes/Internal/Pipes.cs
```diff
@@ -108,17 +108,48 @@ public static void CreateInheritablePipe(PipeInheritance inheritance, PipeFlags
...
-                pipeName,
-                FileDesiredAccess.GenericRead,
-                FileShare.None,
-                FileMode.Open,
-                readSideFlags,
-                out readHandle);
+            int maxRetry = 3;
 
-            if (!openReadSideResult.Succeeded)
+            while (true)
```
almiliMSFT: don't we have some generic utility method for retrying an arbitrary action?

almiliMSFT: maybe not, or at least I couldn't find anything

Kr1sso: Actually `RetryOnFailure` from the IO helpers could be used here?

narasamdya: The retry logic here is a bit different from `RetryOnFailure`. We are doing wait for at most 5 sec in a loop, and then retry.

PR #1046, opened 2019-10-15
Implement slow reconciliation

JuanCarlosGI:
Reconciliation when resetting epoch is overloading the master in some cases. Implementing slow reconciliation so that we do this over a large period of time and take load off the master.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
@@ -1210,7 +1210,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
                             var allLocalStoreContent = allLocalStoreContentInfos
                                 .Select(c => (hash: new ShortHash(c.ContentHash), size: c.Size))
                                 .OrderBy(c => c.hash)
-                                .SkipWhile(hashWithSize => hashWithSize.hash < lastProcessedHash)
+                                .SkipWhile(hashWithSize => lastProcessedHash.HasValue && hashWithSize.hash < lastProcessedHash.Value)
                                 .ToList();
 
                             allLocalStoreContentCount = allLocalStoreContent.Count;
@@ -1244,6 +1244,8 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 
                             Counters[ContentLocationStoreCounters.Reconcile_AddedContent].Add(addedContent.Count);
                             Counters[ContentLocationStoreCounters.Reconcile_RemovedContent].Add(removedContent.Count);
+                            totalAddedContent += addedContent.Count;
+                            totalRemovedContent += removedContent.Count;
 
                             // Only call reconcile if content needs to be updated for machine
                             if (addedContent.Count != 0 || removedContent.Count != 0)
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
@@ -1187,67 +1187,95 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 
                     token.ThrowIfCancellationRequested();
 
-                    // Pause events in main event store while sending reconciliation events via temporary event store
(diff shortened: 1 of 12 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1187,67 +1187,95 @@ public async Task<ReconciliationResult> ReconcileAsync(OperationContext context)
...
+                        using (EventStore.PauseSendingEvents())
+                        {
+                            var allLocalStoreContentInfos = await _localContentStore.GetContentInfoAsync(token);
+                            token.ThrowIfCancellationRequested();
 
-                        var allLocalStoreContent = allLocalStoreContentInfos.Select(c => (hash: new ShortHash(c.ContentHash), size: c.Size)).OrderBy(c => c.hash).ToList();
+                            var allLocalStoreContent = allLocalStoreContentInfos
+                                .Select(c => (hash: new ShortHash(c.ContentHash), size: c.Size))
+                                .OrderBy(c => c.hash)
+                                .SkipWhile(hashWithSize => hashWithSize.hash < lastProcessedHash)
```
ljcollins25: >hashWithSize.hash < lastProcessedHash [](start = 59, length = 37)

Not sure, but we need to make sure the comparison to nullable here is correct.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1187,67 +1187,95 @@ public async Task<ReconciliationResult> ReconcileAsync(OperationContext context)
...
+                    var totalRemovedContent = 0;
+                    var allLocalStoreContentCount = 0;
+                    ShortHash? lastProcessedHash = null;
+                    var isFinished = false;
+
+                    while (!isFinished)
                     {
-                        var allLocalStoreContentInfos = await _localContentStore.GetContentInfoAsync(token);
-                        token.ThrowIfCancellationRequested();
+                        var delayTask = Task.Delay(_configuration.ReconciliationCycleFrequency);
```
ljcollins25: >var delayTask = Task.Delay(_configuration.ReconciliationCycleFrequency); [](start = 24, length = 72)

Why create the delay task here?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1187,67 +1187,95 @@ public async Task<ReconciliationResult> ReconcileAsync(OperationContext context)
...
-                        MarkReconciled();
-                        return new ReconciliationResult(addedCount: addedContent.Count, removedCount: removedContent.Count, totalLocalContentCount: allLocalStoreContent.Count);
+                        if (!isFinished)
+                        {
+                            await delayTask;
+                        }
                     }
+
+                    MarkReconciled();
+                    return new ReconciliationResult(addedCount: totalAddedContent, removedCount: totalRemovedContent, totalLocalContentCount: allLocalStoreContentCount);
```
ljcollins25: >totalRemovedContent [](start = 97, length = 19)

Doesn't seem to be set.

ljcollins25: >totalAddedContent [](start = 64, length = 17)

Doesn't seem to be set.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
JuanCarlosGI: Wanted to make sure that we started the cycle exacly every {frequency}, instead of every {frequency + operation duration} in the case when I create the task after operation completion. I realize that probably the operation duration is fairly quick, but still I took the precaution.

---
In reply to: [335053567](https://github.com/microsoft/BuildXL/pull/1046#discussion_r335053567) [](ancestors = 335053567)

ljcollins25:
>         public async Task<ReconciliationResult> ReconcileAsync(OperationContext context)

Might want to write a test for this.

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs:1168 in bc14017. [](commit_id = bc1401717356ee2d3fbd3f9e8a3d1702fdda31c2, deletion_comment = False)

PR #1047, opened 2019-10-15
Add back a previously removed SandboxedProcessInfo constructor

almiliMSFT:
This fixes some compilation errors in an external project that already took a dependency on this constructor.


Public/Src/Engine/Processes/SandboxedProcessInfo.cs
```diff
@@ -76,6 +76,22 @@ public sealed class SandboxedProcessInfo
         /// </summary>
         public ContainerConfiguration ContainerConfiguration { get; }
 
+        /// <remarks>
+        /// This constructor is never used in this project, but there exist external projects that
```
almiliMSFT: > external projects

i.e., CloudBuild

PR #1053, opened 2019-10-16
Fallback logic for IsSharedOpaqueOutput on macOS

almiliMSFT:
The fallback logic means checking creation timestamp in addition to checking extended attributes.

This should accommodate smoothly transitioning between older versions (which used creation timestamps only) and newer versions (which use extended attributes).

### DIFF
--- a/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
+++ b/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
@@ -104,7 +104,7 @@ namespace BuildXL.Processes
             private static extern int SetXattr(
                 [MarshalAs(UnmanagedType.LPStr)] string path,
                 [MarshalAs(UnmanagedType.LPStr)] string name,
-                void* value,
+                void *value,
                 ulong size,
                 uint position,
                 int options);
@@ -113,7 +113,7 @@ namespace BuildXL.Processes
             private static extern long GetXattr(
                 [MarshalAs(UnmanagedType.LPStr)] string path,
                 [MarshalAs(UnmanagedType.LPStr)] string name,
-                void* value,
+                void *value,
                 ulong size,
                 uint position,
                 int options);
--- a/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
+++ b/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
@@ -160,6 +160,7 @@ namespace BuildXL.Processes
             /// </summary>
             public static bool IsSharedOpaqueOutput(string expandedPath) => IsSharedOpaqueOutputWithFallback(expandedPath, checkFallback: true);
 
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -104,7 +104,7 @@ private static unsafe class Unix
             private static extern int SetXattr(
                 [MarshalAs(UnmanagedType.LPStr)] string path,
                 [MarshalAs(UnmanagedType.LPStr)] string name,
-                void *value,
+                void* value,
```
almiliMSFT:
```suggestion
                void *value,
```

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -113,7 +113,7 @@ private static unsafe class Unix
             private static extern long GetXattr(
                 [MarshalAs(UnmanagedType.LPStr)] string path,
                 [MarshalAs(UnmanagedType.LPStr)] string name,
-                void *value,
+                void* value,
```
almiliMSFT:
```suggestion
                void *value,
```

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -158,12 +158,24 @@ public static void SetPathAsSharedOpaqueOutput(string expandedPath)
             /// Checks if the given path is an output under a shared opaque by checking if
             /// it contains extended attribute by <see cref="MY_XATTR_NAME"/> name.
             /// </summary>
-            public static bool IsSharedOpaqueOutput(string expandedPath)
+            public static bool IsSharedOpaqueOutput(string expandedPath) => IsSharedOpaqueOutputWithFallback(expandedPath, checkFallback: true);
+
+            private static bool IsSharedOpaqueOutputWithFallback(string expandedPath, bool checkFallback)
```
almiliMSFT:
```suggestion
            // TODO: delete the fallback logic after a successful transition from old to new logic
            private static bool IsSharedOpaqueOutputWithFallback(string expandedPath, bool checkFallback)
```

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -94,7 +94,7 @@ private static unsafe class Unix
             private const string MY_XATTR_NAME = "com.microsoft.buildxl:shared_opaque_output";
 
             // arbitrary value; in the future, we could store something more useful here (e.g., the producer PipId or something)
-            private const long MY_XATTR_VALUE = 42; 
+            private const long MY_XATTR_VALUE = 42;
```
Kr1sso: Can we rename this to bxl_sharedopaque_* and get rid of the my_ prefix?

PR #1055, opened 2019-10-16
Update msbuild

cdmihai:


Public/Src/Tools/Tool.MsBuildGraphBuilder/MsBuildGraphBuilder.cs
```diff
@@ -17,7 +17,7 @@
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Exceptions;
 using Microsoft.Build.Execution;
-using Microsoft.Build.Experimental.Graph;
+using Microsoft.Build.Graph;
```
smera: you should also bump the version defined in MSBuildAssemblyLoader, so we require the right version as the minimum one accepted.

PR #1057, opened 2019-10-16
Restore checkpoint in the background when current checkpoint is not old enough

JuanCarlosGI:
This should reduce startup time when a machine is being restarted.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
@@ -375,11 +375,11 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             }
             catch (Exception e)
             {
-                _tracer.Debug(context, $"Failed to write latest checkpoint state to disk: {e}");
+                _tracer.Warning(context, $"Failed to write latest checkpoint state to disk: {e}");
             }
         }
 
-        public (string id, DateTime checkpointTime)? GetLatestCheckpointInfo(OperationContext context)
+        public (string checkpointId, DateTime checkpointTime)? GetLatestCheckpointInfo(OperationContext context)
         {
             try
             {
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
@@ -42,6 +42,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         private readonly AbsolutePath _checkpointStagingDirectory;
         private readonly AbsolutePath _incrementalCheckpointDirectory;
         private readonly AbsolutePath _incrementalCheckpointInfoFile;
+        private readonly AbsolutePath _lastCheckpointFile;
 
         private CounterCollection<ContentLocationStoreCounters> Counters { get; }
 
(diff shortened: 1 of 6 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
```diff
@@ -344,7 +347,14 @@ public Task<BoolResult> RestoreCheckpointAsync(OperationContext context, string
                             }
 
                             // Restoring the checkpoint
-                            return _database.RestoreCheckpoint(context, extractedCheckpointDirectory);
+                            var result = _database.RestoreCheckpoint(context, extractedCheckpointDirectory);
+
+                            if (result)
+                            {
```
dcwuser: Comments please on what this is for.

SergeyTeplyakov:
>                         result = await RestoreCheckpointStateAsync(context, checkpointState);

Lets be explicit here and pass 'force: false'.

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs:429 in 373c47e. [](commit_id = 373c47ece75d8db3d5a27879d8ace0621d449095, deletion_comment = False)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -598,15 +598,34 @@ internal async Task<BoolResult> CreateCheckpointAsync(OperationContext context)
...
         {
             var token = context.Token;
+
+            if (!force)
+            {
+                var latestCheckpoint = _checkpointManager.GetLatestCheckpointInfo(context);
+                var latestCheckpointAge = _clock.UtcNow - latestCheckpoint?.checkpointTime;
+                var shouldRestoreInBackground = latestCheckpointAge < _configuration.Checkpoint.RestoreCheckpointAgeThreshold;
+
+                if (shouldRestoreInBackground)
```
SergeyTeplyakov: >shouldRestoreInBackground [](start = 20, length = 25)

Is it ok, that we trace here but still do the same thing regardless of the check?

SergeyTeplyakov: I think we need to explicitely trace that we skipped the eager checkpoint restore with the data about the checkpoint.

Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
```diff
@@ -356,6 +367,36 @@ public Task<BoolResult> RestoreCheckpointAsync(OperationContext context, string
...
 
+        private void WriteLatestCheckpointToFile(OperationContext context, CheckpointState checkpointState)
+        {
+            try
+            {
+                _fileSystem.WriteAllText(_lastCheckpointFile, $"{checkpointState.CheckpointId},{checkpointState.CheckpointTime}");
+            }
+            catch (Exception e)
+            {
+                _tracer.Debug(context, $"Failed to write latest checkpoint state to disk: {e}");
```
SergeyTeplyakov: nit: I think it should be a warning. Technically, this should never happen...

Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
```diff
@@ -356,6 +367,36 @@ public Task<BoolResult> RestoreCheckpointAsync(OperationContext context, string
...
+            }
+        }
+
+        public (string id, DateTime checkpointTime)? GetLatestCheckpointInfo(OperationContext context)
```
SergeyTeplyakov: >GetLatestCheckpointInfo [](start = 53, length = 23)

nit: checkpointId?

The naming looks a bit inconsistent. It should be either 'checkpointId and checkpointTime' or 'id' and 'time', imo.

PR #1060, opened 2019-10-17
Add tracing and propagate cancellationToken in reconciliation cycle

JuanCarlosGI:
Add tracing and propagate cancellationToken in reconciliation cycle

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
@@ -1195,10 +1195,25 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 
                     while (!isFinished)
                     {
-                        var delayTask = Task.Delay(_configuration.ReconciliationCycleFrequency);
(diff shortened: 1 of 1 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1195,10 +1195,25 @@ public async Task<ReconciliationResult> ReconcileAsync(OperationContext context)
 
                     while (!isFinished)
                     {
-                        var delayTask = Task.Delay(_configuration.ReconciliationCycleFrequency);
+                        var delayTask = Task.Delay(_configuration.ReconciliationCycleFrequency, context.Token);
```
SergeyTeplyakov: One minor thing: does it make sense to remove 'delayTask' local altogether and just have:

```
if (!isFinished) await Task.Delay(...)
```

JuanCarlosGI: I just wanted to make sure that we really did cycles every 30 minutes, instead of every (30 minutes + cycle duration)

PR #1061, opened 2019-10-17
Fix Crash in ProcessDumper. Only dump pids under same username

mpysson:
Fixes [AB#1613553](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1613553)
Fixes [AB#1616824](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1616824)

### DIFF
--- a/Public/Src/Engine/Processes/ProcessDumper.cs
+++ b/Public/Src/Engine/Processes/ProcessDumper.cs
@@ -11,6 +11,7 @@ using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Management;
+using System.Net;
 using System.Runtime.InteropServices;
 using BuildXL.Native.IO;
 using BuildXL.Native.Processes.Windows;
@@ -28,7 +29,7 @@ namespace BuildXL.Processes
         /// </summary>
         private static readonly object s_dumpProcessLock = new object();
 
-        private static readonly string[] s_skipProcesses = {
+        private static readonly HashSet<string> s_skipProcesses = new HashSet<string>() {
             "conhost", // Conhost dump causes native error 0x8007012b (Only part of a ReadProcessMemory or WriteProcessMemory request was completed) - Build 1809
         };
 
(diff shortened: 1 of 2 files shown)


Public/Src/Engine/Processes/ProcessDumper.cs
```diff
@@ -199,7 +199,21 @@ public static bool TryDumpProcessAndChildren(int parentProcessId, string dumpDir
...
+                }
+                catch (InvalidOperationException)
+                {
+                    if (!p.HasExited)
+                    {
+                        throw;
+                    }
+                }
+
+                if (processName == null || s_skipProcesses.Contains(processName, StringComparer.InvariantCultureIgnoreCase) || p.StartTime > treeDumpInitiateTime)
```
narasamdya: Can we change `s_skipProcess` to a set with `StringComparer.OrdinalIgnoreCase`?

Public/Src/Engine/Processes/ProcessDumper.cs
```diff
@@ -199,7 +199,21 @@ public static bool TryDumpProcessAndChildren(int parentProcessId, string dumpDir
...
+                string processName = null;
+                try
+                {
+                    processName = p.ProcessName;
+                }
+                catch (InvalidOperationException)
+                {
+                    if (!p.HasExited)
+                    {
+                        throw;
```
narasamdya: >throw; [](start = 24, length = 6)

Instead of throwing an exception that can crash BuildXL, would it be better to skip the process, but notifies (log or write a dummy file) user that dumping the process failed?

Public/Src/Engine/Processes/ProcessDumper.cs
```diff
@@ -254,53 +268,54 @@ public static bool TryDumpProcessAndChildren(int parentProcessId, string dumpDir
...
-            }
-
-            return result;
+            return GetChildProcessTreeIds(parentProcessId, maxTreeDepth, isRootQuery: true);
         }
 
-        internal static List<KeyValuePair<string, int>> GetChildProcessTreeIds(int parentProcessId, int maxTreeDepth)
+        private static List<KeyValuePair<string, int>> GetChildProcessTreeIds(int idToQuery, int maxTreeDepth, bool isRootQuery)
         {
             List<KeyValuePair<string, int>> result = new List<KeyValuePair<string, int>>();
```
narasamdya: May be not related to this change. Instead of creating a new instance of `List` for each call of `GetChildProcessTreeIds`,  I'm just wondering why `result` is not an argument of `GetChildProcessTreeIds`, and is used to populate the processes?

mpysson: I guess it's due to the logic in the recursion. The outer list is populated with a modified version of the inner recursive call. That could be reworked to have inner call get passed the state from the outer context so the full name could be constructed by the inner recursive call. That would indeed be more efficient.

That said, this code is not performance critical and it works. I don't think it's worth the effort

---
In reply to: [336642280](https://github.com/microsoft/BuildXL/pull/1061#discussion_r336642280) [](ancestors = 336642280)

PR #1062, opened 2019-10-17
Reset stream before retrying to add content to CAS

narasamdya:
[AB#1600682](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1600682)

### DIFF
--- a/Public/Src/Engine/Cache.Plugin.CacheCore/CacheCoreArtifactContentCache.cs
+++ b/Public/Src/Engine/Cache.Plugin.CacheCore/CacheCoreArtifactContentCache.cs
@@ -370,18 +370,19 @@ namespace BuildXL.Engine.Cache.Plugin.CacheCore
                 async cache =>
                 {
                     Contract.Assert(content.CanSeek);
-                    int addCount = 0;
+                    long initialPos = content.Position;
+                    bool attempted = false;
 
                     var result = await Helpers.RetryOnFailureAsync(
                         async lastAttempt =>
                         {
-                            if (addCount > 0)
+                            if (attempted)
                             {
-                                // Reset stream.
-                                content.Seek(0, SeekOrigin.Begin);
+                                // Reset stream to initial position.
+                                content.Seek(initialPos, SeekOrigin.Begin);
                             }
 
-                            ++addCount;
+                            attempted = true;
                             return await cache.AddToCasAsync(content, new CasHash(contentHash));
                         });
                     return result;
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/Cache.Plugin.CacheCore/CacheCoreArtifactContentCache.cs
```diff
@@ -369,9 +369,19 @@ public async Task<string> GetDiagnosticsAsync()
...
                     var result = await Helpers.RetryOnFailureAsync(
                         async lastAttempt =>
                         {
+                            if (addCount > 0)
+                            {
+                                // Reset stream.
+                                content.Seek(0, SeekOrigin.Begin);
+                            }
+
+                            ++addCount;
```
dcwuser: Is this the right ordering for incrementing addCount? If this lambda will only be called after the first failure, then addCount needs to be incremented before the test because we want to re-set the stream on the first re-try? Actually, if this lambed runs only after a failure, you don't need addCount at all and can just always re-set the stream.

dcwuser: Never mind, looking at RetryOnFailure, I see that's not how it works. Your lambda gets called even the first time.

---
In reply to: [336236874](https://github.com/microsoft/BuildXL/pull/1062#discussion_r336236874) [](ancestors = 336236874)

Public/Src/Engine/Cache.Plugin.CacheCore/CacheCoreArtifactContentCache.cs
```diff
@@ -369,9 +369,19 @@ public async Task<string> GetDiagnosticsAsync()
             Possible<CasHash, Failure> maybeStored = await maybeOpen.ThenAsync(
                 async cache =>
                 {
+                    Contract.Assert(content.CanSeek);
+                    int addCount = 0;
+
                     var result = await Helpers.RetryOnFailureAsync(
                         async lastAttempt =>
                         {
+                            if (addCount > 0)
```
dcwuser: You might want to consider always re-setting to origin. If start at origin, this first re-set is of course unnecessary. But if you don't start at origin, then the start point will change to 0 on retries. Alternatively, you could ensure consistency by noting initial Position and always re-setting to it.

Public/Src/Engine/Cache.Plugin.CacheCore/CacheCoreArtifactContentCache.cs
```diff
@@ -369,9 +369,19 @@ public async Task<string> GetDiagnosticsAsync()
             Possible<CasHash, Failure> maybeStored = await maybeOpen.ThenAsync(
                 async cache =>
                 {
+                    Contract.Assert(content.CanSeek);
+                    int addCount = 0;
```
dcwuser: Minor: Really, a bool is enough, and might make code logic slightly clearer.

PR #1067, opened 2019-10-18
wait time between copy retries

minghinmatthewlam:
- keep track of wait time for each location we are trying to copy from
- if location has already waited long enough, try copying straight away, otherwise wait the remaining time
- between every cycle of copying from list of locations, still has increasing wait times


Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -137,8 +139,12 @@ protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
                 PutResult putResult = null;
                 var badContentLocations = new HashSet<MachineLocation>();
                 var missingContentLocations = new HashSet<MachineLocation>();
+                var lastFailureTimes = new List<DateTime>();
                 int attemptCount = 0;
+                TimeSpan waitDelay = new TimeSpan();
```
kewilli: >new TimeSpan() [](start = 37, length = 14)

Nit: `TimeSpan.Zero` ?

Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -489,6 +509,17 @@ public Task<BoolResult> RequestCopyFileAsync(OperationContext context, ContentHa
                 }
                 finally
                 {
+                    // If the replicaIndex hasn't been tried before it won't have a value in lastFailureTimes so add it.
+                    // Otherwise replace the old failure time with the current time.
+                    if(lastFailureTimes.Count <= replicaIndex)
```
kewilli: >if( [](start = 20, length = 3)

Nit: `if (`

PR #1072, opened 2019-10-18
Add constructor (ResultBase, string) to implementations of ResultBase

kewilli:
Extension method `ErrorResult.AsResult<TResult>` expects to convert any given ErrorResult to any implementation of ResultBase. To do this, it uses reflection to find a constructor of TResult which takes a ResultBase and a string. This method is throwing in production because PutBlobResult does not include the necessary constructor. I added a test to ensure that all implementations include this constructor.

### DIFF
--- a/Public/Src/Cache/ContentStore/DistributedTest/TestErrorResultConverter.cs
+++ b/Public/Src/Cache/ContentStore/DistributedTest/TestErrorResultConverter.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using BuildXL.Cache.ContentStore.Interfaces.Results;
+using Xunit;
+
+namespace BuildXL.Cache.ContentStore.Distributed.Test
+{
+    public class TestErrorResultConverter
+    {
+        /// <summary>
+        /// Validate that all implementations of ResultBase include the necessary constructor to use the AsResult extension method.
+        /// </summary>
+        [Fact]
+        public void ValidateAsResult()
+        {
+            var resultBaseType = typeof(ResultBase);
+            IEnumerable<Type> resultTypes = AppDomain.CurrentDomain.GetAssemblies()
+                .Where(assembly => assembly.FullName.Contains("BuildXL"))
+                .SelectMany(assembly => assembly.GetTypes())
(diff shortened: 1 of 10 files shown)


Public/Src/Cache/ContentStore/DistributedTest/TestErrorResultConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+                .SelectMany(assembly => assembly.GetTypes())
+                .Where(t => resultBaseType.IsAssignableFrom(t) && !t.IsInterface && !t.IsAbstract && !t.IsGenericType && t.FullName != resultBaseType.FullName && t.Name != "CustomError");
+
+            var testError = new ErrorResult(new Exception(nameof(ValidateAsResult)));
+
+            foreach (Type resultType in resultTypes)
+            {
+                MethodInfo asResultMethod = typeof(ErrorResult).GetMethod(nameof(ErrorResult.AsResult)).MakeGenericMethod(new Type[] { resultType });
+
+                // testError.AsResult<resultType>();
```
SergeyTeplyakov: Remove?

Public/Src/Cache/ContentStore/DistributedTest/TestErrorResultConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace BuildXL.Cache.ContentStore.Distributed.Test
+{
+    public class TestErrorResultConverter
+    {
+        /// <summary>
+        /// Validate that all implementations of ResultBase include the necessary constructor to use the AsResult extension method.
+        /// </summary>
+        [Fact]
+        public void ValidateAsResult()
```
SergeyTeplyakov: Nit, should this test case be in ErrorResultConverterTests?
Minor, but it maybe easier to have one test suite instead of more then one.

Public/Src/Cache/ContentStore/DistributedTest/TestErrorResultConverter.cs
kewilli: This was left to explain the reflection magic in the next line. This isn't code which was removed.

Public/Src/Cache/ContentStore/DistributedTest/TestErrorResultConverter.cs
kewilli: ErrorResultConverterTests is in BuildXL.Cache.ContentStore.Interfaces.Test, which does not include our distributed bits. To validate this feature, we need to cover more ground.

PR #1073, opened 2019-10-21
Make the ProcessRunScriptAnalyzer x-plat compatible

Kr1sso:


Public/Src/Tools/Execution.Analyzer/Analyzers.Core/ProcessRunScriptAnalyzer.cs
```diff
@@ -251,20 +272,22 @@ private IEnumerable<AbsolutePath> GetOutputDirectories(Process pip)
...
             {
-                writer.WriteLine("IF EXIST \"{0}\" del \"{0}\"", fileOutput.Path.ToString(PathTable));
+                var path = fileOutput.Path.ToString(PathTable);
+                writer.WriteLine(ScriptIfPathExistsConditionWithCommand(path, $"{FileDeleteOperation} \"{path}\""));
             }
             writer.WriteLine();
 
-            writer.WriteLine(":: Delete Directories");
+            writer.WriteLine($"{CommentPrefix} Delete Directories");
             foreach (var directory in GetOutputDirectories(pip))
```
mpysson: >oreach (var directory in GetOutputDirectories(pip)) [](start = 13, length = 51)

This comes from a time before shared opaque directories. It's a bit sketchy now since there may actually be inputs in these directories. Embedding more precise file deletion logic in this script is overkill. I'm not sure what the best way to go is. Maybe some sort of "proceed with caution" warning is in order. What do you think?

Public/Src/Engine/Processes/PipEnvironment.cs
```diff
@@ -49,14 +50,27 @@ public PipEnvironment()
...
+#if PLATFORM_WIN
                     SpecialFolderUtilities.GetFolderPath(Environment.SpecialFolder.Windows),
-                    Path.Combine(SpecialFolderUtilities.SystemDirectory, "wbem"));
-            var pathExt = ".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC";
+                    Path.Combine(SpecialFolderUtilities.SystemDirectory, "wbem")
+#else
+                    MacPaths.UsrBin,
+                    MacPaths.UsrSbin,
+                    MacPaths.Bin,
+                    MacPaths.Sbin
```
mpysson: This is a behavioral change to how we run pips. Were people manually adding them to PATH previously? Is there any concern that in some circumstances you might not want these included?

Kr1sso: When starting a new shell those PATH components are set by default. It's were the basic OS tools life. In a lot of scenarios we whitelist those paths when running sandboxed, this thus makes sure that we have a sane environment where basic tools are available for a pip to execute.

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/ProcessRunScriptAnalyzer.cs
Kr1sso: I've added a warning to this is could be an issue.

PR #1074, opened 2019-10-21
SandboxedProcessMac: don't cancel CancellableTimedAction twice

almiliMSFT:


Public/Src/Engine/Processes/SandboxedProcessMac.cs
```diff
@@ -185,8 +185,6 @@ protected override System.Diagnostics.Process CreateProcess()
             process.StartInfo.Arguments = string.Empty;
             process.StartInfo.RedirectStandardInput = true;
 
-            process.Exited += (o, e) => m_perfCollector.Cancel();
```
almiliMSFT: can lead to
```
System.ObjectDisposedException: The CancellationTokenSource has been disposed.
   at void System.Threading.CancellationTokenSource.ThrowObjectDisposedException()
   at Process BuildXL.Processes.SandboxedProcessMac.CreateProcess()+(object o, EventArgs e) => { }
   at void System.Diagnostics.Process.RaiseOnExited()
   at void System.Diagnostics.Process.CompletionCallback(object waitHandleContext, bool wasSignaled)
   at void System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
```

PR #1076, opened 2019-10-21
Improvements in memory-based scheduler throttling

semihokur:
- Reduced the default memory usage of process pips in case of no historical ram usage info.
- Master is now aware of workers' actual available memory, and commit memory usage.
- Measure working set, pagefile usage of process pips besides virtual memory
- Record working set in the historical perf data table instead of the virtual memory to accurately estimate the available memory in the next builds.

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcess.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcess.cs
@@ -69,8 +69,9 @@ namespace BuildXL.Processes
         private TaskSourceSlim<bool> m_standardInputTcs;
         private readonly PathTable m_pathTable;
         private readonly string[] m_allowedSurvivingChildProcessNames;
-        private ulong m_maxPeakWorkingSet;
-        private ulong m_maxPeakPagefileUsage;
+
+        private readonly PerformanceCollector.Aggregation m_peakWorkingSet = new PerformanceCollector.Aggregation();
+        private readonly PerformanceCollector.Aggregation m_peakPagefileUsage = new PerformanceCollector.Aggregation();
 
         [SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope", Justification = "We own these objects.")]
         internal SandboxedProcess(SandboxedProcessInfo info)
@@ -236,18 +237,16 @@ namespace BuildXL.Processes
                         {
                             currentPeakWorkingSet = (currentPeakWorkingSet ?? 0) + memoryUsage.PeakWorkingSetSize;
                             currentPeakPagefileUsage = (currentPeakPagefileUsage ?? 0) + memoryUsage.PeakPagefileUsage;
-
                         }
                     }
                 }
 
-                m_maxPeakWorkingSet = Math.Max(m_maxPeakWorkingSet, currentPeakWorkingSet ?? 0);
-                m_maxPeakPagefileUsage = Math.Max(m_maxPeakPagefileUsage, currentPeakPagefileUsage ?? 0);
+                m_peakWorkingSet.RegisterSample(currentPeakWorkingSet ?? 0);
+                m_peakPagefileUsage.RegisterSample(currentPeakPagefileUsage ?? 0);
 
                 return currentPeakWorkingSet;
             }
         }
-
 
         /// <inheritdoc />
         public long GetDetoursMaxHeapSize()
@@ -502,7 +501,7 @@ namespace BuildXL.Processes
             JobObject jobObject = m_detouredProcess.GetJobObject();
             if (jobObject != null)
             {
-                jobAccountingInformation = jobObject.GetAccountingInformation(m_maxPeakWorkingSet, m_maxPeakPagefileUsage);
+                jobAccountingInformation = jobObject.GetAccountingInformation(Convert.ToUInt64(m_peakWorkingSet.Maximum), Convert.ToUInt64(m_peakPagefileUsage.Maximum));
             }
 
             ProcessTimes primaryProcessTimes = m_detouredProcess.GetTimesForPrimaryProcess();
--- a/Public/Src/Engine/Processes/SandboxedProcess.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcess.cs
@@ -69,6 +69,8 @@ namespace BuildXL.Processes
         private TaskSourceSlim<bool> m_standardInputTcs;
         private readonly PathTable m_pathTable;
         private readonly string[] m_allowedSurvivingChildProcessNames;
+        private ulong m_maxPeakWorkingSet;
+        private ulong m_maxPeakPagefileUsage;
 
         [SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope", Justification = "We own these objects.")]
         internal SandboxedProcess(SandboxedProcessInfo info)
@@ -181,10 +183,7 @@ namespace BuildXL.Processes
         }
 
         /// <inheritdoc />
-        /// <remarks>
-        /// Gets the peak memory usage for the job object associated with the detoured process while the process is active.
-        /// </remarks>
-        public ulong? GetActivePeakMemoryUsage()
+        public ulong? GetActivePeakWorkingSet()
         {
             using (m_queryJobDataLock.AcquireReadLock())
             {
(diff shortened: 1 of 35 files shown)


Public/Src/Engine/Processes/SandboxedProcess.cs
```diff
@@ -197,10 +196,59 @@ public async Task<SandboxedProcessResult> GetResultAsync()
...
+                }
+
+                m_maxPeakWorkingSet = Math.Max(m_maxPeakWorkingSet, currentPeakWorkingSet ?? 0);
+                m_maxPeakPagefileUsage = Math.Max(m_maxPeakPagefileUsage, currentPeakPagefileUsage ?? 0);
+
+                return currentPeakWorkingSet;
             }
         }
 
+
```
almiliMSFT: nit: remove empty line
```suggestion
```

Public/Src/Engine/Processes/SandboxedProcess.cs
```diff
@@ -197,10 +196,59 @@ public async Task<SandboxedProcessResult> GetResultAsync()
...
+                            // we are too late: process id got reused by another process
+                            continue;
+                        }
+
+                        var memoryUsage = Interop.Dispatch.GetMemoryUsageCounters(processHandle.DangerousGetHandle());
+                        if (memoryUsage != null)
+                        {
+                            currentPeakWorkingSet = (currentPeakWorkingSet ?? 0) + memoryUsage.PeakWorkingSetSize;
+                            currentPeakPagefileUsage = (currentPeakPagefileUsage ?? 0) + memoryUsage.PeakPagefileUsage;
+
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Processes/SandboxedProcess.cs
```diff
@@ -197,10 +196,59 @@ public async Task<SandboxedProcessResult> GetResultAsync()
                     return null;
                 }
 
-                return detouredProcess.GetJobObject()?.GetPeakMemoryUsage();
+                ulong? currentPeakWorkingSet = null;
```
almiliMSFT: Why `ulong?` with `null` as default instead of `ulong` with `0` as default?  It seems the same to me and it would make some of the code below syntactically simpler

Public/Src/Engine/Processes/SandboxedProcess.cs
```diff
@@ -69,6 +69,8 @@ public sealed class SandboxedProcess : ISandboxedProcess
         private TaskSourceSlim<bool> m_standardInputTcs;
         private readonly PathTable m_pathTable;
         private readonly string[] m_allowedSurvivingChildProcessNames;
+        private ulong m_maxPeakWorkingSet;
```
almiliMSFT: Did you consider using `PerformanceCollector.Aggregation` here instead of `ulong`?  In that case you could make this field `readonly` and you wouldn't have to manually compute the maximum every time (the `Aggregation` object would take care of that)

Public/Src/Engine/Scheduler/Distribution/Worker.cs
```diff
@@ -172,6 +172,22 @@ protected set
...
+        /// Percentage of available commit used.
+        /// </summary>
+        public int? ActualCommitPercent;
+
+        /// <summary>
+        /// The machine's total commit in MB
+        /// </summary>
+        public int? ActualCommitTotalMB;
+
+
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Scheduler/ProcessResourceManager.cs
```diff
@@ -139,6 +140,7 @@ public void TryFreeResources(int requiredRamMb)
             }
 
             using (scope)
+            using (new StoppableTimer(() => scope.RefreshRamUsage(), 0, 2000))
```
ljcollins25: >using (new StoppableTimer(() => scope.RefreshRamUsage(), 0, 2000)) [](start = 12, length = 66)

Why is this needed?

Public/Src/Engine/Processes/SandboxedProcess.cs
```diff
@@ -197,10 +196,59 @@ public async Task<SandboxedProcessResult> GetResultAsync()
...
+                ulong? currentPeakWorkingSet = null;
+                ulong? currentPeakPagefileUsage = null;
+
+                var jobObject = detouredProcess.GetJobObject();
+                if (jobObject == null || !jobObject.TryGetProcessIds(out uint[] childProcessIds) || childProcessIds.Length == 0)
+                {
+                    return null;
+                }
+
+                foreach (uint processId in childProcessIds)
```
narasamdya: >foreach (uint processId in childProcessIds) [](start = 16, length = 43)

Would this affect perf.?

Public/Src/Engine/Scheduler/RunnablePip.cs
```diff
@@ -470,7 +470,9 @@ public static PipResult CreatePipResultFromExecutionResult(DateTime start, Execu
                         ioCounters: performanceInformation.IO,
                         userTime: performanceInformation.UserTime,
                         kernelTime: performanceInformation.KernelTime,
-                        peakMemoryUsage: performanceInformation.PeakMemoryUsage,
+                        peakVirtualMemoryUsage: performanceInformation.PeakVirtualMemoryUsage,
+                        peakWorkingSet: performanceInformation.PeakWorkingSet,
+                        peakPagefileUsage: performanceInformation.PeakPagefileUsage,
```
narasamdya: Would it be better to put these data into a struct?

Public/Src/Engine/Processes/SandboxedProcess.cs
semihokur: it seems negligible in my test builds. I already measure how much time we spend as a part of QueryRamUsage counter.

Public/Src/Engine/Scheduler/ProcessResourceManager.cs
semihokur: There is no way to get the memory usage of the child processes after they die. That's why, we iterate through the child processes every 2 seconds to get the total workingset usage.

semihokur: We need to iterate through the child processes every 2 seconds to get the peak memory usages. When the child process dies, there is no way to get the memory usage. That's why, we do sampling here. If there is a child process created and terminated under 2 seconds, we might not catch the memory usage of that process.

---
In reply to: [337223679](https://github.com/microsoft/BuildXL/pull/1076#discussion_r337223679) [](ancestors = 337223679)

Public/Src/Engine/Processes/SandboxedProcess.cs
semihokur: It is more related to the callers of that method. In potential future scenarios, we would want to understand whether we were able to get the memory usage at all. However, 0 can mean that as well. I will think more

---
In reply to: [337179878](https://github.com/microsoft/BuildXL/pull/1076#discussion_r337179878) [](ancestors = 337179878)

Public/Src/Engine/Scheduler/ProcessResourceManager.cs
ljcollins25: Doesn't the schedule already query the ram usage every 2 seconds?

---
In reply to: [338328220](https://github.com/microsoft/BuildXL/pull/1076#discussion_r338328220) [](ancestors = 338328220,337223679)

ljcollins25: Not a big deal, but how about having the scheduler do this as a part of UpdateResourceAvailability. I'm slightly concerned about having so many timers.

---
In reply to: [338328220](https://github.com/microsoft/BuildXL/pull/1076#discussion_r338328220) [](ancestors = 338328220,337223679)

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3647,6 +3661,27 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
...
+
+                            int peakVirtualMemoryUsageMb = executionResult.PerformanceInformation.MemoryCounters.PeakVirtualMemoryUsageMb;
+                            int peakWorkingSetMb = executionResult.PerformanceInformation.MemoryCounters.PeakWorkingSetMb;
+                            int peakPagefileUsageMb = executionResult.PerformanceInformation.MemoryCounters.PeakPagefileUsageMb;
+
+                            Logger.Log.PeakMemoryUsage(
+                                operationContext, 
+                                runnablePip.Description,
+                                (int)executionResult.PerformanceInformation.NumberOfProcesses,
+                                (int)executionResult.PerformanceInformation.ProcessExecutionTime.TotalSeconds,
```
ljcollins25: >TotalSeconds [](start = 97, length = 12)

It would be interesting to log estimated execution time as well.

semihokur:
> I'd like to have a talk to you after merging this to clarify what we understand those three memory usage types are and how they influence scheduling decisions. Mostly due to the fact of emulating it as close as possible on macOS too.

Sure. We already use the workingset memory for macOS, so I just made the same thing for Windows. The behavior is currently the same. I measure the virtual and pagefile usage just for logging; those counters serve no purpose in our scheduler. In that sense, memory throttling behavior should be the same after the code is merged. happy to talk though.

PR #1078, opened 2019-10-21
Use checked-in package-lock.json file for npm install

almiliMSFT:
This avoids generating package-lock.json upon every build; given that `npm install` without a lock file is not necessarily deterministic, every build can in theory generate a new package-lock.json.  Instead, when the lock file is explicitly provided, the behavior of `npm install` is fixed and deterministic.

### DIFF
--- a/Shared/Scripts/bxl.ps1
+++ b/Shared/Scripts/bxl.ps1
@@ -560,7 +560,7 @@ if (!$skipFilter){
 
     if ($Minimal) {
         # filtering by core deployment.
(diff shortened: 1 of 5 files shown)


Shared/Scripts/bxl.ps1
```diff
@@ -560,7 +560,7 @@ if (!$skipFilter){
 
     if ($Minimal) {
         # filtering by core deployment.
-        $AdditionalBuildXLArguments +=  "/f:(output='$($useDeployment.buildDir)\*'or(output='out\bin\$DeployConfig\Sdk\*')or($CacheOutputFilter))and~($CacheLongRunningFilter)ortag='protobufgenerator'oroutput='cg\*'"
```
olkononenko: >'oroutput='cg\*'" [](start = 198, length = 17)

I might be missing something, why can we remove this filter?

almiliMSFT: because there are no build outputs under the `cg` folder any longer

PR #1080, opened 2019-10-21
Update AzDev packages and fix breaking changes

JuanCarlosGI:
Drop.RemotableClient does not exist anymore, so I removed it.

Everything else is just fixing whatever breaking changes were introduced.

Will coordinate will CloudBuild so that the packages match.


Public/Src/Tools/SymbolDaemon/VsoSymbolClient.cs
```diff
@@ -46,7 +47,7 @@ public sealed class VsoSymbolClient : ISymbolClient
 
         private VssCredentials GetCredentials() =>
             new VsoCredentialHelper(m => m_logger.Verbose(m))
-                .GetCredentials(m_config.Service, true, null);
+                .GetCredentials(m_config.Service, true, null, null, PromptBehavior.Never);
```
dcwuser: Is PromptNever the desired behavior? I thought the fix Christian needed was to prompt

JuanCarlosGI: His fix is in the CB repo. Here, I'm just leaving us with he same behavior as before.

PR #1081, opened 2019-10-22
Support proper recovery from Redis data loss for cluster state keys.

ljcollins25:


Public/Src/Cache/ContentStore/Distributed/NuCache/ReplicatedRedisHashKey.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// </summary>
+        public async Task<Result<T>> UseReplicatedHashAsync<T>(OperationContext context, RedisOperation operation, Func<RedisBatch, string, Task<T>> addOperations, [CallerMemberName] string caller = null)
+        {
+            // Query to see which db has highest version number
+            (var primaryVersionedResult, var secondaryVersionedResult) = await _redis.ExecuteRaidedAsync(context, async redisDb =>
+            {
+                var operationResult = await redisDb.ExecuteBatchAsync(context,
+                    async batch =>
+                    {
+                        var versionTask = batch.AddOperation(_key, b => b.HashIncrementAsync(_key, nameof(ReplicatedHashVersionNumber)));
```
SergeyTeplyakov: Minor, but this may fail with unobserved task exception: if both addOpations and versionTask fail, then versionTask is unobserved.

Public/Src/Cache/ContentStore/Distributed/NuCache/ReplicatedRedisHashKey.cs
```diff
@@ -0,0 +1,210 @@
...
+                    await target.ExecuteBatchAsync(context,
+                        b =>
+                        {
+                            var deleteTask = b.AddOperation(_key, b => b.KeyDeleteAsync(_key));
+                            var restoreTask = b.AddOperation(_key, b => b.KeyRestoreAsync(_key, sourceDump).WithResultAsync(Unit.Void));
+                            return Task.WhenAll(deleteTask, restoreTask).WithResultAsync(Unit.Void);
+                        },
+                        RedisOperation.HashDeleteAndRestore);
+
+                    if (postMirrorSourceVersion.HasValue)
```
SergeyTeplyakov: Just curious: should the version update be par tof the second execute batch to make the update "atomic"?

ljcollins25: The second update batch is to the target. This is to the source. We only want to modify the version if the target update succeeded.

---
In reply to: [337752360](https://github.com/microsoft/BuildXL/pull/1081#discussion_r337752360) [](ancestors = 337752360)

Public/Src/Cache/ContentStore/Distributed/NuCache/ReplicatedRedisHashKey.cs
ljcollins25: Will fix in subsequent changes.

---
In reply to: [337751017](https://github.com/microsoft/BuildXL/pull/1081#discussion_r337751017) [](ancestors = 337751017)

PR #1085, opened 2019-10-22
Extend the waiting period to create crash dumps

semihokur:
It might take a couple of minutes to create crash dumps in CloudBuild. That's why, I increased the timeout from 1min to 5min. This will hopefully reduce the number of watson dumps.

### DIFF
--- a/Public/Src/App/Bxl/BuildXLApp.cs
+++ b/Public/Src/App/Bxl/BuildXLApp.cs
@@ -118,9 +118,9 @@ namespace BuildXL
     /// </summary>
     internal sealed class BuildXLApp : IDisposable
     {
-        // We give the failure completion logic a generous 60 seconds to complete since in some cases taking a crash dump
+        // We give the failure completion logic a generous 300 seconds to complete since in some cases taking a crash dump
         // can take quite a while
-        private const int FailureCompletionTimeoutMs = 60 * 1000;
+        private const int FailureCompletionTimeoutMs = 300 * 1000;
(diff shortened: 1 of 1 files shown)


mpysson:
>                 Thread.Sleep(FailureCompletionTimeoutMs);

So let's say the second crashing thread gets parked here. The crash handler on the first thread completes before the timeout. Does that cause the application to exit immediately or is this second thread allowed to complete and ExceptionUtilities.FailFast()?

If the second thread is not allowed to fail fast I think bumping up the timeout should be sufficient to address the additional AzureWatson entries we're seeing. If the second thread can still call ExceptionUtilities.FailFast() I don't think this is a good change.

---
Refers to: Public/Src/App/Bxl/BuildXLApp.cs:1768 in c4feeee. [](commit_id = c4feeee53b3ae9cac7e0275332a90f79f98ff176, deletion_comment = False)

semihokur:
>                 Thread.Sleep(FailureCompletionTimeoutMs);

Environment.Exit causes all threads being terminated; so the problem was mostly related to the insufficient timeout. 

---
In reply to: [545204930](https://github.com/microsoft/BuildXL/pull/1085#issuecomment-545204930) [](ancestors = 545204930)

---
Refers to: Public/Src/App/Bxl/BuildXLApp.cs:1768 in c4feeee. [](commit_id = c4feeee53b3ae9cac7e0275332a90f79f98ff176, deletion_comment = False)

PR #1091, opened 2019-10-23
Remove emptyFileHashShortcut and useTrustedHash options

SergeyTeplyakov:
The option is true everywhere.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentStoreSettings.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentStoreSettings.cs
@@ -104,10 +104,6 @@ namespace BuildXL.Cache.ContentStore.Distributed.Stores
         /// </summary>
         public bool UseTrustedHash { get; set; } = false;
 
-        /// <summary>
-        /// Whether the shortcuts for streaming, placing, and pinning the empty file are used.
-        /// </summary>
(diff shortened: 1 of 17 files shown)


Public/Src/Cache/ContentStore/Library/Stores/ContentStoreSettings.cs
```diff
@@ -20,7 +20,7 @@ public sealed class ContentStoreSettings
         /// <summary>
         /// Whether the shortcuts for streaming, placing, and pinning the empty file are used.
         /// </summary>
-        public bool UseEmptyFileHashShortcut { get; set; } = true;
+        public const bool UseEmptyFileHashShortcut = true;
```
kewilli: >public const bool UseEmptyFileHashShortcut = true; [](start = 8, length = 50)

Can this just be removed?

Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentStoreSettings.cs
```diff
@@ -104,10 +104,6 @@ public int AssumeAvailableReplicaCount
         /// </summary>
         public bool UseTrustedHash { get; set; } = false;
```
kewilli: >UseTrustedHash [](start = 20, length = 14)

I find it odd that we can't do the same with TrustedHash yet. It predates the empty file hash optimization by a good bit.

PR #1092, opened 2019-10-23
Emit change affected inputs with full paths

norahuang:
[AB#1609039](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1609039)

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
@@ -2672,7 +2672,7 @@ namespace BuildXL.Processes
                     m_pip.ChangeAffectedInputListWrittenFilePath,
                     () => string.Join(
                         Environment.NewLine,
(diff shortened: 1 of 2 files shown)


Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2672,7 +2672,7 @@ private async Task<bool> PrepareDirectoryOutputsAsync(HashSet<AbsolutePath> pres
                     m_pip.ChangeAffectedInputListWrittenFilePath,
                     () => string.Join(
                         Environment.NewLine,
-                        changeAffectedInputs.Select(i => i.GetName(m_pathTable).ToString(m_pathTable.StringTable)).Distinct().OrderBy(n => n)),
+                        changeAffectedInputs.Select(i => i.ToString(m_pathTable)).Distinct().OrderBy(n => n)),
```
narasamdya: >Distinct().OrderBy [](start = 82, length = 18)

Do we need to dedupe and order the entries? I don't think so.

PR #1095, opened 2019-10-24
Add so-called engine state id in incremental scheduling state

narasamdya:
The bug causes an underbuild when incremental scheduling state is reused from an old the engine state due to using different server processes.

[AB#1623765](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1623765)


Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/IncrementalSchedulingTests/GraphChangesTests.cs
```diff
@@ -1264,7 +1263,42 @@ public void GAISSCannotBeReusedFromEngineStateWhenGraphChange()
             result = RunScheduler(schedulerState: result.SchedulerState).AssertScheduled(q.PipId);
 
             AssertVerboseEventLogged(EventId.IncrementalSchedulingReuseState);
-            AssertLogContains(false, "Attempt to reuse existing incremental scheduling state: " + ReuseKind.ChangedGraph);
+            AssertLogContains(false, "Attempt to reuse existing incremental scheduling state from engine state: " + ReuseFromEngineStateKind.ChangedGraph);
```
mpysson: >Attempt to reuse existing incremental scheduling state from engine state [](start = 38, length = 72)

Consider replacing this with the DX code so the text is flexible if we ever change message strings again

Public/Src/Engine/Scheduler/IncrementalScheduling/GraphAgnosticIncrementalSchedulingState.cs
```diff
@@ -1262,40 +1274,66 @@ private void Save(FileEnvelopeId atomicSaveToken, string incrementalSchedulingSt
...
+                ? ReuseFromEngineStateKind.ChangedGraph
                 : (!m_incrementalSchedulingStateId.IsAsSafeOrSaferThan(newId)
-                    ? ReuseKind.MismatchedId
-                    : ReuseKind.Reusable);
+                    ? ReuseFromEngineStateKind.MismatchedId
+                    : ReuseFromEngineStateKind.Reusable);
+
+            if (reuseKind == ReuseFromEngineStateKind.Reusable)
+            {
+                Guid engineStateId = default;
```
mpysson: >engineStateId [](start = 21, length = 13)

Suggest logging the engineStateId used every build for sake of debugging

Public/Src/Engine/UnitTests/Scheduler.IntegrationTest/IncrementalSchedulingTests/GraphChangesTests.cs
narasamdya: This test tests the reason of ISS' reusability. We verify the change kind.

PR #1098, opened 2019-10-24
Lightweight trie implementation

almiliMSFT:
This PR adds an alternative trie implementation, namely using a linked list to represent node children instead of a preallocated array.  When using this new implementation, the kext memory footprint is reduced significantly, even 10x in some large builds.

Two additional sysctl variables are introduced:
  1. `bxl_enable_cache` to control whether access reports are cached inside the kext (on by default)
  2. `bxl_enable_light_trie` to control whether to use this lightweight trie implementation or the old one (on by default)

### DIFF
--- a/Public/Src/Sandbox/MacOs/Sandbox/Src/Utilities/Trie.cpp
+++ b/Public/Src/Sandbox/MacOs/Sandbox/Src/Utilities/Trie.cpp
@@ -529,7 +529,7 @@ void Trie::forEach(void *callbackArgs, for_each_fn callback)
 {
     typedef struct { for_each_fn callback; void *args; } State;
     State state = { .callback = callback, .args = callbackArgs };
-    traverse(/*computeKey*/ kind_ == kUintTrie, /*callbackArgs*/ &state, [](void *s, uint64_t key, Node *node)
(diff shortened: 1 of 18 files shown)


Public/Src/Sandbox/MacOs/Sandbox/Src/Utilities/Trie.cpp
```diff
@@ -2,72 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 #include "Trie.hpp"
+#include "Monitor.hpp"
 
 #define super OSObject
 
-// ================================== class Node ==================================
```
almiliMSFT: moved to `TrieNode.(cpp|hpp)`

Public/Src/Sandbox/MacOs/Sandbox/Src/Utilities/Trie.cpp
```diff
@@ -114,43 +55,6 @@ void Trie::free()
     super::free();
 }
 
-bool Trie::findChildNode(Node *node, int idx, bool createIfMissing)
```
almiliMSFT: moved to `Node`

PR #1099, opened 2019-10-25
Child processes can be configured to breakaway from the sandbox

smera:
Enable to configure a collection of child process names which will breakaway from the sandbox when spawned by the main process. When a process breakaways, its accesses are not monitored and it can outlive the parent lifespan.
For now this is a Windows-only feature. We may consider adding the equivalent functionality on Mac in some future round.


Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -785,6 +836,7 @@ public ArraySegment<byte> GetPayloadBytes(FileAccessSetup setup, MemoryStream st
             {
                 WriteDebugFlagBlock(writer, ref debugFlagsMatch);
                 WriteInjectionTimeoutBlock(writer, timeoutMins);
+                WriteChildProcessesToBreakAwayFromSandbox(writer, ChildProcessesToBreakawayFromSandbox);
```
almiliMSFT: you may want to do this conditionally `#if PLATFORM_WIN`; otherwise you'll have to fix the FAM parser for Mac

Public/Src/Engine/UnitTests/Processes/SandboxedProcessBreakawayTest.cs
```diff
@@ -0,0 +1,108 @@
...
+            var info = ToProcessInfo(
+                ToProcess(
+                    Operation.SpawnExe(
+                        Context.PathTable, 
+                        CreateFileArtifactWithName(InfiniteWaiterToolName, TestDeploymentDir))),
+                fileAccessManifest: fam);
+
+            // Let's shorten the default time to wait for nested processes, since we are spawning 
+            // a process that never ends and we don't want this test to wait for that long
+            info.NestedProcessTerminationTimeout = TimeSpan.FromMilliseconds(10);
```
almiliMSFT: I expected all these breakaway processes to be automatically added to `allowedSurvivingChildProcessNames`; when those are specified, after the main process exits SandboxedProcess doesn't for those surviving child processes to exit.

almiliMSFT: Never mind, I see that there is a case when the test doesn't let the infinite waiter survive, in which case there is no pointing waiting longer than 10ms.

Public/Src/Engine/Processes/FileAccessManifest.cs
smera: So if I'm getting this right, I think I fixed the manifest parser already on the mac side (to just ignore the list for now)

Public/Src/Engine/UnitTests/Processes/SandboxedProcessBreakawayTest.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var result = RunProcess(info).GetAwaiter().GetResult();
+            XAssert.AreEqual(0, result.ExitCode);
+
+            if (!letInfiniteWaiterSurvive)
+            {
+                // If we didn't let infinite waiter escape, we should have killed it when the job object was finalized
+                XAssert.IsTrue(result.Killed);
+                result.SurvivingChildProcesses.Any(process => process?.Path != null && string.Equals(System.IO.Path.GetFileName(process.Path), InfiniteWaiterToolName, StringComparison.InvariantCultureIgnoreCase));
```
almiliMSFT: Don't you want to assert what this call to `Any()` should return?

Instead of asserting whether this `Any` returns true or false, you could use something like

```cs
XAssert.Contains(
    result.SurvivingChildProcesses.Select(p => p?.Path).Where(p => p != null).Select(p => Path.GetFileName(p).ToUpperInvariant()),
    InfiniteWaiterToolName.ToUpperInvariant());
```

Public/Src/Engine/UnitTests/Processes/SandboxedProcessBreakawayTest.cs
```diff
@@ -0,0 +1,108 @@
...
+                        Operation.ReadFile(srcFile2))),
+                fileAccessManifest: fam);
+
+            var result = RunProcess(info).GetAwaiter().GetResult();
+            XAssert.AreEqual(0, result.ExitCode);
+
+            // We should see the access that happens on the main test process
+            XAssert.IsTrue(result.FileAccesses.Any(fa => !string.IsNullOrEmpty(fa.Path) && AbsolutePath.TryCreate(Context.PathTable, fa.Path, out AbsolutePath result) && result == srcFile1.Path));
+            // We shouldn't see the access that happens on the spawned process
+            XAssert.IsFalse(result.FileAccesses.Any(fa => !string.IsNullOrEmpty(fa.Path) && AbsolutePath.TryCreate(Context.PathTable, fa.Path, out AbsolutePath result) && result == srcFile2.Path));
```
almiliMSFT: consider using `XAssert.Contains` and `XAssert.ContainsNot` instead

Public/Src/Engine/Processes/FileAccessManifest.cs
almiliMSFT: yes, sorry, I hadn't reached that part yet when I wrote this, and from the description I assumed you hadn't changed anything on the Mac side

Public/Src/Sandbox/Windows/DetoursServices/DataTypes.h
```diff
@@ -360,6 +360,26 @@ typedef struct ManifestTranslatePathsStrings_t
...
 typedef const ManifestTranslatePathsStrings_t * PManifestTranslatePathsStrings;
 
+// ==========================================================================
+// == ManifestChildProcessesToBreakAwayFromJob
+// ==========================================================================
+typedef struct ManifestChildProcessesToBreakAwayFromJob_t
+{
+    GENERATE_TAG("ChildProcessesToBreakAwayFromJob", 0xABCDEF03)
+
+        inline size_t GetSize() const
```
narasamdya: >     [](start = 4, length = 4)

nit: remove extra space

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
```diff
@@ -1850,6 +1852,40 @@ BOOL WINAPI Detoured_CreateProcessW(
...
+        // In this case we just let the regular detoured create process take control
+        if (imageName != L"")
+        {
+            std::unordered_set<wstring, CaseInsensitiveStringHasher, CaseInsensitiveStringComparer>::iterator result;
+            result = g_processNamesToBreakAwayFromJob->find(imageName);
+            
+            if (result != g_processNamesToBreakAwayFromJob->end())
+            {
+                Dbg(L"Allowing process to breakaway from job object. Image name: '%s'", imageName.c_str());
+
```
narasamdya: Do you intentionally want to send this debugging message to the log during production?

Public/Src/Pips/Dll/Operations/Process.cs
```diff
@@ -718,6 +723,16 @@ public sealed partial class Process : Pip
...
 
+
+        /// <summary>
+        /// Process names that will break away from the sandbox when spawned by the main process
+        /// </summary>
+        /// <remarks>
+        /// The accesses of processes that break away from them sandbox won't be observed.
+        /// Processes that breakaway can survive the lifespan of the sandbox
+        /// </remarks>
+        public ReadOnlyArray<PathAtom> ChildProcessesToBreakawayFromSandbox { get; }
```
narasamdya: >ChildProcessesToBreakawayFromSandbox [](start = 39, length = 36)

Should this be part of the fingerprint? I think it should because the pip will have a different behavior.

Public/Src/Utilities/Native/Processes/IProcessUtilities.cs
```diff
@@ -105,6 +105,9 @@ internal interface IProcessUtilities
         /// <summary><see cref="ProcessUtilities.OSSupportsNestedJobs"/></summary>
         bool OSSupportsNestedJobs();
 
+        /// <summary><see cref="ProcessUtilities.SandboxSupportsProcessBreakaway"></see></summary>
+        bool SandboxSupportsProcessBreakaway();
```
narasamdya: Is this used?

Public/Src/Utilities/UnitTests/Executables/InfiniteWaiter/Program.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// InfiniteWaiter.exe - Executable which launches and blocks forever.
```
narasamdya: > InfiniteWaiter.exe  [](start = 2, length = 20)

Why creates a new exe instead of generalizing (and placing in a proper namespace) DummyWaiter?

Public/Src/Engine/UnitTests/Processes/SandboxedProcessBreakawayTest.cs
```diff
@@ -0,0 +1,108 @@
...
+                result.SurvivingChildProcesses.Any(process => process?.Path != null && string.Equals(System.IO.Path.GetFileName(process.Path), InfiniteWaiterToolName, StringComparison.InvariantCultureIgnoreCase));
+            }
+            else
+            {
+                // If we did let it escape, then nothing should have been killed (nor tried to survive and later killed, from the job object point of view)
+                XAssert.IsFalse(result.Killed);
+                XAssert.IsTrue(result.SurvivingChildProcesses == null);
+
+                // Let's retrieve the child process and confirm it survived
+                var dummyWaiterInfo = RetrieveChildProcessesCreatedBySpawnExe(result).Single();
```
narasamdya: >dummyWaiterInfo [](start = 20, length = 15)

This sounds like you can generalize DummyWaiter instead of creating a new InfiniteWaiter.

narasamdya:
LGTM. It would be better if you can reuse DummyWaiter.

Public/Src/Engine/UnitTests/Processes/SandboxedProcessBreakawayTest.cs
smera: Sure, will change

---
In reply to: [339159918](https://github.com/microsoft/BuildXL/pull/1099#discussion_r339159918) [](ancestors = 339159918)

Public/Src/Sandbox/Windows/DetoursServices/DetouredFunctions.cpp
smera: I'll enclose it in a SUPER_VERBOSE #if

---
In reply to: [339169069](https://github.com/microsoft/BuildXL/pull/1099#discussion_r339169069) [](ancestors = 339169069)

Public/Src/Engine/UnitTests/Processes/SandboxedProcessBreakawayTest.cs
smera: The variable name is a leftover of a previous attempt. Changed that.

---
In reply to: [339177903](https://github.com/microsoft/BuildXL/pull/1099#discussion_r339177903) [](ancestors = 339177903)

Public/Src/Utilities/UnitTests/Executables/InfiniteWaiter/Program.cs
smera: Too much drama for marginal gains... I'd have to deal with argument passing, etc. to identify the desired behavior. Considering this is a two line implementation, I'd prefer to leave it as is

---
In reply to: [339177541](https://github.com/microsoft/BuildXL/pull/1099#discussion_r339177541) [](ancestors = 339177541)

Public/Src/Pips/Dll/Operations/Process.cs
smera: Good point. Will add.

---
In reply to: [339171834](https://github.com/microsoft/BuildXL/pull/1099#discussion_r339171834) [](ancestors = 339171834)

PR #1101, opened 2019-10-25
Proactive copy even when blob in redis

JuanCarlosGI:
Expiry time for blobs in Redis is short (30 mins), so it's possible to put it and, in the same build, be unable to retrieve it from Redis, because it already expired.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
@@ -236,8 +236,15 @@ namespace BuildXL.Cache.ContentStore.Distributed.Sessions
 
                 if (result && recorder != null)
                 {
-                    // Fire and forget since this step is optional.
-                    ContentLocationStore.PutBlobAsync(context, result.ContentHash, recorder.RecordedBytes).FireAndForget(context);
+                    if (Settings.ShouldInlinePutBlob)
+                    {
+                        var putBlobResult = await ContentLocationStore.PutBlobAsync(context, result.ContentHash, recorder.RecordedBytes);
+                    }
+                    else
+                    {
+                        // Fire and forget since this step is optional.
+                        ContentLocationStore.PutBlobAsync(context, result.ContentHash, recorder.RecordedBytes).FireAndForget(context);
+                    }
                 }
             }
             else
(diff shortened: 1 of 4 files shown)


Public/Src/Cache/ContentStore/DistributedTest/Sessions/RedisDistributedContentTests.cs
```diff
@@ -397,10 +397,12 @@ public Task RepeatedBlobIsSkipped()
                     var fileString = Encoding.Default.GetString(file);
 
                     await session0.PutContentAsync(context, fileString).ShouldBeSuccess();
+                    await Task.Delay(500); // Wait for put blob
```
kewilli: >await Task.Delay(500); // Wait for put blob [](start = 20, length = 43)

Rather than wait for a static set of time, could you query for success in a loop? This test will be run on machines which are maxed at 100%, and 500ms might not be enough time for it to complete the FireAndForget task.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -256,7 +254,7 @@ private Task<PutResult> PerformPutFileGatedOperationAsync(OperationContext opera
             var registerResult = await RegisterPutAsync(context, UrgencyHint.Nominal, result);
 
             // Only perform proactive copy to other machines if we didn't put the blob into Redis and we succeeded in registering our location.
```
yash256: >// Only perform proactive copy to other machines if we didn't put the blob into Redis and we succeeded in registering our location. [](start = 12, length = 131)

Is this comment still valid?

Public/Src/Cache/ContentStore/DistributedTest/Sessions/RedisDistributedContentTests.cs
ljcollins25: Maybe add a config (or event static variable) that tells it to PutBlob inline.

---
In reply to: [340240813](https://github.com/microsoft/BuildXL/pull/1101#discussion_r340240813) [](ancestors = 340240813)

JuanCarlosGI:
\azp run Mac BuildXL PR Validation

JuanCarlosGI:
\azp run Primary BuildXL PR Validation

PR #1102, opened 2019-10-25
Add retries around SetPathAsSharedOpaqueOutput

almiliMSFT:
Retries are needed because: (Win|Unix).SetPathAsSharedOpaqueOutput does something like:
  - check if file has write access rights
  - if it doesn't, give it those rights
  - mark the file as shared opaque output
  - ...

There is a race between checking and setting access rights, so it is possible that here we check its rights, we see that it has the correct rights, then someone else (e.g., the cache) revokes those rights, and so we fail to mark the file as shared opaque output.

### DIFF
--- a/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
+++ b/Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
@@ -181,7 +181,12 @@ namespace BuildXL.Processes
         }
 
         private const int MaxNumberOfAttemptsForMarkingSharedOpaqueOutputs = 3;
-        private static readonly TimeSpan SleepDurationBetweenMarkingAttempts = TimeSpan.FromMilliseconds(100);
+        private static readonly TimeSpan SleepDurationBetweenMarkingAttempts = TimeSpan.FromMilliseconds(10);
+
+        private static TimeSpan Multiply(TimeSpan timeSpan, int multiplier)
+        {
+            return TimeSpan.FromMilliseconds(multiplier * timeSpan.TotalMilliseconds);
+        }
 
         /// <summary>
         /// Marks a given path as "shared opaque output"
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -180,19 +180,56 @@ private static bool IsSharedOpaqueOutputWithFallback(string expandedPath, bool c
             }
         }
 
+        private const int MaxNumberOfAttemptsForMarkingSharedOpaqueOutputs = 3;
+        private static readonly TimeSpan SleepDurationBetweenMarkingAttempts = TimeSpan.FromMilliseconds(100);
```
smera: Isn't 100ms a little bit too much? I'd have said 10ms to begin with

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
```diff
@@ -180,19 +180,56 @@ private static bool IsSharedOpaqueOutputWithFallback(string expandedPath, bool c
...
         /// <summary>
         /// Marks a given path as "shared opaque output"
         /// </summary>
+        /// <remarks>
+        /// Retries are needed because: (Win|Unix).SetPathAsSharedOpaqueOutput does something like
+        ///   - check if file has write access rights
+        ///   - if it doesn't, give it those rights
+        ///   - mark the file as shared opaque output
+        ///   - ...
+        /// There is a race between checking and setting access rights, so it is possible that
```
smera: I'm not denying that this is happening, and retries are always good :). But I'm curious about what's really going on. From my understanding the cache cannot be racing with this (the pip ran, so we actually set the marker before pushing this to the cache, or it is a cache hit and then we change the marker during materialization - if the marker is not there yet). And any race between pips mean a double write anyway.

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
almiliMSFT: Will change to 10s and will add a linear backoff

Public/Src/Engine/Processes/SharedOpaqueOutputHelper.cs
almiliMSFT: All good questions.  I don't know what exactly happened.  In one build I saw, a file was just produced by a COPYFILE pip, and then marking it subsequently failed with EACCESS.  So it is possible that retrying will not help in this case, but right now that's all we can do.

PR #1104, opened 2019-10-26
Use POSIX delete for casaas to ensure that files are deleted properly even if there are open handles (with FileShare.Delete).

ljcollins25:
Use POSIX delete for casaas to ensure that files are deleted properly even if there are open handles (with FileShare.Delete).


Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -417,6 +421,12 @@ private BoolResult GarbageCollectContentCore(OperationContext context)
                 totalContentSize += entry.ContentSize * replicaCount;
                 totalContentCount += replicaCount;
 
+                int logSize = (int)Math.Log(Math.Max(1, entry.ContentSize), 2);
+                countsByLogSize[logSize]++;
+                totalSizeByLogSize[logSize] += entry.ContentSize * replicaCount;
+                uniqueSizeByLogSize[logSize] += entry.ContentSize;
```
SergeyTeplyakov: nit: extra empty line after this one.

PR #1105, opened 2019-10-26
Skip checkpoint restore rather than performing restore asynchronously which could lead to two concurrent restore operations which is not safe.

ljcollins25:
Skip checkpoint restore rather than performing restore asynchronously which could lead to two concurrent restore operations which is not safe.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
@@ -600,17 +600,20 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
             {
                 var latestCheckpoint = _checkpointManager.GetLatestCheckpointInfo(context);
                 var latestCheckpointAge = _clock.UtcNow - latestCheckpoint?.checkpointTime;
-                var shouldRestoreInBackground = latestCheckpointAge < _configuration.Checkpoint.RestoreCheckpointAgeThreshold;
+
+                // Only skip if this is the first restore and it is sufficiently recent
+                var shouldSkipRestore = _lastRestoreTime == default 
+                    && latestCheckpoint != null 
+                    && latestCheckpoint?.checkpointTime.IsRecent(_clock.UtcNow, _configuration.Checkpoint.RestoreCheckpointAgeThreshold) == true;
 
                 if (latestCheckpointAge > _configuration.LocationEntryExpiry)
                 {
                     Tracer.Debug(context, $"Checkpoint {latestCheckpoint.Value.checkpointId} age is {latestCheckpointAge}, which is larger than location expiry {_configuration.LocationEntryExpiry}");
                 }
 
-                if (shouldRestoreInBackground)
+                if (shouldSkipRestore)
                 {
-                    Tracer.Debug(context, $"Checkpoint {latestCheckpoint.Value.checkpointId} will be restored in the background. Age=[{latestCheckpointAge}], Threshold=[{_configuration.Checkpoint.RestoreCheckpointAgeThreshold}]");
-                    RestoreCheckpointStateAsync(context, checkpointState, force: true).FireAndForget(context);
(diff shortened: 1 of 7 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -600,17 +600,20 @@ private async Task<BoolResult> RestoreCheckpointStateAsync(OperationContext cont
             {
                 var latestCheckpoint = _checkpointManager.GetLatestCheckpointInfo(context);
                 var latestCheckpointAge = _clock.UtcNow - latestCheckpoint?.checkpointTime;
-                var shouldRestoreInBackground = latestCheckpointAge < _configuration.Checkpoint.RestoreCheckpointAgeThreshold;
+
+                // Only skip if this is the first restore and it is sufficiently recent
+                var shouldSkipRestore = _lastRestoreTime == default
```
SergeyTeplyakov: >_lastRestoreTime [](start = 40, length = 16)

I'm not sure if this is intentional, but _lastRestoreTime is set only for the successful case. So it is possible to hit this case in a normal case (not on startup) as well.

Does it make sense to just pass 'force: true' only when we call this method from the startup?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -600,17 +600,20 @@ private async Task<BoolResult> RestoreCheckpointStateAsync(OperationContext cont
             {
                 var latestCheckpoint = _checkpointManager.GetLatestCheckpointInfo(context);
                 var latestCheckpointAge = _clock.UtcNow - latestCheckpoint?.checkpointTime;
-                var shouldRestoreInBackground = latestCheckpointAge < _configuration.Checkpoint.RestoreCheckpointAgeThreshold;
+
+                // Only skip if this is the first restore and it is sufficiently recent
+                var shouldSkipRestore = _lastRestoreTime == default 
+                    && latestCheckpoint != null 
+                    && latestCheckpoint?.checkpointTime.IsRecent(_clock.UtcNow, _configuration.Checkpoint.RestoreCheckpointAgeThreshold) == true;
```
SergeyTeplyakov: very nit: either 'lastCheckpont != null' or the question mark in 'lastCheckpoint?.checkpointTime' are redundant.

SergeyTeplyakov:
>                         if (!store.TryGetValue(nameof(ClusterStateKeys.MaxMachineId), out var maxMachinesString, nameof(Columns.ClusterState)) ||

nit: does it make sense to use 'TryGetGlobalEntry' here?

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs:378 in 17fa364. [](commit_id = 17fa36449ab17338a814ade6f9f9bd8047a61eb7, deletion_comment = False)

ljcollins25:
>                         if (!store.TryGetValue(nameof(ClusterStateKeys.MaxMachineId), out var maxMachinesString, nameof(Columns.ClusterState)) ||

It does but that has other implications since TryGetGlobalEntry call _keyValueStore.Use. In theory this should all be moved out of keyValueStore.Use and into the base type, but that's a bit out of scope for what I was trying to achieve.

---
In reply to: [547042551](https://github.com/microsoft/BuildXL/pull/1105#issuecomment-547042551) [](ancestors = 547042551)

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs:378 in 17fa364. [](commit_id = 17fa36449ab17338a814ade6f9f9bd8047a61eb7, deletion_comment = False)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
ljcollins25: Added comment.

---
In reply to: [339677607](https://github.com/microsoft/BuildXL/pull/1105#discussion_r339677607) [](ancestors = 339677607)

PR #1106, opened 2019-10-28
[macOS] Low memory resource throttling improvements

Kr1sso:
Adjust some resource throttling mechanisms to better model low memory pressure situations on macOS (jetsam interaction). Also fixes a bug recently introduced with a polling timer for memory usage observation.

### DIFF
--- a/Public/Src/Engine/Scheduler/ProcessResourceManager.cs
+++ b/Public/Src/Engine/Scheduler/ProcessResourceManager.cs
@@ -243,6 +243,7 @@ namespace BuildXL.Scheduler
             private Func<int> m_queryRamUsageMb;
             private bool m_completed;
             private bool m_cancelled;
+            private readonly object m_refreshLock = new object();
 
             public readonly PipId PipId;
             public readonly int ExpectedRamUsageMb;
@@ -286,12 +287,10 @@ namespace BuildXL.Scheduler
 
             public int QueryRamUsageMb()
             {
-                if (!m_cancellation.IsCancellationRequested && !m_cancelled && !m_completed && !m_isDisposed)
+                lock(m_refreshLock)
                 {
                     return m_queryRamUsageMb?.Invoke() ?? 0;
                 }
-
-                return 0;
             }
 
             public void RefreshRamUsage()
@@ -326,9 +325,13 @@ namespace BuildXL.Scheduler
 
             public void Dispose()
             {
-                Contract.Assert(!m_completed || m_queryRamUsageMb != null, "Must register query ram usage before completion");
-                m_queryRamUsageMb = null;
-                m_isDisposed = true;
+                lock(m_refreshLock)
+                {
+                    Contract.Assert(!m_completed || m_queryRamUsageMb != null, "Must register query ram usage before completion");
+                    m_queryRamUsageMb = null;
+                    m_isDisposed = true;
+                }
+
                 m_cancellation.Dispose();
                 m_resourceManager.FreeExecution(PipId);
             }
--- a/Public/Src/Engine/Scheduler/ProcessResourceManager.cs
+++ b/Public/Src/Engine/Scheduler/ProcessResourceManager.cs
@@ -96,6 +96,23 @@ namespace BuildXL.Scheduler
             }
         }
 
+        /// <summary>
+        /// Updates the ram usage indicators for all cancelable resource scopes
+        /// </summary>
+        public void UpdateRamUsageForResourceScopes()
+        {
+            lock (m_syncLock)
+            {
+                foreach (var scope in m_pipResourceScopes.Values)
+                {
+                    if (scope.CanCancel)
+                    {
+                        scope.RefreshRamUsage();
+                    }
+                }
+            }
+        }
+
         private void FreeResourcesByPreference(
             IComparer<ResourceScope> scopeComparer,
             ref int requiredRam,
@@ -140,7 +157,6 @@ namespace BuildXL.Scheduler
             }
 
             using (scope)
-            using (new StoppableTimer(() => scope.RefreshRamUsage(), 0, 2000))
             {
                 var result = await execute(scope.Token, registerQueryRamUsageMb: scope.RegisterQueryRamUsageMb);
                 scope.Complete();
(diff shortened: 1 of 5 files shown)


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2173,9 +2173,22 @@ private void UpdateResourceAvailability(PerformanceCollector.MachinePerfInfo per
...
-            resourceAvailable &= !(result && (pressureLevel > m_configuration.Schedule.MaximumAllowedMemoryPressureLevel));
+            if (!resourceAvailable)
+            {
+                Memory.PressureLevel pressureLevel = Memory.PressureLevel.Normal;
+                var result = Memory.GetMemoryPressureLevel(ref pressureLevel) == Dispatch.MACOS_INTEROP_SUCCESS;
+                resourceAvailable = !(result && (pressureLevel > m_configuration.Schedule.MaximumAllowedMemoryPressureLevel));
+
+                if (!result)
+                {
+                    Logger.Log.UnableToGetMemoryPressureLevel(
```
almiliMSFT: why would you ever not be able to get memory pressure level?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2173,9 +2173,22 @@ private void UpdateResourceAvailability(PerformanceCollector.MachinePerfInfo per
             }
 
 #if PLATFORM_OSX
-            Memory.PressureLevel pressureLevel = Memory.PressureLevel.Normal;
-            var result = Memory.GetMemoryPressureLevel(ref pressureLevel) == Dispatch.MACOS_INTEROP_SUCCESS;
-            resourceAvailable &= !(result && (pressureLevel > m_configuration.Schedule.MaximumAllowedMemoryPressureLevel));
+            if (!resourceAvailable)
```
almiliMSFT: maybe leave a comment that the idea here is that even if resources are not available by the `exceededMaxRamUtilizationPercentage && underMinimumAvailableRam` metric, on macOS we can override that decision if the pressure level is under a configured threshold

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2173,9 +2173,22 @@ private void UpdateResourceAvailability(PerformanceCollector.MachinePerfInfo per
...
 
 #if PLATFORM_OSX
-            Memory.PressureLevel pressureLevel = Memory.PressureLevel.Normal;
-            var result = Memory.GetMemoryPressureLevel(ref pressureLevel) == Dispatch.MACOS_INTEROP_SUCCESS;
-            resourceAvailable &= !(result && (pressureLevel > m_configuration.Schedule.MaximumAllowedMemoryPressureLevel));
+            if (!resourceAvailable)
+            {
+                Memory.PressureLevel pressureLevel = Memory.PressureLevel.Normal;
+                var result = Memory.GetMemoryPressureLevel(ref pressureLevel) == Dispatch.MACOS_INTEROP_SUCCESS;
+                resourceAvailable = !(result && (pressureLevel > m_configuration.Schedule.MaximumAllowedMemoryPressureLevel));
```
almiliMSFT: This implies that if `result` is false (i.e., we were not able to obtain memory pressure level), we set `resourceAvailable` to `true`.

That looks strange to me.  It would make more sense to me that if we are unable to calculate memory pressure, we just use the previous metric for computing `resourceAvailable` (which in this case already determined that resources are not available)

Public/Src/Engine/Scheduler/Scheduler.cs
Kr1sso: I hope we won't - if the backing sysctl key or implementation changes this at least emits an error so we don't fly blind.

Public/Src/Engine/Scheduler/Scheduler.cs
Kr1sso: Will add a comment.

Public/Src/Engine/Scheduler/Scheduler.cs
Kr1sso: My idea is to have unbounded resource usage here to identify issues because we get jetsam'ed, this way we could end up with a lot of premature cancelation (in the case we can't read memory pressure) and very long running builds. I guess the latter is a better indicator that something is wrong than getting killed by the system.

Public/Src/Engine/Scheduler/ProcessResourceManager.cs
```diff
@@ -270,7 +286,12 @@ public void RegisterQueryRamUsageMb(Func<int> queryRamUsageMb)
 
             public int QueryRamUsageMb()
             {
-                return m_queryRamUsageMb?.Invoke() ?? 0;
```
ljcollins25: >m_queryRamUsageMb [](start = 23, length = 17)

I think there needs to be lock here so we prevent calling this concurrently with dispose. Maybe capture the variable, check if not null, and lock on captured variable. Also do the same in dispose.

Public/Src/Engine/Scheduler/ProcessResourceManager.cs
```diff
@@ -270,7 +286,12 @@ public void RegisterQueryRamUsageMb(Func<int> queryRamUsageMb)
 
             public int QueryRamUsageMb()
             {
-                return m_queryRamUsageMb?.Invoke() ?? 0;
+                if (!m_cancellation.IsCancellationRequested && !m_cancelled && !m_completed && !m_isDisposed)
```
ljcollins25: >!m_cancellation.IsCancellationRequested && !m_cancelled && !m_completed && !m_isDisposed [](start = 20, length = 88)

These checks shouldn't be needed.

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2173,9 +2172,27 @@ private void UpdateResourceAvailability(PerformanceCollector.MachinePerfInfo per
...
+
+                if (result)
+                {
+                    // If the memory pressure level is not above the configured level but we've infered resources are not available earlier,
+                    // we reset the resource availability and override the decision by looking at the current pressure level only!
+                    resourceAvailable = !(pressureLevel > m_configuration.Schedule.MaximumAllowedMemoryPressureLevel);
+                }
+                else
+                {
+                    Logger.Log.UnableToGetMemoryPressureLevel(
```
semihokur: > Logger.Log.UnableToGetMemoryPressureLevel( [](start = 19, length = 43)

"resource cancelation will be skipped all together"

Does not it mean that we should set 'resourceAvailable' to true?

PR #1112, opened 2019-10-28
Remove internal DHelp link and simplify help argument parsing

mpysson:


Public/Src/Utilities/ToolSupport/HelpWriter.cs
```diff
@@ -214,19 +214,14 @@ public enum HelpLevel : byte
...
         /// The standard help
         /// </summary>
-        Standard = 2,
+        Standard = 1,
 
         /// <summary>
         /// The full help including verbose obscure options
         /// </summary>
-        Verbose = 3,
+        Verbose = 1,
```
norahuang: Verbose = 2,

PR #1113, opened 2019-10-28
Fix the NRE when logging mem usage

semihokur:
There are some NRE happening with random line numbers. I suspect that those are related to logging mem usage. 

System.NullReferenceException: Object reference not set to an instance of an object.
   at async Task<PipExecutionStep> BuildXL.Scheduler.Scheduler.ExecutePipStep(RunnablePip runnablePip) in \.\Public\Src\Engine\Scheduler\Scheduler.cs:line 3769

Fixes [AB#1634391](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1634391)

Fixes [AB#1634458](https://dev.azure.com/mseng/1ES/_boards/board/t/MichaelP%20-%20Team/Stories/?workitem=1634458)

### DIFF
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -3582,9 +3582,7 @@ namespace BuildXL.Scheduler
                         {
                             var perfInfo = executionResult.PerformanceInformation;
 
-                            if (perfInfo != null)
-                            {
-                                m_groupedPipCounters.AddToCounters(
+                            m_groupedPipCounters.AddToCounters(
                                     processRunnable.Process,
                                     new[]
                                     {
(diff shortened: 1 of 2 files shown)


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3582,9 +3582,7 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
                         {
                             var perfInfo = executionResult.PerformanceInformation;
 
-                            if (perfInfo != null)
```
mpysson: >perfInfo [](start = 32, length = 8)

I guess this is guarded against being null by virtue of EnsureSealed() in the getter. Worth leaving an assert in there for the comfort of future readers?

semihokur: we already check whether this is null just above: 

 (executionResult.PerformanceInformation != null)

---
In reply to: [339821083](https://github.com/microsoft/BuildXL/pull/1113#discussion_r339821083) [](ancestors = 339821083)

PR #1114, opened 2019-10-28
Always delete temporary file after copying

kewilli:
When copying content from the L2 cache, it is copied locally to a temporary file, then Put into the CAS by moving it. If the Put was unsuccessful, then the temporary file is deleted. If the Put was successful, then we assume that the temporary file no longer exists due to the Put's move.

Suppose two copies occur simultaneously. In this case, two temporary files are created locally. The first to complete will successfully Put the content into the CAS and the temp file will be moved into the CAS, leaving nothing in the Temp folder. The second will successfully Put, but only because CAS already has this content (satisfied by the first Put). The second temp file will not be deleted, because the Put is successful, so the temp file will continue to exist there.

This change ensures that we attempt to delete the temp file whether the put succeeded or failed.


dcwuser:
Not  a fan of this change in logic. Doing a move instead of copy+delete was specifically chosen to avoid two calls into the disk system. Couldn't we only do the delete if the second move fails?

dcwuser:
While I'm not surprised this can occur in principal, I'm surprised that it occurs at a big enough rate that we care. The fact that it does makes me wonder (1) if understand why it does occur and (2) if we shouldn't de-duplicate copying so that we don't pay the cost of moving the bytes off disk, across the wire, and onto new disk multiple times simultaneously.

Public/Src/Cache/ContentStore/Library/Stores/FileSystemContentStoreInternal.cs
```diff
@@ -1482,7 +1486,7 @@ protected override void DisposeCore()
 
             if (failed)
             {
-                return false;
+                return (false, contentExistsInCache);
```
SergeyTeplyakov: nit: could you please use tuple's name here?

Public/Src/Cache/ContentStore/Interfaces/Results/PutResult.cs
```diff
@@ -15,12 +15,13 @@ public class PutResult : BoolResult, IEquatable<PutResult>
         /// <summary>
         ///     Initializes a new instance of the <see cref="PutResult" /> class.
         /// </summary>
-        public PutResult(ContentHash contentHash, long contentSize)
+        public PutResult(ContentHash contentHash, long contentSize, bool contentAlreadyExistsInCache = false)
```
dcwuser: This is the break-risking change that Sergey warned about. If you add this optional argument, then customers (e.g. ADO) that link to us and use the constrcutor will break unless they re-compile against the new bits. To avoid this, you need to add a separate new constructor.

dcwuser:
>         public override string ToString()

This is minor and not necessary, but if you are doing another iteration, please add the ContentAlreadyExistsInCache to ToString (and ensure ToString is used in log message), so that we can see how often this happens.

---
Refers to: Public/Src/Cache/ContentStore/Interfaces/Results/PutResult.cs:123 in 1f0d32b. [](commit_id = 1f0d32bf63e63581ee385e753ff3abb36e502b4e, deletion_comment = False)

PR #1119, opened 2019-10-29
Increase weight by unit tests that timeout under heavy concurrency

mpysson:
Mitigation for tests that timeout

### DIFF
--- a/Public/Src/Engine/UnitTests/FingerprintStore/Test.BuildXL.FingerprintStore.dsc
+++ b/Public/Src/Engine/UnitTests/FingerprintStore/Test.BuildXL.FingerprintStore.dsc
@@ -36,6 +36,6 @@ namespace Test.BuildXL.FingerprintStore {
         runtimeContent: [
             importFrom("BuildXL.Utilities.UnitTests").TestProcess.deploymentDefinition
         ],
-        runTestArgs: { weight: 2 },
+        runTestArgs: { weight: 10 },
     });
 }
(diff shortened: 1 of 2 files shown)


ljcollins25:
>     export const dll = BuildXLSdk.test({

Should we instead split up the tests using the parallel bucketization? See here: [http://index/?rightProject=BuildXL.Cache.MemoizationStoreAdapter.Test&file=Test.BuildXL.Cache.MemoizationStoreAdapter.dsc&line=12](http://index/?rightProject=BuildXL.Cache.MemoizationStoreAdapter.Test&file=Test.BuildXL.Cache.MemoizationStoreAdapter.dsc&line=12)

---
Refers to: Public/Src/Tools/UnitTests/Analyzers/Test.Tool.Analyzers.dsc:11 in 9705efb. [](commit_id = 9705efb5be34b6e76f490a815c932a3ed3fa9331, deletion_comment = False)

mpysson:
> > ```
> > export const dll = BuildXLSdk.test({
> > ```
> 
> Should we instead split up the tests using the parallel bucketization? See here: http://index/?rightProject=BuildXL.Cache.MemoizationStoreAdapter.Test&file=Test.BuildXL.Cache.MemoizationStoreAdapter.dsc&line=12
> 
> Refers to: Public/Src/Tools/UnitTests/Analyzers/Test.Tool.Analyzers.dsc:11 in [9705efb](https://github.com/microsoft/BuildXL/commit/9705efb5be34b6e76f490a815c932a3ed3fa9331). [](commit_id = [9705efb](https://github.com/microsoft/BuildXL/commit/9705efb5be34b6e76f490a815c932a3ed3fa9331), deletion_comment = False)

I'm not sure that will help. I believe these tests are timing out because the RocksDB based tests are saturating I/O on the machine. I don't think running more of them in parallel will make them run faster if they're blocked on the same resource. There are already enough variants with qualifiers for them to be able to land on different machines

PR #1126, opened 2019-10-30
[macOS] Forward all signal handling masks to spawned child processes

Kr1sso:
Fix child process mask forwarding and bump interop lib, this allows spawned processes (pips) to do their own signal / dump handling instead of BuildXL taking care of it all.


Kr1sso:
The sandbox kernel extension has been republished and not touched (2.4.99).

PR #1127, opened 2019-10-30
Add fast pass for re-pinning connect from hibernated sessions.

SergeyTeplyakov:


Public/Src/Cache/DistributedCache.Host/Configuration/DistributedContentSettings.cs
```diff
@@ -562,6 +562,9 @@ public int[] RetryIntervalForCopiesMs
         [DataMember]
         public bool TraceFileSystemContentStoreDiagnosticMessages { get; set; } = false;
 
+        [DataMember]
+        public bool SkipTouchAndLockAcquisitionWhenPinningFromHibernation { get; set; } = false;
```
ljcollins25: >SkipTouchAndLockAcquisitionWhenPinningFromHibernation [](start = 20, length = 53)

This name is a bit overly verbose. Maybe something like UseFastHibernationPin

PR #1138, opened 2019-11-01
Failure in making outputs private should not fail builds

narasamdya:
When making output directory private, instead of failing the build, we default to the normal behavior where we wipe out the directory. This is consistent with the treatment of preserved output file.

We also include the error code in the exception and log message when move-file failed.

[AB#1640665](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1640665)

### DIFF
--- a/Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
+++ b/Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
@@ -313,11 +313,7 @@ namespace BuildXL.Engine.Cache.Artifacts
                 var updatedDestination = destination.Substring(0, destination.Length - requiredFileName.Length) + requiredFileName;
 
                 // Move file to temporary location
-                var changeCasing = await FileUtilities.MoveFileAsync(destination, updatedDestination, replaceExisting: false);
-                if (!changeCasing)
-                {
-                    return new Failure<string>(I($"Could not move file '{destination}' to change file name casing to '{requiredFileName}'."));
-                }
+                await FileUtilities.MoveFileAsync(destination, updatedDestination, replaceExisting: false);
 
                 return Unit.Void;
             }
--- a/Public/Src/Engine/Dll/Recovery/CorruptedMemosDbRecovery.cs
+++ b/Public/Src/Engine/Dll/Recovery/CorruptedMemosDbRecovery.cs
@@ -112,12 +112,7 @@ namespace BuildXL.Engine.Recovery
             //         RenameMemoDb();
             //     }
             var corruptedMemosDbFile = PrepareCorruptedMemosDbBackupFile(cacheDirectory);
-            var result = FileUtilities.MoveFileAsync(memosDbFile, corruptedMemosDbFile, replaceExisting: true).Result;
-
-            if (!result)
-            {
-                return new Failure<string>(I($"Failed to rename '{memosDbFile}' to '{corruptedMemosDbFile}'"));
-            }
+            FileUtilities.MoveFileAsync(memosDbFile, corruptedMemosDbFile, replaceExisting: true).Wait();
 
             // Delete marker file in case of successful recovery.
             FileUtilities.DeleteFile(MarkerFile(cacheDirectory));
--- a/Public/Src/Engine/Dll/Recovery/CorruptedMemosDbRecovery.cs
+++ b/Public/Src/Engine/Dll/Recovery/CorruptedMemosDbRecovery.cs
@@ -111,11 +111,18 @@ namespace BuildXL.Engine.Recovery
             //         Assert(cacheIntegrityResult == NoIssueFound);
             //         RenameMemoDb();
             //     }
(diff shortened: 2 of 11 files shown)


Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
```diff
@@ -313,11 +313,7 @@ private Possible<Unit> CreateSymlinkIfNotExistsOrTargetMismatch(AbsolutePath sym
                 var updatedDestination = destination.Substring(0, destination.Length - requiredFileName.Length) + requiredFileName;
 
                 // Move file to temporary location
-                var changeCasing = await FileUtilities.MoveFileAsync(destination, updatedDestination, replaceExisting: false);
-                if (!changeCasing)
-                {
-                    return new Failure<string>(I($"Could not move file '{destination}' to change file name casing to '{requiredFileName}'."));
```
mpysson: >    return new Failure<string>(I($"Could not move file '{destination}' to change file name casing to '{requiredFileName}'.")); [](start = 16, length = 126)

You changed the behavior here from returning a Failure<> to throwing an exception. Is that safe to do? I think it's unexpected for a Possible<>returning method to throw sometimes and return a Failure other times

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2621,13 +2622,21 @@ private async Task<bool> PrepareDirectoryOutputsAsync(HashSet<AbsolutePath> pres
...
                                     if (!await m_makeOutputPrivate(path))
                                     {
-                                        throw new BuildXLException("Failed to create a private, writeable copy of an output file from a previous invocation.");
+                                        allOutputsPrivate = false;
+                                        break;
                                     }
                                 }
+
+                                if (!allOutputsPrivate)
+                                {
```
mpysson: Suggest we log something to indicate preserving outputs failed and we had to fall back on deleting them instead

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -2621,13 +2622,21 @@ private async Task<bool> PrepareDirectoryOutputsAsync(HashSet<AbsolutePath> pres
                                         }
                                     });
 
+                                bool allOutputsPrivate = true;
+
                                 foreach (var path in filePaths)
                                 {
                                     if (!await m_makeOutputPrivate(path))
```
mpysson: >m_makeOutputPrivate [](start = 47, length = 19)

Depending on how TryMakeExclusiveLinkAsync fails there may be dangling temp files. Should we clean them up?

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
narasamdya: Previously, the failure code paths in all occurrences of `MoveFileAsycn` have never been exercised. `MoveFileAsync` either returns true (not `Possible` nor `Failure`) or throws an exception. Also, this consistent with `MoveFile` method in .NET. 

I couldn't change `CopyFileAsync` because our version take a predicate as additional argument.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
narasamdya: For preserved output file, they will eventually be scrubbed.
For preserved output directory, the directory will be wiped out, and so the dangling temp file should get deleted as well.

Public/Src/Engine/Cache/Artifacts/LocalDiskContentStore.cs
mpysson: I was referring to `Task<Possible<Unit, Failure>> EnsureFileNameCasingMatchAsync()` now throwing rather than returning a Failure. Now that I look closer I guess it won't actually throw because `MoveFileAsync()` wraps its `NativeWin32Exception` in a `BuildXLException` which `EnsureFileNameCasingMatchAsync` then handles. So nevermind on that one

Public/Src/Engine/Dll/Recovery/CorruptedMemosDbRecovery.cs
```diff
@@ -112,12 +112,7 @@ public override Possible<Unit> Recover()
             //         RenameMemoDb();
             //     }
             var corruptedMemosDbFile = PrepareCorruptedMemosDbBackupFile(cacheDirectory);
-            var result = FileUtilities.MoveFileAsync(memosDbFile, corruptedMemosDbFile, replaceExisting: true).Result;
-
-            if (!result)
-            {
-                return new Failure<string>(I($"Failed to rename '{memosDbFile}' to '{corruptedMemosDbFile}'"));
```
mpysson: Here you did change the behavior of `Possible<Unit> Recover()` to now throw instead of returning a Failure

PR #1139, opened 2019-11-01
Use a single lock per trie (instead of per node)

almiliMSFT:


Public/Src/Sandbox/MacOs/Sandbox/Src/Utilities/Trie.hpp
```diff
@@ -96,6 +96,9 @@ class Trie : public OSObject
     /*! Payload for the 'onChangeCallback_' function */
     void *onChangeData_;
 
+    /*! Used only when modifying this node's list of children */
```
almiliMSFT:
```suggestion
    /*! Used only when modifying a node's list of children */
```

Public/Src/Engine/UnitTests/Processes/ProcessDumperTests.cs
```diff
@@ -174,9 +174,9 @@ public void DumpProcessTreeTest()
         [DllImport("libc", SetLastError = true, EntryPoint = "kill")]
         private static extern unsafe int SendSignal(int pid, int signal);
 
-        private const int SIG_ABRT = 6;
+        private const int SIG_ILL = 4;
```
almiliMSFT: for some reason SIGABRT doesn't kill the process

PR #1141, opened 2019-11-01
Grpc push copy supporting touch-less puts

JuanCarlosGI:
Implementing a gRPC push file, to eventually replace the current way of copyiong to a machine of requesting them to request a copy from us.

Also, making sure that these copies do not bump the lastAccessTimes of the content.

Note about how the push copy is implemented. There is an initial "handshake" between Machine A trying to send the content to Machine B (This is done via request/response headers):
- A: I want to send you hash X
- B: "Start sending bytes" or "Do not copy (either because I already have the content or I'm currently copying the content)"
- A starts or aborts the copy
- B reports back the status of the operation (Only when the copy was initiated)

### DIFF
--- a/Public/Src/Cache/ContentStore/App/CopyFileTo.cs
+++ b/Public/Src/Cache/ContentStore/App/CopyFileTo.cs
@@ -1,0 +1,78 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Threading;
+using BuildXL.Cache.ContentStore.Exceptions;
+using BuildXL.Cache.ContentStore.Hashing;
+using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
+using BuildXL.Cache.ContentStore.Interfaces.Tracing;
+using BuildXL.Cache.ContentStore.Service;
+using BuildXL.Cache.ContentStore.Service.Grpc;
+using BuildXL.Cache.ContentStore.Tracing.Internal;
+using BuildXL.Cache.ContentStore.Utils;
+using CLAP;
+using Microsoft.Practices.TransientFaultHandling;
+
+namespace BuildXL.Cache.ContentStore.App
+{
+    internal sealed partial class Application
+    {
+        /// <summary>
+        /// Run the CopyFileTo verb.
+        /// </summary>
+        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
+        [Verb(Description = "Copy file to another CASaaS")]
+        internal void CopyFileTo(
+            [Required, Description("Machine to copy to")] string host,
+            [Required, Description("Path to source file")] string sourcePath,
+            [Description("File name where the GRPC port can be found when using cache service. 'CASaaS GRPC port' if not specified")] string grpcPortFileName,
+            [Description("The GRPC port"), DefaultValue(0)] int grpcPort)
+        {
+            Initialize();
+
+            var context = new Context(_logger);
+            var operationContext = new OperationContext(context, CancellationToken.None);
+            var retryPolicy = new RetryPolicy(
+                new TransientErrorDetectionStrategy(),
+                new FixedInterval("RetryInterval", (int)_retryCount, TimeSpan.FromSeconds(_retryIntervalSeconds), false));
+
+            if (grpcPort == 0)
+            {
+                grpcPort = Helpers.GetGrpcPortFromFile(_logger, grpcPortFileName);
+            }
+
+            var hasher = ContentHashers.Get(HashType.MD5);
+            var bytes = File.ReadAllBytes(sourcePath);
+            var hash = hasher.GetContentHash(bytes);
+
+            try
+            {
+                using var clientCache = new GrpcCopyClientCache(context);
+                using var rpcClientWrapper = clientCache.CreateAsync(host, grpcPort, useCompression: false).GetAwaiter().GetResult();
+                var rpcClient = rpcClientWrapper.Value;
+                var path = new AbsolutePath(sourcePath);
+
+                using var stream = File.OpenRead(path.Path);
+
+                // This action is synchronous to make sure the calling application doesn't exit before the method returns.
+                var copyFileResult = retryPolicy.ExecuteAsync(() => rpcClient.PushFileAsync(operationContext, hash, stream)).Result;
+                if (!copyFileResult.Succeeded)
+                {
+                    _logger.Error($"{copyFileResult}");
+                    throw new CacheException(copyFileResult.ErrorMessage);
+                }
+                else
+                {
+                    _logger.Info($"Copy of {sourcePath} was successful");
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new CacheException(ex.ToString());
+            }
+        }
+    }
+}
--- a/Public/Src/Cache/ContentStore/App/CopyFileTo.cs
+++ b/Public/Src/Cache/ContentStore/App/CopyFileTo.cs
@@ -2,9 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Threading;
+using System.Threading.Tasks;
 using BuildXL.Cache.ContentStore.Exceptions;
 using BuildXL.Cache.ContentStore.Hashing;
 using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
@@ -54,10 +54,10 @@ namespace BuildXL.Cache.ContentStore.App
                 var rpcClient = rpcClientWrapper.Value;
                 var path = new AbsolutePath(sourcePath);
 
-                using var stream = File.OpenRead(path.Path);
+                using Stream stream = File.OpenRead(path.Path);
 
                 // This action is synchronous to make sure the calling application doesn't exit before the method returns.
-                var copyFileResult = retryPolicy.ExecuteAsync(() => rpcClient.PushFileAsync(operationContext, hash, stream)).Result;
+                var copyFileResult = retryPolicy.ExecuteAsync(() => rpcClient.PushFileAsync(operationContext, hash, () => Task.FromResult(stream))).Result;
                 if (!copyFileResult.Succeeded)
                 {
                     _logger.Error($"{copyFileResult}");
--- a/Public/Src/Cache/ContentStore/App/CopyFileTo.cs
+++ b/Public/Src/Cache/ContentStore/App/CopyFileTo.cs
@@ -23,7 +23,6 @@ namespace BuildXL.Cache.ContentStore.App
         /// <summary>
         /// Run the CopyFileTo verb.
         /// </summary>
-        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
         [Verb(Description = "Copy file to another CASaaS")]
         internal void CopyFileTo(
             [Required, Description("Machine to copy to")] string host,
--- a/Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
@@ -71,6 +71,11 @@ namespace BuildXL.Cache.ContentStore.Distributed
         /// <param name="hash">The hash of the file to be copied.</param>
         /// <param name="targetMachine">The machine that should copy the file</param>
         Task<BoolResult> RequestCopyFileAsync(OperationContext context, ContentHash hash, MachineLocation targetMachine);
+
+        /// <summary>
+        /// Pushes content to a target machine.
+        /// </summary>
+        Task<BoolResult> PushFileAsync(OperationContext context, ContentHash hash, Stream source, MachineLocation targetMachine);
     }
 
     /// <summary>
--- a/Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
@@ -60,16 +60,13 @@ namespace BuildXL.Cache.ContentStore.Distributed
     }
 
     /// <summary>
-    /// Requests another machine to copy from the current machine.
+    /// Copies files to another machine.
     /// </summary>
-    public interface ICopyRequester
+    public interface ICopySender
     {
         /// <summary>
         /// Requests another machine to copy a file.
         /// </summary>
-        /// <param name="context">The context of the operation</param>
-        /// <param name="hash">The hash of the file to be copied.</param>
-        /// <param name="targetMachine">The machine that should copy the file</param>
         Task<BoolResult> RequestCopyFileAsync(OperationContext context, ContentHash hash, MachineLocation targetMachine);
 
         /// <summary>
--- a/Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
@@ -1,13 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using BuildXL.Cache.ContentStore.Hashing;
 using BuildXL.Cache.ContentStore.Interfaces.FileSystem;
 using BuildXL.Cache.ContentStore.Interfaces.Results;
-using BuildXL.Cache.ContentStore.Interfaces.Tracing;
 using BuildXL.Cache.ContentStore.Tracing.Internal;
 
 // ReSharper disable All
@@ -72,7 +72,7 @@ namespace BuildXL.Cache.ContentStore.Distributed
         /// <summary>
         /// Pushes content to a target machine.
         /// </summary>
-        Task<BoolResult> PushFileAsync(OperationContext context, ContentHash hash, Stream source, MachineLocation targetMachine);
+        Task<BoolResult> PushFileAsync(OperationContext context, ContentHash hash, Func<Task<Stream>> source, MachineLocation targetMachine);
     }
 
     /// <summary>
--- a/Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
@@ -62,7 +62,7 @@ namespace BuildXL.Cache.ContentStore.Distributed
     /// <summary>
     /// Copies files to another machine.
     /// </summary>
-    public interface ICopySender
+    public interface IProactiveCopier
     {
         /// <summary>
         /// Requests another machine to copy a file.
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventData.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventData.cs
@@ -28,6 +28,9 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache.EventStreaming
 
         /// <nodoc />
         Reconcile,
+
+        /// <nodoc />
+        AddLocationWithoutTouching,
     }
 
     /// <summary>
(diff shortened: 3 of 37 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventData.cs
```diff
@@ -203,45 +204,51 @@ public sealed class AddContentLocationEventData : ContentLocationEventData
         /// </summary>
         public IReadOnlyList<long> ContentSizes { get; }
 
+        /// <summary>
+        /// Whether the operation should cuase the content last access time to be updated.
+        /// </summary>
+        public bool ShouldTouch { get; }
```
ljcollins25: >ShouldTouch [](start = 20, length = 11)

Changing this introduces a breaking change to the serialization protocol for event hub. This will cause errors. You need to make this change in a backward compatible manner.

JuanCarlosGI: Very well. I'll add a new event, something like "AddWithoutTouching"

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
```diff
@@ -33,6 +33,8 @@ service ContentServer {
 
     rpc CopyFile(CopyFileRequest) returns (stream CopyFileResponse) {}
 
+    rpc PushFile(stream PushFileRequest) returns (stream PushFileResponse) {}
```
dcwuser: 1. Why is this a bi-directional stream? 2. Why don't we use the same streaming chunk type for pushing and pulling? Here is proto I used in prototype:
service Copier {
  rpc Read (ReadRequest) returns (stream Chunk) {}
  rpc Write (stream Chunk) returns (WriteReply) {}
}

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
```diff
@@ -196,6 +198,19 @@ enum CopyCompression {
     gzip = 1;
 }
 
+// PushFile
+message PushFileRequest {
```
dcwuser: >PushFileRequest [](start = 8, length = 15)

Not happy that we are adding extra data to every chunk. Stuff transmitted just once should go in header.

Public/Src/Cache/ContentStore/App/CopyFileTo.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.Practices.TransientFaultHandling;
+
+namespace BuildXL.Cache.ContentStore.App
+{
+    internal sealed partial class Application
+    {
+        /// <summary>
+        /// Run the CopyFileTo verb.
+        /// </summary>
+        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
```
SergeyTeplyakov: Do we need this?

Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
```diff
@@ -60,17 +60,19 @@ public interface ITraceableAbsolutePathFileCopier : IAbsolutePathFileCopier, ITr
     }
 
     /// <summary>
-    /// Requests another machine to copy from the current machine.
+    /// Copies files to another machine.
     /// </summary>
-    public interface ICopyRequester
+    public interface ICopySender
```
SergeyTeplyakov: It seems that this interface supports both, the push and the pull models of copying files to the machine.

So, maybe the name should be reflect that...
Like IProactiveCopier? (this one is bad as well, but at least it is vague enough)... Happy to brainstorm the name:)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -933,7 +933,7 @@ private Task<BoolResult> EnsureInitializedAsync()
         /// <summary>
         /// Notifies a central store that content represented by <paramref name="contentHashes"/> is available on a current machine.
         /// </summary>
-        public async Task<BoolResult> RegisterLocalLocationAsync(OperationContext context, IReadOnlyList<ContentHashWithSize> contentHashes)
+        public async Task<BoolResult> RegisterLocalLocationAsync(OperationContext context, IReadOnlyList<ContentHashWithSize> contentHashes, bool touch = true)
```
SergeyTeplyakov: nit: you can keep the default value only on the interface declaration level and remove it here.

Public/Src/Cache/ContentStore/Distributed/NuCache/TransitioningContentLocationStore.cs
```diff
@@ -136,12 +136,12 @@ public Task<GetBulkLocationsResult> GetBulkAsync(Context context, IReadOnlyList<
         }
 
         /// <inheritdoc />
-        public Task<BoolResult> RegisterLocalLocationAsync(Context context, IReadOnlyList<ContentHashWithSize> contentHashes, CancellationToken cts, UrgencyHint urgencyHint)
+        public Task<BoolResult> RegisterLocalLocationAsync(Context context, IReadOnlyList<ContentHashWithSize> contentHashes, CancellationToken cts, UrgencyHint urgencyHint, bool touch = true)
```
SergeyTeplyakov: the same nit: I think I'll prefer not to pollute the code with defaults.

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventData.cs
```diff
@@ -90,6 +93,8 @@ public static ContentLocationEventData Deserialize(BuildXLReader reader, DateTim
                     return new TouchContentLocationEventData(sender, hashes, eventTimeUtc);
                 case EventKind.Reconcile:
                     return new ReconcileContentLocationEventData(sender, reader.ReadString());
+                case EventKind.AddLocationWithoutTouching:
+                    return new AddContentLocationWithoutTouchingEventData(sender, hashes, reader.ReadReadOnlyList(r => r.ReadInt64Compact()));
```
SergeyTeplyakov: To avoid any runtime issues we should make sure that we won't send this events until all the masters are updated.

The bits rollout is happening randomly, so if the feature that causes this events to be sent are on by default it is possible that some workers will be updated first, generate this event and will send it to the old master.

JuanCarlosGI: It will just be a matter of keeping this disabled until rollout is complete. It's disabled by default, so I don't think there is much risk, as long as we leave a reasonable period between release and enabling the feature

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -264,11 +264,20 @@ private Task<PutResult> PerformPutFileGatedOperationAsync(OperationContext opera
             if (registerResult && Settings.ProactiveCopyMode != ProactiveCopyMode.Disabled)
             {
                 // Since the rest of the operation is done asynchronously, create new context to stop cancelling operation prematurely.
-                WithOperationContext(
+                var proactiveCopyTask = WithOperationContext(
                     context,
                     CancellationToken.None,
-                    operationContext => RequestProactiveCopyIfNeededAsync(operationContext, result.ContentHash, path)
```
SergeyTeplyakov: >RequestProactiveCopyIfNeededAsync [](start = 40, length = 33)

In this case you're awaiting the result but ignoring the errors.

So you can trace explicitly, or do our normal:

if (!proactiveCopyResult.Success) {return ...}

SergeyTeplyakov: Or you can explicitely ignore the result and add a comment saying that the error is traced.

---
In reply to: [342737938](https://github.com/microsoft/BuildXL/pull/1141#discussion_r342737938) [](ancestors = 342737938)

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventData.cs
JuanCarlosGI: It will just be a matter of keeping this disabled until rollout is complete. It's disabled by default, so I don't think there is much risk, as long as we leave a reasonable period between release and enabling the feature

---
In reply to: [342735439](https://github.com/microsoft/BuildXL/pull/1141#discussion_r342735439) [](ancestors = 342735439)

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventData.cs
JuanCarlosGI: Very well. I'll add a new event, something like "AddWithoutTouching"

---
In reply to: [341698926](https://github.com/microsoft/BuildXL/pull/1141#discussion_r341698926) [](ancestors = 341698926)

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
JuanCarlosGI: I used a bidirectional stream to keep the whole transaction in a single request. In order to transmit the response headers as soon as we write them, it is required that the response type is a stream. Otherwise, the headers will be transmitted with the final response, making it impossible for us to complete the initial handshake to determine if copies are needed.

---
In reply to: [342196589](https://github.com/microsoft/BuildXL/pull/1141#discussion_r342196589) [](ancestors = 342196589)

Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -245,7 +245,7 @@ public Task<BoolResult> RequestCopyFileAsync(OperationContext context, ContentHa
             return _proactiveCopyIoGate.GatedOperationAsync(ts =>
                 {
                     var cts = new CancellationTokenSource();
```
SergeyTeplyakov: we need to wrap cts into a using block (or use using var cts).

This is very important, because currently, we'll trigger the timer even when the operation finishes successfully.

THis can be pretty bad if we have hundreds of operations like this every let's say a minute.

SergeyTeplyakov: but one very important point here. The method should be made async and we should do 'return await context.PerformOperationAsync()'

to avoid early release of cancellation token source.

---
In reply to: [342741588](https://github.com/microsoft/BuildXL/pull/1141#discussion_r342741588) [](ancestors = 342741588)

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventStore.cs
```diff
@@ -125,7 +125,21 @@ protected async Task DispatchAsync(OperationContext context, ContentLocationEven
                             context,
                             addContent.Sender,
                             addContent.ContentHashes.SelectList((hash, index) => new ShortHashWithSize(hash, addContent.ContentSizes[index])),
-                            eventData.Reconciling);
+                            eventData.Reconciling,
+                            updateLastAccessTime: true);
+                    }
+
+                    break;
+                case AddContentLocationWithoutTouchingEventData addContentWithoutTouching:
```
ljcollins25: >case AddContentLocationWithoutTouchingEventData addContentWithoutTouching: [](start = 16, length = 74)

Can we combine this case statement with the case statement above and just have a ternary expression for computing updateLastAccessTime?

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
```diff
@@ -196,6 +198,15 @@ enum CopyCompression {
...
```
dcwuser: Should be possible to re-use CopyFileResponse type. Just re-name to FileTransferChunk or similiar.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcContentServer.cs
```diff
@@ -344,6 +360,94 @@ private async Task<RequestCopyFileResponse> RequestCopyFileAsync(RequestCopyFile
...
+        public async Task HandlePushFileAsync(IAsyncStreamReader<PushFileRequest> requestStream, IServerStreamWriter<PushFileResponse> responseStream, ServerCallContext callContext)
+        {
+            OperationStarted();
+
+            var startTime = DateTime.UtcNow;
+
+            byte[] hashBytes = null;
+            var hashType = HashType.Unknown;
+            Guid traceId = default;
+            foreach (var header in callContext.RequestHeaders)
```
dcwuser: Rather than using headers directly in server and client, I recommend creating a type (e.g. PushRequest) that stores then, and knows how to construct itself from headers via a static FromHeaders method and how to convert itself to headers via an instance ToHeaders method. Then use the type in both client and server. This keeps all the serialization/deserialization logic (e.g. header names) in one place.

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -257,6 +258,79 @@ public async Task<BoolResult> RequestCopyFileAsync(OperationContext context, Con
...
+                    { "hash-bin", hash.ToHashByteArray() }, // -bin suffix required to send bytes directly
+                    { "hash_type", hash.HashType.ToString() },
+                    { "trace_id", context.TracingContext.Id.ToString() }
+                };
+
+                var call = _client.PushFile(headers, cancellationToken: context.Token);
+                var requestStream = call.RequestStream;
+
+                var responseHeaders = await call.ResponseHeadersAsync;
+                foreach (var header in responseHeaders)
```
dcwuser: Same comment here. Create a type to encapsulate this rather than manipulating headers directly.

Public/Src/Cache/ContentStore/Distributed/FileSystem/IFileCopier.cs
```diff
@@ -60,17 +60,19 @@ public interface ITraceableAbsolutePathFileCopier : IAbsolutePathFileCopier, ITr
...
         /// </summary>
-        /// <param name="context">The context of the operation</param>
-        /// <param name="hash">The hash of the file to be copied.</param>
-        /// <param name="targetMachine">The machine that should copy the file</param>
         Task<BoolResult> RequestCopyFileAsync(OperationContext context, ContentHash hash, MachineLocation targetMachine);
+
+        /// <summary>
+        /// Pushes content to a target machine.
+        /// </summary>
+        Task<BoolResult> PushFileAsync(OperationContext context, ContentHash hash, Stream source, MachineLocation targetMachine);
```
dcwuser: Cancellable?

JuanCarlosGI: It is. CancellationToken is in OperationContext

---
In reply to: [343825436](https://github.com/microsoft/BuildXL/pull/1141#discussion_r343825436) [](ancestors = 343825436)

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventData.cs
```diff
@@ -310,6 +318,55 @@ public override int GetHashCode()
...
+            ContentSizes = contentSizes;
+        }
 
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            var sb = new StringBuilder();
+            for (int i = 0; i < ContentHashes.Count; i++)
+            {
+                sb.Append($"Hash={ContentHashes[i]}, Size={ContentSizes[i]}");
```
dcwuser: As written, these are going to run together, e.g. "Hash=ABC, Size=1Hash=DEF, Size=2Hash=123, Size=3"

Public/Src/Cache/ContentStore/Distributed/NuCache/EventStreaming/ContentLocationEventData.cs
```diff
@@ -310,6 +318,55 @@ public override int GetHashCode()
...
+    public sealed class AddContentLocationWithoutTouchingEventData : ContentLocationEventData
+    {
+        /// <summary>
+        /// A list of content sizes associated with <see cref="ContentLocationEventData.ContentHashes"/>.
+        /// </summary>
+        public IReadOnlyList<long> ContentSizes { get; }
+
+        /// <nodoc />
+        public AddContentLocationWithoutTouchingEventData(MachineId sender, IReadOnlyList<ShortHashWithSize> addedContent)
+            : this(sender, addedContent.SelectList(c => c.Hash), addedContent.SelectList(c => c.Size))
```
dcwuser: This is probably too much to undertake now, but not a fan of doing all this list allocation just to make calls work. Can we refactor use co-variant list types or iterators?

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreBase.cs
```diff
@@ -86,7 +86,7 @@ protected override Task<BoolResult> ShutdownCoreAsync(OperationContext context)
         }
 
         /// <nodoc />
-        public Task<BoolResult> RegisterLocalLocationAsync(Context context, IReadOnlyList<ContentHashWithSize> contentHashes, CancellationToken cts, UrgencyHint urgencyHint)
+        public Task<BoolResult> RegisterLocalLocationAsync(Context context, IReadOnlyList<ContentHashWithSize> contentHashes, CancellationToken cts, UrgencyHint urgencyHint, bool touch = true)
```
dcwuser: As long as we are changing API, can we drop urgency hint?

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
```diff
@@ -407,6 +422,24 @@ private Task<ProactiveCopyResult> RequestProactiveCopyIfNeededAsync(OperationCon
                 });
         }
 
+        private async Task<BoolResult> RequestOrPushContentAsync(OperationContext context, ContentHash hash, MachineLocation target, bool isInsideRing)
+        {
+            if (Settings.PushProactiveCopies)
+            {
+                var streamResult = await Inner.OpenStreamAsync(context, hash, context.Token);
```
dcwuser: Need using around the stream to ensure disposal.

JuanCarlosGI: Not sure about this. Wrapping on "using" here would mean that we would dispose it as soon as we return it, wouldn't it?

---
In reply to: [343831893](https://github.com/microsoft/BuildXL/pull/1141#discussion_r343831893) [](ancestors = 343831893)

Public/Src/Cache/ContentStore/Distributed/Stores/DistributedContentCopier.cs
```diff
@@ -259,13 +257,42 @@ public Task<BoolResult> RequestCopyFileAsync(OperationContext context, ContentHa
...
+                    Tracer,
+                    operation: () => _copyRequester.PushFileAsync(innerContext, hash, source, targetLocation),
+                    traceOperationStarted: false,
+                    extraEndMessage: result =>
+                        $"ContentHash={hash.ToShortString()} " +
+                        $"TargetLocation=[{targetLocation}] " +
+                        $"InsideRing={isInsideRing} " +
+                        $"IOGate.OccupiedCount={_settings.MaxConcurrentProactiveCopyOperations - _proactiveCopyIoGate.CurrentCount} " +
+                        $"IOGate.Wait={ts.TotalMilliseconds}ms." +
+                        $"Timeout={_timeoutForProactiveCopies}" +
```
dcwuser: Missing space.

Public/Src/Cache/ContentStore/Distributed/Sessions/DistributedContentSession.cs
dcwuser: Sorry, you are right. I didn't see that this lambda returned a stream. So change to: please be sure to analyze when this stream will be disposed, including error & cancellation paths. I will look too, but your knowledge of this code is deeper.

---
In reply to: [343832763](https://github.com/microsoft/BuildXL/pull/1141#discussion_r343832763) [](ancestors = 343832763,343831893)

Public/Src/Cache/ContentStore/Library/Service/Grpc/GrpcCopyClient.cs
```diff
@@ -257,6 +258,83 @@ public async Task<BoolResult> RequestCopyFileAsync(OperationContext context, Con
...
+
+                var stream = await source();
+
+                if (stream == null)
+                {
+                    await requestStream.CompleteAsync();
+                    return new BoolResult("Failed to retrieve source stream.");
+                }
+
+                using (stream)
```
dcwuser: Okay, here is using. This looks right to me, including for errors and cancellations.

Public/Src/Cache/ContentStore/Grpc/ContentStore.proto
JuanCarlosGI: In here I'm not using indexes. Besides, I think there is value in being able to modify both independently, particularly since there is virtually no cost in having these as separate response type.

---
In reply to: [343805723](https://github.com/microsoft/BuildXL/pull/1141#discussion_r343805723) [](ancestors = 343805723)

PR #1143, opened 2019-11-01
Prevent crashes caused by FP store logging

olkononenko:


Public/Src/Engine/Scheduler/Tracing/MultiExecutionLogTarget.cs
```diff
@@ -87,7 +90,15 @@ protected override void ReportUnhandledEvent<TEventData>(TEventData data)
         {
             foreach (var target in m_targets)
             {
-                data.Metadata.LogToTarget(data, target);
+                try
+                {
```
mpysson: Can we make this more targeted to just the cachemiss implementation so we don't potentially sweep other errors under the rug?

Public/Src/Engine/Scheduler/Tracing/MultiExecutionLogTarget.cs
```diff
@@ -87,7 +90,15 @@ protected override void ReportUnhandledEvent<TEventData>(TEventData data)
...
-                data.Metadata.LogToTarget(data, target);
+                try
+                {
+                    data.Metadata.LogToTarget(data, target);
+                }
+                catch (Exception e)
+                {
+                    // since the exception only affects our logging, log it as a warning
+                    Events.Log.WarningEvent($"An exception occurred while logging '{data.GetType()}' to target '{target.GetType()}':{Environment.NewLine}{e.ToStringDemystified()}");
+                }
```
mpysson: There's a special event, UnexpectedCondition(), that you can use for this sort of thing so we get the exception details to go to telemetry without it crashing the application. Then we can watch for that to see how often this happens at a broad scale

PR #1147, opened 2019-11-01
[Execution Analyzer] Don't crash on invalid XLG format

almiliMSFT:


Public/Src/Tools/Execution.Analyzer/Args.cs
```diff
@@ -382,72 +393,82 @@ public int Analyze()
                 return 0;
             }
 
-            m_analyzer.Prepare();
-            bool dataIsComplete = true;
-            if (m_analysisInput.ExecutionLogPath != null)
+            try
```
almiliMSFT: just added try/catch/finally, nothing else

PR #1151, opened 2019-11-04
[macOS] Memory pressure cancels pips only

Kr1sso:
Use vm pressure events only to decide if pips should be canceled instead of overriding resource availability.

### DIFF
--- a/Public/Src/Engine/Scheduler/Scheduler.cs
+++ b/Public/Src/Engine/Scheduler/Scheduler.cs
@@ -2172,30 +2172,6 @@ namespace BuildXL.Scheduler
                 }
             }
 
-#if PLATFORM_OSX
-            if (!resourceAvailable)
-            {
-                Memory.PressureLevel pressureLevel = Memory.PressureLevel.Normal;
-                var result = Memory.GetMemoryPressureLevel(ref pressureLevel) == Dispatch.MACOS_INTEROP_SUCCESS;
-
-                if (result)
-                {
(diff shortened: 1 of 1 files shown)


Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -2218,7 +2194,30 @@ private void UpdateResourceAvailability(PerformanceCollector.MachinePerfInfo per
...
+                    Logger.Log.UnableToGetMemoryPressureLevel(
+                            m_executePhaseLoggingContext,
+                            availableRam: perfInfo.AvailableRamMb.Value,
+                            minimumAvailableRam: m_configuration.Schedule.MinimumTotalAvailableRamMb,
+                            ramUtilization: perfInfo.RamUsagePercentage.Value,
+                            maximumRamUtilization: m_configuration.Schedule.MaximumRamUtilizationPercentage);
+                }
+
+                if (!m_scheduleConfiguration.DisableProcessRetryOnResourceExhaustion && startCancellingPips)
+#else
```
almiliMSFT: It looks kind of ugly to have these dangling `if`s inside of if/def sections.

Instead, I'd write it like
```csharp
bool startCancellingPips = !resourceAvailable;
#if PLATFORM_OSX
    // special OSX logic for computing whether to start cancelling pips
    startCancellingPips = ...;
#endif

if (!m_scheduleConfiguration.DisableProcessRetryOnResourceExhaustion && startCancellingPips)
{
...
}
```

PR #1152, opened 2019-11-04
Disable System.GC.* features for Unix builds

Kr1sso:
Disable System.GC.Server and System.GC.RetainVM in Unix builds to avoid big BuildXL process footprints for involved builds (~12.5% overall footprint reduction).

### DIFF
--- a/Public/Sdk/Public/Managed/runtimeConfigFiles.dsc
+++ b/Public/Sdk/Public/Managed/runtimeConfigFiles.dsc
@@ -211,7 +211,8 @@ namespace RuntimeConfigFiles {
             }
         } : {};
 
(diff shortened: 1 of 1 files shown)


Public/Sdk/Public/Managed/runtimeConfigFiles.dsc
```diff
@@ -211,11 +211,11 @@ namespace RuntimeConfigFiles {
             }
         } : {};
 
-        // when not using Server GC, in large builds the front end is likely to get completely bogged
+        // when not using Server GC, in large builds the front end is likely to get completely bogged, on Unix we are currently disabling this
```
almiliMSFT: It's clear from below that on non-"win" we are disabling this; instead, say here why

PR #1154, opened 2019-11-04
Add info to contract assertion when retrieving the process id list

semihokur:


Public/Src/Engine/Processes/JobObject.cs
```diff
@@ -321,7 +321,7 @@ private bool TryGetProcessIds(uint length, out uint[] processIds, out uint requi
                     return false;
                 }
 
-                Contract.Assume(bytesWritten <= bufferSizeInBytes);
+                Contract.Assume(bytesWritten <= bufferSizeInBytes, $"More bytes written than the buffer size: {bytesWritten} > {bufferSizeInBytes}.");
```
narasamdya: Can you put it inside an if-clause to avoid creating interpolated string?

```
if (bytesWritten > bufferSizeInBytes) 
{
    Contract.Assume(false, ...);
}
```

PR #1155, opened 2019-11-04
Add a ceiling for the number of idle HashAlgorithm instances held by ContentHasher

mpysson:
Right now we accumulate a lot of HashAlgorithm instances. Each instance of the VsoHashAlgorithm takes 2MB so this accumulation adds up. The maximum number of concurrently used instances ends up being the number that gets pooled.

This change sets a cap for the maximum number of reserve instances that can exist idle in the queue. The total number of instances participating in the pool can still be larger than that cap since instances are not in the queue while in use. So the number in the pool can still spike up quite a bit higher than the limit while there are many instances in use. This just limits the number it idle instances above what is current in use.

### DIFF
--- a/Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
+++ b/Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
@@ -25,7 +25,7 @@ namespace BuildXL.Cache.ContentStore.Hashing
         ///     Object pool that holds all instantiated hash algorithms for each hash type.
         /// </summary>
         /// <remarks>
-        ///     Cap the number of idel reserve instances in the pool so as to not unnecessarily hold large amounts of memory
+        ///     Cap the number of idle reserve instances in the pool so as to not unnecessarily hold large amounts of memory
         /// </remarks>
         private readonly Pool<HashAlgorithm> _algorithmsPool = new Pool<HashAlgorithm>(() => new T(), maxReserveInstances: 10);
 
--- a/Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
+++ b/Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
@@ -24,7 +24,10 @@ namespace BuildXL.Cache.ContentStore.Hashing
         /// <summary>
         ///     Object pool that holds all instantiated hash algorithms for each hash type.
         /// </summary>
-        private readonly Pool<HashAlgorithm> _algorithmsPool = new Pool<HashAlgorithm>(() => new T());
+        /// <remarks>
+        ///     Cap the number of idel reserve instances in the pool so as to not unnecessarily hold large amounts of memory
+        /// </remarks>
+        private readonly Pool<HashAlgorithm> _algorithmsPool = new Pool<HashAlgorithm>(() => new T(), maxReserveInstances: 10);
 
         private readonly ByteArrayPool _bufferPool = new ByteArrayPool(FileSystemConstants.FileIOBufferSize);
 
--- a/Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
+++ b/Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
@@ -27,7 +27,7 @@ namespace BuildXL.Cache.ContentStore.Hashing
         /// <remarks>
         ///     Cap the number of idle reserve instances in the pool so as to not unnecessarily hold large amounts of memory
         /// </remarks>
-        private readonly Pool<HashAlgorithm> _algorithmsPool = new Pool<HashAlgorithm>(() => new T(), maxReserveInstances: 10);
+        private readonly Pool<HashAlgorithm> _algorithmsPool = new Pool<HashAlgorithm>(() => new T(), maxReserveInstances: HashInfoLookup.ContentHasherIdlePoolSize);
 
         private readonly ByteArrayPool _bufferPool = new ByteArrayPool(FileSystemConstants.FileIOBufferSize);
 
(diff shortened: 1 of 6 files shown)


mpysson:
Capping the number of idle VSOHashers to 10 yields a 19% decrease in avg working set of bxl.exe for our self-host build (comparing the No ServerGC numbers)

**Control:**
Run1
Execute.DurationMs=167301
Execute.ProcessAverageThreadCount=290
Execute.ProcessMaximumPrivateMB=3420
Execute.ProcessMaximumWorkingSetMB=2384

Run2
Execute.DurationMs=168384
Execute.ProcessAverageThreadCount=295
Execute.ProcessMaximumPrivateMB=1894
Execute.ProcessMaximumWorkingSetMB=929

**Control No Serve****GC:**
Run1
Execute.DurationMs=168464
Execute.ProcessAverageThreadCount=182
Execute.ProcessMaximumPrivateMB=1445
Execute.ProcessMaximumWorkingSetMB=714
Avg working set 655MB


Run2
Execute.DurationMs=170848
Execute.ProcessAverageThreadCount=186
Execute.ProcessMaximumPrivateMB=1368
Execute.ProcessMaximumWorkingSetMB=707
Avg working set 650MB

**No ServerGC. Max ContentHasher._algorithmsPool idle instances capped to 10**
Run1
Execute.DurationMs=168163
Execute.ProcessAverageThreadCount=186
Execute.ProcessMaximumPrivateMB=724
Execute.ProcessMaximumWorkingSetMB=624
Avg working set 585MB

Run2
Execute.DurationMs=171012
Execute.ProcessAverageThreadCount=198
Execute.ProcessMaximumPrivateMB=605
Execute.ProcessMaximumWorkingSetMB=501
Avg working set 471 Avg MB

Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
```diff
@@ -24,7 +24,10 @@ public class ContentHasher<T> : IContentHasher
         /// <summary>
         ///     Object pool that holds all instantiated hash algorithms for each hash type.
         /// </summary>
-        private readonly Pool<HashAlgorithm> _algorithmsPool = new Pool<HashAlgorithm>(() => new T());
+        /// <remarks>
+        ///     Cap the number of idel reserve instances in the pool so as to not unnecessarily hold large amounts of memory
```
almiliMSFT:
```suggestion
        ///     Cap the number of idle reserve instances in the pool so as to not unnecessarily hold large amounts of memory
```

Public/Src/Cache/ContentStore/Hashing/Pool.cs
```diff
@@ -86,8 +88,11 @@ public virtual PoolHandle Get()
 
         private void Return(T item)
         {
-            _reset?.Invoke(item);
-            _queue.Enqueue(item);
+            if (Size < _maxReserveInstances)
```
almiliMSFT: don't you still need to reset the item even if you decide not to add it to the pool?

mpysson: Fair point. Most of our usages just used the _reset as a mechanism to clean the state and make the object valid for reuse. If we aren't reusing it that is unnecessary work. But one might also want to use the _reset as a mechanism for running logic with other side effects. So I suppose it would be safer to still run it even if we aren't going to be holding a reference to the object. I'll update that.

Public/Src/Cache/ContentStore/Hashing/ContentHasher.cs
```diff
@@ -24,7 +24,10 @@ public class ContentHasher<T> : IContentHasher
         /// <summary>
         ///     Object pool that holds all instantiated hash algorithms for each hash type.
         /// </summary>
-        private readonly Pool<HashAlgorithm> _algorithmsPool = new Pool<HashAlgorithm>(() => new T());
+        /// <remarks>
+        ///     Cap the number of idle reserve instances in the pool so as to not unnecessarily hold large amounts of memory
+        /// </remarks>
+        private readonly Pool<HashAlgorithm> _algorithmsPool = new Pool<HashAlgorithm>(() => new T(), maxReserveInstances: 10);
```
SergeyTeplyakov: Such a low number could have a perf impact in the cache codebase.

CASaaS will definitely have more then 10 instances, and this logic can unpredictably increase churn.

I definitely agree with the premise of this change, but I'm not sure that this should be done unconditionally.

An alternative solution may be to use 10 from bxl but still use the default value from casaas.

mpysson:
Some more perf stats from CB. With server GC _enabled_ this saved about 1GB of worker memory out of 16 - 20GB depending on how you count

Control clean 700k Process pip build:

1. Worker Max PrivateMB (avg across workers): 21199 MB
1. Worker Max Working Set (avg across workers): 19401 MB
1. Average Bxl process working set across lifetime of build (avg across workers): 16.8 GB

Modified:
1. 20403 MB
1. 18477 MB
1. 15.85 GB

I wasn't able to notice any regression in build time due to this change

SergeyTeplyakov:
This is an absolutely awesome result, but my main concern is not about bxl in CloudBuild, but regarding CASaaS.
CASaaS uses multiple simultaneous copies and in steady state it uses way more then 10 hashers at a time. So the concern is that CASaaS will aggresively remove the hashers and will recreate them pretty soon.

That's why I suggested to make the cap configurable and pass 10 from bxl but pass a higher number for CASaaS.

---
In reply to: [550405914](https://github.com/microsoft/BuildXL/pull/1155#issuecomment-550405914) [](ancestors = 550405914)

PR #1157, opened 2019-11-04
Don't process std out/err of canceled pips

almiliMSFT:


Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -1521,7 +1521,7 @@ private void LogCommandLineTooLong(SandboxedProcessInfo info)
                 }
             }
 
-            bool shouldPersistStandardError = errorOrWarnings || m_pip.StandardError.IsValid;
+            bool shouldPersistStandardError = !canceled && (errorOrWarnings || m_pip.StandardError.IsValid);
```
mpysson: >!canceled [](start = 46, length = 9)

Do we differentiate resource exhaustion cancelled vs. timeout cancelled? I think we'd only want to mask standard output stream logging for the first.

almiliMSFT: From line 1340:
```csharp
bool canceled = result.Killed && cancellationToken.IsCancellationRequested;
```
When a pip times out, `cancellationToken.IsCancellationRequested` is false; it is true only when the pip is explicitly canceled, either due to resource exhaustion or Ctrl+C (and in neither case I don't think we need to save std out/err)

Kr1sso: Agreed, we also make sure to set result.killed to `true` in both of those cases.

mpysson: Cool. As long as we don't skip processing the output streams on timeout I think the change is good. Thanks for checking.

PR #1160, opened 2019-11-04
Implement additional retry telemetry and custom pip property telemetry

breidikl:
Track additional pip retry information, including pips that succeed after failure and those failing after all retries.  

Also allow logging a custom pip property that gets aggregated for the build and reported vi a processpattern event (count of pip properties hit, and csv if impacted pips).  

Fixes [AB#1612096](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1612096)

### DIFF
--- a/Public/Src/Engine/Scheduler/PipRetryInfo.cs
+++ b/Public/Src/Engine/Scheduler/PipRetryInfo.cs
@@ -1,5 +1,3 @@
-using System.Collections.Concurrent;
-using System.Collections.Generic;
 using BuildXL.Pips;
 using Logger = BuildXL.Scheduler.Tracing.Logger;
 using BuildXL.Utilities.Collections;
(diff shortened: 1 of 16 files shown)


Public/Src/Engine/Scheduler/PipRetryInfo.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Collections.Generic;
+using BuildXL.Pips;
+using Logger = BuildXL.Scheduler.Tracing.Logger;
+using BuildXL.Utilities.Collections;
+using BuildXL.Utilities.Instrumentation.Common;
+using BuildXL.Utilities.Tracing;
+
+namespace BuildXL.Scheduler
+{
+    internal class PipRetryInfo
```
narasamdya: This is no only for pip retry, but also for pip properties in general. That's why I suggested that we need two separate classes.

breidikl: Split out into a PipPropertyInfo class.

PR #1161, opened 2019-11-05
Fix race condition in QuotaKeeper

SergeyTeplyakov:


Public/Src/Cache/ContentStore/Library/Stores/QuotaManagement/QuotaKeeper.cs
```diff
@@ -164,10 +164,11 @@ protected override async Task<BoolResult> ShutdownCoreAsync(OperationContext con
                 await _processReserveRequestsTask;
             }
 
-            if (_purgeTask != null)
+            var purgeTask = _purgeTask;
+            if (purgeTask != null)
             {
                 context.TraceDebug($"{_tracer.Name}: waiting for purge task.");
-                return await _purgeTask;
+                return await purgeTask;
```
jbayardo: nit: wouldn't it be better for _purgeTask to default to `Task.CompletedTask`? would prevent any further subtle issues like this one.

SergeyTeplyakov: That is a good idea. I'll change it to be non-null.

jbayardo:
New changes LGTM

PR #1162, opened 2019-11-05
Enable non-nullable reference types in QuotaKeeper

SergeyTeplyakov:


dcwuser:
This change appears to change the type of a field that can currently semantically be null with a nullable wrapper around a non-nullable reference type. There doesn't seem to be much point in that. Should be be concentrating on replacing fields that can currently not semantically be null with a straight-up non-nullable type?

PR #1163, opened 2019-11-05
Fix race condition in ContentLocationDatabase

jbayardo:


Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -659,43 +659,25 @@ internal bool ForceCacheFlush(OperationContext context, ContentLocationDatabaseC
...
-                task.GetAwaiter().GetResult();
-            }
-            else
-            {
-                task.FireAndForget(context);
+                _flushTask.GetAwaiter().GetResult();
             }
 
-            return true;
+            return renewed || blocking;
```
SergeyTeplyakov: Right now the logic is interesting.
- If memory cache is off the result is off even when blocking is true
- if blocking is true then the result is true regardless whether we flushed or not
- if blocking is false, then the result indicates whether we actually flushed or not.

This kind of complicated imo:)

jbayardo: This is exactly what it's supposed to be. The result is supposed to be whether the call effectively performed a flush, modulo blocking flushes (in which case we consider it the same as we are waiting for it to finish). The return value is only used for testing

---
In reply to: [342705873](https://github.com/microsoft/BuildXL/pull/1163#discussion_r342705873) [](ancestors = 342705873)

SergeyTeplyakov: Given that the result is used in tests only and the false was never checked, I doubt that the current logic is good.

For instance, you can change this to return true only when it is blocked and it was flushed and the system will work as it is right now.

---
In reply to: [342730347](https://github.com/microsoft/BuildXL/pull/1163#discussion_r342730347) [](ancestors = 342730347,342705873)

SergeyTeplyakov: I actually think that the result should be true when the flush happened synchronously and when actually happened.

---
In reply to: [342731585](https://github.com/microsoft/BuildXL/pull/1163#discussion_r342731585) [](ancestors = 342731585,342730347,342705873)

PR #1170, opened 2019-11-06
Fix GRPC distribution of source affected inputs issue

norahuang:
[AB#1640977](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1640977)


Private/DistributedIntegrationTest/config.dsc
```diff
@@ -21,5 +21,12 @@ config({
             isWritable: false,
             isReadable: true,
         },
+        {
```
narasamdya: >{ [](start = 8, length = 1)

Why do we need a new mount here?

Private/DistributedIntegrationTest/BuildDistributed.cmd
```diff
@@ -30,10 +30,21 @@ set SMDB.CACHE_TEMPLATE_PATH=%TEST_SOLUTION_ROOT%\Out\SharedCache
...
 set BUILDXL_MASTER_ARGS=/maxProc:2 %BUILDXL_MASTER_ARGS%
 set BUILDXL_WORKER_ARGS=/maxProc:6 %BUILDXL_WORKER_ARGS%
+
+REM The following variables are for ChangeAffectedInputTest
+set SOURCE_FILE=%TEST_SOLUTION_ROOT%\Src\ChangeAffectedInputTest\inputfile.txt
+set CHANGEAFFECTEDINPUTTEST_DIR=%TEST_SOLUTION_ROOT%\Src\ChangeAffectedInputTest
+set OUTPUT_DIR_NAME=output
+set OUTPUT_FILE_NAME=outputfile.txt
+set OUTPUT_FILE=%CHANGEAFFECTEDINPUTTEST_DIR%\%OUTPUT_DIR_NAME%\%OUTPUT_FILE_NAME%
+set EXPECTED_WRITTEN_FILE=%TEST_SOLUTION_ROOT%\Src\ChangeAffectedInputTest\expectedChangeAffectedInputListWrittenFile.txt
```
narasamdya: Why do we need set them as environment variables?

norahuang: I need them here and in the test case.

Private/DistributedIntegrationTest/config.dsc
norahuang: My test case has a copyfile pip writing to this directory. If I don't mount it, it claims file wirte in unmounted directory and untrack it.

PR #1175, opened 2019-11-07
Missing accesses on breakaway processes can be compensated by talking to detours without performing IO

smera:
This PR adds an API available to internal Bxl tools which allows to report arbitrary file accesses to detours without actually performing any IO. This enables compensating for missing accesses when processes are allowed to breakaway from the sandbox.
The API is for now intended for trusted internal Bxl tools only and the actual mechanism used to talk to detours is hidden behind it. 
An upcoming combined engine related PR will add a tool that will actually use this API to properly handle the csc compiler service.


Public/Src/Engine/Processes/FileAccessReportLine.cs
```diff
@@ -0,0 +1,292 @@
...
+using BuildXL.Native.IO;
+using static BuildXL.Utilities.FormattableStringEx;
+using JetBrains.Annotations;
+
+namespace BuildXL.Processes
+{
+    /// <summary>
+    /// Low-level parser for file access report lines
+    /// </summary>
+    internal static class FileAccessReportLine
```
smera: Moved this out from SandboxedProcessReports to its own file. TryParse is left untouched. The only addition is GetReportLineForAugmentedFileAccess

Public/Src/Engine/Processes/AugmentedManifestReporter.cs
```diff
@@ -0,0 +1,204 @@
...
+    /// processes may need to compensate for some unobserved accesses.
+    /// In order for processes to use this reporter, they must run in the context of a pip which has a non-empty set of processes that 
+    /// are allowed to breakaway. Otherwise, attempting to report an access will fail.
+    /// Observe this reporter is only intended for internal (to BuildXL) tools at this point, since it requires tools to link against
+    /// this particular BuildXL library. In the future we may consider a more general approach for arbitrary (trusted) tools to perform these actions
+    /// </remarks>
+    public sealed class AugmentedManifestReporter
+    {
+        [CanBeNull]
+        private readonly SafeFileHandle m_detoursReportHandle = null;
```
almiliMSFT: nit: It might be a little cleaner exclusively initialize readonly variables in the constructor (and there assign `null` when appropriate)

Public/Src/Engine/Processes/AugmentedManifestReporter.cs
```diff
@@ -0,0 +1,204 @@
...
+        private readonly UnicodeEncoding m_encoding;
+
+        /// <nodoc/>
+        public static AugmentedManifestReporter Instance = new AugmentedManifestReporter();
+
+        private AugmentedManifestReporter()
+        {
+            m_encoding = new UnicodeEncoding(bigEndian: false, byteOrderMark: false);
+
+            // Keep variable name in sync with DetoursServices on the C++ side
```
almiliMSFT: In other places we use the `CODESYNC` moniker for these kind of comments, which I think does a better job of catching the reader's eye.

Public/Src/Engine/Processes/AugmentedManifestReporter.cs
```diff
@@ -0,0 +1,204 @@
...
+            // when reporting an access
+            if (string.IsNullOrEmpty(handleAsString))
+            {
+                return;
+            }
+
+            var success = int.TryParse(handleAsString, NumberStyles.Integer, CultureInfo.InvariantCulture, out int handlePtr);
+            Contract.Assert(success);
+
+            // We shouldn't try to close this handle. Detours takes care of that.
```
almiliMSFT: maybe move this comment up where this field is declared (and make it a doc comment)

Public/Src/Engine/Processes/SandboxedProcess.cs
```diff
@@ -344,15 +344,14 @@ public void Start()
...
-
                 try
                 {
                     Pipes.CreateInheritablePipe(
                         Pipes.PipeInheritance.InheritWrite,
                         Pipes.PipeFlags.ReadSideAsync,
                         readHandle: out reportHandle,
                         writeHandle: out childHandle);
-
+
```
almiliMSFT: nit: trailing spaces

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -522,7 +529,77 @@ private void AddLookupEntryForProcessExit(uint processId, ReportedProcess report
...
+                out desiredAccess, 
+                out shareMode, 
+                out creationDisposition, 
+                out flagsAndAttributes, 
+                out manifestPath, 
+                out path, 
+                out enumeratePattern, 
+                out processArgs, 
+                out errorMessage);
+
```
almiliMSFT: don't you have to check if `result` is `true` before proceeding?

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -690,11 +767,14 @@ private bool FileAccessReportLineReceived<T>(ref T data, FileAccessReportProvide
                 return true;
             }
 
-            FileAccessStatusMethod method = FileAccessStatusMethod.PolicyBased;
+            // If this is an augmented file access, the method was not based on policy, but a trusted tool reported the access
+            FileAccessStatusMethod method = isAnAugmentedFileAccess? FileAccessStatusMethod.TrustedTool : FileAccessStatusMethod.PolicyBased;
```
almiliMSFT:
```suggestion
            FileAccessStatusMethod method = isAnAugmentedFileAccess ? FileAccessStatusMethod.TrustedTool : FileAccessStatusMethod.PolicyBased;
```

Public/Src/Engine/Processes/AugmentedManifestReporter.cs
smera: Will do

---
In reply to: [343769663](https://github.com/microsoft/BuildXL/pull/1175#discussion_r343769663) [](ancestors = 343769663)

Public/Src/Engine/Processes/SandboxedProcessReports.cs
smera: I could, but It's not really needed, since here I'm just forwarding the result of parsing to the caller. The post-processing that you see after FileAccessReportLine.TryParse(...) is only acting when the returned path can be interpreted as an absolute path. That always make sense even if result is false.

---
In reply to: [343778501](https://github.com/microsoft/BuildXL/pull/1175#discussion_r343778501) [](ancestors = 343778501)

narasamdya:
Newton and I discuss a similar issue some time ago. It's about a tool that runs a Detours/Tracer inside. We cannot Detour that tool. But, we can create a wrapper that spits a file containing what the tool is accessing. `SandboxedProcessPipExecutor` then simply reads that special file and replies/reports it.

Public/Src/Engine/Processes/AugmentedManifestReporter.cs
```diff
@@ -0,0 +1,206 @@
...
+    /// are allowed to breakaway. Otherwise, attempting to report an access will fail.
+    /// Observe this reporter is only intended for internal (to BuildXL) tools at this point, since it requires tools to link against
+    /// this particular BuildXL library. In the future we may consider a more general approach for arbitrary (trusted) tools to perform these actions
+    /// </remarks>
+    public sealed class AugmentedManifestReporter
+    {
+        /// <summary>
+        /// We shouldn't try to close this handle. Detours takes care of that.
+        /// </summary>
+        [CanBeNull] private readonly SafeFileHandle m_detoursReportHandle;
```
narasamdya: >  [](start = 19, length = 1)

nit: Should there be a new line after `[CanBeNull]`?

Public/Src/Engine/Processes/SandboxedProcessReports.cs
```diff
@@ -522,7 +529,77 @@ private void AddLookupEntryForProcessExit(uint processId, ReportedProcess report
...
+
+            // Augmented file accesses never have the manifest path set, since there is no easy access to the manifest for 
+            // processes to use
+            // Let's recreate the manifest path based on the current path and manifest
+            if (string.IsNullOrEmpty(path) || !AbsolutePath.TryCreate(m_pathTable, path, out var absolutePath))
+            {
+                return result;
+            }
+            
+            var success = m_manifest.TryFindManifestPathFor(absolutePath, out AbsolutePath computedManifestPath, out FileAccessPolicy policy);
```
narasamdya: >m_manifest.TryFindManifestPathFor [](start = 26, length = 33)

This unfortunately won't work on VM because the manifest is in the form of array of bytes instead of a tree. 
Either
- Rehydrate the tree
- Have the same path. Please add comment for the reason. Better, add a unit test. There's an integration unit test for this: `Public\Src\Engine\UnitTests\Scheduler.IntegrationTest\ExternalToolExecutionTests.cs`

Public/Src/Engine/Processes/SandboxedProcessReports.cs
narasamdya: nit: I don't like Lance's ternary-code style.
Either this
```
FileAccessStatusMethod method = isAnAugmentedFileAccess ?  FileAccessStatusMethod.TrustedTool  : FileAccessStatusMethod.PolicyBased;
```
or
```
FileAccessStatusMethod method = isAnAugmentedFileAccess 
    ?  FileAccessStatusMethod.TrustedTool  
    : FileAccessStatusMethod.PolicyBased;
```

---
In reply to: [343779106](https://github.com/microsoft/BuildXL/pull/1175#discussion_r343779106) [](ancestors = 343779106)

Public/Src/Engine/Processes/SandboxedProcessReports.cs
smera: Went with the first option. Will add a test.

---
In reply to: [343856730](https://github.com/microsoft/BuildXL/pull/1175#discussion_r343856730) [](ancestors = 343856730)

PR #1179, opened 2019-11-07
Stop logger from throwing when receiving template strings without arguments

jbayardo:
Doing this:

``_logger.Error("{0}")``

Would throw because we'd attempt to `string.Format` the thing. When part of what you're printing is uncontrolled (i.e. an exception from a different subsystem), calls like this:

``_logger.Error($"{exception}")``

Will throw if the string representation of the exception contains any `{}` as substring. This PR stops that from happening, since the latter case is much more common than the former.

### DIFF
--- a/Public/Src/Cache/ContentStore/Library/Logging/Logger.cs
+++ b/Public/Src/Cache/ContentStore/Library/Logging/Logger.cs
@@ -152,7 +152,12 @@ namespace BuildXL.Cache.ContentStore.Logging
         /// <inheritdoc />
         public void Fatal(Exception exception, string messageFormat, params object[] messageArgs)
         {
-            var messageIn = string.Format(CultureInfo.CurrentCulture, messageFormat, messageArgs);
+            var messageIn = messageFormat;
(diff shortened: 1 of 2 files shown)


Public/Src/Cache/ContentStore/Library/Logging/Logger.cs
```diff
@@ -221,7 +231,12 @@ public void Log(Severity severity, string message)
         /// <inheritdoc />
         public void LogFormat(Severity severity, string messageFormat, params object[] messageArgs)
         {
-            LogString(severity, string.Format(CultureInfo.InvariantCulture, messageFormat, messageArgs));
+            if (messageArgs != null && messageArgs.Length > 0)
```
SergeyTeplyakov: The code is duplicated 3 times. I think it makes sense to extract a helper.

Like FormatStringIfNeeded.

jbayardo: Turns out: if I do this, our tests start timing out with resource usage exceeded. I think it should be left as-is

PR #1181, opened 2019-11-07
Keep track of commit memory as a resource

semihokur:
As the linkers pip occasionally fail due to unavailable commit memory in CloudBuild, we started keeping track of commit memory as another resource. For MacOS, as we do not keep track of swapfile usage, I set CommitTotal to a very high number; so that there is no throttling based on commit memory in MacOS.

### DIFF
--- a/Public/Src/Engine/UnitTests/Scheduler/PipRunningTimeTableTests.cs
+++ b/Public/Src/Engine/UnitTests/Scheduler/PipRunningTimeTableTests.cs
@@ -27,7 +27,9 @@ namespace Test.BuildXL.Scheduler
                         };
             var spans = new[] { TimeSpan.Zero, TimeSpan.FromSeconds(1), TimeSpan.MaxValue };
             var ints = new[] {0, int.MaxValue};
-            var ulongs = new ulong[] {0, 1, ulong.MaxValue};
+
+            // Let's say that the highest possible memory usage is currently 1TB.
+            var ulongs = new ulong[] {0, 1, (ulong)1024 * 1024 * 1024 * 1024}; 
             var uints = new uint[] {0, 1, uint.MaxValue};
 
             foreach (var executionStart in times)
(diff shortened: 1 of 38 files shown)


mpysson:
>                         var commitTotal = SafeConvert.ToInt32(CommitTotalMB.Latest);

If you're renaming it to CommitUsedMb I suggest you do it all the way down to make it consistent

---
Refers to: Public/Src/Utilities/Utilities/PerformanceCollector.cs:767 in d1aebc1. [](commit_id = d1aebc1a1ff82d505d9fd9f3f38b3a985dc9e2bd, deletion_comment = False)

Public/Src/Engine/UnitTests/Scheduler/PipRunningTimeTableTests.cs
```diff
@@ -27,7 +27,9 @@ public void PipHistoricPerfDataConstructorDoesntCrash()
                         };
             var spans = new[] { TimeSpan.Zero, TimeSpan.FromSeconds(1), TimeSpan.MaxValue };
             var ints = new[] {0, int.MaxValue};
-            var ulongs = new ulong[] {0, 1, ulong.MaxValue};
+
+            // Let's say that the highest possible memory usage is currently 1TB.
```
mpysson: >Let's say that the highest possible memory usage is currently 1TB. [](start = 15, length = 66)

Everyone's going to laugh at us in 15 years :-0

mpysson:
>                                     Logger.Log.CancellingProcessPipExecutionDueToResourceExhaustion(

It seems like this should be expanded to include commit details and which resource caused the cancellation

---
Refers to: Public/Src/Engine/Scheduler/PipExecutor.cs:1543 in d1aebc1. [](commit_id = d1aebc1a1ff82d505d9fd9f3f38b3a985dc9e2bd, deletion_comment = False)

mpysson:
>         public void Serialize(BuildXLWriter writer)

Is there some serialization version number to bump so you don't try to load previously created historic runtime entries?

---
Refers to: Public/Src/Engine/Scheduler/PipHistoricPerfData.cs:84 in d1aebc1. [](commit_id = d1aebc1a1ff82d505d9fd9f3f38b3a985dc9e2bd, deletion_comment = False)

PR #1182, opened 2019-11-08
Release FileAccessMemory after starting the process

almiliMSFT:
Looks better to me, especially when there is a lot of contention for memory
Screen Shot 2019-11-09 at 7 15 29 AM
Averages: 
  - before: 8513
  - after: 8115

### DIFF
--- a/Public/Src/Engine/Processes/FileAccessManifest.cs
+++ b/Public/Src/Engine/Processes/FileAccessManifest.cs
@@ -852,6 +852,32 @@ namespace BuildXL.Processes
         }
 
         /// <summary>
+        /// Explicitly releases most of its memory.
+        /// </summary>
+        internal void Release()
+        {
+            m_normalizedFragments?.Clear();
+            var workList = new Stack<Node>();
+            workList.Push(m_rootNode);
+            while (workList.Any())
+            {
+                var node = workList.Pop();
+                if (node == null)
+                {
+                    continue;
+                }
+                if (node.Children != null)
+                {
+                    foreach (var child in node.Children)
+                    {
+                        workList.Push(child);
+                    }
+                }
+                node.ReleaseChildren();
+            }
+        }
+
+        /// <summary>
         /// Serializes this manifest.
         /// </summary>
         public void Serialize(Stream stream)
@@ -1146,6 +1172,14 @@ namespace BuildXL.Processes
             /// Returns children nodes.
             /// </summary>
             internal IEnumerable<Node> Children => m_children?.Values;
+
+            /// <summary>
+            /// Releases all children nodes, allowing all that memory to be reclaimed by the garbage collector.
+            /// </summary>
+            internal void ReleaseChildren()
+            {
+                m_children?.Clear();
+            }
 
             /// <summary>
             /// The path ID as understood by the owning path table.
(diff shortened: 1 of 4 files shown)


Public/Src/Engine/Processes/SandboxedProcessMac.cs
```diff
@@ -236,6 +242,14 @@ private async Task OnProcessStartedAsync()
                 LogProcessState($"IOException caught while feeding the standard input: {e.ToString()}");
                 await KillAsync();
             }
+            finally
+            {
+                // release the FileAccessManifest memory
+                // NOTE: just by not keeping any references to 'info' should make the FileAccessManifest object 
+                //       unreachable and thus available for garbage collection.  We call Release() here explicitly 
+                //       just to emphasise the importance of reclaiming this memory.
+                info.FileAccessManifest.Release();
```
almiliMSFT: in `SandboxedProcess` (for Windows) there is already code that simply sets its `m_fileAccessManifest` field to null so that it can be garbage collected

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -851,6 +851,32 @@ public ArraySegment<byte> GetPayloadBytes(FileAccessSetup setup, MemoryStream st
             }
         }
 
+        /// <summary>
+        /// Explicitly releases most of its memory.
+        /// </summary>
+        internal void Release()
```
almiliMSFT: No noticeable difference on selfhost, but FAMs are not big there.  I'll check bigger builds later.

Public/Src/Engine/Processes/SandboxedProcessMac.cs
Kr1sso: Does adding an explicit `info = null` hurt?

mpysson: You predicted my question

mpysson:
>         private readonly Dictionary<StringId, NormalizedPathString> m_normalizedFragments = new Dictionary<StringId, NormalizedPathString>();

Do you think its worth using an object pool for these as well to avoid growth allocations?

---
Refers to: Public/Src/Engine/Processes/FileAccessManifest.cs:25 in 30bf45a. [](commit_id = 30bf45a641b3934227d0343585db35011155ada2, deletion_comment = False)

Public/Src/Engine/Processes/FileAccessManifest.cs
```diff
@@ -851,6 +851,32 @@ public ArraySegment<byte> GetPayloadBytes(FileAccessSetup setup, MemoryStream st
...
 
+        /// <summary>
+        /// Explicitly releases most of its memory.
+        /// </summary>
+        internal void Release()
+        {
+            m_normalizedFragments?.Clear();
+            var workList = new Stack<Node>();
+            workList.Push(m_rootNode);
+            while (workList.Any())
```
narasamdya: >workList.Any() [](start = 19, length = 14)

Use `Count`. This will create possibly tons of `Enumerator`, and since it's called from interface, that enumerator gets boxed.

Public/Src/Engine/Processes/FileAccessManifest.cs
almiliMSFT: But it does seem noticeable in bigger builds

Public/Src/Engine/Processes/SandboxedProcessMac.cs
almiliMSFT: I don't think it makes a difference; `info` is a local variable and assigning null to it just before that local variable is going to go out of scope seems pointless to me

Public/Src/Engine/Processes/FileAccessManifest.cs
almiliMSFT: changed

PR #1183, opened 2019-11-08
Enable self-host build on K drive in CB

narasamdya:


Public/Src/Engine/Dll/Engine.cs
```diff
@@ -2452,6 +2459,16 @@ private static bool IsJournalPossiblyAvailableForPath(PathTable pathTable, Absol
...
                     {
+                        try
+                        {
+                            string finalPathStr = FileUtilities.GetFinalPathNameByHandle(handle);
+                            finalPath = AbsolutePath.Create(pathTable, finalPathStr);
+                        }
+                        catch (NativeWin32Exception)
+                        {
+                            finalPath = path;
+                        }
```
mpysson: Is this legit?

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -2409,12 +2409,17 @@ private bool VerifyJournalAvailableForPath(LoggingContext loggingContext, Absolu
         {
             Contract.Requires(path.IsValid);
 
-            var translatedPath = m_translator?.Translate(Context.PathTable, path) ?? path;
-            bool possiblyEnabled = IsJournalPossiblyAvailableForPath(Context.PathTable, translatedPath);
+            // var translatedPath = m_translator?.Translate(Context.PathTable, path) ?? path;
```
mpysson: >// var translatedPath = m_translator?.Translate(Context.PathTable, path) ?? path; [](start = 12, length = 81)

remove

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -2409,12 +2409,17 @@ private bool VerifyJournalAvailableForPath(LoggingContext loggingContext, Absolu
...
             if (!possiblyEnabled)
             {
-                var drive = translatedPath.GetRoot(Context.PathTable).ToString(Context.PathTable).TrimEnd('\\');
-                Logger.Log.JournalRequiredOnVolumeError(loggingContext, drive, GetConfigureJournalCommand(drive));
+                var drive = finalPath.GetRoot(Context.PathTable).ToString(Context.PathTable).TrimEnd('\\');
+                Logger.Log.JournalRequiredOnVolumeError(
+                    loggingContext,
+                    drive,
+                    path.ToString(Context.PathTable),
+                    path != finalPath ? finalPath.ToString(Context.PathTable) : string.Empty,
```
mpysson: >string.Empty, [](start = 80, length = 13)

I wonder if the log will be confusing with this empty. Would it be better to just include the final path even when it's the same as the requested path? Or maybe you could have a canned string like "PathIsFinal"

Public/Src/Engine/Dll/Engine.cs
narasamdya: It is legit because GetFinalPathNameByHandle can throw NativeWin32Exception.

---
In reply to: [344283054](https://github.com/microsoft/BuildXL/pull/1183#discussion_r344283054) [](ancestors = 344283054)

PR #1185, opened 2019-11-08
Do not print pathsToLog in log message for if actual error/warning is not truncated.

norahuang:
AB#3753846


Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1378,7 +1378,8 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
                                     vmInitializer: environment.VmInitializer,
                                     tempDirectoryCleaner: environment.TempCleaner,
                                     incrementalTools: configuration.IncrementalTools,
-                                    changeAffectedInputs: changeAffectedInputs);
+                                    changeAffectedInputs: changeAffectedInputs,
+                                    inCloudBuild: configuration.InCloudBuild);
```
narasamdya: >inCloudBuild [](start = 36, length = 12)

Rename to `suppressStdOutPathInLog`

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -3660,7 +3664,9 @@ private static string ComputePipTimeoutDumpDirectory(ISandboxConfiguration sandb
             outputToLog = (standardOutEmpty ? string.Empty : standardOut) +
                 (!standardOutEmpty && !standardErrorEmpty ? Environment.NewLine : string.Empty) +
                 (standardErrorEmpty ? string.Empty : standardError);
-            pathsToLog = (standardOutEmpty ? string.Empty : standardOutPath) +
+            pathsToLog = m_inCloudBuild
```
narasamdya: >m_inCloudBuild [](start = 25, length = 14)

`suppressStdPathsInLog`

mpysson:
>             // Also use the emptiness as a hit for whether to show the path to the file on disk

I'd reframe the problem slightly to be about not displaying the paths when they don't provide any additional information. In Office's case they configure the logging to not truncate so repeating the path of the file is not necessary. That's true for dev builds as well as in CB, so we shouldn't gate it on whether it's running in CB.

It may be applicable in other scenarios even if they are run in CB if the logging is configured to truncate.

---
Refers to: Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs:3656 in 397d8f0. [](commit_id = 397d8f0f94942bb9ba8811c605974bae5e363db5, deletion_comment = True)

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -3660,7 +3660,9 @@ private static string ComputePipTimeoutDumpDirectory(ISandboxConfiguration sandb
             outputToLog = (standardOutEmpty ? string.Empty : standardOut) +
                 (!standardOutEmpty && !standardErrorEmpty ? Environment.NewLine : string.Empty) +
                 (standardErrorEmpty ? string.Empty : standardError);
-            pathsToLog = (standardOutEmpty ? string.Empty : standardOutPath) +
+            pathsToLog = m_sandboxConfig.OutputReportingMode == OutputReportingMode.FullOutputAlways
```
mpysson: >pathsToLog = m_sandboxConfig.OutputReportingMode == OutputReportingMode.FullOutputAlways [](start = 12, length = 88)

We should have some UTs for this output rendering. Please expand the set with a test case for this functionality.

Public/Src/Engine/UnitTests/Scheduler/PipExecutorTest.cs
```diff
@@ -980,6 +980,15 @@ public async Task FailingProcessWithErrorRegex(OutputReportingMode outputReporti
                             "Non-truncated error message was not found in error event. Full output:" + log);
                     }
 
+                    if (outputReportingMode == OutputReportingMode.FullOutputAlways)
```
mpysson: >outputReportingMode == OutputReportingMode.FullOutputAlways [](start = 24, length = 59)

I'm not sure this is the right condition. There are some others that may come into play like FullOutputOnError.

I didn't notice this in the previous iteration of the PR but I think it would be better to make the condition whether the message was truncated rather than the configuration that may impact whether message is truncated. For example even if you have truncation enabled (the default) the error might only be 1 line long. Then you wouldn't have a need to open up the .err file since all of its contents would already be in the error message.

mpysson:
>             public LogErrorResult(bool success, bool errorWasTruncated)

Here you can see the result object tracks whether errors were truncated. Can you use the same data to gate whether to print the path to the stderr/stdout files? The pattern for the source data would need to be expanded to cover normal output rather than just errors

---
Refers to: Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs:3672 in f6763cc. [](commit_id = f6763ccc42e74f013ea9aa1782e4b26131d9af1b, deletion_comment = False)

Public/Src/Engine/UnitTests/Scheduler/PipExecutorTest.cs
```diff
@@ -1033,6 +1033,94 @@ public async Task FailingProcessWithChunkSizeErrorAndErrorRegex()
...
+                GetFullPath(".cache"),
+                async env =>
+                { 
+                    string workingDirectory = GetFullPath("work");
+                    AbsolutePath workingDirectoryAbsolutePath = AbsolutePath.Create(env.Context.PathTable, workingDirectory);
+
+                    string destination = GetFullPath("dest");
+                    AbsolutePath destinationAbsolutePath = AbsolutePath.Create(env.Context.PathTable, destination);
+
+                    File.WriteAllText(destination, BadContents);
```
mpysson: >File.WriteAllText(destination, BadContents); [](start = 20, length = 44)

I'm not seeing why its necessary to write content to this file prior to launching the pip

Public/Src/Engine/UnitTests/Scheduler/PipExecutorTest.cs
```diff
@@ -1033,6 +1033,94 @@ public async Task FailingProcessWithChunkSizeErrorAndErrorRegex()
                 pathTable => GetConfiguration(pathTable, enableLazyOutputs: false, outputReportingMode: OutputReportingMode.FullOutputOnError));
         }
 
+        [Theory]
+        [InlineData(true, EventId.PipProcessError)]
+        [InlineData(false, EventId.PipProcessError)]
+        [InlineData(false, EventId.PipProcessError, 10 * SandboxedProcessPipExecutor.OutputChunkInLines)]
```
mpysson: >[InlineData(false, EventId.PipProcessError, 10 * SandboxedProcessPipExecutor.OutputChunkInLines)] [](start = 8, length = 97)

What about the case when the regex matches everything but the message gets truncated due to it being too long?

Public/Src/Engine/UnitTests/Scheduler/PipExecutorTest.cs
```diff
@@ -1033,6 +1033,94 @@ public async Task FailingProcessWithChunkSizeErrorAndErrorRegex()
...
+                    string log = EventListener.GetLog();
+                    string relatedLog = GetRelatedLog(log, eventId);
+                    if (regexMatchesEverything)
+                    {
+                        XAssert.IsFalse(relatedLog.Contains(destination));
+                    }
+                    else
+                    {
+                        XAssert.IsTrue(relatedLog.Contains(destination));
+                    }
```
mpysson: I'd expect the condition to make your asserts wrong if the regex matches everything but the error message is still truncated due to the length being too long. Please add a test case for that.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -4022,13 +4033,24 @@ public async Task<bool> TryLogWarningAsync(SandboxedProcessOutput standardError,
                 return false;
             }
 
-            FormatOutputAndPaths(
+            bool errorWasTruncated = false;
```
mpysson: >error [](start = 17, length = 5)

This function deals with warnings, not errors

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -4022,13 +4033,24 @@ public async Task<bool> TryLogWarningAsync(SandboxedProcessOutput standardError,
...
+
+                FormatOutputAndPaths(
                 warningsOutput,
                 warningsError,
                 standardOutput.FileName,
                 standardError.FileName,
                 out string outputTolog,
-                out string outputPathsToLog);
+                out string outputPathsToLog,
+                errorWasTruncated);
```
mpysson: >errorWasTruncated [](start = 16, length = 17)

Do we even allow truncating warnings? There wouldn't be any backing file with just the filtered warnings to refer to.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -4022,13 +4033,24 @@ public async Task<bool> TryLogWarningAsync(SandboxedProcessOutput standardError,
...
             }
 
-            FormatOutputAndPaths(
+            bool errorWasTruncated = false;           
+            // Ignore empty lines
+            var standardErrorInResult = await standardError.ReadValueAsync();
+            var standardOutputInResult = await standardOutput.ReadValueAsync();
+            if (warningsError.Replace(Environment.NewLine, string.Empty).Trim().Length != standardErrorInResult.Replace(Environment.NewLine, string.Empty).Trim().Length ||
+                warningsOutput.Replace(Environment.NewLine, string.Empty).Trim().Length != standardOutputInResult.Replace(Environment.NewLine, string.Empty).Trim().Length)
+            {
```
mpysson: I think these are always going to be true as long as build output has anything else in addition to warnings.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: As to the test case, I do get the errorWasTruncated == true.
So I guess that is the case.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: Changed to warningWasTruncated

Public/Src/Engine/UnitTests/Scheduler/PipExecutorTest.cs
norahuang: I think I covered this case with                         

> if (errorReader.Peek() == -1 && outReader.Peek() == -1)
> {
>         if (stdOutTotalLength != result.StandardOutput.Length || stdErrTotalLength != result.StandardError.Length)
>     {
>         errorWasTruncated = true;
>     }
>  }

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
norahuang: That's true. But that is what we have today. It will still eliminate the printing of the log path in some cases when we only have warning.

Public/Src/Engine/UnitTests/Scheduler/PipExecutorTest.cs
norahuang: And I added this case         
`[InlineData(true, EventId.PipProcessError, 10 * SandboxedProcessPipExecutor.OutputChunkInLines)]`

mpysson: On further inspection I see on line 1092 that the test case changes the Pip's outputReportingMode dependeing on the errorMessageLength. So the scenario of the message getting truncated due to length still isn't exercised with that test case.

The test case I'm concerned about has the following conditions:
1. No regex filtering is applied
1. OutputReportingMode is set to TruncatedOutputOnError
1. The standard output stream exceeds the length at which it would be truncated

In that case the message will be truncated due to length, not regex matching.

PR #1186, opened 2019-11-08
Reduce weight of heavy unit tests to evaluate K Drive

narasamdya:
[AB#1643778](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1643778)

### DIFF
--- a/Public/Src/Engine/UnitTests/FingerprintStore/Test.BuildXL.FingerprintStore.dsc
+++ b/Public/Src/Engine/UnitTests/FingerprintStore/Test.BuildXL.FingerprintStore.dsc
@@ -36,6 +36,6 @@ namespace Test.BuildXL.FingerprintStore {
         runtimeContent: [
             importFrom("BuildXL.Utilities.UnitTests").TestProcess.deploymentDefinition
         ],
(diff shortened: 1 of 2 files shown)


Public/Src/Engine/UnitTests/FingerprintStore/Test.BuildXL.FingerprintStore.dsc
```diff
@@ -36,6 +36,6 @@ namespace Test.BuildXL.FingerprintStore {
         runtimeContent: [
             importFrom("BuildXL.Utilities.UnitTests").TestProcess.deploymentDefinition
         ],
-        runTestArgs: { weight: 10 },
+        runTestArgs: { weight: 2 },
```
mpysson: >  runTestArgs: { weight: 2 }, [](start = 6, length = 29)

Why not just remove them completely?

narasamdya: Let's go with this weight first. If timeout doesn't happen, we can start reconsidering all weights that we set in our self-host.

PR #1187, opened 2019-11-08
Stop ContentLocationDatabase from over-flushing

jbayardo:


Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -726,11 +726,10 @@ Task forceCacheFlushAsync(OperationContext context, ContentLocationDatabaseCount
                     Tracer,
                     async () =>
                     {
-                        long flushedEntries = 0;
+                        long flushedEntries = _cacheUpdatesSinceLastFlush;
```
SergeyTeplyakov: I think it deserves a comment.

SergeyTeplyakov:
As we discussed, lets try to add a test for this logic.

SergeyTeplyakov:
>         internal bool ForceCacheFlush(OperationContext context, ContentLocationDatabaseCounters? counter = null, bool blocking = true)

I think we should flush during shutdown.
And lets have a test there as well.

---
Refers to: Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs:691 in c71420b. [](commit_id = c71420b0b6d3e08543b11981be9e45a2100501b5, deletion_comment = False)

jbayardo:
As discussed via Teams, flush during shutdown is unnecessary. Have added a comment to reflect that

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -77,6 +77,8 @@ public abstract class ContentLocationDatabase : StartupShutdownSlimBase
         /// </summary>
         private long _cacheUpdatesSinceLastFlush = 0;
 
+        internal long CacheUpdatesSinceLastFlush => _cacheUpdatesSinceLastFlush;
```
SergeyTeplyakov: nit: nodoc or a small comment that explains that this thing is used by tests only.

Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -90,7 +92,7 @@ public abstract class ContentLocationDatabase : StartupShutdownSlimBase
 
         private readonly object _flushTaskLock = new object();
 
-        private Task _flushTask = BoolResult.SuccessTask;
+        internal Task FlushTask { get; private set; } = BoolResult.SuccessTask;
```
SergeyTeplyakov: The same here.

PR #1188, opened 2019-11-09
Log checkpoint size

jbayardo:


Public/Src/Cache/ContentStore/Distributed/NuCache/ContentLocationDatabase.cs
```diff
@@ -995,5 +995,7 @@ public bool HasMachineId(byte[] bytes, int machineId)
                     return MachineIdSet.HasMachineId(reader, localIndex);
                 });
         }
+
+        internal abstract Result<long> GetContentDatabaseSizeBytes();
```
SergeyTeplyakov: A comment? And why it is internal and not public? Just curious, because we usually using internal members because they're used in tests only...

Public/Src/Cache/ContentStore/Distributed/NuCache/RocksDbContentLocationDatabase.cs
```diff
@@ -893,6 +894,14 @@ protected override BoolResult GarbageCollectMetadataCore(OperationContext contex
             }).ToBoolResult();
         }
 
+        internal override Result<long> GetContentDatabaseSizeBytes()
```
SergeyTeplyakov: inheritdoc?

Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
```diff
@@ -90,17 +93,24 @@ public Task<BoolResult> CreateCheckpointAsync(OperationContext context, EventSeq
                         // Creating a working temporary directory
                         using (new DisposableDirectory(_fileSystem, _checkpointStagingDirectory))
                         {
+                            // NOTE(jubayard): this needs to be done previous to checkpointing, because we always
+                            // fetch the latest version's size in this way. This implies there may be some difference
+                            // between the reported value and the actual size on disk: updates will get in in-between.
+                            // The better alternative is to actually open the checkpoint and ask, but it seems like too
+                            // much.
+                            checkpointSize = _database.GetContentDatabaseSizeBytes().ThrowIfFailure();
```
SergeyTeplyakov: I don't think we should fail if we won't get the size. Maybe just print -1 as the size in this case?

Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
```diff
@@ -112,10 +122,10 @@ public Task<BoolResult> CreateCheckpointAsync(OperationContext context, EventSeq
                     }
                 },
                 extraStartMessage: $"SequencePoint=[{sequencePoint}]",
-                extraEndMessage: _ => $"SequencePoint=[{sequencePoint}]");
+                extraEndMessage: result => $"SequencePoint=[{sequencePoint}] Id=[{checkpointId}] Size=[{checkpointSize}]");
```
SergeyTeplyakov: Another minor comment. Does it make sense to print the size in megabytes instead? bytes is too low level imo.

Public/Src/Cache/ContentStore/Distributed/NuCache/CheckpointManager.cs
```diff
@@ -90,17 +93,24 @@ public Task<BoolResult> CreateCheckpointAsync(OperationContext context, EventSeq
                         // Creating a working temporary directory
                         using (new DisposableDirectory(_fileSystem, _checkpointStagingDirectory))
                         {
+                            // NOTE(jubayard): this needs to be done previous to checkpointing, because we always
+                            // fetch the latest version's size in this way. This implies there may be some difference
```
SergeyTeplyakov: And thanks a lot for the detailed comment here!

PR #1189, opened 2019-11-09
Support existing file probe in incremental scheduling

narasamdya:
[AB#1614279](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1614279)

### DIFF
--- a/Public/Src/Engine/Scheduler/IncrementalScheduling/GraphAgnosticIncrementalSchedulingState.cs
+++ b/Public/Src/Engine/Scheduler/IncrementalScheduling/GraphAgnosticIncrementalSchedulingState.cs
@@ -107,9 +107,14 @@ namespace BuildXL.Scheduler.IncrementalScheduling
         private ConcurrentBigMap<AbsolutePath, PipGraphSequenceNumber> m_cleanSourceFiles;
 
         /// <summary>
-        /// Mappings from paths to the <see cref="PipStableId"/>'s of pips that dynamically observed them.
+        /// Mappings from paths to the <see cref="PipStableId"/>'s of pips that dynamically read them.
         /// </summary>
         private readonly IncrementalSchedulingPathMapping<PipStableId> m_dynamicallyObservedFiles;
+
+        /// <summary>
+        /// Mappings from paths to the <see cref="PipStableId"/>'s of pips that dynamically probed them.
+        /// </summary>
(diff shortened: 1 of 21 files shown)


Public/Src/Engine/Scheduler/IncrementalScheduling/GraphAgnosticIncrementalSchedulingState.cs
```diff
@@ -649,13 +658,43 @@ private void ProcessDynamicallyObservedChangedPath(string changedPathStr, PathCh
...
+            {
+                if ((changeReasons & PathChanges.Removed) != 0)
+                {
+                    ProcessDynamicallyObservedChangedPath(
+                        changedPathStr, 
+                        changeReasons, 
+                        path, 
+                        DynamicObservationType.ProbedFile, 
+                        m_dynamicallyProbedFiles);
+                }
```
erickulcyk: should created also go through this path?

narasamdya: It's only for existing file probe. If the file is recreated, then the resulting probe will be the same.

---
In reply to: [344813695](https://github.com/microsoft/BuildXL/pull/1189#discussion_r344813695) [](ancestors = 344813695)

ljcollins25:
>         public ReadOnlyArray<AbsolutePath> DynamicallyObservedFiles

I wonder if this should be a list of tuple with AbsolutePath and DynamicObservationKind rather than 3 separate lists. Just a thought.

---
Refers to: Public/Src/Engine/Scheduler/ExecutionResult.cs:324 in bc0a506. [](commit_id = bc0a5061e885d7da17ca3ebccb1d56e6a503051c, deletion_comment = False)

narasamdya:
>         public ReadOnlyArray<AbsolutePath> DynamicallyObservedFiles

When the kinds grow unmanageably, I'll make it a list of tuples.

---
In reply to: [552612497](https://github.com/microsoft/BuildXL/pull/1189#issuecomment-552612497) [](ancestors = 552612497)

---
Refers to: Public/Src/Engine/Scheduler/ExecutionResult.cs:324 in bc0a506. [](commit_id = bc0a5061e885d7da17ca3ebccb1d56e6a503051c, deletion_comment = False)

PR #1191, opened 2019-11-10
[kext] More robust report sending

almiliMSFT:
When enqueuing into `IOSharedDateQueue` fails (because the queue is full), instead of failing we can just re-queue into our own batching queue which, unlike the `IOSharedDataQueue`, can grow dynamically to accommodate high volumes of reports.


Public/Src/Sandbox/MacOs/Sandbox/Src/ConcurrentSharedDataQueue.cpp
```diff
@@ -291,19 +291,30 @@ bool ConcurrentSharedDataQueue::enqueueWithLocking(const EnqueueArgs &args)
...
+    else if (enableBatching_)
+    {
+        // failed to enqueue into IOSharedDataQueue and batching is enabled
+        //   --> re-queue into our own batching queue and keep chugging along
+        return enqueueWithBatching({
+            .report = report,
+            .cacheRecord = nullptr
+        });
+    }
+    else
```
Kr1sso: Nit: Let's just get rid of the else clause all together, writing its code as default.

almiliMSFT:
reports are expected to arrive in order, and this would break that assumption

PR #1193, opened 2019-11-11
Proactive copy on pin

JuanCarlosGI:
Moved ProactiveCopyIfNeededAsync to ReadOnlyDistributedContentSession. Alothough it shows as a lot of read and green in the PR, all the code remained the same.

Main change is that, after copying because of a Pin, we will attempt a proactive copy.


Public/Src/Cache/ContentStore/Distributed/Sessions/ReadOnlyDistributedContentSession.cs
```diff
@@ -964,6 +1030,30 @@ private class RemotePinning
                 BoolResult updated = await UpdateContentTrackerWithNewReplicaAsync(operationContext, new[] { new ContentHashWithSize(remote.ContentHash, copy.ContentSize) }, cancel, UrgencyHint.Nominal);
                 if (updated.Succeeded)
                 {
+                    if (Settings.ProactiveCopyOnPin && Settings.ProactiveCopyMode != ProactiveCopyMode.Disabled)
```
dcwuser: Short comment here like "// Initiate a proactive copy if just pinned content is under-replicated"

PR #1194, opened 2019-11-11
Don't use Assert.All()

almiliMSFT:
In the past, Assert.All used to be flaky on macOS


Public/Src/Utilities/UnitTests/TestUtilities.XUnit/XAssert.cs
```diff
@@ -356,6 +356,60 @@ private static string RenderContainer<T>(IEnumerable<T> container)
             return I($"[{nl}{string.Join("," + nl, elems)}{nl}]");
         }
 
+        /// <summary>
+        /// Asserts that <paramref name="condition"/> holds for every element in <paramref name="container"/>.
```
almiliMSFT:
```suggestion
        /// Asserts that <paramref name="condition"/> holds for every element in <paramref name="container"/>
```

PR #1195, opened 2019-11-11
Cap number of UnexpectedCondition events sent to telemetry

mpysson:


Public/Src/Utilities/Instrumentation/Tracing/Log.cs
```diff
@@ -108,13 +108,55 @@ public abstract partial class Logger
...
+                if (s_currentSessionId != loggingContext.GetRootContext().Session.Id)
+                {
+                    s_currentSessionId = loggingContext.GetRootContext().Session.Id;
+                    s_unexpectedConditionsLogged = 0;
+                }
+
+                s_unexpectedConditionsLogged++;
+
+                // Only log to telemetry fir the first 5 unexpected conditions per build session
+                if (s_unexpectedConditionsLogged <= 5)
```
olkononenko: >5 [](start = 52, length = 1)

should we make this a const, and then use it in the test as well?

Public/Src/Utilities/Instrumentation/Tracing/Log.cs
```diff
@@ -108,13 +108,55 @@ public abstract partial class Logger
...
+            lock (s_currentSessionId)
+            {
+                // Check and reset the context for what build session this in case the server process is reused
+                if (s_currentSessionId != loggingContext.GetRootContext().Session.Id)
+                {
+                    s_currentSessionId = loggingContext.GetRootContext().Session.Id;
+                    s_unexpectedConditionsLogged = 0;
+                }
+
+                s_unexpectedConditionsLogged++;
```
olkononenko: >s_unexpectedConditionsLogged++; [](start = 16, length = 31)

I know it's highly unlikely, but what do you think about adding some check against an overflow? maybe move the increment inside `if (...) {}` ?

Public/Src/Utilities/Instrumentation/Tracing/Log.cs
```diff
@@ -108,13 +108,55 @@ public abstract partial class Logger
...
 
+        /// <summary>
+        /// Logs an UnexpectedCondition both locally as a warning and a capped number of times per build session to telemetry.
+        /// Always call this wrapper function
+        /// </summary>
+        /// <param name="loggingContext"></param>
+        /// <param name="description"></param>
+        public void UnexpectedCondition(LoggingContext loggingContext, string description)
+        {
+            lock (s_currentSessionId)
```
olkononenko: >lock (s_currentSessionId) [](start = 12, length = 25)

I think it's a potential race condition: 
-- thread#1 takes a lock on an instance of `s_currentSessionId`
-- thread#1 reaches line 124 (`s_currentSessionId` is assigned a new variable) 
-- thread#2 enters `UnexpectedCondition` and it will successfully acquire another lock (because `s_currentSessionId` is a new instance now)

The worst thing that might happen is that we'll send more events to the telemetry; though it will only happen if we attempt to log multiple events from multiple threads at roughly the same time, and `s_currentSessionId` is not initialized. In other words, I'm not sure whether it warrants a change.

Public/Src/Utilities/Instrumentation/Tracing/Log.cs
mpysson: Good idea. We should expect the unexpected for what is unexpected

Public/Src/App/UnitTests/Bxl/AppLevelLoggingTests.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        [Fact]
+        public void UnexpectedConditionTelemetryCapTest()
+        {
+            using (var listener = new TrackingEventListener(Events.Log))
+            {
+                listener.RegisterEventSource(global::BuildXL.Tracing.ETWLogger.Log);
+                LoggingContext context = new LoggingContext("Test");
+
+                for (int i=0; i < 6; i++)
```
olkononenko: >i=0 [](start = 25, length = 3)

nit: spaces

Public/Src/Utilities/Instrumentation/Tracing/Log.cs
mpysson: Good catch. We share the same LoggingContext for the entire session so we shouldn't hit it. Easy enough to guard against though by using a separate object for the lock.

PR #1197, opened 2019-11-11
Make resolver initialization async

almiliMSFT:


Public/Src/FrontEnd/Core/FrontEndHostController.cs
```diff
@@ -1154,14 +1155,15 @@ public bool TryInitializeFrontEndsAndResolvers(IConfiguration configuration, Qua
...
                 resolvers.Add(resolver);
+                initResolverTasks[i] = resolver.InitResolverAsync(resolverConfiguration, maybeWorkspaceResolver.Result);
+            }
+
+            var results = await TaskUtilities.SafeWhenAll(initResolverTasks);
+            if (results.Any(r => !r))
+            {
+                // Error should have been reported.
+                return false;
             }
```
mpysson: What motivated the change to make the resolvers async? If it was perf for allowing them to init in parallel, what are the results?

almiliMSFT: That's what motivated it.  The results are inconclusive.  I know we sometimes see long resolver initialization times, but that can be due to NugetResolver and DownloadResolver downloading stuff from the internet.  With this change, download times are not going to improve, but in some cases more parallelism might speed things up a bit

Public/Src/FrontEnd/Core/FrontEndHostController.cs
```diff
@@ -1108,7 +1108,7 @@ private string GetRootFrontEndFolder()
         /// <summary>
         /// Initializes front-ends.
         /// </summary>
-        public bool TryInitializeFrontEndsAndResolvers(IConfiguration configuration, QualifierId[] requestedQualifiers)
+        public async Task<bool> TryInitializeFrontEndsAndResolvers(IConfiguration configuration, QualifierId[] requestedQualifiers)
```
smera: >TryInitializeFrontEndsAndResolvers [](start = 32, length = 34)

you may want to append 'Async' to the name

Public/Src/FrontEnd/Core/FrontEndHostController.cs
smera: I concur. And in my own personal experience the download resolver is what's going on most of the time. I guess the problem is that there is no progress indication on that case, and you just see 'evaluating blah...'

---
In reply to: [345330786](https://github.com/microsoft/BuildXL/pull/1197#discussion_r345330786) [](ancestors = 345330786)

PR #1204, opened 2019-11-12
Trim the contents of IpcClientForwardedMessage events

semihokur:
DX12005_IpcClientForwardedMessage_DurationMs is pretty high in many selfhost builds. The average duration is around 3min. With these changes, I dramatically reduce the content of the messages. 

For Office builds, that duration is pretty low; less than 30sec.

### DIFF
--- a/Public/Src/Engine/Scheduler/ApiServer.cs
+++ b/Public/Src/Engine/Scheduler/ApiServer.cs
@@ -93,7 +93,7 @@ namespace BuildXL.Scheduler
             });
         }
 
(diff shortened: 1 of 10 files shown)


olkononenko:
I'm concerned that with this PR we won't have any way to link `RequestId` to a request. If we ever need to do some investigations into what happened during a build, all we'd have is the list of IDs... Should we log at least one full message per request?

semihokur:
> I'm concerned that with this PR we won't have any way to link `RequestId` to a request. If we ever need to do some investigations into what happened during a build, all we'd have is the list of IDs... Should we log at least one full message per request?

I passed the request id to the deamon so that we will have the id and full message in the deamon log.

PR #1205, opened 2019-11-13
Better handling of augmented FPs

olkononenko:
Under rare conditions, we might have multiple augmented pathsets, as a result we might have ArgumentExceptions in FP Store. This PR provides a better handling of such scenarios. In addition, it also changes how we classify/log cache misses in case of FP augmentation.

### DIFF
--- a/Public/Src/Engine/Scheduler/PipExecutor.cs
+++ b/Public/Src/Engine/Scheduler/PipExecutor.cs
@@ -4953,6 +4953,7 @@ namespace BuildXL.Scheduler
             return new PipExecutorCounter[]
             {
                 PipExecutorCounter.CacheMissesForDescriptorsDueToWeakFingerprints,
+                PipExecutorCounter.CacheMissesForDescriptorsDueToAugmentedWeakFingerprints,
                 PipExecutorCounter.CacheMissesForDescriptorsDueToStrongFingerprints,
                 PipExecutorCounter.CacheMissesForDescriptorsDueToArtificialMissOptions,
                 PipExecutorCounter.CacheMissesForCacheEntry,
--- a/Public/Src/Engine/Scheduler/PipExecutor.cs
+++ b/Public/Src/Engine/Scheduler/PipExecutor.cs
@@ -1879,17 +1879,40 @@ namespace BuildXL.Scheduler
         ///   a non-null result is returned.
         /// - If cache lookup fails (i.e., the result is inconclusive due to failed hashing, etc.), a null result is returned.
         /// </summary>
-        public static Task<RunnableFromCacheResult> TryCheckProcessRunnableFromCacheAsync(
+        public static async Task<RunnableFromCacheResult> TryCheckProcessRunnableFromCacheAsync(
             ProcessRunnablePip processRunnable,
             PipExecutionState.PipScopeState state,
             CacheableProcess cacheableProcess)
         {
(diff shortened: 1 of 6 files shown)


Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1879,28 +1879,54 @@ private static void ReportFileAccesses(ExecutionResult processExecutionResult, F
...
+                pipCacheMiss,
+                canAugmentWeakFingerprint: processRunnable.Process.AugmentWeakFingerprintPathSetThreshold(processRunnable.Environment.Configuration.Cache) > 0,
+                isWeakFingerprintAugmented: false);
+
+            if (!runnableFromCacheResult.CanRunFromCache)
+            {
+                Contract.Assert(pipCacheMiss.Value.CacheMissType != PipCacheMissType.Invalid, "Must have valid cache miss reason");
+                processRunnable.Environment.Counters.IncrementCounter((PipExecutorCounter)pipCacheMiss.Value.CacheMissType);
+
+                Logger.Log.ScheduleProcessPipCacheMiss(
```
mpysson: >ScheduleProcessPipCacheMiss [](start = 27, length = 27)

We already have counters for how many pips are misses due to the various reasons. Should we just add the CacheMissType to this log message and have the one log event? The discrete events below don't really add any more information and slimming down redundant information from our log file is goodness.

Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -2319,22 +2347,40 @@ async Task<RunnableFromCacheResult> innerCheckRunnableFromCacheAsync(List<BoxRef
...
                             {
-                                pipCacheMiss.Value.CacheMissType = PipCacheMissType.MissForDescriptorsDueToWeakFingerprints;
-                                Logger.Log.TwoPhaseCacheDescriptorMissDueToWeakFingerprint(
-                                    operationContext,
-                                    description,
-                                    weakFingerprint.ToString());
+                                if (isWeakFingerprintAugmented)
+                                {
+                                    pipCacheMiss.Value.CacheMissType = PipCacheMissType.CacheMissesForDescriptorsDueToAugmentedWeakFingerprints;
+                                    Logger.Log.TwoPhaseCacheDescriptorMissDueToAugmentedWeakFingerprint(
```
mpysson: >TwoPhaseCacheDescriptorMissDueToAugmentedWeakFingerprint [](start = 47, length = 56)

Suggest removing this, TwoPhaseCacheDescriptorMissDueToWeakFingerprint(), and TwoPhaseCacheDescriptorMissDueToStrongFingerprints(). See other comment above

Public/Src/Engine/Scheduler/PipExecutor.cs
olkononenko: Discussed offline

---
In reply to: [347055628](https://github.com/microsoft/BuildXL/pull/1205#discussion_r347055628) [](ancestors = 347055628)

Public/Src/Engine/Scheduler/PipExecutor.cs
olkononenko: Discussed offline


---
In reply to: [347055775](https://github.com/microsoft/BuildXL/pull/1205#discussion_r347055775) [](ancestors = 347055775)

PR #1207, opened 2019-11-13
Add support for shared compilation for MSBuild scheduled pips

smera:
This PR adds shared compilation support for csc/vbc compilations for the MSBuild frontend.
This is achieved by:
- Configuring VBCSCompiler.exe so it can breakaway from the sandbox (if that feature is supported by the underlying sandbox)
- Attaching an MSBuild logger in that case that can compensate for missing accesses and report them as augmented accesses. This is done by intercepting the csc/vbc MSBuild task command line arguments and using the Roslyn code analysis functionality to statically predict inputs/outputs.

### DIFF
--- a/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
+++ b/Public/Src/FrontEnd/MsBuild/PipConstructor.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections;
+using System.Reflection;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
@@ -11,6 +11,7 @@ using System.Text;
 using BuildXL.FrontEnd.Sdk;
 using BuildXL.FrontEnd.Utilities;
 using BuildXL.FrontEnd.Workspaces.Core;
+using BuildXL.Native.Processes;
 using BuildXL.Pips;
 using BuildXL.Pips.Builders;
 using BuildXL.Pips.Operations;
@@ -45,7 +46,6 @@ namespace BuildXL.FrontEnd.MsBuild
         {
             "/NoLogo",
             "/p:TrackFileAccess=false", // Turns off MSBuild's Detours based tracking. Required always to prevent Detouring a Detour which is not supported
-            "/p:UseSharedCompilation=false", //  Turn off new MSBuild flag to reuse VBCSCompiler.exe (a feature of Roslyn csc compiler) to compile C# files from different projects
             "/m:1", // Tells MsBuild not to create child processes for building, but instead to simply execute the specified project file within a single process
             "/IgnoreProjectExtensions:.sln", // Tells MSBuild to avoid scanning the local file system for sln files to build, but instead to simply use the provided project file.
             "/ConsoleLoggerParameters:Verbosity=Minimal", // Minimize the console logger
@@ -53,9 +53,12 @@ namespace BuildXL.FrontEnd.MsBuild
             "/nodeReuse:false" // Even though we are already passing /m:1, when an MSBuild task is requested with an architecture that doesn't match the one of the host process, /nodeReuse will be true unless set otherwise
         };
 
+        // Keep in sync with the bxl deployment
+        internal const string VBCSCompilerLogger = "VBCSCompilerLogger.dll";
+
         private AbsolutePath Root => m_resolverSettings.Root;
 
-        private readonly AbsolutePath m_msBuildPath;
+        private readonly AbsolutePath m_msBuildPath; 
         private readonly AbsolutePath m_dotnetExePath;
         private readonly string m_frontEndName;
         private readonly IEnumerable<KeyValuePair<string, string>> m_userDefinedEnvironment;
@@ -75,7 +78,9 @@ namespace BuildXL.FrontEnd.MsBuild
         // All projects should contain this property since the build graph is created by MSBuild under the /graph option
         // TODO: it would be better if MSBuild provided the property name
         internal const string s_isGraphBuildProperty = "IsGraphBuild";
-
+        
+        private bool UseSharedCompilation => ProcessUtilities.SandboxSupportsProcessBreakaway() && m_resolverSettings.UseManagedSharedCompilation != false;
+        
         /// <nodoc/>
         public PipConstructor(
             FrontEndContext context,
@@ -590,11 +595,25 @@ namespace BuildXL.FrontEnd.MsBuild
             //
             // All child processes: Don't wait to kill the processes.
             // CODESYNC: CloudBuild repo TrackerExecutor.cs "info.NestedProcessTerminationTimeout = TimeSpan.Zero"
-            processBuilder.AllowedSurvivingChildProcessNames = ReadOnlyArray<PathAtom>.FromWithoutCopy(
+            var allowedSurvivingChildProcessNames = new List<PathAtom>() { 
                 PathAtom.Create(m_context.StringTable, "mspdbsrv.exe"),
                 PathAtom.Create(m_context.StringTable, "vctip.exe"),
-                PathAtom.Create(m_context.StringTable, "conhost.exe"),
-                PathAtom.Create(m_context.StringTable, "VBCSCompiler.exe"));
+                PathAtom.Create(m_context.StringTable, "conhost.exe")};
+
+            // If the sandbox supports process breakaway and shared compilation is configured to run, we configure VBCSCompiler as such
+            // Otherwise, we add it as a process that is safe to kill when it survives
+            var vbcsCompiler = PathAtom.Create(m_context.StringTable, "VBCSCompiler.exe");
+            if (UseSharedCompilation)
+            {
+                processBuilder.ChildProcessesToBreakawayFromSandbox = ReadOnlyArray<PathAtom>.FromWithoutCopy(vbcsCompiler);
+            }
+            else
+            {
+                allowedSurvivingChildProcessNames.Add(vbcsCompiler);
+            }
+            
+            processBuilder.AllowedSurvivingChildProcessNames = allowedSurvivingChildProcessNames.ToReadOnlyArray();
+
             // There are some cases (e.g. a 64-bit MSBuild launched as a child process from a 32-bit MSBuild instance) where
             // processes need a little bit more time to finish. Increasing the timeout does not affect job objects where no child
             // processes survive, or job object where the only surviving processes are the ones explicitly allowed to survive (which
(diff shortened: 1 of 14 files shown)


Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
```diff
@@ -0,0 +1,191 @@
...
+                    default:
+                        return;
+                }
+
+                // We were able to split the compiler invocation from its arguments. This is the indicator
+                // that something didn't go as expected. Since failing to parse the command line means we
+                // are not catching all inputs/outputs properly, we have no option but to fail the corresponding pip
+                if (arguments == null)
+                {
+                    throw new ArgumentException($"Unexpected tool name in command line. Expected '{CscToolName}' or '{VbcToolName}', but got: {commandLine.CommandLine}");
```
cdmihai: nit: method makes assumptions on the root cause of the received null.

smera: Fair, I'll refactor that


---
In reply to: [346063932](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346063932) [](ancestors = 346063932)

Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                return null;
+            }
+
+            return commandLine.Substring(index + toolToTrim.Length);
+        }
+
+        private void RegisterAccesses(CommandLineArguments args)
+        {
+            // All inputs
```
cdmihai: Why aren't the references considered as inputs?

cdmihai: Think it's this one [https://github.com/dotnet/roslyn/blob/a51ee294a07b04cca9d307e50772c87b04f2309d/src/Compilers/Core/Portable/CommandLine/CommandLineArguments.cs#L56](https://github.com/dotnet/roslyn/blob/a51ee294a07b04cca9d307e50772c87b04f2309d/src/Compilers/Core/Portable/CommandLine/CommandLineArguments.cs#L56)

---
In reply to: [346065602](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346065602) [](ancestors = 346065602)

smera: Good catch. I think you found a copy & paste error, observe I'm adding AnalyzerReferences twice, but not ReferencePaths :)

---
In reply to: [346065905](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346065905) [](ancestors = 346065905,346065602)

Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
```diff
@@ -0,0 +1,191 @@
...
+            RegisterInputs(args.SourceFiles.Select(source => source.Path));
+            RegisterInputs(args.AnalyzerReferences.Select(reference => reference.FilePath));
+            RegisterInputs(args.EmbeddedFiles.Select(embedded => embedded.Path));
+            RegisterInput(args.Win32ResourceFile);
+            RegisterInput(args.Win32Icon);
+            RegisterInput(args.Win32Manifest);
+            RegisterInputs(args.AdditionalFiles.Select(additional => additional.Path));
+            RegisterInputs(args.MetadataReferences.Select(metadata => metadata.Reference).Where(pathOrAssemblyName => FileUtilities.FileSystem.IsPathRooted(pathOrAssemblyName)));
```
cdmihai: Why only add the rooted paths?

Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
```diff
@@ -0,0 +1,191 @@
...
+    /// This MSBuild logger is used by the MSBuild frontend of BuildXL when scheduling pips: when process breakaway is supported, VBSCompiler.exe is allowed to escape the sandbox
+    /// and outlive the originating pip. This logger is used as a way to compensate for the missing accesses and the MSBuild frontend attaches it to every MSBuild invocation.
+    /// </remarks>
+    public class CompilerFileAccessLogger : Logger
+    {
+        private const string CscTaskName = "Csc";
+        private const string VbcTaskName = "Vbc";
+        private const string CscToolName = "csc.exe";
+        private const string VbcToolName = "vbc.exe";
+        private AugmentedManifestReporter m_augmentedReported;
```
cdmihai: typo: name should end in Reporter right?

Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        private void RegisterOutput(string filePath)
+        {
+            RegisterAccess(filePath, m_augmentedReported.TryReportFileCreations);
+        }
+
+        private void RegisterInput(string filePath)
+        {
+            RegisterAccess(filePath, m_augmentedReported.TryReportFileReads);
```
cdmihai: No idea on the perf / asyncness of TryReportFileReads. Should inputs be batched rather than sent one by one? Or it doesn't really matter?

Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
smera: yes

---
In reply to: [346068547](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346068547) [](ancestors = 346068547)

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -590,11 +595,25 @@ private void SetProcessEnvironmentVariables(IReadOnlyDictionary<string, string>
...
                 PathAtom.Create(m_context.StringTable, "mspdbsrv.exe"),
                 PathAtom.Create(m_context.StringTable, "vctip.exe"),
-                PathAtom.Create(m_context.StringTable, "conhost.exe"),
-                PathAtom.Create(m_context.StringTable, "VBCSCompiler.exe"));
+                PathAtom.Create(m_context.StringTable, "conhost.exe")};
+
+            // If the sandbox supports process breakaway and shared compilation is configured to run, we configure VBCSCompiler as such
+            // Otherwise, we add it as a process that is safe to kill when it survives
+            var vbcsCompiler = PathAtom.Create(m_context.StringTable, "VBCSCompiler.exe");
+            if (UseSharedCompilation)
```
cdmihai: Wonder if there's a nenvironment or msbuild property to turn shared compilation off as well. If it's turned off, then bxl will duplicate the file accesses right?

Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
smera: It shouldn't matter. This writes into an already opened pipe. There is in fact logic on bxl side (the other end of the pipe) to do async reads

---
In reply to: [346069206](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346069206) [](ancestors = 346069206)

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
```diff
@@ -613,6 +632,29 @@ private bool TryAddMsBuildArguments(ProjectWithPredictions project, PipDataBuild
...
 
+            // If process breakaway is supported and shared compilation is configured to run, set up the logger that will mimic all the proper accesses
+            // Otherwise, disable shared compilation
+            if (UseSharedCompilation)
+            {
+                var vbcsCompilerLoggerPath = AbsolutePath.Create(PathTable, Assembly.GetAssembly(typeof(PipConstructor)).Location)
+                    .GetParent(PathTable)
+                    // Depending on the framework of the MSBuild we are using, we should provide the corresponding logger
+                    // Keep in sync with BuildXL deployment
+                    .Combine(PathTable, m_resolverSettings.ShouldRunDotNetCoreMSBuild()? "dotnetcore" : "net472")
```
cdmihai: Would it make sense to layout the logger dll right next to this dll so you don't have to guess the pass?

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
smera: You mean a mechanism that overrides passing /p:UseSharedCompilation=false to MSBuild?
I was assuming passing that global property will trump any other configuration at this respect. I'm not 100% sure though.
However, even in that case duplicating the accesses is not really a big deal. The cost of actually reporting those accesses should be not significant. And having duplicates is not a problem.

---
In reply to: [346069926](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346069926) [](ancestors = 346069926)

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
smera: But the decision is not made based on the location of this dll, but on whether we should run netcore or full framework MSBuild. And each version needs to live in its own folder, otherwise there might be some double-deployment problems when dlls clash because of tfm disagreement

---
In reply to: [346070749](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346070749) [](ancestors = 346070749)

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
cdmihai: The msbuild build could decide to turn off shared compilation (e.g. turning it off in a `directory.build.props`), in which case bxl can skip shared compilation as well. Or maybe that already happens and I missed it.

cdmihai: But if there's no real consequence on logging the IO regardless of what msbuild chooses to do then it doesn't matter.
Also, there could be build time logic to change shared compilation, so I guess it's not really knowable what the build decides to do

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
cdmihai: Then wouldn't it be better if the dsc files did the output directory checks and copy things rather than having the c# code do it? But it's more of a nit really, doesn't matter that much :)

smera: But dsc files take care of the deployment of bxl itself. And therefore they need to consider both cases, since we need both versions of the logger (full and netcore). The decision of which one to use happens at runtime, for a particular bxl invocation. dsc files are not involved at runtime.

---
In reply to: [346074810](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346074810) [](ancestors = 346074810)

Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
smera: Given the spec that string can be either an absolute path or an assembly display name. Looking at that more carefully, I don't really know how to handle the assembly display name since it needs some resolution that doesn't look like a static thing. Looking...

---
In reply to: [346068155](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346068155) [](ancestors = 346068155)

Public/Src/FrontEnd/MsBuild/PipConstructor.cs
cdmihai: Ah, okay, I thought bxl had two deployments, one for full framework, with the full framework binaries copied, and one netcore, with the netcore binaries copies (some of which will be the same, i.e. netstandard).

Public/Src/Tools/VBCSCompilerLogger/VBCSCompilerLogger.cs
smera: Ended up implementing a logic similar to the one the compiler does. SearchPaths are a collection of directories that are used for resolving relative paths

---
In reply to: [346076348](https://github.com/microsoft/BuildXL/pull/1207#discussion_r346076348) [](ancestors = 346076348,346068155)

PR #1215, opened 2019-11-15
Add eager fingerprint incorporation

SergeyTeplyakov:
AD#1644324

### DIFF
--- a/Public/Src/Cache/MemoizationStore/Vsts/BuildCacheServiceConfiguration.cs
+++ b/Public/Src/Cache/MemoizationStore/Vsts/BuildCacheServiceConfiguration.cs
@@ -120,10 +120,11 @@ namespace BuildXL.Cache.MemoizationStore.Vsts
         /// <nodoc />
         public const int DefaultEagerFingerprintIncorporationNagleBatchSize = 100;
 
-        /// <summary>
-        /// todoc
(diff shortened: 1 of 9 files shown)


Public/Src/Cache/MemoizationStore/Vsts/BuildCacheServiceConfiguration.cs
```diff
@@ -108,6 +109,22 @@ public class BuildCacheServiceConfiguration
...
+        public const bool DefaultEnableEagerFingerprintIncorporation = false;
+
+        /// <nodoc />
+        public static readonly TimeSpan DefaultEagerFingerprintIncorporationNagleInterval = TimeSpan.FromMinutes(5);
+
+        /// <nodoc />
+        public const int DefaultEagerFingerprintIncorporationNagleBatchSize = 100;
+
+        /// <nodoc />
+        public static TimeSpan DefaultEagerFingerprintIncorporationExpiry = TimeSpan.FromDays(1);
```
dcwuser: Technically these defaults violate the rule that, when shipped with no configuration, the new feature is off. For that to be true, they should both be set to infinity. The difference seems low risk, but I wouldn't want to explain myself in front of RCA if it goes wrong.

PR #1216, opened 2019-11-15
Add /fireForgetMaterializeOutput feature

semihokur:
I introduced /fireForgetMaterializeOutput so that we can do A/B testing. I added comments, so it should be easy to review the changes.

[AB#1611984](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1611984)
Fixes [AB#1617127](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1617127)


Public/Src/Engine/Scheduler/ExecutionResult.cs
```diff
@@ -610,6 +610,18 @@ public static ExecutionResult GetFailureNotRunResult(LoggingContext loggingConte
             return result;
         }
 
+        /// <summary>
+        /// Gets an empty success result for Mate
```
olkononenko: >Mate [](start = 45, length = 4)

nit: materialization ?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -344,6 +344,26 @@ private void StartWorkers(LoggingContext loggingContext)
             m_allWorker = new AllWorker(m_workers.ToArray());
 
             ExecutionLog?.WorkerList(new WorkerListEventData { Workers = m_workers.SelectArray(w => w.Name) });
+        }
+
+        private readonly TaskSourceSlim<bool> m_schedulerCompletionExceptMaterializeOutputs = TaskSourceSlim.Create<bool>();
```
olkononenko: > pr [](start = 7, length = 3)

nit: empty line

olkononenko: >ate [](start = 12, length = 3)

nit: why do we have a mix of fields and methods in this class? shouldn't we keep all the fields at the top followed by the methods?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -344,6 +344,26 @@ private void StartWorkers(LoggingContext loggingContext)
             m_allWorker = new AllWorker(m_workers.ToArray());
 
             ExecutionLog?.WorkerList(new WorkerListEventData { Workers = m_workers.SelectArray(w => w.Name) });
+        }
+
+        private readonly TaskSourceSlim<bool> m_schedulerCompletionExceptMaterializeOutputs = TaskSourceSlim.Create<bool>();
+        private bool AnyPendingPipsExceptMaterializeOutputs()
+        {
+            long numRunningOrQueued = m_pipQueue.NumRunningOrQueued;
+            long numRunningOrQueuedExceptMaterialize = numRunningOrQueued - m_pipQueue.GetNumRunningByKind(DispatcherKind.Materialize) - m_pipQueue.GetNumQueuedByKind(DispatcherKind.Materialize);
```
olkononenko: >DispatcherKind.Materialize [](start = 107, length = 26)

nit: should we make this an argument to make the methods more 'generic'

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -344,6 +344,26 @@ private void StartWorkers(LoggingContext loggingContext)
...
+
+        private readonly TaskSourceSlim<bool> m_schedulerCompletionExceptMaterializeOutputs = TaskSourceSlim.Create<bool>();
+        private bool AnyPendingPipsExceptMaterializeOutputs()
+        {
+            long numRunningOrQueued = m_pipQueue.NumRunningOrQueued;
+            long numRunningOrQueuedExceptMaterialize = numRunningOrQueued - m_pipQueue.GetNumRunningByKind(DispatcherKind.Materialize) - m_pipQueue.GetNumQueuedByKind(DispatcherKind.Materialize);
+
+            if (numRunningOrQueuedExceptMaterialize == 0)
+            {
+                RetrievePipStateCounts(out long totalPips, out long readyPips, out long waitingPips, out long runningPips, out long donePips, out long failedPips, out long skippedPips, out long ignoredPips);
```
olkononenko: >RetrievePipStateCounts(out long totalPips, out long readyPips, out long waitingPips, out long runningPips, out long donePips, out long failedPips, out long skippedPips, out long ignoredPips); [](start = 16, length = 191)

why do we need this? isn't `m_pipQueue.NumRunningOrQueued` already capturing all the pips?

olkononenko:
can we add a test that would check that the change (well, at least some parts of it) is working as intended?

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -109,19 +111,20 @@ internal sealed class RemoteWorker : RemoteWorkerBase
...
-                // Limit number of concurrently attaching workers
-                async token => await m_masterService.WorkerAttachSemaphore.AcquireAsync(token));
+                OnConnectionTimeOutAsync);
 
             // Depending on how long send requests take. It might make sense to use the same thread between all workers. 
             m_sendThread = new Thread(SendBuildRequests);
         }
 
-        public override void Initialize(PipGraph pipGraph, IExecutionLogTarget executionLogTarget)
+        private TaskSourceSlim<bool> m_schedulerCompletion;
```
olkononenko: >private TaskSourceSlim<bool> m_schedulerCompletion; [](start = 8, length = 51)

nit: can you put all the fields together?

Public/Src/Engine/Scheduler/Scheduler.cs
```diff
@@ -3383,7 +3403,13 @@ private async Task<PipExecutionStep> ExecutePipStep(RunnablePip runnablePip)
                     }
 
                 case PipExecutionStep.MaterializeOutputs:
-                    {
+                    {
+                        if (m_configuration.Distribution.FireForgetMaterializeOutput && !AnyPendingPipsExceptMaterializeOutputs())
+                        {
+                            // There is no pips running anything except materializeOutputs. 
+                            m_schedulerCompletionExceptMaterializeOutputs.TrySetResult(false);
```
olkononenko: >false [](start = 87, length = 5)

if one is looking only at this block of code, it's not clear / confusing why we are setting `m_schedulerCompletionExceptMaterializeOutputs` to false; i.e., essentially we have a bool that says 'completed but some materialization is still running', and once the engine enters this state, we set that bool to `!true`

olkononenko: I understand that you setting it to false so you can put it through the existing logic (`Task.WhenAny(m_attachCompletion.Task, m_schedulerCompletion.Task)`) and I appreciate that you added that detailed comment there; but I think we need to make it clear what is going on here (in this method). Maybe we need to rename the variable?

Alternatively, we can set it here to 'true' so the var name aligns with the assigned value, and change the `RemoteWorker` :
```
var attachCompletionResult = await Task.WhenAny(m_attachCompletion.Task, m_schedulerCompletion.Task);

if (attachCompletionResult == m_schedulerCompletion.Task || !await attachCompletionResult) 
{
...
} 
``` 

---
In reply to: [348756422](https://github.com/microsoft/BuildXL/pull/1216#discussion_r348756422) [](ancestors = 348756422)

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -486,12 +503,10 @@ private async Task DisconnectAsync(string buildFailure = null)
                 m_sendThread.Join();
             }
 
-            CancellationTokenSource exitCancellation = new CancellationTokenSource();
-
             // Only wait a short amount of time for exit (15 seconds) if worker is not successfully attached.
             if (m_attachCompletion.Task.Status != TaskStatus.RanToCompletion || !await m_attachCompletion.Task)
             {
-                exitCancellation.CancelAfter(TimeSpan.FromSeconds(15));
+                m_exitCancellation.CancelAfter(TimeSpan.FromSeconds(15));
```
olkononenko: >15 [](start = 68, length = 2)

nit: make it a const?

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -648,7 +670,10 @@ public async Task SendToRemote(OperationContext operationContext, RunnablePip ru
             }
             catch (InvalidOperationException)
             {
-                if (m_isConnectionLost)
+                // If the connection is lost or the worker is not running at all since the beginning 
+                // (due to ValidateCacheConnection failure), then fail the pips.
+                // Otherwise, it is an unexpected exception, so throw it.
+                if (m_isConnectionLost || !m_everAvailable)
```
olkononenko: >m_everAvailable [](start = 43, length = 15)

shouldn't it be `EverAvailable` (i.e. Volatile.Read()) ?

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
```diff
@@ -657,7 +682,7 @@ public async Task SendToRemote(OperationContext operationContext, RunnablePip ru
                     // and before we add those build requests to blocking collection(m_buildRequests), 
                     // we will try to add the build request to the blocking colletion which is marked as completed 
                     // It will throw InvalidOperationException. 
-                    FailRemotePip(pipCompletionTask, "Connection was lost");
+                    FailRemotePip(pipCompletionTask, $"Connection was lost. Was the worker ever available: {m_everAvailable}");
```
olkononenko: >m_everAvailable [](start = 108, length = 15)

`EverAvailable` ?

olkononenko:
Approving the PR, but I'd appreciate if you can find somebody else to review it as well since I'm not very familiar with our 'distribution' code.

semihokur:
> Approving the PR, but I'd appreciate if you can find somebody else to review it as well since I'm not very familiar with our 'distribution' code.

There is currently none :)

Public/Src/Engine/Scheduler/Scheduler.cs
semihokur: no particular reason. as Scheduler class gets bigger, people wanted to put the new private members close to where they used it. 

---
In reply to: [348743741](https://github.com/microsoft/BuildXL/pull/1216#discussion_r348743741) [](ancestors = 348743741)

Public/Src/Engine/Scheduler/Scheduler.cs
semihokur: the if branch of the method is already related to MaterializeOutput part. Materialize step is pretty embedded in this method; so we do not need to make it generic.

---
In reply to: [348745577](https://github.com/microsoft/BuildXL/pull/1216#discussion_r348745577) [](ancestors = 348745577)

Public/Src/Engine/Scheduler/Scheduler.cs
semihokur: Added comment.

---
In reply to: [348746468](https://github.com/microsoft/BuildXL/pull/1216#discussion_r348746468) [](ancestors = 348746468)

Public/Src/Engine/Dll/Distribution/RemoteWorker.cs
semihokur: good catch!

---
In reply to: [348763766](https://github.com/microsoft/BuildXL/pull/1216#discussion_r348763766) [](ancestors = 348763766)

PR #1218, opened 2019-11-15
RocksDB tuning for macOS

Kr1sso:
This PR updates the RocksDB packages and introduces some performance tuning adjustments for the macOS version. Flushing out + compacting memtables more frequently reduces the BuildXL memory footprint and using LZ4 does so at a minimal cost. The RocksDB repo has been updated to build RocksDB for macOS with dynamically linked Snappy and LZ4 libraries, the build steps patch the RocksDB dylib to look for its dependencies in the @loader_path when loading.

(NEW)

ServerOffRAM-J

vs. (OLD)

ServerOffRAM-H

### DIFF
--- a/Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
+++ b/Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
@@ -172,8 +172,17 @@ namespace BuildXL.Engine.Cache.KeyValueStores
                     // The background compaction threads run in low priority, so they should not hamper the rest of
                     // the system. The number of cores in the system is what we want here according to official docs,
                     // and we are setting this to the number of logical processors, which may be higher.
+#if !PLATFORM_OSX
                     .SetMaxBackgroundCompactions(Environment.ProcessorCount)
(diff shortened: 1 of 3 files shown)


Public/Src/Utilities/KeyValueStore/RocksDb/RocksDbStore.cs
```diff
@@ -172,8 +172,17 @@ private struct DefaultOptions
                     // The background compaction threads run in low priority, so they should not hamper the rest of
                     // the system. The number of cores in the system is what we want here according to official docs,
                     // and we are setting this to the number of logical processors, which may be higher.
+#if !PLATFORM_OSX
                     .SetMaxBackgroundCompactions(Environment.ProcessorCount)
                     .SetMaxBackgroundFlushes(1)
+#else
+                    // The memtable uses significant chunks of available system memory on macOS, we increase the number
```
almiliMSFT: All this is probably true for Windows too, which is why it would be better if this was configurable (e.g., via some sort of RocksDB config file) instead of hardcoding it for the two platforms.  (of course, this can be done in a separate PR)

PR #1220, opened 2019-11-15
Proactive replication

JuanCarlosGI:
Implement proactive replication.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
@@ -1,0 +1,137 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.ContractsLight;
+using System.Linq;
+using System.Text;
+using BuildXL.Cache.ContentStore.Hashing;
+using BuildXL.Cache.ContentStore.Utils;
+
+namespace BuildXL.Cache.ContentStore.Distributed.NuCache
+{
+    public class BinManager
+    {
+        private const int MaxBins = 64 * 1024;
+
+        private readonly MachineLocation[] _bins;
+        private readonly int _locationsPerBin;
+        private readonly int _entriesPerMachine;
+
+        private MachineLocation[][] _machinesForBins;
+
+        private readonly IContentHasher _contentHasher = ContentHashers.Get(HashType.MD5);
+
+        public BinManager(int machinesPerBin, int entriesPerMachine = 4, int amountOfBins = MaxBins)
+        {
+            Contract.Assert(entriesPerMachine <= byte.MaxValue);
+            Contract.Assert(AmountOfBinsIsValid(amountOfBins), $"{nameof(amountOfBins)} should be in range [1, {MaxBins}] and be a power of 2.");
+            _locationsPerBin = machinesPerBin;
+            _entriesPerMachine = entriesPerMachine;
+            _bins = new MachineLocation[amountOfBins];
+        }
+
+        private static bool AmountOfBinsIsValid(int amount)
+        {
+            return amount > 0 && amount <= MaxBins &&
+                ((amount & (amount - 1)) == 0); // Is power of 2.
+        }
+
+        public void AddLocation(MachineLocation location) => ProcessLocation(location, valueToSet: location);
+
+        public void RemoveLocation(MachineLocation location) => ProcessLocation(location, valueToSet: default);
+
+        private void ProcessLocation(MachineLocation location, MachineLocation valueToSet)
+        {
+            var keyBytes = new byte[location.Data.Length + 1];
+            location.Data.CopyTo(keyBytes, 0);
+            for (byte entry = 0; entry < _entriesPerMachine; entry++)
+            {
+                keyBytes[keyBytes.Length - 1] = entry;
+                var key = _contentHasher.GetContentHash(keyBytes);
+                var index = BitConverter.ToUInt16(key.ToByteArray(), 1);
+                _bins[index % _bins.Length] = valueToSet;
+            }
+
+            // Invalidate current configuration.
+            _machinesForBins = null;
+        }
+
+        public MachineLocation[] GetLocations(ContentHash hash)
+        {
+            _machinesForBins ??= ComputeBins();
+            var index = BitConverter.ToUInt16(hash.ToByteArray(), 1);
+            return _machinesForBins[index % _bins.Length];
+        }
+
+        private MachineLocation[][] ComputeBins()
+        {
+            var result = new MachineLocation[_bins.Length][];
+            var first = new List<MachineLocation>();
+            foreach (var bin in _bins)
+            {
+                if (!bin.Equals(default) && !first.Contains(bin))
+                {
+                    first.Add(bin);
+                    if (first.Count == _locationsPerBin)
+                    {
+                        break;
+                    }
+                }
+            }
+
+            if (first.Count < _locationsPerBin)
+            {
+                var entry = first.ToArray();
+                for (var i = 0; i < _bins.Length; i++)
+                {
+                    result[i] = entry;
+                }
+
+                return result;
+            }
+
+            result[0] = first.ToArray();
+            var prev = result[0];
+
+            for (var bin = _bins.Length - 1; bin > 0; bin--)
+            {
+                if (!_bins[bin].Equals(default))
+                {
+                    prev = insertFirst(prev, _bins[bin]);
+                }
+
+                result[bin] = prev;
+            }
+
+            return result;
+
+            static MachineLocation[] insertFirst(MachineLocation[] from, MachineLocation withValue)
+            {
+                var result = new MachineLocation[from.Length];
+                result[0] = withValue;
+                var repeated = false;
+                for (var i = 1; i < result.Length; i++)
+                {
+                    if (!repeated && withValue.Equals(from[i - 1]))
+                    {
+                        repeated = true;
+                    }
+
+                    result[i] = from[i - (repeated ? 0 : 1)];
+                }
+
+                return result;
+            }
+        }
+
+        public override string ToString()
+        {
+            _machinesForBins ??= ComputeBins();
+
+            var bins = string.Join(", ", _bins.Select(bin => bin.Path));
+            return bins + "\n" + string.Join("\n", _machinesForBins.Select(machines => string.Join(",", machines.Select(machine => machine.Path))));
+        }
+    }
+}
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
 using System.Linq;
-using System.Text;
 using BuildXL.Cache.ContentStore.Hashing;
 using BuildXL.Cache.ContentStore.Utils;
 
@@ -23,7 +22,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 
         private readonly IContentHasher _contentHasher = ContentHashers.Get(HashType.MD5);
 
-        public BinManager(int machinesPerBin, int entriesPerMachine = 4, int amountOfBins = MaxBins)
+        public BinManager(int machinesPerBin, int entriesPerMachine = 10, int amountOfBins = MaxBins)
         {
             Contract.Assert(entriesPerMachine <= byte.MaxValue);
             Contract.Assert(AmountOfBinsIsValid(amountOfBins), $"{nameof(amountOfBins)} should be in range [1, {MaxBins}] and be a power of 2.");
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
@@ -10,7 +10,7 @@ using BuildXL.Cache.ContentStore.Utils;
 
 namespace BuildXL.Cache.ContentStore.Distributed.NuCache
 {
-    public class BinManager
+    internal class BinManager
     {
         private const int MaxBins = 64 * 1024;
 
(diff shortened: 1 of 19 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1380,6 +1410,47 @@ public Task<BoolResult> RegisterLocalLocationAsync(OperationContext context, IRe
...
         }
 
+        private Task<BoolResult> ProactiveReplicationAsync(OperationContext context)
+        {
+            return context.PerformOperationAsync(
+                Tracer,
+                async () =>
+                {
+                    IEnumerable<ContentInfo> localContent = await _localContentStore.GetContentInfoAsync(context.Token);
+                    localContent = localContent.OrderByDescending(info => info.LastAccessTimeUtc);
```
dcwuser: This orders by local last access time, not global effective last access time.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1380,6 +1410,47 @@ public Task<BoolResult> RegisterLocalLocationAsync(OperationContext context, IRe
             }
         }
 
+        private Task<BoolResult> ProactiveReplicationAsync(OperationContext context)
```
dcwuser: Please lots more comments here saying what each part is doing.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -647,6 +654,29 @@ private async Task<BoolResult> RestoreCheckpointStateAsync(OperationContext cont
                         Task.Run(() => ReconcileAsync(context), context.Token).FireAndForget(context);
                     }
                 }
+
+                // Make sure to only have one thread doing proactive replication, in case last proactive replication is still happening.
+                if (_configuration.EnableProactiveReplication && Interlocked.CompareExchange(ref _replicating, 1, 0) == 0)
```
dcwuser: If previous proactive replication task is still running, this skips current cycle and lets old one run. It would be better to stop old task and start a new one, since we now have better information than old cycle.

SergeyTeplyakov: I think we should use another patter here, when we have a task field and a lock and we just returning an existing task if its still running.

I think this place is not a perf critical enough to deserve lock free impl. I would suggest using more readable solution.

Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// <nodoc />
+        public MachineLocation[] GetLocations(ContentHash hash)
+        {
+            _machinesForBins ??= ComputeBins();
+            var index = BitConverter.ToUInt16(hash.ToByteArray(), 1);
+            return _machinesForBins[index % _bins.Length];
+        }
+
+        private MachineLocation[][] ComputeBins()
+        {
```
SergeyTeplyakov: This stuff looks complicated (a lot of code) with no comments.

Plus I would like to see binManagerTests. This stuff is fully isolated and should be very testable.

Public/Src/Cache/ContentStore/Distributed/NuCache/ClusterState.cs
```diff
@@ -162,5 +208,10 @@ public Result<MachineLocation> GetRandomMachineLocation(IReadOnlyList<MachineLoc
                 return new Result<MachineLocation>("Could not select a machine location.");
             }
         }
+
+        /// <summary>
+        /// Gets the set of locations designated for a hash. This is relevant for proactive copies and eviction.
+        /// </summary>
+        public MachineLocation[] GetDesignatedLocations(ContentHash hash) => _binManager.GetLocations(hash);
```
SergeyTeplyakov: Will fail with NRE if _binManager is null.

Does it make sense to add a NullBinManager? or just create it all the time?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1380,6 +1410,47 @@ public Task<BoolResult> RegisterLocalLocationAsync(OperationContext context, IRe
...
+                async () =>
+                {
+                    IEnumerable<ContentInfo> localContent = await _localContentStore.GetContentInfoAsync(context.Token);
+                    localContent = localContent.OrderByDescending(info => info.LastAccessTimeUtc);
+
+                    var proactiveReplicationBlock = new ActionBlock<ContentHash>(
+                        async hash => await _proactiveReplicationTaskFactory(context, hash),
+                        new ExecutionDataflowBlockOptions()
+                        {
+                            MaxDegreeOfParallelism = _configuration.ProactiveReplicationConcurrencyLimit
```
SergeyTeplyakov: ProactiveReplicationDegreeOfParallelism? For consistentcy's sake.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1380,6 +1410,47 @@ public Task<BoolResult> RegisterLocalLocationAsync(OperationContext context, IRe
...
+                    var contentCopied = 0;
+                    foreach (var content in localContent)
+                    {
+                        if (Database.TryGetEntry(context, content.ContentHash, out var entry))
+                        {
+                            if (entry.Locations.Count < _configuration.ProactiveCopyLocationsThreshold)
+                            {
+                                await proactiveReplicationBlock.SendAsync(content.ContentHash);
+
+                                contentCopied++;
```
SergeyTeplyakov: We need a special tracing here similar to one we have in garbage collection with some stats: how may items were replicated, how many successes and failures etc.
And we need special counters for each individual replication.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -1380,6 +1410,47 @@ public Task<BoolResult> RegisterLocalLocationAsync(OperationContext context, IRe
...
+
+                    var proactiveReplicationBlock = new ActionBlock<ContentHash>(
+                        async hash => await _proactiveReplicationTaskFactory(context, hash),
+                        new ExecutionDataflowBlockOptions()
+                        {
+                            MaxDegreeOfParallelism = _configuration.ProactiveReplicationConcurrencyLimit
+                        });
+
+                    var contentCopied = 0;
+                    foreach (var content in localContent)
```
SergeyTeplyakov: Maybe we need a delay between enumerations.

We can easily overwhelm the system with all the proactive replication requests happening in the entire stamp.

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
SergeyTeplyakov: Most likely it should be 1, because we need (most likely) to have less then 1 active proactive replication per hash at a time, otherwise we can overwhelm the system.

---
In reply to: [346986211](https://github.com/microsoft/BuildXL/pull/1220#discussion_r346986211) [](ancestors = 346986211)

SergeyTeplyakov: We have a limit right now. So maybe we don't have to have a delay between copies.

---
In reply to: [346987150](https://github.com/microsoft/BuildXL/pull/1220#discussion_r346987150) [](ancestors = 346987150,346986211)

SergeyTeplyakov: But I'm not sure that we need to do this in parallel anyways.

---
In reply to: [346987563](https://github.com/microsoft/BuildXL/pull/1220#discussion_r346987563) [](ancestors = 346987563,346987150,346986211)

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
JuanCarlosGI: Wouldn't playing with max degree of parallelism achieve the same thing?

---
In reply to: [346986895](https://github.com/microsoft/BuildXL/pull/1220#discussion_r346986895) [](ancestors = 346986895)

dcwuser: Not unless there is a delay inside the proactive replication task. Suppose we want each machine to generate only 1 proactive replication per minute. You can't achieve that by setting MaxDegreeOfParallelism = 1.

---
In reply to: [347010635](https://github.com/microsoft/BuildXL/pull/1220#discussion_r347010635) [](ancestors = 347010635,346986895)

Public/Src/Cache/ContentStore/Distributed/NuCache/TransitioningContentLocationStore.cs
```diff
@@ -252,6 +252,17 @@ public Result<MachineLocation> GetRandomMachineLocation(IReadOnlyList<MachineLoc
             return _localLocationStore.ClusterState.GetRandomMachineLocation(except);
         }
 
+        /// <inheritdoc />
+        public Result<MachineLocation[]> GetDesignatedLocations(ContentHash hash)
+        {
+            if (_configuration.HasReadMode(ContentLocationMode.LocalLocationStore))
```
dcwuser: Minor: This organization breaks both conventions for if/else. Convention 1:
  if (exception) { do exceptional thing; }
  do normal thing;
Convention 2:
  if (exception) {
    do exceptional thing;
  } else {
    do normal thing;
  }
But nobody does:
  if (!exception) { do normal thing; }
  do exceptional thing;

Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.ContractsLight;
+using BuildXL.Cache.ContentStore.Hashing;
+using BuildXL.Cache.ContentStore.Utils;
+
+namespace BuildXL.Cache.ContentStore.Distributed.NuCache
+{
+    internal class BinManager
+    {
+        private const int MaxBins = 64 * 1024;
```
ljcollins25: >64 * 1024; [](start = 36, length = 10)

ushort.MaxValue

Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
```diff
@@ -0,0 +1,146 @@
...
+        private readonly MachineLocation[] _bins;
+        private readonly int _locationsPerBin;
+        private readonly int _entriesPerMachine;
+
+        private MachineLocation[][] _machinesForBins;
+
+        private readonly IContentHasher _contentHasher = ContentHashers.Get(HashType.MD5);
+
+        /// <nodoc />
+        public BinManager(int machinesPerBin, int entriesPerMachine, int amountOfBins)
```
ljcollins25: >int entriesPerMachine, int amountOfBins [](start = 46, length = 39)

At least in the scheme, I had in mind, these values would not need to be specified. Rather, they would be computed based on the number of active machines.

Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            // Invalidate current configuration.
+            _machinesForBins = null;
+        }
+
+        /// <nodoc />
+        public MachineLocation[] GetLocations(ContentHash hash)
+        {
+            _machinesForBins ??= ComputeBins();
+            var index = BitConverter.ToUInt16(hash.ToByteArray(), 1);
```
ljcollins25: >BitConverter.ToUInt16(hash.ToByteArray(), 1) [](start = 24, length = 44)

Maybe just convert manually by grabbing the first two bytes from the hash.

Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
```diff
@@ -0,0 +1,146 @@
...
+            return _machinesForBins[index % _bins.Length];
```
ljcollins25: >_machinesForBins [](start = 19, length = 16)

Race condition here with _machineForBins being marked null. You should capture into a local or lock around mutations to _machinesForBins.

Public/Src/Cache/ContentStore/Distributed/NuCache/ClusterState.cs
```diff
@@ -57,6 +58,19 @@ public sealed class ClusterState
         /// </summary>
         public IReadOnlyList<MachineId> InactiveMachines { get; private set; } = CollectionUtilities.EmptyArray<MachineId>();
 
+        private readonly BinManager _binManager;
```
ljcollins25: >_binManager [](start = 36, length = 11)

Can we expose this so the master can log the bins?

Public/Src/Cache/ContentStore/Distributed/NuCache/LocalLocationStore.cs
```diff
@@ -223,14 +228,16 @@ private CentralStorage CreateCentralStorage(CentralStoreConfiguration configurat
         }
 
         /// <nodoc />
-        public void PreStartupInitialize(Context context, ILocalContentStore localStore, IDistributedContentCopier copier)
+        public void PreStartupInitialize(Context context, ILocalContentStore localStore, IDistributedContentCopier copier, Func<OperationContext, ContentHash, Task<ResultBase>> proactiveCopyTaskFactory)
```
ljcollins25: >Func<OperationContext, ContentHash, Task<ResultBase>> proactiveCopyTaskFactory [](start = 123, length = 78)

This seems a little strange. Likely, it is a result of doing proactive replication inside of LLS. I would think that it should instead occur inside of the DistributedContentStore.

Public/Src/Cache/ContentStore/Distributed/NuCache/BinManager.cs
JuanCarlosGI: Not the same. ushort.MaxValue == (64 * 1024) - 1

---
In reply to: [347646590](https://github.com/microsoft/BuildXL/pull/1220#discussion_r347646590) [](ancestors = 347646590)

PR #1222, opened 2019-11-15
Reduce scrubber logs

narasamdya:
In some WDG logs, we found ~74,000 lines of 
```
[3:44.874] verbose DX0858: Scrubber deletes file...
```
Most of them coming from deleting sideband files.

[AB#1647405](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1647405)

### DIFF
--- a/Public/Src/Engine/Dll/DirectoryScrubber.cs
+++ b/Public/Src/Engine/Dll/DirectoryScrubber.cs
@@ -280,7 +280,7 @@ namespace BuildXL.Engine
                                                 // File is not in the build, delete it.
                                                 if (TryDeleteFile(pm.LoggingContext, fullPath))
                                                 {
-                                                    Interlocked.Increment(ref filesRemoved);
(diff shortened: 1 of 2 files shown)


Public/Src/Engine/Dll/DirectoryScrubber.cs
```diff
@@ -408,7 +408,6 @@ private bool TryDeleteFile(LoggingContext loggingContext, string path)
             try
             {
                 FileUtilities.DeleteFile(path, waitUntilDeletionFinished: true, tempDirectoryCleaner: m_tempDirectoryCleaner);
-                Tracing.Logger.Log.ScrubbingFile(loggingContext, path);
```
almiliMSFT: instead, please add a `bool` argument to this method which denotes whether or not to log deleted file; then only suppress logging when deleting sideband files themselves, but please don't suppress logging which shared opaque outputs got deleted (but feel free to redirect those logs into a separate log file)

PR #1223, opened 2019-11-15
Change the solution name when it's created via -vsnew argument

olkononenko:
Currently, if one uses '-vsnew' to create a solution, a generated solution will be named 'BuildXLNew'. If someone only has a single solution, it's fine. However, it becomes very confusing if you have multiple enlistments on your machine.

This PR changes how the name is constructed (essentially it's the old logic but we add 'new' at the end)

### DIFF
--- a/Public/Src/IDE/Generator/IdeGenerator.cs
+++ b/Public/Src/IDE/Generator/IdeGenerator.cs
@@ -85,6 +85,12 @@ namespace BuildXL.Ide.Generator
                     solutionNameCandidate = "ideGenerated";
                 }
 
+                // If a new solution generator is used, change the solution name so users can distinguish between the old and new one.
+                if (ideConfiguration.IsNewEnabled)
+                {
+                    solutionNameCandidate += "New";
+                }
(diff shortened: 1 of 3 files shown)


Public/Src/IDE/Generator/IdeGenerator.cs
```diff
@@ -85,6 +85,12 @@ public static void Configure(ConfigurationImpl config, IStartupConfiguration sta
                     solutionNameCandidate = "ideGenerated";
                 }
 
+                // If a new solution generator is used, change the solution name so users can distinguish between the old and new one.
+                if (ideConfiguration.IsNewEnabled)
+                {
+                    solutionNameCandidate += "New";
+                }
```
mpysson: Are we ready to make "new" the standard and instead start labeling the other implementation as "old" instead?

olkononenko: I don't think enough people have tried it so we could declare that we can deprecate the old one

PR #1224, opened 2019-11-15
Add /vsTargetFramework for constraining flavors of generated VS projects

almiliMSFT:
Three additional switches are added to `bxl.ps1`:
  * `-vsNewNet472`
  * `-vsNewNetCore`
  * `-vsNewAll`

Those switches are implemented via the `/vsTargetFramework` BuildXL command-line argument.

The idea is to be able to explicitly limit generated CS projects to a set of target frameworks.

Right now, the VS generator looks at the pip graph, goes through all csc pips, groups them by their assembly name, then for each group generates a .csproj file where its `<TargetFrameworks>` property contains the union of `qualifier.targetFramework` values of all the pips in the group.

The problem is, even when a qualifier is explicitly specified (e.g., /q:DebugNet472), the pip graph will still contain some projects whose qualifier's target framework is something else (e.g., netcoreapp3.0), and therefore, the generated VS projects will contain multiple target frameworks for those projects.  That, among other things, causes "find references" in VS to return references for each target framework, which can be annoying. 

With this change, passing `-vsNewNet472` causes the VS generator to ignore pips whose target framework is different from Net472; `-vsNewNetCore` is the same, except only pips targeting netcoreapp/netstandard are accepted.  In either case, generated VS projects will all target a single framework.  On the other hand, `-vsNewAll` does not make any restrictions and so target frameworks of generated projects will reflect what was in the pip graph when `/vs` was invoked.

### DIFF
--- a/Public/Src/App/Bxl/Args.cs
+++ b/Public/Src/App/Bxl/Args.cs
@@ -17,7 +17,6 @@ using BuildXL.Utilities;
 using BuildXL.Utilities.CLI;
 using BuildXL.Utilities.Configuration;
 using BuildXL.Utilities.Tracing;
-using static BuildXL.Interop.MacOS.Memory;
 using static BuildXL.Utilities.FormattableStringEx;
 using HelpLevel = BuildXL.Utilities.Configuration.HelpLevel;
 using Strings = bxl.Strings;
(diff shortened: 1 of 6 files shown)


Public/Src/App/Bxl/Args.cs
```diff
@@ -17,7 +17,6 @@
 using BuildXL.Utilities.CLI;
 using BuildXL.Utilities.Configuration;
 using BuildXL.Utilities.Tracing;
-using static BuildXL.Interop.MacOS.Memory;
 using static BuildXL.Utilities.FormattableStringEx;
```
almiliMSFT:
```suggestion
using static BuildXL.Interop.MacOS.Memory;
using static BuildXL.Utilities.FormattableStringEx;
```

Public/Src/IDE/Generator/CsprojFile.cs
```diff
@@ -54,6 +54,13 @@ internal override void VisitProcess(Process process, ProcessType pipCategory)
                 return;
             }
 
+            // if requested, only generate project flavors that match the currently specified qualifiers
```
almiliMSFT:
```suggestion
            // if requested, only generate project flavors that match the currently specified qualifiers.
```

PR #1229, opened 2019-11-18
Reset Redis connection multiplexer due to connectivity issues.

SergeyTeplyakov:
This is a bit of WIP code review to show the idea. The remaining part is to configure and enable this feature.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreConfiguration.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreConfiguration.cs
@@ -117,6 +117,15 @@ namespace BuildXL.Cache.ContentStore.Distributed.Redis
         public bool AreBlobsSupported => BlobExpiryTimeMinutes > 0 && MaxBlobCapacity > 0 && MaxBlobSize > 0;
 
         /// <summary>
+        /// The number of consecutive redis connection errors that will trigger reconnection.
+        /// </summary>
+        /// <remarks>
+        /// Due to a bug in StackExchange.Redis, the client may fail to reconnect to Redis Azure Cache in some cases.
+        /// This configuration allows us to recreate a connection if there is no successful calls to redis and all the calls are failing with connectivity issues.
+        /// </remarks>
+        public int RedisConnectionErrorLimit { get; set; } = int.MaxValue;
+
+        /// <summary>
         /// Indicates the mode used when writing content locations
         /// </summary>
         public ContentLocationMode WriteMode { get; set; } = ContentLocationMode.Redis;
--- a/Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
@@ -179,19 +179,19 @@ namespace BuildXL.Cache.ContentStore.Distributed.Redis
 
         // In some cases, Redis.StackExchange library may fail to reset the connection to Azure Redis Cache.
         // To work around this problem we reset the connection multiplexer if all the calls are failing with RedisConnectionException.
-        private readonly int _connectionErrorCountLimit;
+        private readonly int _redisConnectionErrorLimit;
         private int _connectionErrorCount;
         private readonly object _resetConnectionsLock = new object();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RedisDatabaseAdapter"/> class.
         /// </summary>
-        public RedisDatabaseAdapter(RedisDatabaseFactory databaseFactory, string keySpace, int connectionErrorCountLimit = int.MaxValue)
+        public RedisDatabaseAdapter(RedisDatabaseFactory databaseFactory, string keySpace, int redisConnectionErrorLimit = int.MaxValue)
         {
             _databaseFactory = databaseFactory;
             KeySpace = keySpace;
-            _connectionErrorCountLimit = connectionErrorCountLimit;
-            _redisRetryStrategy = new RetryPolicy(new RedisRetryPolicy(HandleRedisException), RetryStrategy.DefaultExponential);
+            _redisConnectionErrorLimit = redisConnectionErrorLimit;
+            _redisRetryStrategy = new RetryPolicy(new RedisRetryPolicy(OnRedisException), RetryStrategy.DefaultExponential);
         }
 
         /// <summary>
@@ -300,7 +300,7 @@ namespace BuildXL.Cache.ContentStore.Distributed.Redis
                 catch (Exception ex)
                 {
                     batch.NotifyConsumersOfFailure(ex);
-                    OnRedisException(context, ex);
+                    HandleRedisExceptionAndResetMultiplexerIfNeeded(context, ex);
                     return new BoolResult(ex);
                 }
             }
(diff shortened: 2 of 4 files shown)


Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
```diff
@@ -286,16 +293,65 @@ public async Task<BoolResult> ExecuteBatchOperationAsync(Context context, IRedis
...
 
+        private void RedisOperationSucceeded()
+        {
+            // If the operation is successful, then reset the error count to 0.
+            Interlocked.Exchange(ref _connectionErrorCount, 0);
+        }
+
+        private void OnRedisException(Context context, Exception exception)
+        {
+            if (IsRedisConnectionException(exception))
```
dcwuser: Add comment explaining reasoning behind this logic.

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
```diff
@@ -286,16 +293,65 @@ public async Task<BoolResult> ExecuteBatchOperationAsync(Context context, IRedis
...
+                            context.Warning($"Reset redis connection due to connectivity issues. ConnectionErrorCount={_connectionErrorCountLimit}.");
+                            Interlocked.Exchange(ref _connectionErrorCount, 0);
+                            _databaseFactory.ResetConnectionMultiplexer();
+                        }
+                    }
+                }
+            }
+        }
+
+        private void HandleRedisException(Exception exception)
```
dcwuser: Unclear to me when HandleRedisException is the code path vs when OnRedisException is the code path. Should we only have one? If we need both, shouldn't they both route to same logic (e.g. both could trigger reset?)

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
```diff
@@ -286,16 +293,81 @@ public async Task<BoolResult> ExecuteBatchOperationAsync(Context context, IRedis
...
+
+        /// <summary>
+        /// This method "handles" redis connectivity errors and reset the connection multiplexer if there are no successful operations happening.
+        /// </summary>
+        private void HandleRedisExceptionAndResetMultiplexerIfNeeded(Context context, Exception exception)
+        {
+            if (IsRedisConnectionException(exception))
+            {
+                // Using double-checked locking approach to reset the connection multiplexer only once.
+                if (Interlocked.Increment(ref _connectionErrorCount) >= _redisConnectionErrorLimit)
```
jbayardo: >if (Interlocked.Increment(ref _connectionErrorCount) >= _redisConnectionErrorLimit) [](start = 16, length = 83)

I'd use == instead of >=. If there's a high volume of concurrent calls, we could end up having the same issue we had with the ContentLocationDatabase's flush.

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
```diff
@@ -286,16 +293,81 @@ public async Task<BoolResult> ExecuteBatchOperationAsync(Context context, IRedis
...
+        {
+            // Lets be very pessimistic here: reset the connectivity errors every time when the operation succeed or any other exception
+            // but redis connection exception occurs.
+            if (IsRedisConnectionException(exception))
+            {
+                Interlocked.Increment(ref _connectionErrorCount);
+            }
+            else
+            {
+                Interlocked.Exchange(ref _connectionErrorCount, 0);
```
jbayardo: >Interlocked.Exchange(ref _connectionErrorCount, 0); [](start = 16, length = 51)

Question: why are we resetting the error count to 0 when an exception isn't a connection exception? isn't it possible for them to be interleaved with other Redis exceptions? in such a case, we could potentially never reach the required limit

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
dcwuser: There is another code path incrementing the counter, so if we test for == here we may miss it. (Say threshold is 5. We might see 3 and then 6 because 4 and 5 were incremented by other code path.) Given the double-checked-locking, I don't see how we can get multiple threads trying to reset simultaneously.

---
In reply to: [347636197](https://github.com/microsoft/BuildXL/pull/1229#discussion_r347636197) [](ancestors = 347636197)

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
dcwuser: Intuitively, I agree with Julian here: not clear why another exception should re-assure us about the one we care about. But I don't think my thinking is ironclad and could be easily swayed.

---
In reply to: [347636933](https://github.com/microsoft/BuildXL/pull/1229#discussion_r347636933) [](ancestors = 347636933)

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreConfiguration.cs
```diff
@@ -116,6 +116,15 @@ public class RedisContentLocationStoreConfiguration : LocalLocationStoreConfigur
...
         public bool AreBlobsSupported => BlobExpiryTimeMinutes > 0 && MaxBlobCapacity > 0 && MaxBlobSize > 0;
 
+        /// <summary>
+        /// The number of consecutive redis connection errors that will trigger reconnection.
+        /// </summary>
+        /// <remarks>
+        /// Due to a bug in StackExchange.Redis, the client may fail to reconnect to Redis Azure Cache in some cases.
+        /// This configuration allows us to recreate a connection if there is no successful calls to redis and all the calls are failing with connectivity issues.
+        /// </remarks>
+        public int RedisConnectionErrorLimit { get; set; } = int.MaxValue;
```
dcwuser: This isn't a big deal, but using MaxValue to almost-certainly disable instead of null to really truely disable seems likely to generate occasional mysterious resets if we roll this out and then decide to turn it off.

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
SergeyTeplyakov: I can add a comment here that will explain why there is >=. David gave a perfect explanation here.

---
In reply to: [347641974](https://github.com/microsoft/BuildXL/pull/1229#discussion_r347641974) [](ancestors = 347641974,347636197)

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
SergeyTeplyakov: I wanted to be safe here. My idea is that if the error is sticky, then there is no other errors happening besides this one, so it is ok to reset the multiplexer when 10 (or even 5) connectivity issues happened.

If another exception does not reset the count, then the overall connection reset counter should be higher.

On the other hand, successful operation should definitely reset it anyways, so maybe reset is just fine after lets say 20 connectivity issues even if another issues were happening in the process.

---
In reply to: [347642715](https://github.com/microsoft/BuildXL/pull/1229#discussion_r347642715) [](ancestors = 347642715,347636933)

Public/Src/Cache/ContentStore/Distributed/Redis/RedisContentLocationStoreConfiguration.cs
SergeyTeplyakov: Ok. switch to nullable.

---
In reply to: [347643577](https://github.com/microsoft/BuildXL/pull/1229#discussion_r347643577) [](ancestors = 347643577)

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
SergeyTeplyakov: stopped resetting the error count.

---
In reply to: [347660296](https://github.com/microsoft/BuildXL/pull/1229#discussion_r347660296) [](ancestors = 347660296,347642715,347636933)

Public/Src/Cache/ContentStore/Distributed/Redis/RedisDatabaseAdapter.cs
```diff
@@ -286,16 +293,81 @@ public async Task<BoolResult> ExecuteBatchOperationAsync(Context context, IRedis
...
+        /// <summary>
+        /// This method "handles" redis connectivity errors and reset the connection multiplexer if there are no successful operations happening.
+        /// </summary>
+        private void HandleRedisExceptionAndResetMultiplexerIfNeeded(Context context, Exception exception)
+        {
+            if (IsRedisConnectionException(exception))
+            {
+                // Using double-checked locking approach to reset the connection multiplexer only once.
+                // Checking for greater then or equals because another thread can increment _connectionErrorCount.
+                if (Interlocked.Increment(ref _connectionErrorCount) >= _redisConnectionErrorLimit)
```
ljcollins25: >Interlocked.Increment(ref _connectionErrorCount) [](start = 20, length = 48)

I think we should not increment here since multiple threads can call this concurrently. Instead, we should only increment after resetting. Further, if another thread is currently resetting the connection, the current thread should wait, but not reset the connection.

PR #1233, opened 2019-11-18
Cancellation for slower redis

minghinmatthewlam:
- Wait until one of the redis instances completes, if succeeded, there is a retry window for second instance to complete as well, otherwise, cancel second instance from retrying.
- If the first completed instance fails, wait for second instance to complete.
- Retry window is defaulted to null, meaning both instance will be waited for (previous implementation) currently until configured.

### DIFF
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
@@ -158,14 +158,15 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         }
 
         /// <nodoc />
-        public async Task<TResult> ExecuteRedisFallbackAsync<TResult>(OperationContext context, Func<RedisDatabaseAdapter, Task<TResult>> executeAsync, [CallerMemberName]string caller = null)
+        public async Task<TResult> ExecuteRedisFallbackAsync<TResult>(OperationContext context, Func<RedisDatabaseAdapter, CancellationToken, Task<TResult>> executeAsync, [CallerMemberName]string caller = null)
             where TResult : ResultBase
         {
-            var primaryResult = await ExecuteAndCaptureRedisErrorsAsync(PrimaryRedisDb, executeAsync);
+            using var cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(context.Token);
+            var primaryResult = await ExecuteAndCaptureRedisErrorsAsync(PrimaryRedisDb, executeAsync, cancellationTokenSource.Token);
             if (!primaryResult.Succeeded && HasSecondary)
             {
                 Tracer.Info(context, $"{Tracer.Name}.{caller}: Error in {GetDbName(PrimaryRedisDb)} redis db falling back to secondary redis db: {primaryResult}");
-                return await ExecuteAndCaptureRedisErrorsAsync(SecondaryRedisDb, executeAsync);
+                return await ExecuteAndCaptureRedisErrorsAsync(SecondaryRedisDb, executeAsync, cancellationTokenSource.Token);
             }
 
             return primaryResult;
--- a/Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
+++ b/Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
@@ -70,11 +70,12 @@ namespace BuildXL.Cache.ContentStore.Distributed.NuCache
         }
 
         /// <nodoc />
-        public async Task<BoolResult> ExecuteRedisAsync(OperationContext context, Func<RedisDatabaseAdapter, CancellationToken, Task<BoolResult>> executeAsync, [CallerMemberName]string caller = null)
+        public async Task<BoolResult> ExecuteRedisAsync(OperationContext context, Func<RedisDatabaseAdapter, CancellationToken, Task<BoolResult>> executeAsync, TimeSpan? retryWindow, [CallerMemberName]string caller = null)
         {
             (var primaryResult, var secondaryResult) = await ExecuteRaidedAsync(
                 context,
                 executeAsync,
+                retryWindow,
                 concurrent: true);
 
             if (!HasSecondary)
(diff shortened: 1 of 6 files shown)


Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
```diff
@@ -84,10 +87,12 @@ public async Task<BoolResult> ExecuteRedisAsync(OperationContext context, Func<R
         }
 
         /// <nodoc />
-        public async Task<(TResult primary, TResult secondary)> ExecuteRaidedAsync<TResult>(OperationContext context, Func<RedisDatabaseAdapter, Task<TResult>> executeAsync, bool concurrent = true, [CallerMemberName]string caller = null)
+        public async Task<(TResult primary, TResult secondary)> ExecuteRaidedAsync<TResult>(OperationContext context, Func<RedisDatabaseAdapter, CancellationToken, Task<TResult>> executeAsync, TimeSpan? retryWindow, bool concurrent = true, [CallerMemberName]string caller = null)
```
ljcollins25: >retryWindow [](start = 203, length = 11)

I think we should get rid of the retry window and just cancel retries once one of the operations finishes.

SergeyTeplyakov: The retry window is effectively a feature flag here.

Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
```diff
@@ -101,8 +106,44 @@ public async Task<(TResult primary, TResult secondary)> ExecuteRaidedAsync<TResu
...
+                {
+                    // The second task is not done within a given timeout.
+                    cancellationTokenSource.Cancel();
+
+                    var failingRedisDb = GetDbName(fasterResultTask == primaryResultTask ? SecondaryRedisDb : PrimaryRedisDb);
+                    Tracer.Info(context, $"{Tracer.Name}.{caller}: Error in {failingRedisDb} redis db using result from other redis db: {fasterResultTask}");
+
+                    // Avoiding task unobserved exception if the second task will fail.
+                    // TODO ST: pass the severity into this operation to trace the failure as a debug severity message (not available yet).
+                    slowerResultTask.FireAndForget(context);
```
jbayardo: >slowerResultTask.FireAndForget [](start = 20, length = 30)

Question: shouldn't this happen before you cancel the token?

Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
minghinmatthewlam: In standup today we discussed  to leave the retryWindow, and configure the amount later on to test its effects.

Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
```diff
@@ -84,10 +87,12 @@ public async Task<BoolResult> ExecuteRedisAsync(OperationContext context, Func<R
         }
 
         /// <nodoc />
-        public async Task<(TResult primary, TResult secondary)> ExecuteRaidedAsync<TResult>(OperationContext context, Func<RedisDatabaseAdapter, Task<TResult>> executeAsync, bool concurrent = true, [CallerMemberName]string caller = null)
+        public async Task<(TResult primary, TResult secondary)> ExecuteRaidedAsync<TResult>(OperationContext context, Func<RedisDatabaseAdapter, CancellationToken, Task<TResult>> executeAsync, TimeSpan? retryWindow, bool concurrent = true, [CallerMemberName]string caller = null)
             where TResult : BoolResult
         {
-            var primaryResultTask = ExecuteAndCaptureRedisErrorsAsync(PrimaryRedisDb, executeAsync);
+            using var cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(context.Token);
```
dcwuser: This using needs a {} block.

Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
dcwuser: It shouldn't matter. The Task won't go out of scope until we leave the method, so there is no chance of an unobserved exception getting to the GC. Organizationally, it does make sense to me to do it first, but I can't think of any reason it should have to.

---
In reply to: [348685761](https://github.com/microsoft/BuildXL/pull/1233#discussion_r348685761) [](ancestors = 348685761)

Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
minghinmatthewlam: Based on https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement, I thought it'd be fine to do "using var" since calls Dispose() on the token source at the end of the method, unless we'd like to restrict to only a portion of this method.

Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
minghinmatthewlam: I will change the ordering, because the ordering makes more logical sense to me too, and it'll help me rerun the build pipelines.

Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
SergeyTeplyakov: This is a new feature in C# 8:)

---
In reply to: [348761467](https://github.com/microsoft/BuildXL/pull/1233#discussion_r348761467) [](ancestors = 348761467)

Public/Src/Cache/ContentStore/Distributed/NuCache/RaidedRedisDatabase.cs
```diff
@@ -101,8 +106,44 @@ public async Task<(TResult primary, TResult secondary)> ExecuteRaidedAsync<TResu
...
+                    Tracer.Info(context, $"{Tracer.Name}.{caller}: Error in {failingRedisDb} redis db using result from other redis db: {fasterResultTask}");
+
+                    if (fasterResultTask == primaryResultTask)
+                    {
+                        return (await fasterResultTask, default);
+                    }
+                    else
+                    {
+                        return (default, await fasterResultTask);
+                    }
```
ljcollins25: This is not safe. The '|' (or) operator for BoolResult is not null safe, so returning default will cause a null reference exception.

PR #1235, opened 2019-11-19
Report number of physically running processes to status.csv

almiliMSFT:
The new column name is `RunningProcesses`.  This is similar to `ExternalProcesses`, except that `ExternalProcesses` counts the number of processes executing under `PipExecutor`, whereas `RunningProcesses` counts the number of actual user-specified processes physically running (any child processes spawned by those main pip processes are not included here).


Public/Src/Engine/Scheduler/PipExecutor.cs
```diff
@@ -1134,7 +1134,7 @@ private static void MakeSharedOpaqueOutputIfNeeded(IPipExecutionEnvironment envi
         /// <param name="state">the pip scoped execution state</param>
         /// <param name="pip">The pip to execute</param>
         /// <param name="fingerprint">The pip fingerprint</param>
-        /// <param name="processIdListener">Callback to call when the process is actually started</param>
+        /// <param name="processIdListener">Callback to call when the process is actually started (PID is passed to it) and when the process exit (negative PID is passed to it)</param>
```
almiliMSFT:
```suggestion
        /// <param name="processIdListener">Callback to call when the process is actually started (PID is passed to it) and when the process exited (negative PID is passed to it)</param>
```

Public/Src/Engine/UnitTests/Processes/SandboxedProcessTest.cs
```diff
@@ -1085,6 +1085,31 @@ public async Task TempAccessesAreUnexpectedByDefault()
                 string.Join(Environment.NewLine, result.AllUnexpectedFileAccesses.Select(p => p.GetPath(Context.PathTable))));
         }
 
+        [Fact]
+        public async Task ProcessStartExitIsReportedToListener()
+        {
+            var srcFile = CreateSourceFile();
+
+            var fam = new FileAccessManifest(Context.PathTable);
+            fam.ReportFileAccesses = true; // We explicitly do not set ReportAccess (testing catchall reporting)
```
almiliMSFT:
```suggestion
            fam.ReportFileAccesses = false;
```

Public/Src/Tools/Execution.Analyzer/Analyzers.Core/Xldb/XldbProtobufExtensions.cs
```diff
@@ -7,6 +7,7 @@
 using BuildXL.Scheduler.Tracing;
 using BuildXL.Utilities;
 using BuildXL.Xldb.Proto;
+using Microsoft.VisualBasic;
```
almiliMSFT:
```suggestion
```

Public/Src/Engine/Scheduler/Tracing/ExecutionLog.Events.cs
```diff
@@ -1280,6 +1285,7 @@ public void Serialize(BinaryLogger.EventWriter writer)
             writer.Write(LookupRunning);
 
             writer.Write(ExternalProcesses);
```
semihokur: We might rename ExternalProcesses to PipExecutorProcesses. your call?

Public/Src/Engine/Scheduler/Distribution/LocalWorker.cs
```diff
@@ -21,6 +23,13 @@ public sealed class LocalWorker : Worker
         /// </summary>
         public ConcurrentDictionary<PipId, Unit> CurrentlyExecutingPips = new ConcurrentDictionary<PipId, Unit>();
 
+        /// <summary>
+        /// The number of pips that are currently running (i.e., the associated pip process is still alive and running)
+        /// </summary>
+        public int CurrentlyRunningPipCount => m_currentlyRunningPipCount;
```
semihokur: can we rename it to RunningProcesses to be more consistent with the column name?

narasamdya:
Why do we need `RunningProcesses` counter? Is it not the same (or almost the same) as the number of executing process pip?

Public/Src/Engine/Processes/SandboxedProcess.cs
```diff
@@ -492,6 +492,8 @@ private async Task OnProcessExitingAsync()
 
         private async Task OnProcessExited()
         {
+            m_processIdListener?.Invoke(-ProcessId);
```
narasamdya: >m_processIdListener [](start = 12, length = 19)

Does it have to go as low as `SandboxedProcess`? Can it be tracked in `SandboxedProcessPipExecutor`?

Public/Src/Engine/Scheduler/Distribution/LocalWorker.cs
narasamdya: >m_currentlyRunningPipCount [](start = 47, length = 26)

I think we need `Volatile.Read` for reading currently running pip count.

almiliMSFT:
> Why do we need `RunningProcesses` counter? Is it not the same (or almost the same) as the number of executing process pip?

That's why: to be able to see if it is indeed close to number of executing process pips (because it seems that in some cases it's not)

Public/Src/Engine/Scheduler/Distribution/LocalWorker.cs
almiliMSFT: will change

Public/Src/Engine/Processes/SandboxedProcess.cs
almiliMSFT: Good point.  It's more accurate if it's here, but I don't think it needs to be that accurate, so I'll move it.

Public/Src/Engine/Scheduler/Distribution/LocalWorker.cs
almiliMSFT: will do

Public/Src/Engine/Scheduler/Tracing/ExecutionLog.Events.cs
almiliMSFT: sounds good

Public/Src/Engine/Scheduler/Tracing/ExecutionLog.Events.cs
```diff
@@ -1215,7 +1215,12 @@ public struct StatusEventData : IExecutionLogEventData<StatusEventData>
         /// <summary>
         /// Number of externally running processes
```
almiliMSFT:
```suggestion
        /// Number of processes running under PipExecutor
```

Public/Src/Engine/Processes/SandboxedProcess.cs
almiliMSFT: done.

Public/Src/Engine/Processes/SandboxedProcessPipExecutor.cs
```diff
@@ -736,8 +740,6 @@ private bool SandboxedProcessNeedsExecuteExternal
 
         private bool ShouldSandboxedProcessExecuteExternal
             => SandboxedProcessNeedsExecuteExternal
-               // Process does not talk to BuildXL server.
-               && m_processIdListener == null
```
narasamdya: >&& m_processIdListener == null [](start = 15, length = 30)

I like this!

PR #1236, opened 2019-11-19
Remove duplicates from ObservedAccessedFileNames

olkononenko:
ObservedAccessedFileNames are a part of a PathSet. Under some conditions this list might contain multiple values with different casing. As a result we might have "different" lists in subsequent runs. This PR removes any duplicates from that list.

[AB#1648383](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1648383)

### DIFF
--- a/Public/Src/Engine/Scheduler/Fingerprints/ObservedInputProcessor.cs
+++ b/Public/Src/Engine/Scheduler/Fingerprints/ObservedInputProcessor.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.ContractsLight;
-using System.IO;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
@@ -892,10 +891,10 @@ namespace BuildXL.Scheduler.Fingerprints
             where TTarget : struct, IObservedInputProcessingTarget<TObservation>
         {
             using (var pooledPathSet = Pools.GetAbsolutePathSet())
-            using (var pooledStringIdSet = Pools.GetStringIdSet())
             {
-                HashSet<StringId> accessedFileNames = pooledStringIdSet.Instance;
                 HashSet<AbsolutePath> visitedPaths = pooledPathSet.Instance;
+
+                HashSet<StringId> accessedFileNames = new HashSet<StringId>(pathTable.StringTable.CaseInsensitiveEqualityComparer);
 
                 bool addFileNamesFromObservations = false;
                 if (!observedAccessedFileNames.IsValid)
@@ -942,7 +941,7 @@ namespace BuildXL.Scheduler.Fingerprints
                 {
                     observedAccessedFileNames = SortedReadOnlyArray<StringId, CaseInsensitiveStringIdComparer>.SortUnsafe(
                                                 accessedFileNames.ToArray(),
-                                                new CaseInsensitiveStringIdComparer(pathTable.StringTable));
+                                                pathTable.StringTable.CaseInsensitiveStringIdComparer);
                 }
 
                 if (searchPaths.Count == 0)
--- a/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprintingVersion.cs
+++ b/Public/Src/Engine/Scheduler/Fingerprints/PipFingerprintingVersion.cs
@@ -41,7 +41,8 @@ namespace BuildXL.Scheduler.Fingerprints
         /// 67: Added SourceChangeAffectedContents
         /// 68: Added ChildProcessesToBreakawayFromSandbox
         /// 69: Added dynamic existing probe.
+        /// 70: Removed duplicates from ObservedAccessedFileNames.
         /// </remarks>
-        TwoPhaseV2 = 69,
+        TwoPhaseV2 = 70,
     }
 }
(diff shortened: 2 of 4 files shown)


olkononenko:
Question: Should we go one step further and cast file names to lower (upper?) case?

narasamdya:
>             ref SortedReadOnlyArray<StringId, CaseInsensitiveStringIdComparer> observedAccessedFileNames,

I wonder why `observedAccessedFileName` is not an array of `PathAtom`. If it were, then you would not have this uniqueness problem.

---
Refers to: Public/Src/Engine/Scheduler/Fingerprints/ObservedInputProcessor.cs:883 in 4dfcdd4. [](commit_id = 4dfcdd40cd38e85b4d7636d3b8ab96b319b4b43a, deletion_comment = False)

narasamdya:
Why not use `PathAtom`? Then you have, for free, case-sensitive for Unix and case-insensitive for Win?

---
In reply to: [555299442](https://github.com/microsoft/BuildXL/pull/1236#issuecomment-555299442) [](ancestors = 555299442)

Public/Src/Engine/Scheduler/Fingerprints/ObservedInputProcessor.cs
```diff
@@ -892,11 +891,11 @@ private static bool IsPathWithinAny(PathTable pathTable, AbsolutePath path, IEnu
             where TTarget : struct, IObservedInputProcessingTarget<TObservation>
         {
             using (var pooledPathSet = Pools.GetAbsolutePathSet())
-            using (var pooledStringIdSet = Pools.GetStringIdSet())
             {
-                HashSet<StringId> accessedFileNames = pooledStringIdSet.Instance;
                 HashSet<AbsolutePath> visitedPaths = pooledPathSet.Instance;
 
+                HashSet<StringId> accessedFileNames = new HashSet<StringId>(pathTable.StringTable.CaseInsensitiveEqualityComparer);
```
narasamdya: >pathTable.StringTable.CaseInsensitiveEqualityComparer [](start = 76, length = 53)

Is it OK for non-Windows?

narasamdya:
Please add a unit test. A unit test can make clear what the code is trying to accomplish. Waiting on this.

Public/Src/Engine/Scheduler/Fingerprints/ObservedInputProcessor.cs
almiliMSFT: for now it is, and if we ever get to port this to Linux, we'll have to deal with this in a whole bunch of places

mpysson:
What was the externally observable behavior prior to this change? Was it just causing is to proliferate equivalent pathsets or were we getting misses due to the behavior?

mpysson:
Please add a unit test

Public/Src/Engine/Scheduler/Fingerprints/ObservedInputProcessor.cs
semihokur: That'd be awesome if you try to use the pooled HashSet objects. You can add that to ObservedInputProcessingState:

```
        private static volatile ObjectPool<ObservedInputProcessingState> s_pool;
        public readonly HashSet<StringId> AccessedFileNames;
        private readonly IEqualityComparer<StringId> m_comparer;
        private ObservedInputProcessingState(IEqualityComparer<StringId> comparer)
        {
            m_comparer = comparer;
            AccessedFileNames = new HashSet<StringId>(comparer);
        }
        public static ObservedInputProcessingState GetInstance(IEqualityComparer<StringId> comparer)
        {
            if (s_pool == null)
            {
                s_pool = new ObjectPool<ObservedInputProcessingState>(
                    () => new ObservedInputProcessingState(comparer),
                    state => state.Clear());
            }

            return s_pool.GetInstance().Instance;
        }
```

In theory, this is not fully thread-safe as you can overwrite s_pool variable even though s_pool is volatile. However, that's still perfectly fine.

olkononenko: I thought about doing this (though in a little bit different way, i.e., by maintaining a map from pathtable/comparer to a HashSet pool); however, pooling hashsets won't give us any perf benefits because at the end of the day we are still creating a new hashset in `SearchPathDirectoryMembershipFilter` method. So instead of making it there, I moved the logic to this method and stopped using pooled `stringid` hashset

---
In reply to: [348068346](https://github.com/microsoft/BuildXL/pull/1236#discussion_r348068346) [](ancestors = 348068346)

olkononenko: this is the same behavior as we currently have (see `SearchPathDirectoryMembershipFilter`)

---
In reply to: [347749353](https://github.com/microsoft/BuildXL/pull/1236#discussion_r347749353) [](ancestors = 347749353)

olkononenko:
in wdg builds we are seeing pathsets with differences essentially in how accessednames list is sorted "foo, Foo, FOO", "FOO, foo, Foo", etc

---
In reply to: [555617398](https://github.com/microsoft/BuildXL/pull/1236#issuecomment-555617398) [](ancestors = 555617398)

olkononenko:
Currently we treat all paths as case-insensitive (on both platforms)

---
In reply to: [555353405](https://github.com/microsoft/BuildXL/pull/1236#issuecomment-555353405) [](ancestors = 555353405,555299442)

PR #1238, opened 2019-11-19
Enable earlyWorkerRelease by default

semihokur:


Public/Src/Engine/Dll/Engine.cs
```diff
@@ -1154,6 +1154,12 @@ public static void PopulateLoggingAndLayoutConfiguration(CommandLineConfiguratio
                 mutableConfig.Distribution.EarlyWorkerRelease = false;
             }
 
+            if (mutableConfig.Logging.Environment == ExecutionEnvironment.OfficeAPEXLab)
+            {
+                // Disable EarlyWorkerRelease for APEX builds for now.
+                mutableConfig.Distribution.EarlyWorkerRelease = false;
+            }
```
mpysson: Let's remove this hardcoded special case. APEX isn't exactly using distribution today anyway. When we get there we can just disable it on the command line if there are some issues.

Public/Src/Engine/Dll/Engine.cs
```diff
@@ -1154,6 +1154,12 @@ public static void PopulateLoggingAndLayoutConfiguration(CommandLineConfiguratio
                 mutableConfig.Distribution.EarlyWorkerRelease = false;
             }
 
+            if (mutableConfig.Logging.Environment == ExecutionEnvironment.OfficeAPEXLab)
+            {
+                // Disable EarlyWorkerRelease for APEX builds for now.
+                mutableConfig.Distribution.EarlyWorkerRelease = false;
```
mpysson: Let's remove this hardcoded special case. APEX isn't exactly using distribution today anyway. When we get there we can just disable it on the command line if there are some issues.

PR #1239, opened 2019-11-19
Console cleanup

norahuang:
Lower the distribution related logs level to Verbose. So that they won't spam the console output.
[AB#1641018](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1641018)
[AB#1647366](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1647366)


Public/Src/FrontEnd/Core/FrontEndHostController.cs
```diff
@@ -112,7 +112,7 @@ public new Workspace Workspace
 
         private readonly PerformanceCollector m_collector;
 
-        private static readonly TimeSpan EvaluationProgressReportingPeriod = TimeSpan.FromMilliseconds(500);
+        private static readonly TimeSpan EvaluationProgressReportingPeriod = TimeSpan.FromMilliseconds(1000);
```
mpysson: Can we use `BuildXLEngine.GetTimerUpdatePeriodInMs` here for consistency across all other updating messages like this?
https://github.com/microsoft/BuildXL/blob/cc80caf8f8556801eb34ea8f90452067848b48fc/Public/Src/Engine/Dll/Engine.cs#L3335

norahuang: The evaluation usually takes a few seconds(10ish from a few builds I looked at) to finish and the GetTimerUpdatePeriodInMs is 10 secs now. So if we use GetTimerUpdatePeriodInMs, we won't actually get the update messages of the evaluation progress.

mpysson: I think that's ok to for there not to be frequent status updates for sub 10 second evaluations. The idea of those console status updates is to communicate details when something very long running so the user doesn't think the app is hung.

PR #1245, opened 2019-11-20
QTest: Add support for Flaky Suppression File

plade:
QTest flaky test suppression feature requires access to the suppression file stored in source repository.
This change adds this file to as an artifact to QTest.

[AB#1638043](https://dev.azure.com/mseng/708e929f-6bd5-415a-8daf-25b1dac08dd8/_workitems/edit/1638043)

Testing: Confirmed that the flaky suppression file gets passed to QTestExe and marks a flaky test as flaky as in in the QTest log:
`2019-11-20 12:47:35,655 DEBUG [QTestSourceRepoFlakyTestManagementClient] In target: Public\Src\FrontEnd\UnitTests\Core_release_X64, marking test: Test.BuildXL.FrontEnd.Core.FrontEndHostControllerTests.DownloadFileTestWithHash as flaky`

### DIFF
--- a/Public/Sdk/Public/Tools/QTest/Tool.QTestRunner.dsc
+++ b/Public/Sdk/Public/Tools/QTest/Tool.QTestRunner.dsc
@@ -112,9 +112,9 @@ function validateArguments(args: QTestArguments): void {
  * Find Flaky Supression file from the .config directory of source code
  */
 function findFlakyFile(): File {
-    let configDir = d`${Context.getMount("SourceRoot").path}\.config`;
-    let flakyDir = d`${configDir}\flakytests`;
-    let flakyFileName = 'CloudBuild.FlakyTests.json';
+    let configDir = d`${Context.getMount("SourceRoot").path}/.config`;
+    let flakyDir = d`${configDir}/flakytests`;
+    let flakyFileName = a`CloudBuild.FlakyTests.json`;
 
     if (File.exists(f`${flakyDir}/${flakyFileName}`)) {
         return f`${flakyDir}/${flakyFileName}`;
@@ -123,6 +123,7 @@ function findFlakyFile(): File {
     if (File.exists(f`${configDir}/${flakyFileName}`)) {
         return f`${configDir}/${flakyFileName}`;
     }
+    return undefined;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


Public/Sdk/Public/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -108,6 +108,18 @@ function validateArguments(args: QTestArguments): void {
     }
 }
 
+/**
+ * Find Flaky Supression file from the .config directory of source code
+ */
+function findFlakyFile(): File {
+    let configDir = d`${Context.getMount("SourceRoot").path}\.config`;
+    let flakyFileName = 'CloudBuild.FlakyTests.json';
+    let flakyFiles = globR(configDir, flakyFileName);
```
mpysson: > let flakyFiles = globR(configDir, flakyFileName); [](start = 3, length = 50)

It's best to avoid the glob. Can you just construct the absolute path directly?

plade: The path for the file inside .config folder is up to customer's discretion.
The .config directory is quite sparse, so with a full filename specified the globR should return quite quickly.

mpysson: Do we need to give customers the flexibility to place that file at any arbitrary path under their config directory? That makes us add that glob, which could have some performance implications on graph caching if other files in that directory churn. It also creates the problem of error handling if you find 2 files with the same name. What's the upside for BuildXL customers? (I know there probably isn't consistency for QuickBuild customers)

plade: Decided to restrict the file location to one of two places. I have pushed the change.

narasamdya:
Add "AB#1638043" to the PR description so that it links this PR automatically to AzDev work items.

Public/Sdk/Public/Tools/QTest/Tool.QTestRunner.dsc
```diff
@@ -108,6 +108,23 @@ function validateArguments(args: QTestArguments): void {
     }
 }
 
+/**
+ * Find Flaky Supression file from the .config directory of source code
+ */
+function findFlakyFile(): File {
+    let configDir = d`${Context.getMount("SourceRoot").path}\.config`;
+    let flakyDir = d`${configDir}\flakytests`;
+    let flakyFileName = 'CloudBuild.FlakyTests.json';
```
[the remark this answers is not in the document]
surthumm: Iman, The file name is fixed. We dont allow user to change the name of the file. Also, the location, we restricted it based on the suggestion from Mike to two locations. Either in .config folder or under a subfolder .config\flakytests. Hope this helps clarify.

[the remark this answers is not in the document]
narasamdya: I have concern with using `SourceRoot`. Office, for example, has their source root set in the build directory where the metabuild produces the specs for the product build. But users may want to have the file in their source `dev` directory. 

My preference is, as mentioned above, the file can be specified through `QTestArguments`. The caller of `runQTest` is the one who should find where the flaky file is. For example, for our selfhost, the BuildXLSDK.dsc can be the one that specifies the file should be located to the two locations Mike suggested. While for Office, they can control it with environment variable. But the QTestRunner itself is agnostic on the shape of the repo.

Or have I overlooked something?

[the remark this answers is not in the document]
plade: Discussed offline. Currently there is no direct way for QTest SDK to access files from enlistment root (in case it is different from `SourceRoot `) without running into file monitoring violations. If there is work required where we need to access enlistment root we will handle it in a later PR.

PR #1252, opened 2019-11-21
IsDirectorySymlinkOrJunction / TryGetReparsePointType cleanup for macOS

Kr1sso:
Get rid of extensive try/catch blocks on the hot path when determining ReparsePoint types.

**TryGetReparsePointType**
Call interop implementation directly instead of going through the System.File.IO layers and avoid potential exception handling all together.

**IsDirectorySymlinkOrJunction**
When directories don't exist, no exception is thrown, otherwise call into TryGetReparsePointType and see if the path is a symlink via calling our interop implementation and not throwing unnecessarily.

Before vs. After (**58.45% time savings**)

SandboxedProcess.DirectorySymlinkCheckingDurationMs=**13710080**
SandboxedProcess.DirectorySymlinkCheckingDurationMs=**5696262**

### DIFF
--- a/Public/Src/Utilities/Native/IO/FileUtilities.cs
+++ b/Public/Src/Utilities/Native/IO/FileUtilities.cs
@@ -560,8 +560,8 @@ namespace BuildXL.Native.IO
             }
         }
 
-        /// <see cref="IFileSystem.GetFileAttributes(string)"/>
-        public static FileAttributes GetFileAttributes(string path) => s_fileSystem.GetFileAttributes(path);
+        /// <see cref="IFileSystem.GetFileAttributes(string, bool)"/>
+        public static FileAttributes GetFileAttributes(string path, bool throwOnFailure = true) => s_fileSystem.GetFileAttributes(path, throwOnFailure);
 
         /// <see cref="IFileSystem.SetFileAttributes(string, FileAttributes)"/>
         public static void SetFileAttributes(string path, FileAttributes attributes)
@@ -727,13 +727,16 @@ namespace BuildXL.Native.IO
         /// </summary>
         public static bool IsDirectorySymlinkOrJunction(string path)
         {
-            if (Directory.Exists(path))
+            if (OperatingSystemHelper.IsUnixOS)
             {
                 var reparsePointType = FileUtilities.TryGetReparsePointType(path);
                 return reparsePointType.Succeeded && reparsePointType.Result == ReparsePointType.SymLink;
             }
 
-            return false;
+            FileAttributes dirSymlinkOrJunction = FileAttributes.ReparsePoint | FileAttributes.Directory;
+            FileAttributes attributes = FileUtilities.GetFileAttributes(path, throwOnFailure: false);
+
+            return (attributes & dirSymlinkOrJunction) == dirSymlinkOrJunction;
         }
 
 #endregion
--- a/Public/Src/Utilities/Native/IO/FileUtilities.cs
+++ b/Public/Src/Utilities/Native/IO/FileUtilities.cs
@@ -131,9 +131,9 @@ namespace BuildXL.Native.IO
 
         /// <see cref="IFileUtilities.DeleteDirectoryContents(string, bool, Func{string, bool}, ITempCleaner, CancellationToken?)"/>
         public static void DeleteDirectoryContents(
-            string path, 
-            bool deleteRootDirectory = false, 
-            Func<string, bool> shouldDelete = null, 
+            string path,
+            bool deleteRootDirectory = false,
+            Func<string, bool> shouldDelete = null,
             ITempCleaner tempDirectoryCleaner = null,
             CancellationToken? cancellationToken = default) =>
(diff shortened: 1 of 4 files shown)


Public/Src/Utilities/Native/IO/FileUtilities.cs
```diff
@@ -560,8 +560,8 @@ public static unsafe FileAttributes GetFileAttributesByHandle(SafeFileHandle fil
             }
         }
 
-        /// <see cref="IFileSystem.GetFileAttributes(string)"/>
-        public static FileAttributes GetFileAttributes(string path) => s_fileSystem.GetFileAttributes(path);
+        /// <see cref="IFileSystem.GetFileAttributes(string, bool)"/>
+        public static FileAttributes GetFileAttributes(string path, bool throwOnFailure = true) => s_fileSystem.GetFileAttributes(path, throwOnFailure);
```
narasamdya: >bool throwOnFailure = true [](start = 68, length = 26)

Having this `bool throwOnFailure` looks icky. Just like we have `DeleteFile` and `TryDeleteFile`, why don't we add `TryGetFileAttributes` ?

Public/Src/Utilities/Native/IO/FileUtilities.cs
```diff
@@ -727,23 +727,16 @@ public static Possible<string> TryGetReparsePointTarget(SafeFileHandle handle, s
...
-
-                return (attributes & dirSymlinkOrJunction) == dirSymlinkOrJunction;
-            }
-            catch (NativeWin32Exception)
-            {
-                // FileSystem.Win.
-                return false;
-            }
-            catch (BuildXLException)
+            if (OperatingSystemHelper.IsUnixOS)
```
narasamdya: >if (OperatingSystemHelper.IsUnixOS) [](start = 12, length = 35)

Instead of branching, we can add `IsDirectorySymlinkOrJunction` into `IFileSystem`, and then here we simply call `s_fileSystem.IsDirectorySymlinkOrJunction(path)`.

For Win, in `FileSystem.Win.cs`, instead of calling `GetFileAttributes`, you call `TryGetFileAttributes`, which already existed. Thus, you don't need to introduce `TryGetFileAttributes` in this class.

What do you think about that?

Public/Src/Utilities/Native/IO/Windows/FileSystem.Win.cs
```diff
@@ -3004,11 +3004,14 @@ public unsafe uint GetHardLinkCountByHandle(SafeFileHandle handle)
         }
 
         /// <inheritdoc />
-        public FileAttributes GetFileAttributes(string path)
+        public FileAttributes GetFileAttributes(string path, bool throwOnFailure = true)
         {
             if (!TryGetFileAttributes(path, out FileAttributes attributes, out int hr))
             {
-                ThrowForNativeFailure(hr, "FindFirstFileW", nameof(GetFileAttributes));
+                if (throwOnFailure)
```
narasamdya: >if (throwOnFailure) [](start = 16, length = 19)

We have TryGetFileAttributes, so we don't need condition whether to throw or not. If you follow my suggestion before, you don't need this change.

Kr1sso: TryGetFileAttributes is private. With the suggestion above we don't need it, without changing the interface it was necessary.

Public/Src/Utilities/Native/IO/IFileSystem.cs
```diff
@@ -243,12 +243,12 @@ public interface IFileSystem
         FileFlagsAndAttributes GetFileFlagsAndAttributesForPossibleReparsePoint(string expandedPath);
 
         /// <summary>
-        /// Thin wrapper for native GetFileAttributesW that throws an exception on failure
+        /// Thin wrapper for native GetFileAttributesW that throws an exception on failure by default
```
almiliMSFT: what does it do when an error happens but `throwOnFailure` is `false`?

Kr1sso: In both implementation it returns `FileAttributes.Normal` - the Windows implementation does that since the beginning.

Public/Src/Utilities/Native/IO/Unix/FileSystem.Unix.cs
```diff
@@ -1246,6 +1238,12 @@ public Possible<string> TryResolveReparsePointRelativeTarget(string path, string
 
         /// <inheritdoc />
         public bool TryWriteFileSync(SafeFileHandle handle, byte[] content, out int nativeErrorCode) => throw new NotImplementedException();
-      
+
+        /// <inheritdoc />
+        public bool IsDirectorySymlinkOrJunction(string path)
+        {
+            var reparsePointType = TryGetReparsePointType(path);
+            return reparsePointType.Succeeded && reparsePointType.Result == ReparsePointType.SymLink;
```
almiliMSFT: It looks like this returns `true` for file symlinks too, no?

PR #1256, opened 2019-11-21
Add /guard:cf and /dynamicbase to linker for all unmanaged code

erikma:
Part 2 of SDL fixes

### DIFF
--- a/Public/Sdk/Experimental/Msvc/Native/Tools/Link/Link.dsc
+++ b/Public/Sdk/Experimental/Msvc/Native/Tools/Link/Link.dsc
@@ -42,6 +42,10 @@ export const defaultLinkArguments = <Arguments>{
     enableTypeProp: true,
     enableSpecDevirt: true,
     enableVectorization: true,
+
+    // Control Flow Guard required by SDL. It needs /DYNAMICBASE (randomizedBaseAddress).
+    cfGuard: true,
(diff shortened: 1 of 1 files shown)


Public/Sdk/Experimental/Msvc/Native/Tools/Link/Link.dsc
```diff
@@ -42,6 +42,10 @@ export const defaultLinkArguments = <Arguments>{
     enableTypeProp: true,
     enableSpecDevirt: true,
     enableVectorization: true,
+
+    // Control Flow Guard required by SDL. It needs /DYNAMICBASE (randomizedBaseAddress).
+    cfGuard: true,
+    randomizedBaseAddress: true,
```
mpysson: Do we have to worry about this exposing latent memory management bugs in our code? I'm not saying we should avoid enabling it due to that, just want to know if we should watch out after deploying

erikma: Yes, it could expose something stack-related - these (required SDL 7) flags are related to stack overflow attacks previously encountered in the wild. They add magic numbers onto the stack that are checked for corruption and throw an exception if so. If there's something awry in the Detours code we'll start to see new failures.

PR #1258, opened 2019-11-21
Fix ValueCache: the factory may be called at most once for any given key

almiliMSFT:


Public/Src/FrontEnd/UnitTests/Core/EvaluationSchedulerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            : base(output)
+        { }
+
+        [Theory]
+        [InlineData(100)]
+        public void ValueCacheGetOrAddInvokesFactoryOnlyOnce(int count)
+        {
+            var evalScheduler = new EvaluationScheduler(degreeOfParallelism: 1);
+            var counter = 0;
+            var key = "some-key";
```
almiliMSFT:
```suggestion
```

Public/Src/FrontEnd/UnitTests/Core/EvaluationSchedulerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            var result = Enumerable
+                .Range(1, count)
+                .ToArray()
+                .AsParallel()
+                .Select(i => evalScheduler.ValueCacheGetOrAdd(key, () => Interlocked.Increment(ref counter)))
```
almiliMSFT:
```suggestion
                .Select(i => evalScheduler.ValueCacheGetOrAdd("some-key", () => Interlocked.Increment(ref counter)))
```

PR #1264, opened 2019-11-22
Allow null process sandbox stdout/err options

erikma:
To specify no file redirection for output streams. Typically used when using stream observers instead.

### DIFF
--- a/Public/Src/Engine/Processes/SandboxedProcessOutputBuilder.cs
+++ b/Public/Src/Engine/Processes/SandboxedProcessOutputBuilder.cs
@@ -11,6 +11,12 @@ using BuildXL.Utilities;
 
 namespace BuildXL.Processes
 {
+    /// <summary>
+    /// Handler for stdout and stderr incremental output from a sandboxed process.
+    /// Calls a stream observer if configured, and stores the first maxMemoryLength characters
+    /// in a memory buffer (StringBuilder). If stream redirection is configured and the memory
+    /// buffer overflows, writes the stream to a backing file.
+    /// </summary>
     internal sealed class SandboxedProcessOutputBuilder : IDisposable
     {
         private readonly SandboxedProcessFile m_file;
@@ -36,7 +42,6 @@ namespace BuildXL.Processes
         {
             Contract.Requires(encoding != null);
             Contract.Requires(maxMemoryLength >= 0);
-            Contract.Requires(fileStorage != null);
 
             m_stringBuilderWrapper = Pools.GetStringBuilder();
             m_stringBuilder = m_stringBuilderWrapper.Instance;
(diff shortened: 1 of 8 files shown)


smera:
>             ISandboxedProcessFileStorage fileStorage,

maybe add a [CanBeNull] ?

---
Refers to: Public/Src/Engine/Processes/SandboxedProcessInfo.cs:101 in db68319. [](commit_id = db68319351eac3f8cc55f6650f1ef00efde8954b, deletion_comment = False)

smera:
>             ISandboxedProcessFileStorage fileStorage,

And same here

---
Refers to: Public/Src/Engine/Processes/SandboxedProcessInfo.cs:137 in db68319. [](commit_id = db68319351eac3f8cc55f6650f1ef00efde8954b, deletion_comment = False)

Public/Src/Engine/Processes/SandboxedProcessStandardFiles.cs
```diff
@@ -44,6 +44,17 @@ public void Serialize(BuildXLWriter writer)
...
 
+        /// <summary>
+        /// Serializes an empty instance into the given <paramref name="writer"/>.
+        /// </summary>
+        public static void SerializeEmpty(BuildXLWriter writer)
+        {
+            Contract.Requires(writer != null);
+
+            writer.Write(string.Empty);
+            writer.Write(string.Empty);
```
smera: output and error? Maybe a comment about this?

smera:
>         public static SandboxedProcessStandardFiles Deserialize(BuildXLReader reader)

add [CanBeNull] or a comment about the null case?

---
Refers to: Public/Src/Engine/Processes/SandboxedProcessStandardFiles.cs:61 in 8636106. [](commit_id = 863610633f014ec8dcaddf701a8d7d0d5b790c91, deletion_comment = False)

Public/Src/Engine/Processes/SandboxedProcessOutputBuilder.cs
```diff
@@ -102,11 +107,16 @@ public bool AppendLine(string data)
                 {
                     HandleRecoverableIOException(() => m_textWriter.WriteLine(data));
                 }
+                else if (m_fileStorage == null && m_stringBuilder.Length >= m_maxMemoryLength)
+                {
+                    // The caller should have configured an observer, called above. If not and no backing file configured,
+                    // we stop collecting the output stream at the in-memory buffer length, dropping any remainder.
```
smera: I'm maybe missing the context, but this silent drop looks a little dangerous. Shouldn't we return false here, or some other indication that the data was not written?

erikma: Returning false breaks the stream in DetouredProcess. We actually just want to silently drop the data in this case, but note there is a call to the observer above. If there is no file redirection and no observer then the stream effectively goes to devnull

erikma: Added an assert that there should be an observer or and output file or both to help avoid issues.

PR #1266, opened 2019-11-25
ValueCache dictionary should not be static

almiliMSFT:


Public/Src/FrontEnd/Core/EvaluationScheduler.cs
```diff
@@ -33,7 +33,7 @@ public sealed class EvaluationScheduler : IEvaluationScheduler
 
         private readonly ActionBlock<QueueItem> m_queue;
 
-        private static readonly ConcurrentDictionary<string, Lazy<object>> m_valueCache = new ConcurrentDictionary<string, Lazy<object>>();
+        private readonly ConcurrentDictionary<string, Lazy<object>> m_valueCache = new ConcurrentDictionary<string, Lazy<object>>();
```
almiliMSFT: I mistakenly left this static in the first place

PR #1269, opened 2020-01-14
clean up to handle consuming source files

mmiller-msft:
analyzer was throwing an exception on reads from enlistment <source> files.  Also was not reporting directory consumption properly.  Added all consumed files properly and removed some nonfunctioning code


dannyvv:
Worked with @mmiller-msft and the PR has been merged through Azure DevOps with commit: f9fedbafffd93227110f03b860c0e80391261a8e

PR #1271, opened 2020-04-23
Reduce trace volume from Detours plugin export finding algorithm

erikma:
Only trace when a suitable export cannot be found.


erikma:
Very strange git history, dropping this version of the PR.

PR #1287, opened 2021-04-07
Fix FileFlagPosixSemantics value (it was missing a 0 at the end)

yitzhaks:
While browsing the code for this repo, the C# compiler pointed out to me that there are duplicate values in the `FileFlagsAndAttributes` enum. Since that was unexpected, I took a closer look and found one value was wrong, and surprisingly the other duplication is by design (though I haven't fully understood the design there).
So I fixed the wrong value and added a comment to the other value explaining why it is by design.


marcelolynch:
Hello @yitzhaks, thanks a lot for the PR. Unfortunately due to how we are set up right now we can't accept PRs from GitHub. I created an internal PR with your commit and the changes will be pushed soon to this repo, but I made a mistake and squash-committed so your authorship was gone -- so sorry!

yitzhaks:
Thanks @marcelolynch! No worries about the authorship. I'm just glad the value is correct now.

PR #1292, opened 2021-09-15
Fixed Typo (GrpcWorkerServer-> GrpcOrchestratorServer) in Distributed-Builds.md

Pooffick:


mpysson:
Thanks for the contribution! We have some validations that are internal only. We'll commit your change internally once those validations pass and then push it back to the public repo.

Pooffick:
@mpysson , thanks for the answer!
I suppose the request can be closed?

mpysson:
@Pooffick  - Yup, I'll close it now. I was just waiting for the sync of your commit back to GitHub

PR #1294, opened 2021-09-24
Fix typo in Correctness section

jimmys:
Corrected a minor typo in the Correctness section of the WhyBuildXL document.


mpysson:
Thanks for the contribution! We have some validations that are internal only. I'm committed your change internally and it is has cloned back to GitHub.

PR #1298, opened 2022-01-10
Update SendReport.cpp to skip files with newlines in their name

DamGouz:
We have encountered a bug where the reportaccesses example breaks when a file with a newline in its name is reported by DetoursServices
The proposed solution treats any such invalid filename as an empty filename

### DIFF
--- a/Public/Src/Sandbox/Windows/DetoursServices/SendReport.cpp
+++ b/Public/Src/Sandbox/Windows/DetoursServices/SendReport.cpp
@@ -81,7 +81,11 @@ void ReportFileAccess(
     else {
         fileName = policyResult.GetCanonicalizedPath().GetPathString();
     }
-
+    
+    if (wcspbrk((wchar_t*)fileName, L"\u000A\u000B\u000C\u000D\u0085\u2028\u2029")) {
+        fileName = L"";
+    }
+    
     if (fileName == nullptr) {
         fileName = L"";
     }


johnterickson:
@damgouz thank you for the contribution (and welcome to the Microsoft if you are the Damianos Gkouzkouris from Microsoft)!

johnterickson:
@DamGouz @jan11011977  Let me know if you all are using BuildXL and if we can help you onboard. My email is in my profile.

Public/Src/Sandbox/Windows/DetoursServices/SendReport.cpp
```diff
@@ -81,7 +81,11 @@ void ReportFileAccess(
     else {
         fileName = policyResult.GetCanonicalizedPath().GetPathString();
     }
-
+    
+    if (wcspbrk((wchar_t*)fileName, L"\u000A\u000B\u000C\u000D\u0085\u2028\u2029")) {
+        fileName = L"";
```
pgunasekara, line 86: This will cause the bxl engine to ignore this file access since we're ignoring the rest of the filename if a newline character is found by always setting the filename to empty.
https://github.com/microsoft/BuildXL/blob/4663518d2d9afb8ae9884b768816b596054ce176/Public/Src/Engine/Processes/SandboxedProcessReports.cs#L692
This is probably not the behaviour you want.

Public/Src/Sandbox/Windows/DetoursServices/SendReport.cpp
```diff
@@ -81,7 +81,11 @@ void ReportFileAccess(
     else {
         fileName = policyResult.GetCanonicalizedPath().GetPathString();
     }
-
+    
+    if (wcspbrk((wchar_t*)fileName, L"\u000A\u000B\u000C\u000D\u0085\u2028\u2029")) {
```
pgunasekara, line 85: Move this to run after the null check?

pgunasekara (changes requested):

narasamdya (changes requested):

Public/Src/Sandbox/Windows/DetoursServices/SendReport.cpp
DamGouz: I think this is precisely what we want. In case a filename is broken by having a newline character, we would ideally want to skip it, as any attempt to process it through ReportAccesses leads to incorrect parsing of the filename.

Public/Src/Sandbox/Windows/DetoursServices/SendReport.cpp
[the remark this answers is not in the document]
narasamdya: Another thing is, perhaps the only characters that need escaping is `\r\n` as they are used to delimit the end of a single report.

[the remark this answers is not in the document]
DamGouz: Sounds like a plan. I will work on it and update the request. Thank you for your time.

Public/Src/Sandbox/Windows/DetoursServices/SendReport.cpp
```diff
@@ -81,7 +81,11 @@ void ReportFileAccess(
     else {
         fileName = policyResult.GetCanonicalizedPath().GetPathString();
     }
-
+    
+    if (wcspbrk((wchar_t*)fileName, L"\u000A\u000B\u000C\u000D\u0085\u2028\u2029")) {
+        fileName = L"";
+    }
```
narasamdya, line 87: For verification, please add a unit test. You can add a unit test in [PipExecutorDetoursTest.cs](https://github.com/microsoft/BuildXL/blob/master/Public/Src/Engine/UnitTests/Processes.Detours/PipExecutorDetoursTest.cs); see for example this [test](https://github.com/microsoft/BuildXL/blob/4663518d2d9afb8ae9884b768816b596054ce176/Public/Src/Engine/UnitTests/Processes.Detours/PipExecutorDetoursTest.cs#L6348)

The test harness itself is in this [DetoursTest folder](https://github.com/microsoft/BuildXL/tree/5248abbde6962043288ed353f4a9738825c0dc89/Public/Src/Sandbox/Windows/DetoursTests)

narasamdya (changes requested):
Need to modify change so that it won't cause underbuild. Plus, need to add unit test.

Public/Src/Sandbox/Windows/DetoursServices/SendReport.cpp
DamGouz: I worked on it for some time but I am struggling with the unit test, as I don't have a thorough knowledge of the codebase. I was wondering if it would be possible for someone from the BuildXL team to look at this, now that the issue has been identified?
Thank you

pgunasekara: Thanks for reporting this to us and helping with the fix! I will pick up this PR for you to make escape character changes and add the unit test.

pgunasekara: Sorry about the wait @DamGouz! I have pushed a fix for this issue for you: https://github.com/microsoft/BuildXL/commit/e58b9c97323d7620096207cc501a9c689389e2ec

Thank you for reporting this bug to us!

PR #1307, opened 2022-09-19
Keep running if targetDir is already deleted

NotWoods:
Small change to downloadHelper to check if the removed directory existed. If the directory doesn't exist, keep running since there's nothing to delete. 

Fixes first time runs.


mpysson:
Hi @NotWoods,

Since you are a Microsoft employee, we ask that you please make this change to the [Microsoft internal version](https://mseng.visualstudio.com/Domino/_git/BuildXL.Internal) of this repo. The changes will then be synchronized back to GitHub.

Thanks!

PR #1312, opened 2022-11-13
Avoid creating the same directories multiple times

dennis-yemelyanov:
When running builds with a relative large number of files to cache and fewer folders, time spent in `CreateDirectory` can be significant. To avoid calling `Directory.CreateDirectory` more than needed, cache already created folders


dennis-yemelyanov:
Before:
before2

After:
after2

PR #1348, opened 2025-07-02
Fix client update shell command built from environment values

odaysec:
https://github.com/microsoft/BuildXL/blob/ff358811bb34f3b0398b0258f1ec5dbca3f8f841/Public/Src/IDE/VsCode/client/src/update.ts#L124-L124

Dynamically constructing a shell command with values from the local environment, such as file paths, may inadvertently change the meaning of the shell command. Such changes can occur when an environment value contains characters that the shell interprets in a special way, for instance quotes and spaces. This can result in the shell command misbehaving, or even allowing a malicious user to execute arbitrary commands on the system.

fix the `update.ts` the shell command should be constructed in a way that avoids interpretation by the shell. Instead of interpolating the potentially tainted `latestUpdatePath` directly into the command string, the `spawn` function should use its argument array to pass the dynamic values separately. This ensures that special characters in filenames are treated as literal values rather than being interpreted by the shell.
1. Using `spawn` or `execFile` with arguments passed separately instead of interpolating them into the command string.
2. Validating or sanitizing the `latestUpdatePath` and `file` variables to ensure they do not contain unexpected or malicious input.

### DIFF
--- a/Public/Src/IDE/VsCode/client/src/update.ts
+++ b/Public/Src/IDE/VsCode/client/src/update.ts
@@ -121,7 +121,8 @@ export function checkForUpdates() {
                                     if (file.toLowerCase().startsWith("dscriptvscodeextension.")) {
                                         const vscodeProcessPath = path.dirname(process.execPath);
                                         const vsCpdeCommandFile = path.join(vscodeProcessPath, "bin", "code.cmd");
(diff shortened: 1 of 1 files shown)


smera:
Hello @odaysec,
Thanks for your interest in BuildXL! Unfortunately, per our [contributing guide](CONTRIBUTING.md), we ask that contributions are limited to substantial changes as opposed to cleanups/minor fixes.

Thanks,
Serge.

